PR #9, opened 2019-05-20
Tests/coverage added for alerts

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20497]

### DIFF
--- a/test/integration/services/alert.js
+++ b/test/integration/services/alert.js
@@ -7,9 +7,7 @@ context('Alert Service', function() {
       .clock();
 
     cy
-      .window()
-      .should('have.property', 'Radio')
-      .then(Radio => {
+      .getRadio(Radio => {
         Radio.request('alert', 'show:success', 'success');
       });
 
@@ -30,9 +28,7 @@ context('Alert Service', function() {
       });
 
     cy
-      .window()
-      .should('have.property', 'Radio')
-      .then(Radio => {
+      .getRadio(Radio => {
         Radio.request('alert', 'show:info', 'info');
       });
 
@@ -41,9 +37,7 @@ context('Alert Service', function() {
       .contains('info');
 
     cy
-      .window()
-      .should('have.property', 'Radio')
-      .then(Radio => {
+      .getRadio(Radio => {
         Radio.request('alert', 'show:error', 'error');
       });
 
@@ -57,9 +51,7 @@ context('Alert Service', function() {
     };
 
     cy
-      .window()
-      .should('have.property', 'Radio')
-      .then(Radio => {
+      .getRadio(Radio => {
         Radio.request('alert', 'show:undo', options);
       });
 
@@ -72,9 +64,7 @@ context('Alert Service', function() {
       });
 
     cy
-      .window()
-      .should('have.property', 'Radio')
-      .then(Radio => {
+      .getRadio(Radio => {
         Radio.request('alert', 'show:undo', options);
       });
 
--- a/test/integration/services/alert.js
+++ b/test/integration/services/alert.js
@@ -60,7 +60,7 @@ context('Alert Service', function() {
       .find('.js-dismiss')
       .click()
       .then(() => {
-        expect(options.onComplete).to.be.called;
+        expect(options.onComplete).to.be.calledOnce;
       });
 
     cy
@@ -73,7 +73,7 @@ context('Alert Service', function() {
       .find('.js-undo')
       .click()
       .then(() => {
-        expect(options.onUndo).to.be.called;
+        expect(options.onUndo).to.be.calledOnce;
       });
   });
 });
--- a/test/integration/services/alert.js
+++ b/test/integration/services/alert.js
@@ -1,0 +1,89 @@
+context('Alert Service', function() {
+  specify('Displaying', function() {
+    cy
+      .visit('/');
+
+    cy
+      .clock();
+
+    cy
+      .window()
+      .should('have.property', 'Radio')
+      .then(Radio => {
+        Radio.request('alert', 'show:success', 'success');
+      });
+
+    cy
+      .get('.alert-box')
+      .contains('success')
+      // 4000 alert delay + 900 animation fade
+      .tick(4900);
+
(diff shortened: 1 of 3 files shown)


test/integration/services/alert.js
```diff
@@ -0,0 +1,74 @@
+describe('Alert Service', function() {
+  specify('Displaying', function() {
```
paulfalgout: only line missing is the delay timeout.. should be able to set the delay, hard set a wait to the delay + animation time and check `should('not.exist')`

test/integration/services/alert.js
```diff
@@ -0,0 +1,76 @@
...
+      .should('have.property', 'Radio')
+      .then(Radio => {
+        Radio.request('alert', 'show:success', 'success');
+      });
+
+    cy
+      .get('.alert-box')
+      .contains('success')
+      .wait(4800)
+      .should('not.exist');
```
luketlancaster: that work for you @paulfalgout?

paulfalgout: Try using cy.clock here to set the time and move it forward so that the test doesn't delay.

paulfalgout: also just to be sure this test is correct, try `should('exist')` to be sure it fails... sometimes the negative tests can be false passing because maybe `wait` returns something different than we expect or whatever.  _although_... just came to me.. the other place we could test this is with another test using the `onComplete` stub.  That'd be more reliable that `not.exist` come to think of it.

luketlancaster: I've tried various iterations of this:
```js
cy
  .clock()
  .get('.alert-box')
  .contains('success')
  .should('exist')
  .then(() => {
    this.clock.tick(4800);
  })
  // or just tick()
  .should('not.exist');
```

but no dice. Thoughts on what I'm missing?

paulfalgout: the clock is mocking the javascript timer api, so the clock / tick needs to surround the method building the timeout.. so it'd be more like:
```
clock()  // prevent any timers from moving on their own
window()
...
Radio => {
 request('modal')  // creates timer
}
.should('exist')  // timer already created but no "time" has passed
tick(n)  // pass the time
.should('not.exist')

clock.restore()
```

test/integration/services/alert.js
```diff
@@ -0,0 +1,89 @@
+describe('Alert Service', function() {
```
paulfalgout: this is my bad, but this should be `context`  it's an alias of `describe` but `describe`/`its` is for the `when this happens` tdd style tests, where as `context`/`specify` works better generally for our uses.

test/integration/services/alert.js
```diff
@@ -0,0 +1,89 @@
...
+      .should('have.property', 'Radio')
+      .then(Radio => {
+        Radio.request('alert', 'show:success', 'success');
+      });
+
+    cy
+      .get('.alert-box')
+      .contains('success')
+      .should('exist')
+      .tick(4900);
```
paulfalgout: let's add a comment here like `// 4000 alert delay + 900 animation fade`

test/integration/services/alert.js
```diff
@@ -0,0 +1,89 @@
...
+      .window()
+      .should('have.property', 'Radio')
+      .then(Radio => {
+        Radio.request('alert', 'show:success', 'success');
+      });
+
+    cy
+      .get('.alert-box')
+      .contains('success')
+      .should('exist')
```
paulfalgout: we actually don't need this line.. `contains` will fail if it's not true.  as long as line 24 can be swapped to `exist` and fail, we're good.

test/integration/services/alert.js
```diff
@@ -0,0 +1,89 @@
...
+      .then(Radio => {
+        Radio.request('alert', 'show:undo', options);
+      });
+
+    cy
+      .get('.alert-box')
+      .find('.js-dismiss')
+      .click()
+      .then(() => {
+        expect(options.onComplete).to.be.called;
```
paulfalgout: let's be specific with `calledOnce`

test/integration/services/alert.js
```diff
@@ -0,0 +1,89 @@
...
+      .should('have.property', 'Radio')
+      .then(Radio => {
+        Radio.request('alert', 'show:info', 'info');
+      });
+
+    cy
+      .get('.alert-box')
+      .contains('info');
+
+    cy
```
paulfalgout: now that we're testing components a lot more, we're using this pattern everywhere.  I think it's time we make a custom command.  I'm cool with doing it in this PR or we can do it in a next PR.. up to you, but ideally the API would be:
```js
cy
  .getRadio(Radio => {
    Radio.request(...
  });
```

test/integration/services/alert.js
luketlancaster: So looks like there might be an issue with the undo then 🤔. The stub is being called twice...I'll keep looking, but any extra eyes you wanna throw on there would be helpful

test/support/commands.js
```diff
@@ -48,10 +48,11 @@ Cypress.Commands.add('tooltipContainsHtml', { prevSubject: true }, (subject, msg
...
+Cypress.Commands.add('getRadio', (channel, event, options) => {
   cy
-    .get('#alert-region')
-    .find('.alert-box')
-    .contains(msg);
+    .window()
+    .should('have.property', 'Radio')
+    .then(Radio => {
+      Radio.request(channel, event, options);
+    });
```
luketlancaster: @paulfalgout this kinda what you were thinking?

test/integration/services/alert.js
paulfalgout: sure enough.   We need to do `    const alertView = this.showAlert(_.omit(options, 'onUndo', 'onComplete'));` in the alert service.  `triggerMethod` currently looks at passed in options as possible handlers, so it's duplicating the handler.  This is due to be removed in a future version of Mn, so we should pull out the handlers so that they're not passed to the view for now.

luketlancaster:
alrighty @paulfalgout. Should be good to go now?

PR #10, opened 2019-05-24
Test modals

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20498]

thought I'd get this up and try to get a review before the weekend. I know the tests are going to need changes, so fire away

### DIFF
--- a/test/integration/services/modal.js
+++ b/test/integration/services/modal.js
@@ -1,0 +1,194 @@
+// import _ from 'underscore';
+
+context('Modal Service', function() {
+  const smallOnTall = () => {
+    cy
+      .getRadio(Radio => {
+        Radio.request('modal', 'show:tall', {
+          headingText: 'Tall Modal Header',
+          bodyText: 'Tall Modal, which is under the small modal.',
+        });
+      });
+
+    cy
+      .getRadio(Radio => {
+        Radio.request('modal', 'show:small', {
+          headingText: 'Small Modal Header',
+          bodyText: 'Small Modal, this guy shows over the top of the tall modal.',
+        });
+      });
+  };
+  specify('Displaying', function() {
+    const buttonStub = cy.stub();
+    cy
+      .visit('/');
+
+    smallOnTall();
+
+    cy
+      .get('.modal--small')
+      .should('contain', 'Small Modal')
+      .find('.modal-cancel')
+      .click();
+
+    cy
+      .get('.modal--tall')
+      .should('contain', 'Tall Modal')
+      .find('.modal-cancel')
+      .click();
+
+    // Close using the X icon
+    cy
+      .getRadio(Radio => {
+        Radio.request('modal', 'show:small', {
+          headingText: 'Small Modal Header',
+          bodyText: 'Small Modal',
+          onCancel: buttonStub,
+        });
+      });
+
+    cy
+      .get('.modal--small')
+      .find('.modal-close')
+      .click()
+      .then(() => {
+        expect(buttonStub).to.be.calledOnce;
+        buttonStub.resetHistory();
+      });
+
+    cy
+      .getRadio(Radio => {
+        Radio.request('modal', 'show:small', {
+          headingText: 'Close On Save',
+          bodyText: 'Default onSave behavior. Also a cool custom class!',
+          buttonClass: 'save-button',
+          className: 'modal--small custom-class',
+        });
+      });
+
+    cy
+      .get('.modal--small')
+      .should('have.class', 'custom-class')
+      .find('.save-button')
+      .click();
+
+    cy
+      .get('.modal--small')
+      .should('not.exist');
+
+    cy
+      .getRadio(Radio => {
+        Radio.request('modal', 'show', {
+          footerView: false,
+          bodyText: 'No Footer',
+        });
+      });
+
+    cy
+      .get('.modal')
+      .then((modal) => {
+        expect(modal).to.not.have.class('modal-footer');
+      });
+
+    cy
+      .getRadio(Radio => {
+        const modal = Radio.request('modal', 'show', {
+          bodyText: 'Submit Disabled',
+          onSubmit: () => {
+            modal.disableSubmit(true);
+          },
+        });
+      });
+
+    cy
+      .get('.modal')
+      .find('.js-submit')
+      .click()
+      .should('be.disabled');
+
+    cy
+      .getRadio(Radio => {
+        Radio.request('modal', 'show', {
+          onSubmit: buttonStub,
+        });
+      });
+
+    cy
+      .get('.modal')
+      .find('.js-submit')
+      .click()
+      .then(() => {
+        expect(buttonStub).to.be.calledOnce;
+        buttonStub.resetHistory();
+      });
+
+    cy
+      .getRadio(Radio => {
+        Radio.request('modal', 'show', {
+          onCancel: buttonStub,
+        });
+      });
+
+    cy
+      .get('.modal')
+      .find('.js-close')
+      .last() // Close Button
+      .click()
+      .then(() => {
+        expect(buttonStub).to.be.calledOnce;
+        buttonStub.resetHistory();
+      });
+
+    smallOnTall();
+
+    // click the overlay, tall should stll be there
+    cy
+      .get('.fill-window--dark')
+      .last()
+      .click('left')
+      .get('.modal--small')
+      .should('not.exist');
+
+    cy
+      .get('.modal--tall')
+      .contains('Tall Modal')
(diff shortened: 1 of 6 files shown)


test/integration/services/modal.js
```diff
@@ -0,0 +1,144 @@
...
+  };
+  specify('Displaying', function() {
+    cy
+      .visit('/');
+
+    smallOnTall();
+
+    cy
+      .get('.modal--small')
+      .within(() => {
```
paulfalgout: another way this test could be written to get around 'contains' changing the context would be:
```js
cy
 .get('.modal--small')
 .should('contain', 'Small Modal')
 .find('.modal-cancel')
 .click();
```
I'm not a huge fan of `within` unless it's providing a significant reduction in complexity.

test/integration/services/modal.js
```diff
@@ -0,0 +1,144 @@
...
+      })
+      .find('.save-button')
+      .click()
+      .should('not.exist');
+
+    cy
+      .getRadio(Radio => {
+        Radio.request('modal', 'show', {
+          footerView: false,
+          bodyText: 'No 🦶',
```
paulfalgout: what's the square here?

test/integration/services/modal.js
```diff
@@ -0,0 +1,144 @@
+// import _ from 'underscore';
```
paulfalgout: 🎗

test/integration/services/modal.js
```diff
@@ -0,0 +1,144 @@
...
+      .contains('Tall Modal')
+      .get('.fill-window--dark')
+      .click('left');
+
+    cy
+      .getRadio(Radio => {
+        const modal = Radio.request('modal', 'show', {
+          bodyText: 'Hi',
+        });
+        modal.getRegion('body').startPreloader();
```
paulfalgout: We never had this before as it was added later, but why don't we add a sugar function to the modal so that this can just be `modal.startPreloader()`

test/integration/services/modal.js
```diff
@@ -0,0 +1,144 @@
...
+      .find('.js-submit')
+      .click()
+      .should('be.disabled');
+
+    smallOnTall();
+
+    // click the overlay, tall should stll be there
+    cy
+      .get('.fill-window--dark')
+      .first()
```
paulfalgout: would this just be `.last()`?

test/integration/services/modal.js
```diff
@@ -0,0 +1,144 @@
...
+      .next()
+      .click('left')
+      .get('.modal--tall')
+      .contains('Tall Modal')
```
paulfalgout: should probably test that the other one closed

test/integration/services/modal.js
```diff
@@ -0,0 +1,144 @@
...
+      .then((modal) => {
+        expect(modal).to.not.have.class('modal-footer');
+      });
+
+    cy
+      .getRadio(Radio => {
+        const modal = Radio.request('modal', 'show', {
+          bodyText: 'Submit Disabled',
+          onSubmit: () => {
+            modal.disableSubmit(true);
```
paulfalgout: we should also test enabling it as well as that `onCancel` works.

test/integration/services/modal.js
luketlancaster: Oh shoot. Forgot to change that. It's a [foot](https://emojipedia.org/foot/) emoji

test/integration/services/modal.js
luketlancaster: Should I be testing the `onCancel` and `onSubmit` that are on the modal_views already (so just testing that the modal gets destroyed as expected), or "custom" methods like this?

paulfalgout: a yeah, I suppose there are defaults.. either way. you can make `stubs` or just test that they close I suppose.  Really we just need to test all the clicks of things and all the states of things

luketlancaster:
@paulfalgout updated with requested changes. The commit order looks weird here on github...locally it looks like this:
```
* 38287aa (HEAD -> chore/ch20498-test-modals, origin/chore/ch20498-test-modals) Adds test for modal service
* 9b26412 Return view instead of region from modal service
* 038368d Adds fonticon cross and uses it in modals
* 86297b3 Add startPreloader sugar to modal_views
```
but 🤷‍♂.

test/integration/services/modal.js
```diff
@@ -0,0 +1,182 @@
...
+      });
+
+    smallOnTall();
+
+    // click the overlay, tall should stll be there
+    cy
+      .get('.fill-window--dark')
+      .last()
+      .click('left')
+      .then((fillWindow) => {
```
paulfalgout: I think this can be `.should('not.have.class', 'modal--small')`.  same below.

test/integration/services/modal.js
```diff
@@ -0,0 +1,182 @@
...
+        expect(fillWindow).to.not.have.class('modal--small');
+      })
+      .get('.modal--tall')
```
paulfalgout: we typically do a new `cy` when the context changes

test/integration/services/modal.js
```diff
@@ -0,0 +1,182 @@
...
+
+    cy
+      .get('.spinner')
+      .get('.fill-window--dark')
+      .click('right');
+
+    cy
+      .getRadio(Radio => {
+        Radio.request('modal', 'show', {
+          headerView: 'Custom Header',
```
paulfalgout: can we test that these worked?

test/integration/services/modal.js
```diff
@@ -0,0 +1,182 @@
...
+        });
+      });
+
+    cy
+      .get('.modal')
+      .find('.js-submit')
+      .click()
+      .then(() => {
+        expect(buttonStub).to.be.calledOnce;
+        buttonStub.reset();
```
paulfalgout: I think this should be `.resetHistory` if it's available in the Cypress sinon version.  It may not be available yet though..  I think `.reset` is being removed by sinon for more specific API

test/integration/services/modal.js
```diff
@@ -0,0 +1,182 @@
...
+    cy
+      .get('.modal')
+      .find('.js-submit')
+      .click()
+      .should('be.disabled');
+
+    cy
+      .getRadio(Radio => {
+        Radio.request('modal', 'show', {
+          onSubmit: buttonStub,
```
paulfalgout: so while I did say a stub would work here, we do need to add one test to check the default handler.. mainly for code coverage.. and not really because the test is suuuper needed, but more so for watching coverage reports

test/integration/services/modal.js
```diff
@@ -0,0 +1,182 @@
...
+      .visit('/');
+
+    smallOnTall();
+
+    cy
+      .get('.modal--small')
+      .should('contain', 'Small Modal')
+      .find('.modal-cancel')
+      .click()
+      .get('.modal--tall')
```
paulfalgout: add a `cy` here

test/integration/services/modal.js
```diff
@@ -0,0 +1,182 @@
...
+          className: 'modal--small custom-class',
+          onSubmit: () => {
+            modal.destroy();
+          },
+        });
+      });
+
+    cy
+      .get('.modal--small')
+      .then((modal) => {
```
paulfalgout: `should('have.class', 'custom-class')`

test/integration/services/modal.js
```diff
@@ -0,0 +1,182 @@
...
+        expect(modal).to.have.class('custom-class');
+      })
+      .find('.save-button')
+      .click()
+      .should('not.exist');
```
paulfalgout: I think this should be a full `cy  .get('.modal--small')  .should`

test/integration/services/modal.js
paulfalgout: I'm not sure that this works though.. what even is `fillWindow` in this context? `undefined`?  really I think this should just be `get` `modal--small` `should('not.exist')`

test/integration/services/modal.js
```diff
@@ -0,0 +1,182 @@
...
+          buttonClass: 'save-button',
+          className: 'modal--small custom-class',
+        });
+      });
+
+    cy
+      .get('.modal--small')
+      .should('have.class', 'custom-class')
+      .find('.save-button')
+      .click();
```
luketlancaster: changed this one to handle the default, and let the stub take care of the custom

test/integration/services/modal.js
luketlancaster: oh geez. I just left that hanging there didn't I

luketlancaster:
@paulfalgout this look good?

PR #11, opened 2019-05-28
Feature/picklist

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch20496]


src/js/behaviors/picklist-transport.js
```diff
@@ -113,7 +106,7 @@ export default Behavior.extend({
     const nextIndex = $items.index($highlighted) + 1;
 
     if (nextIndex === $items.length) {
-      if (!this.shouldLoopList()) return $highlighted;
+      if (!this.getOptions('shouldLoopList')) return $highlighted;
```
luketlancaster: getting a `this.getOptions is not a function` error when trying to loop the list manually via arrow keys here

paulfalgout: Should be getOption

paulfalgout: Now that I'm looking at this again.. this is an older option copied over.  I have no doubt that it works, but currently there's no way to use it, and we never had and may never have a use case.  In this case I'd be fine with removing the `shouldLoopList` option altogether and simplify the whole bit.

luketlancaster: So behavior would be to not loop the list at all. If you hit the bottom and hit the down arrow, you don't go anywhere. Same with the top and the up arrow?

paulfalgout: correct

luketlancaster:
man...i really should have done a merge instead of a rebase there. old habits die hard though

PR #12, opened 2019-05-30
WIP Feature/components

paulfalgout:
Need to fix some things on selectlist and merge in the other picklist work first, plus fix the commits, but a decent WIP


luketlancaster:
I feel like there's not much I added in testing, but there's at least a bit. The coverage on actionlist in particular is not great

luketlancaster:
@paulfalgout alrighty...I did one of those tests for `onClose`, but it didn't look like the coverage numbers went up? Same for typing into the input. Thoughts on what I have/how it should change?

src/js/components/actionlist/index.js
```diff
@@ -69,9 +69,6 @@ export default Picklist.extend({
   onSelect({ model }) {
     if (model.get('isDisabled')) return;
 
-    const onSelect = model.get('onSelect');
```
paulfalgout: Ah we definitely need to call `onSelect`, we just don't need to check if `onSelect` exists prior to calling it.

paulfalgout:
I'm don't know that codecov modifies if there's a failure on the branch, but you're definitely covering the lines now in the `care-ops-frontend/coverage` reports that get generated via `npm run dev:coverage`

paulfalgout:
@luketlancaster take a look at the last few commits here.  Going to go ahead and merge though since all reports are 💯 and I need this for the menu.

PR #13, opened 2019-06-02
Add initial data layer services and helpers

paulfalgout:
```js
describe('test', function() {
  it('should run', function() {
    cy
      .server()
      .routePatient()
      .visit('/');
  });
});
```
Then `$.ajax('/api/patients/1', Radio.request('auth', 'get:ajaxAuth'))` in the console

### DIFF
--- a/test/support/helpers/json-api.js
+++ b/test/support/helpers/json-api.js
@@ -1,0 +1,53 @@
+import _ from 'underscore';
+
+function getResource(data, type) {
+  if (_.isArray(data)) {
+    return _.map(data, _.partial(getResource, _, type));
+  }
+
+  return {
+    id: data.id,
+    type,
+    attributes: _.omit(data, 'id'),
+    relationships: {},
+  };
+}
+
+function getIncluded(included = [], data, type) {
+  const resource = getResource(data, type);
+
+  if (_.isArray(resource)) {
+    included.push(...resource);
+    return included;
+  }
+
+  included.push(resource);
+  return included;
+}
+
+function getRelationship(resource, type) {
+  if (!resource) return;
+
+  if (_.isArray(resource)) {
+    return _.map(resource, _.partial(getRelationship, _, type));
+  }
+
+  return {
+    id: resource.id,
+    type,
+  };
+}
+
+function getError(detail, key) {
+  return {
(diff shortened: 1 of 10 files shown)


paulfalgout:
`$.ajax('http://localhost:8081/api/patients/c01cb96d-f4ca-4cea-bebb-56f1a1e35a83', Radio.request('auth', 'get:ajaxAuth'));` The auth service will make this work when docker's running

webpack.config.js
```diff
@@ -39,6 +39,12 @@ module.exports = {
     compress: true,
     historyApiFallback: true,
     hot: true,
+    proxy: {
+      '/api': {
+        target: 'http://localhost:8080',
+        pathRewrite: { '^/api': '' },
```
shadowhand: 💯

test/support/helpers/json-api.js
```diff
@@ -0,0 +1,53 @@
...
+
+  return {
+    id: resource.id,
+    type,
+  };
+}
+
+function getError(detail, key) {
+  return {
+    source: { pointer: `/data/attributes/${ key }` },
```
shadowhand: I don't understand what this line is trying to accomplish. We already send a `source` in BE error messages.

paulfalgout: This is just for cypress stubbing of errors.  Particularly when we need to stub an array of errors.  Never used it on mainapp as it was part of the unmerged patient sidebar, but here's an example use case: https://github.com/RoundingWell/RWell/pull/3436/files#diff-cd565288bbc9736faa2e68200bc60792R1041  It's just generating the minimal error json required for cypress to cause and test form errors without having to store it in json.

paulfalgout: BTW this PR was originally just the copied over cypress helpers for building on-the-fly stubbed data in jsonapi.. but then it felt similar enough that both dev environments, Cypress and docker could now access API.  Ideally hoping to replace the fixture files with a service that automates their creation with generic data like faker.js, and any test specific fixturing would be in the tests.. everything in collections will be the context data that needs to be there for the app to work, but isn't test specific.

paulfalgout:
The ❌ here is codecov complaining about coverage..  will be the state of things until we're testing features against the data.

PR #15, opened 2019-06-06
Adds faker generator and various faker factories

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20532]

Couldn't wrap my head around what `unit.js` was doing so I went rogue and did this. Rip 'em up and lemme know what you think. *Should* support adding any other "factories" in there, following whatever the backend has in their various [schema](https://github.com/RoundingWell/care-ops-backend/blob/develop/src/Patient/Schema/PatientSchema.php) files?


test/fixtures/config/patient-fields.js
```diff
@@ -0,0 +1,11 @@
+const faker = require('faker');
+
+module.exports = {
+  generate: () => {
+    return {
+      'id' : faker.random.uuid(),
+      'name': faker.lorem.word(),
+      'value': faker.random.number(),
```
luketlancaster: technically this can be...anything. Array, object, bool, string, number, etc. I figure we can cross that bridge when we get to it though

test/fixtures/config/groups.js
```diff
@@ -0,0 +1,11 @@
+const faker = require('faker');
+
+module.exports = {
+  generate: () => {
+    return {
+      'id' : faker.random.uuid(),
+      'name': `${faker.name.jobDescriptor()} ${faker.name.jobArea()}`,
```
luketlancaster: probably overkill

test/fixtures/README.md
```diff
@@ -3,14 +3,14 @@
...
-Ideally the fixture data (other than tests) can be randomly generated in the future.
 
 The files in the root of `fixtures/` are singular resources that are not requested
 as collections within the app.
 
 ### `collections/`
 
-Collections include arrays of multiple entities.
+Collections include arrays of multiple entities. Collections are automatically
+generated using faker when running tests.
```
luketlancaster: could probably use better documentation...this was the first thing I did last week or whenever I started on this

config/faker-generator.js
```diff
@@ -0,0 +1,18 @@
+const _ = require('underscore');
+const fs = require('fs');
+const files = fs.readdirSync('./test/fixtures/config');
+
+_.each(files, (file) => {
+  const entity_factory = require(`../test/fixtures/config/${file}`);
+  const dest = `./test/fixtures/collections/${file.split('.')[0]}.json`;
+
+  const collection = _.times(1000, () => {
```
luketlancaster: this is...a lot. I bumped it up from 100 to see how long it would take to run, and it was consistently 1 second (either way)

paulfalgout: I think 100 may be enough.  It may not take more time to generate, but Cypress will slow with lots of data.

config/faker-generator.js
```diff
@@ -0,0 +1,18 @@
...
+    return entity_factory.generate();
+  });
+
+  fs.writeFile(dest, JSON.stringify(collection), (e) => {
+    if(e) {
```
paulfalgout: The config files aren't going to fail the linter during build, but I think `npm run lint` should still error for the config files.  Needs to be run as the style is different

package.json
```diff
@@ -9,15 +9,16 @@
   "homepage": "https://RoundingWell.com",
   "scripts": {
     "analyze": "cross-env NODE_ENV=production webpack --profile --json > webpack.stats.json && webpack-bundle-analyzer webpack.stats.json",
-    "build": "cross-env NODE_ENV=production webpack",
-    "build:ci": "cross-env NODE_ENV=test webpack",
+    "build": "npm run fixtures:generate && cross-env NODE_ENV=production webpack",
```
paulfalgout: I don't think this should be tied to the build.  It's only related to Cypress at the moment, and may always be that way.  So it'd be better to call it prior to calling Cypress..  In fact, might be able to just make a Cypress plugin: https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/plugins/state-colors.js
https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/plugins/index.js#L19

And in that case all the `faker-generator.js` code could live over there..

test/fixtures/config/groups.js
```diff
@@ -0,0 +1,11 @@
+const faker = require('faker');
+
+module.exports = {
+  generate: () => {
+    return {
+      'id' : faker.random.uuid(),
```
paulfalgout: This won't fail the linter, but keys should not be `'`'d unless it's necessary to do so, and it should not be for any of these cases.

test/fixtures/config/groups.js
```diff
@@ -0,0 +1,11 @@
+const faker = require('faker');
+
+module.exports = {
+  generate: () => {
```
paulfalgout: this won't fail the linter, but `=>` is giving this function the context of the file, but isn't necessary.  On a POJO, unless context is needed, this can just be `generate() {`

config/faker-generator.js
```diff
@@ -0,0 +1,18 @@
+const _ = require('underscore');
+const fs = require('fs');
+const files = fs.readdirSync('./test/fixtures/config');
+
+_.each(files, (file) => {
+  const entity_factory = require(`../test/fixtures/config/${file}`);
```
paulfalgout: this should fail the linter but `camelCase`

config/faker-generator.js
```diff
@@ -0,0 +1,18 @@
...
+  const dest = `./test/fixtures/collections/${file.split('.')[0]}.json`;
+
+  const collection = _.times(1000, () => {
+    return entity_factory.generate();
+  });
+
+  fs.writeFile(dest, JSON.stringify(collection), (e) => {
+    if(e) {
+      console.log(e);
```
paulfalgout: this could be `console.error(e);` or even `throw e;`  `e` is an error string correct?

config/faker-generator.js
paulfalgout: could also just be `const { generate } = `

config/faker-generator.js
paulfalgout: this could also probably be `_.times(100, entityFactory.generate);`

config/faker-generator.js
```diff
@@ -0,0 +1,18 @@
+const _ = require('underscore');
+const fs = require('fs');
```
paulfalgout: I think what this would look like with the `unit.js` example is:
```js
const req = require.context('../test/fixtures/config/', true, /^(.*\.(js$))[^.]*$/i);

req.keys().forEach(key => {
  const entityFactory = req(key);
  const dest = key.replace('/config'. '/collections').replace('.js', '.json');

  // etc
});
```
That said, this is node so what you have is fine

config/faker-generator.js
```diff
@@ -0,0 +1,18 @@
...
+
+_.each(files, (file) => {
+  const entity_factory = require(`../test/fixtures/config/${file}`);
+  const dest = `./test/fixtures/collections/${file.split('.')[0]}.json`;
+
+  const collection = _.times(1000, () => {
+    return entity_factory.generate();
+  });
+
+  fs.writeFile(dest, JSON.stringify(collection), (e) => {
```
paulfalgout: might want to add `JSON.stringify(collection, null, 2);` here.  It'll make the files readable

test/fixtures/config/patient-fields.js
paulfalgout: yeah interesting.. in the future I could see passing in the key from `_.times` and iterating with that key over an array of types to get the various types needed, but we don't need that right away.  Really we won't want to test the types that way, but having the randomized data look as much like an actual case as possible is always preferred.

test/fixtures/config/patients.js
```diff
@@ -0,0 +1,14 @@
...
+
+module.exports = {
+  generate: () => {
+    return {
+      'id' : faker.random.uuid(),
+      'first_name' : faker.name.firstName(),
+      'last_name' : faker.name.lastName(),
+      'birth_date' : faker.date.past(),
+      'sex' : faker.random.arrayElement(['m', 'f']),
+      'mrn' : faker.random.number(),
```
paulfalgout: is this actually a number?  I think at least it's a string that might be a number so `${ faker.random.number() }`?

test/fixtures/README.md
paulfalgout: Yeah so we should document
```
### `config/`
What this folder includes and what to do about it
```

test/fixtures/config/groups.js
paulfalgout: produces some really great results though 🤣

config/faker-generator.js
luketlancaster: Looks like it's an...error...[object?](https://nodejs.org/api/fs.html#fs_fs_writefile_file_data_options_callback)? Those docs say it's throw-able, so I'll just do that

config/faker-generator.js
luketlancaster: so the first thing I ran into was `context` not being a function...because I was running this through node instead of webpack? Is that a thing?

config/faker-generator.js
luketlancaster: was wondering if there was some magic that could do that, but I only looked at the JSON object, not down into the `stringify` method itself 🙃

test/fixtures/config/patients.js
luketlancaster: yup, looks like it gets converted to a string by the backend 👍

shadowhand:
Just worth noting that [Prism](https://github.com/stoplightio/prism) might be a better option than manually setting up faker. It supports (in v3) automatic mocking from an `openapi.json` file.

config/faker-generator.js
paulfalgout: ah yeah I dunno maybe that's the case.  FWIW, if it's moved to a cypress plugin I believe that'd no longer be a problem anyhow.  It's kind of interesting as a Cypress plugin.. might be worth publishing at some point

luketlancaster: Yeah, when I first started on this I was thinking that with a little work it could be pushed up separately.

luketlancaster:
@paulfalgout ok...I pushed up some changes. Couldn't get the `require.context` to work out when using this as a plugin either, so I left it alone.

luketlancaster:
gave 'em a rebase... 🤞

PR #17, opened 2019-06-10
First pass of adding entity + service boilerplates

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20533]

### DIFF
--- a/src/js/entities-service/entities/clinicians.js
+++ b/src/js/entities-service/entities/clinicians.js
@@ -1,0 +1,22 @@
+import BaseEntity from 'js/base/entity-service';
+import { _Model, Model, Collection } from './entities/clinicians';
+
+const Entity = BaseEntity.extend({
+  Entity: { _Model, Model, Collection },
+  radioRequests: {
+    'clinicians:model': 'getModel',
+    'clinicians:collection': 'getCollection',
+    'fetch:clinician:model': 'fetchClinician',
+  },
+  fetchClinician(id) {
+    const include = [
+      'role',
+    ];
+
+    const data = { include };
+
+    return this.fetchModel(id, { data });
+  },
+});
+
+export default new Entity();
--- a/src/js/entities-service/entities/roles.js
+++ b/src/js/entities-service/entities/roles.js
@@ -1,0 +1,12 @@
+import BaseEntity from 'js/base/entity-service';
+import { _Model, Model, Collection } from './entities/roles';
+
+const Entity = BaseEntity.extend({
+  Entity: { _Model, Model, Collection },
+  radioRequests: {
+    'roles:model': 'getModel',
+    'roles:collection': 'getCollection',
+  },
+});
+
+export default new Entity();
--- a/src/js/entities-service/clinicians.js
+++ b/src/js/entities-service/clinicians.js
@@ -1,0 +1,21 @@
+import Store from 'backbone.store';
+import BaseCollection from 'js/base/collection';
+import BaseModel from 'js/base/model';
+
+const TYPE = 'clinicians';
+
+const _Model = BaseModel.extend({
+  type: TYPE,
+  urlRoot: '/api/clinicians',
+});
+
+const Model = Store(_Model, TYPE);
+const Collection = BaseCollection.extend({
+  model: Model,
+});
+
+export {
+  _Model,
+  Model,
+  Collection,
+};
--- a/src/js/entities-service/entities/groups.js
+++ b/src/js/entities-service/entities/groups.js
@@ -1,0 +1,21 @@
+import Store from 'backbone.store';
+import BaseCollection from 'js/base/collection';
+import BaseModel from 'js/base/model';
+
+const TYPE = 'groups';
+
+const _Model = BaseModel.extend({
+  type: TYPE,
+  urlRoot: '/api/groups',
+});
+
+const Model = Store(_Model, TYPE);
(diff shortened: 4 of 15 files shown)


src/js/entities-service/entities/roles.js
```diff
@@ -0,0 +1,22 @@
+import BaseEntity from 'js/base/entity-service';
+import { _Model, Model, Collection } from './entities/roles';
+
+const Entity = BaseEntity.extend({
+  Entity: { _Model, Model, Collection },
+  radioRequests: {
+    'roles:model': 'getModel',
+    'roles:collection': 'getCollection',
+    'fetch:role:model': 'fetchRole',
```
paulfalgout: I think we can remove `fetchRole` entirely.  Until a use case arises, roles should be treated like `states` Just options available to the organization.

src/js/entities-service/entities/clinicians.js
```diff
@@ -0,0 +1,22 @@
...
+const Entity = BaseEntity.extend({
+  Entity: { _Model, Model, Collection },
+  radioRequests: {
+    'clinicians:model': 'getModel',
+    'clinicians:collection': 'getCollection',
+    'fetch:clinician:model': 'fetchClinician',
+  },
+  fetchClinician(id) {
+    const include = [
+      'role',
```
shadowhand: @luketlancaster @paulfalgout the care ops backend does not support includes, and I don't really want to add support for them because it makes for some really questionable performance.

From my perspective, the JS models should use the `links` attached to responses to generate new promises to fetch related data on demand. For instance (syntax might be different):

```js
let roles = await model.clinicians.get(clinicianId).roles();
```

paulfalgout: just to be clear here.. you're suggesting when we have a list of things that have relationships.. like the action list that relates to the patient and clinician that we should make (number of results * number of relationships + 1) requests to show the list?

shadowhand: Multiple requests in parallel, yes. Something like https://www.npmjs.com/package/swagger-fluent would probably help.

paulfalgout: well we could setup a node server to do the packaging.  that `swagger-fluent` lib is for node.  _technically_ it can be run in the browser but it's over 3M, and the author of `request`, it's dependency suggests, it shouldn't be used in the browser.  To my knowledge most frontend solutions have been working to reduce roundtrips because in addition to response time there's latency and concurrency limitations with browsers.  Perhaps we could find some other resources that suggest this is a good path or a library with more than one contributor and a user-base that does it this way?

paulfalgout: Also even in your example it wouldn't be 5 async requests.  It'd be 1 request vs 1 request + 4 async requests (assuming only 1 level of relationships)

shadowhand: https://hackernoon.com/5-small-steps-to-a-better-web-api-6fa9698178a6
https://softwareengineering.stackexchange.com/a/305875

I assume that the top level item will probably already have a local reference. For instance:

- when the clinician first accesses the page:
  - parallel (clinician, organization)
  - parallel (groups, roles)
- when the clinician views a group:
  - await patients
- when the clinician views a patient:
  - (patient already loaded)
  - parallel (groups, fields, actions, ...)

shadowhand: Also, HTTP/2 support when available makes parallel async *much* faster than a single large request. It also utilizes lambda better, as lambda will automatically scale the server up and down to deal with higher traffic. Lambda will (probably) not scale up as quickly for a small number of requests that take a long time to complete.

paulfalgout: I think it'd be more like:
- when the clinician first accesses the page:
  - parallel (clinician, organizations)
  - parallel (clinician.groups, clinician.roles)
  - parallel (each group's group.clinicians needed across the app for various functionality) - we'll be loading any clinician in multiple groups multiple times.
- when the clinician views a group:
  - (group already loaded via clinician.groups)
  - await group.patients
- when the clinician views a patient:
  - ~~(patient already loaded via group.patients)~~ since we don't know for sure if the patient data is still valid, load it again
  - parallel (patient.groups, patient.fields, patient.actions, ...)

shadowhand: If loading all of the organization groups, do:

- ~~parallel (clinician.groups, clinician.roles)~~ parallel (organization.groups, organization.roles)

shadowhand: Oh and this article (which I was hunting for and finally found):

https://apisyouwonthate.com/blog/lets-stop-building-apis-around-a-network-hack

test/support/api/patient.js
```diff
@@ -14,11 +14,11 @@ Cypress.Commands.add('routePatient', (mutator = _.identity) => {
       let included = [];
 
       included = getIncluded(included, this.fxGroups, 'groups');
-      included = getIncluded(included, this.fxPatientFields, 'patient-fields');
+      included = getIncluded(included, this.fxFields, 'fields');
```
luketlancaster: @paulfalgout should these be getting pushed onto the included array?

luketlancaster: I'm dumb and didn't see that `included` is being passed to `getIncluded`. carry on.

PR #18, opened 2019-06-14
Configure Continuous Delivery for frontend

BenCoffeed:
- [x] Add the Clubhouse Story ID: [ch20603]
- [x] Add QA tasks for anything needing testing _not_ behind a flag
- [x] Update or add any QA tasks on the feature flag card

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,24 +2,40 @@ version: 2.1
 orbs:
   cypress: cypress-io/cypress@1
   codecov: codecov/codecov@1.0.4
+  aws-s3: circleci/aws-s3@1.0.11
+
 jobs:
-  deploy:
+  build:
     executor: cypress/base-10
     steps:
       - attach_workspace:
           at: ~/
-      - run: npm run build
-      - run: echo S3 orb stuff for dist/
+      - run: cd ~/project && npm run build
+      - persist_to_workspace:
+          root: ~/project
+          paths:
+            - dist/
+  deploy:
+    docker:
+      - image: 'circleci/python:2.7'
+    steps:
+      - attach_workspace:
+          at: ~/
+      - aws-s3/sync:
+          from: ~/dist
+          to: s3://$bucket_name
+          overwrite: true
+
 workflows:
-  build:
+  test-deploy:
     jobs:
       - cypress/install:
           pre-steps:
             - run:
-                name: Set registry URL
+                name: Set FontAwesome registry URL
                 command: 'npm config set "@fortawesome:registry" https://npm.fontawesome.com/'
             - run:
-                name: Authenticate with registry
+                name: Authenticate with FontAwesome registry
                 command: 'npm config set "//npm.fontawesome.com/:_authToken" $FONTAWESOME'
           build: 'npm run build:ci'
       - cypress/run:
@@ -33,9 +49,19 @@ workflows:
             - run: npm run coverage:report:ci
             - codecov/upload:
                 file: coverage/lcov.info
+      - build:
+          filters:
+            branches:
+              only: 
+                - develop
+                - feature/ch20603/automateFEbuilds
+          requires:
+            - cypress/run
       - deploy:
           filters:
             branches:
-              only: master
+              only: 
+                - develop
+                - feature/ch20603/automateFEbuilds
           requires:
-            - cypress/run
+            - build


.circleci/config.yml
```diff
@@ -2,24 +2,65 @@ version: 2.1
...
+          name: move dist directory into built_code directory
+          command: mv ./dist built_code/
+      - persist_to_workspace:
+          root: ./built_code
+          paths:
+            - dist
   deploy:
-    executor: cypress/base-10
+    docker:
+      - image: 'circleci/python:2.7'
```
paulfalgout: I don't think this orb requires python, it only lists bash.. is there any reason not to use the previous environment?  It'll be a lot faster as you won't have to run npm install or deal with using caches

.circleci/config.yml
```diff
@@ -2,24 +2,65 @@ version: 2.1
...
     steps:
       - attach_workspace:
-          at: ~/
-      - run: npm run build
-      - run: echo S3 orb stuff for dist/
```
paulfalgout: I _kinda_ think you could drop all of the build steps added and just add this at line 12:
```yml
  - aws-s3/sync:
      from: /dist
      to: s3://$bucket_name
      overwrite: true
```

.circleci/config.yml
```diff
@@ -2,24 +2,65 @@ version: 2.1
...
+          name: Authenticate with registry
+          command: 'npm config set "//npm.fontawesome.com/:_authToken" $FONTAWESOME'
+      - run:
+          name: Install
+          command: npm i --no-dev
+      - run: 
+          name: Run build script
+          command: 'npm run build'
+      - save_cache:
+          key: npm-{{ checksum "composer.lock" }}
```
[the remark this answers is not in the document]
BenCoffeed: ha! Copy pasta caught!

.circleci/config.yml
```diff
@@ -2,24 +2,65 @@ version: 2.1
...
+            - npm-
+      - run:
+          name: Set registry URL
+          command: 'npm config set "@fortawesome:registry" https://npm.fontawesome.com/'
+      - run:
+          name: Authenticate with registry
+          command: 'npm config set "//npm.fontawesome.com/:_authToken" $FONTAWESOME'
+      - run:
+          name: Install
+          command: npm i --no-dev
```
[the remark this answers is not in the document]
BenCoffeed: So, does ci include dev dependencies?

.circleci/config.yml
BenCoffeed: AWS CLI needs python.

.circleci/config.yml
BenCoffeed: If we reuse built items, most likely, yes.

.circleci/config.yml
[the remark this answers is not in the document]
paulfalgout: yes, but that is a good thing.  the dev dependencies are what does the building, the regular dependency are shipped with the build.

.circleci/config.yml
paulfalgout: well I mean the build runs on line 11.  I don't see a need to create a new environment to do the build, the results from line 11 and the new build should be identical.

paulfalgout: well I mean the build runs on line 11.  I don't see a need to create a new environment to do the build.. the results from line 11 and the new build should be identical.

BenCoffeed: 👍 Alsot, unforunately, the aws-s3 orb from CircleCI only includes commands and not jobs. So, it still needs to be run in a defined job context.

.circleci/config.yml
BenCoffeed: See note about aws-s3 orb not having jobs defined, but only commands below.

.circleci/config.yml
```diff
@@ -2,24 +2,65 @@ version: 2.1
...
+  build:
+    docker: 
+      - image: circleci/node:10
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - npm-{{ checksum "package-lock.json" }}
+            - npm-
+      - run:
```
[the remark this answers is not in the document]
BenCoffeed: If you're comfortable with shipping what is built in the ci workflow, I'm all for not building another time.

.circleci/config.yml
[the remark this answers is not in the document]
BenCoffeed: 👍

BenCoffeed:
CodeCov is complaining. I queued up a re-run since the issues reported have nothing to do with the single file I changed.

BenCoffeed:
Yes. I intended on removing in a separate PR since the inclusion of that is really a testing step for this PR.

paulfalgout:
I think it's safe to update since the test is recorded here https://circleci.com/workflow-run/1a13cffa-1149-4d70-b880-eee2001d0143 ?  But we can run a 2nd PR if you'd prefer

PR #19, opened 2019-06-17
Adds fixtures + org and action api test helpers

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20534]

think more need to be added? I wanted to get these up for review, and I can add more in + keep working on that setRelationship helper (was giving me trouble, so put it on pause for now).

### DIFF
--- a/test/fixtures/config/actions.js
+++ b/test/fixtures/config/actions.js
@@ -1,0 +1,20 @@
+const faker = require('faker');
+const moment = require('moment');
+
+module.exports = {
+  generate() {
+    return {
+      id: faker.random.uuid(),
+      name: `${ faker.company.bsBuzz() } ${ faker.company.catchPhraseNoun() }`,
+      details: faker.lorem.sentences(),
+      due_date: faker.date.between(
+        moment().subtract(1, 'week').format(),
+        moment().add(1, 'week').format()
+      ),
+      duration: faker.random.number({
+        min: 1,
+        max: 1200,
+      }),
+    };
+  },
+};
--- a/test/fixtures/config/clinicians.js
+++ b/test/fixtures/config/clinicians.js
@@ -1,0 +1,20 @@
+const faker = require('faker');
+const _ = require('underscore');
+
+module.exports = {
+  generate() {
+    const credentials = [];
+
+    _.times(_.random(1, 3), () => {
+      credentials.push(faker.finance.currencyCode());
+    });
+
+    return {
+      id: faker.random.uuid(),
+      first_name: faker.name.firstName(),
+      last_name: faker.name.lastName(),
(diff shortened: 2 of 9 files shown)


test/fixtures/config/clinicians.js
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = {
+  generate() {
+    return {
+      id: faker.random.uuid(),
+      first_name: faker.name.firstName(),
+      last_name: faker.name.lastName(),
+      email: faker.internet.email(),
+      credentials: [
+        `${ faker.finance.currencyCode() }`,
```
shadowhand: 😆

test/support/index.js
```diff
@@ -19,3 +19,5 @@ import './commands';
 import './coverage';
 
 import './api/patient';
+import './api/actions';
```
paulfalgout: can we alphabetize these?  because why not?

test/support/api/organizations.js
```diff
@@ -0,0 +1,35 @@
+import _ from 'underscore';
+import { getResource, getIncluded, getRelationship } from 'helpers/json-api';
+
+Cypress.Commands.add('routeOrganization', (mutator = _.identity) => {
+  cy
+    .fixture('collections/organizations').as('fxOrganizations')
+    // .fixture('collections/clinicians').as('fxClinicians')
```
paulfalgout: I do suspect at some point, we'll need to be loading the clinicians for the groups as it'll take all of that info to make the assign component, but I'm ok with this for now

test/fixtures/config/clinicians.js
paulfalgout: does this need to be casted to a string? seems like it might be already?

test/fixtures/config/actions.js
```diff
@@ -0,0 +1,16 @@
+const faker = require('faker');
+
+module.exports = {
+  generate() {
+    return {
+      id: faker.random.uuid(),
+      name: `${ faker.company.bsBuzz() } ${ faker.company.catchPhraseNoun() }`,
+      details: faker.lorem.sentences(),
+      due_date: faker.date.future(),
```
paulfalgout: I think maybe we should use `between` with moment here as due dates can be passed due.  not that, we'll want to test that with this fixture, but just having more-like-real data is probably better.

test/support/index.js
luketlancaster: whoops, saw that and thought of it and then didn't remember to do it

test/support/api/organizations.js
luketlancaster: oh yeah...I didn't see this as a relationship on the org endpoint, so I just left this here.

PR #23, opened 2019-06-24
Feature/ch20621 stub bootstrap data

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20621]

### DIFF
--- a/test/integration/globals/error.js
+++ b/test/integration/globals/error.js
@@ -1,14 +1,11 @@
 context('Global Error Page', function() {
   specify('Displaying and navigating the global error page', function() {
     cy
-      .visit('/');
+      .visit('/404');
 
     cy
-      .window()
-      .should('have.property', 'Backbone')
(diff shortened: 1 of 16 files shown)


test/integration/globals/error.js
```diff
@@ -1,7 +1,7 @@
 context('Global Error Page', function() {
   specify('Displaying and navigating the global error page', function() {
     cy
-      .visit('/');
+      .visit('/nothinghere');
```
paulfalgout: actually what's happening here is that the `navigate` trigger is happening prior to the router starting because we introduced an async load before starting the router (a pattern we didn't have in mainapp because the bootstrap came with the initial html payload)   _but_ it looks like the history service still handles the back button correctly even if there's nothing to go back to.. _so_ lines 6-11 should be removed and line 4 should be `/404`

Perhaps the first test should be
```
cy
      .url()
      .should('contain', '404');
```

PR #25, opened 2019-06-26
Basic all patients list

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20619]

### DIFF
--- a/src/js/entities-service/patients.js
+++ b/src/js/entities-service/patients.js
@@ -7,6 +7,7 @@ const Entity = BaseEntity.extend({
     'patients:model': 'getModel',
     'patients:collection': 'getCollection',
     'fetch:patient:model': 'fetchPatient',
+    'fetch:patients:collection': 'fetchPatients',
   },
   fetchPatient(id) {
     const include = [
@@ -17,6 +18,15 @@ const Entity = BaseEntity.extend({
     const data = { include };
 
     return this.fetchModel(id, { data });
+  },
+  fetchPatients() {
+    const include = [
+      'groups',
+    ];
+
+    const data = { include };
+
+    return this.fetchCollection({ data });
   },
 });
 
--- a/src/js/views/patients/list/patients-all_views.js
+++ b/src/js/views/patients/list/patients-all_views.js
@@ -1,9 +1,29 @@
+import _ from 'underscore';
+
+import Radio from 'backbone.radio';
 import { View, CollectionView } from 'marionette';
 
 import hbs from 'handlebars-inline-precompile';
 
 const ItemView = View.extend({
-  template: hbs`{{ name }}`,
+  template: hbs`
+    {{ first_name }} {{ last_name }}
+    {{#each groups}}
+      <strong>{{ this.name }}</strong>
+    {{/each}}
+  `,
+  className: 'patient-list__item',
+  templateContext() {
+    return {
+      groups: _.map(this.model.getGroups().models, 'attributes'),
+    };
+  },
+  triggers: {
+    click: 'click',
+  },
+  onClick() {
+    Radio.trigger('event-router', 'patient:dashboard', this.model.id);
+  },
 });
 
 const ListView = CollectionView.extend({
--- a/test/integration/patients/list/patients-all.js
+++ b/test/integration/patients/list/patients-all.js
@@ -1,0 +1,50 @@
+import _ from 'underscore';
+import 'js/utils/formatting';
+import { getIncluded } from 'helpers/json-api';
+
+context('patient sidebar', function() {
+  specify('display patient data', function() {
+    cy
+      .server()
+      .routePatient()
+      .routePatients((fx) => {
+        _.each(fx.data, (patient, index) => {
+          patient.id = index + 1;
+          patient.attributes = {
+            id: patient.id,
+            first_name: 'First',
+            last_name: 'Last',
+          };
+        });
+
+        fx.data[0].relationships.groups.data = _.collectionOf(['1', '2'], 'id');
+
+        fx.included = getIncluded(fx.included, [
+          {
+            id: '1',
+            name: 'Group One',
+          },
+          {
+            id: '2',
+            name: 'Another Group',
+          },
+        ], 'groups');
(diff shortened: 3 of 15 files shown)


test/support/api/patients.js
```diff
@@ -31,3 +31,33 @@ Cypress.Commands.add('routePatient', (mutator = _.identity) => {
...
     .as('routePatient');
 });
+
+Cypress.Commands.add('routePatients', (mutator = _.identity) => {
+  cy
+    .fixture('collections/patients').as('fxPatients')
+    .fixture('collections/groups').as('fxGroups');
+
+  cy.route({
+    url: '/api/patients**', // probably a regex to match for includes or just copy/past it?
```
luketlancaster: @paulfalgout what should this be?

paulfalgout: probably `/api/patients?*` otherwise you're going to match single patients too

test/support/api/patients.js
```diff
@@ -31,3 +31,33 @@ Cypress.Commands.add('routePatient', (mutator = _.identity) => {
...
+    response() {
+      const data = getResource(_.sample(this.fxPatients, 30), 'patients');
+      const groups = _.sample(this.fxGroups, 2);
```
paulfalgout: I think to make this a little more random we should sample 5 groups or so for the include and in the `_.each` do like `_.sample(groups, _.random(1, 3))`

test/integration/patients/list/patients-all.js
```diff
@@ -0,0 +1,56 @@
+import _ from 'underscore';
+
+context('patient sidebar', function() {
+  specify('display patient data', function() {
+    cy
+      .server()
+      .routePatient((fx) => {
+        fx.data = {
+          id: 1,
```
shadowhand: Patient IDs are UUIDs.

test/support/api/patients.js
```diff
@@ -31,3 +31,33 @@ Cypress.Commands.add('routePatient', (mutator = _.identity) => {
...
+
+      let included = [];
+
+      included = getIncluded(included, groups, 'groups');
+
+      _.each(data, (patient) => {
```
paulfalgout: only need to wrap the argument if there's more than one for `=>`s    Mind adding `arrow-parens: [2, "as-needed"]` to the eslint.. it _should_ auto fix this if we add that.

src/js/entities-service/patients.js
```diff
@@ -18,6 +19,16 @@ const Entity = BaseEntity.extend({
 
     return this.fetchModel(id, { data });
   },
+  fetchPatients() {
+    const include = [
+      'groups',
+      'patient-fields',
```
paulfalgout: we can remove this one

src/js/entities-service/patients.js
```diff
@@ -18,6 +19,16 @@ const Entity = BaseEntity.extend({
...
+    ].join(',');
+
+    const data = { include };
+
+    return this.fetchCollection({ url: '/api/patients', data });
```
paulfalgout: We shouldn't need to pass the url.  It should be defined on the collection

src/js/views/patients/list/patients-all_views.js
```diff
@@ -1,9 +1,29 @@
...
+      <strong>{{ this.name }}</strong>
+    {{/each}}
+  `,
+  className: 'patient-list__item',
+  templateContext() {
+    return {
+      groups: _.map(this.model.getGroups().models, 'attributes'),
+    };
+  },
+  events: {
```
paulfalgout: let's use `triggers` instead of `events` and trigger `click`

test/integration/patients/list/patients-all.js
```diff
@@ -0,0 +1,56 @@
+import _ from 'underscore';
+
+context('patient sidebar', function() {
+  specify('display patient data', function() {
+    cy
+      .server()
+      .routePatient((fx) => {
```
paulfalgout: you shouldn't need to set the data here.  It's not really accomplishing anything

test/integration/patients/list/patients-all.js
```diff
@@ -0,0 +1,56 @@
...
+          },
+        };
+
+        return fx;
+      })
+      .routePatients((fx) => {
+        _.each(fx.data, (patient, index) => {
+          patient.id = index + 1;
+          patient.attributes = {
+            id: patient.id,
```
paulfalgout: we don't set the id in the attributes

test/integration/patients/list/patients-all.js
```diff
@@ -0,0 +1,56 @@
...
+          attributes: {
+            first_name: 'First',
+            last_name: 'Last',
+          },
+        };
+
+        return fx;
+      })
+      .routePatients((fx) => {
+        _.each(fx.data, (patient, index) => {
```
paulfalgout: I think you can get away with either `unshift`ing a patient into the data to test, or just modify the first one.  The rest of the data can be random

src/js/entities-service/patients.js
luketlancaster: yeah, that's what I thought but it didn't seem to work without it. I get a `Uncaught Error: A "url" property or function must be specified`

paulfalgout: define the url on the patient entity collection.  we'll only pass the url when it isn't specific to the resource

test/integration/patients/list/patients-all.js
```diff
@@ -0,0 +1,43 @@
...
+        });
+
+        return fx;
+      })
+      .visit('/patients/all')
+      .wait('@routePatients');
+
+    cy
+      .get('.app-frame__content')
+      .should('contain', 'First Last')
```
paulfalgout: These `should`s should be able to move between 36 & 37.. that might require pushing a new group to includes and the making the relationship on `fx.data[0]`.  It'd be similar to https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/integration/patients/patient-sidebar.js#L98   Then we can just modify the group for the test row and leave the rest random

PR #26, opened 2019-06-27
Add uuids to all new model.save

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch20669]

Not current covered so codecov is gonna complain, but we'll catch this with add action shortly.


paulfalgout:
Screen Shot 2019-06-27 at 10 28 16 PM

PR #32, opened 2019-07-01
Require all access to be authenticated

shadowhand:
- [x] Add the Clubhouse Story ID: [ch20684]

Possibly:

- [ ] store the [`expiresIn`][1] value after authentication
- [ ] use [`checkSession`][2] to get new tokens when necessary

[1]: https://auth0.com/docs/libraries/auth0js/v9#extract-the-authresult-and-get-user-info
[2]: https://auth0.com/docs/libraries/auth0js/v9#using-checksession-to-acquire-new-tokens

### DIFF
--- a/src/js/services/auth.js
+++ b/src/js/services/auth.js
@@ -1,11 +1,23 @@
 import $ from 'jquery';
+import _ from 'underscore';
+import moment from 'moment';
 import auth0 from 'auth0-js';
 import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
-let token = sessionStorage.getItem('auth:token');
-let expires = sessionStorage.getItem('auth:expires');
+let token;
+let expires;
+
+function clearHash() {
+  window.history.replaceState('', document.title, window.location.pathname + window.location.search);
+}
+
+function getSessionExpires() {
+  const sessionExpires = sessionStorage.getItem('auth:expires');
+
+  return sessionExpires && moment.unix(sessionExpires).utc();
+}
 
 export default App.extend({
   channelName: 'auth',
@@ -13,8 +25,9 @@ export default App.extend({
     'ajaxAuth': 'getAjaxAuth',
     'currentUser': 'getCurrentUser',
     'currentOrg': 'getCurrentOrg',
-    'expires': 'getTokenExpiration',
-    'logout': 'removeToken',
+    'isTokenValid': 'isTokenValid',
+    'login': 'login',
+    'logout': 'logout',
     'bootstrap': 'initBootstrap',
   },
   getAjaxAuth() {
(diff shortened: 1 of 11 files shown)


shadowhand:
@paulfalgout not sure what else you want me to do with this. Obviously something needs to change with testing but I'm not sure where to start. Willing to tackle it if you point me in the right direction...

paulfalgout:
I'm not certain the way to get around auth0 for cypress testing at this point.  Either way we can probably sit on this for a little bit, but it's a great model.  I didn't look into it much but what was the difference between this and the auth0-lock lib?  They seemed at first glance to be essentially the same, but, only the lock had visuals for the user experience.. what happens here?  I'm assuming it's a redirect to some authentication URL and then back to roundingwell.com?

shadowhand:
@paulfalgout regarding `auth0-lock`, this warning is at the top of [the docs][1]:

> Embedded login for web uses Cross Origin Authentication. In some browsers this can be unreliable if you do not set up a Custom Domain and host your app on the same domain. Using Custom Domains with Auth0 is a paid feature. If you cannot use Custom Domains, consider migrating to Universal Login.

This restriction for Cross Origin and requiring a custom domain (which is not free) made me avoid that solution for now.

You have a good point about the refresh token... I should address that.

[1]: https://auth0.com/docs/libraries/lock/v11

shadowhand:
@paulfalgout token expiration is going to be a more tricky than relying on auth0 refresh, because we ask for an `idToken`, which has expires based on the social provider settings. To get around all this, we probably need to change the auth setup to:

1. fetch an auth0 `accessToken` using [`audience`][1], then store the token and the `expiresIn` value
2. check the `expiresIn` value periodically and request a new `accessToken` before it expires

[1]: https://auth0.com/docs/api-auth/tutorials/verify-access-token

shadowhand:
@paulfalgout after working through various attempts to do sessions and refresh, there was no easier solution that what I just pushed. The ID token is what we need, and as soon as we have to deal with sessions we lose the ability to pass the user email directly to the API.

The good news is, once the user has authenticated once with Auth0, the token exchange happens without user interaction. So long as long as we check the `Radio.request('auth:expires')` periodically and warn the user when their session is about to expire, then trigger a `auth:logout` to force a new token to be loaded, the flow should be really smooth.

Regarding cypress testing, it seems like it should be possible to use cypress to:

```js
const today = new Date();
sessionStorage.setItem('auth:token', process.env.API_TOKEN)
sessionStorage.setItem('auth:expires', new Date().setDate(today.getDate() + 1).toString());
```

As a `beforeEach()` or something like that.

src/js/services/auth.js
```diff
@@ -33,15 +30,62 @@ export default App.extend({
...
 
-    $.when(Radio.request('entities', 'fetch:clinicians:current')).then(currentUser => {
-      this.currentUser = currentUser;
-      this.currentOrg = this.currentUser.getOrganization();
-      d.resolve(currentUser);
+    // https://auth0.com/docs/libraries/auth0js/v9#webauth-authorize-
+    const auth = new auth0.WebAuth({
+      clientID: '2eg7mz2db0B31gctULMTcgax1vdvgPip',
+      domain: 'roundingwell-care-team.auth0.com',
+      responseType: 'id_token',
```
paulfalgout: So from the meeting I gathered that the `token` is what is needed for the refresh?..  I'm not sure what `responseMode` `fragment` does with this, but `responseType` can be space separated: `'id_token token'` to apparently somehow return both?  Either way, it doesn't seem like we need to worry about this for MVP, but it might be possible to get both?

shadowhand: @paulfalgout we can get both, but access tokens (`token`) is the only thing that can be refreshed and it does not contain any user info. The ID token is generated once per login and cannot be refreshed.

paulfalgout: Seems to be an issue a lot of people are dealing with... most questions go unanswered, and the few that do... not great responses: https://community.auth0.com/t/anyone-refreshing-tokens-id-token-missing-in-refresh-token-response/7067/2 https://community.auth0.com/t/oauth-token-not-returning-id-token/8303

For another day I guess.

shadowhand: Oh I should also mention: ID tokens last for 24 hours, so most likely users would only need to login once a day.

paulfalgout: makes me think maybe we should set the expire to 10 hours or something so that someone logging in at the end of the day is more likely to have to relogin each day instead of having a chance to hit that 24 hour mark the next day

shadowhand: Well, I'm using `sessionStorage` so that if the user closes the tab or browser they will need to login again. I think what we have is fine until proven otherwise.

paulfalgout:
@shadowhand @luketlancaster ok I think this may be ready to go and working with at least patient lists now.  Cypress is happy as well.  I made a few modifications to the auth code.. most notably I moved it to pre `app.start` and separated it from the bootstrap.  I ran into a few smaller issues that I also addressed such as hitting a url like `/patients/all` was causing a redirect error and if there was an issue with the `hash` for some reason, it never cleared, so a refresh just repeated the error.  I also switched to using `moment`, for one to have a single date interface, and two because it needed to be converted to `utc` to do proper calculations and now we can do more complicated tests if we want to do something like "Your session will expire in 10 minutes" or something.

There's still somethings that need to be handled gracefully like if we get a 401/403 from the API, how do we handle it..  currently when you logout, you immediately get the login screen... eventually we'll probably want some roundingwell screen that prompts for login that we can logout too..  but for another epic.

shadowhand:
@paulfalgout I can't approve this PR because I made it.

PR #35, opened 2019-07-03
Add groups filter menu to all patients and actions views

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20685]


src/js/views/patients/view/view_views.js
```diff
@@ -106,7 +108,15 @@ const ListView = CollectionView.extend({
   },
 });
 
+const GroupsDropList = Droplist.extend({
+  template: hbs`{{fas "filter"}} {{ name }}{{#unless name}}All Groups{{/unless}} {{far "angle-down"}}`,
```
paulfalgout: I don't understand the `#unless` here.  When is there not a selected?  it looks to me that `_getSelectedGroup` should always return a value, and if it doesn't it's malformed data and not "All Groups"

luketlancaster: yeah, looks like I don't need it? I think the last iteration of the object passed in here fixed it (maybe removing the template from the picklistOptions). Should be good to go then

test/integration/patients/list/patients-all.js
```diff
@@ -34,6 +53,24 @@ context('patient sidebar', function() {
...
+      .find('.table-list__item')
+      .first()
+      .should('contain', 'First Last');
+
+    cy
+      .get('.button--blue')
+      .should('contain', 'All Groups')
+      .click();
+
+    cy
```
paulfalgout: actually this part of the test doesn't mean anything and should be removed.  The stub can return anything it wants and doesn't care what the filter is set to.  The way to test this will be by seeing if the endpoint is hit with the correct filter data.  So rather than setup the test to look like we're testing the group data base on the filter selection, let's avoid doing that and just test the data from the stub.. then later test the filter data... with nothing list related once the filter is selected.

PR #37, opened 2019-07-09
Introduce action sidebar

paulfalgout:
Adds action sidebar without styling, attributes or editing

- [x] Add the Clubhouse Story ID: [ch20679]


paulfalgout:
It's known that coverage will suffer here.  We'll have to test it once save is fully a thing

PR #38, opened 2019-07-11
Templates and views for various Action Event/Activity types

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20710]

### DIFF
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -60,6 +60,17 @@ careOptsFrontend:
               other {}
             }
           sexLabel: Sex
+    sidebar:
+      actionSidebar:
+        activityViews:
+          added: "<strong>{ first_name } { last_name }</strong> ({ role }) added this Action"
+          clinicianAssigned: '<strong>{ first_name } { last_name }</strong> ({ role }) changed the Owner to { to_first_name } { to_last_name }'
+          detailsUpdated: '<strong>{ first_name } { last_name }</strong> ({ role }) changed the details of this Action'
+          dueDateUpdated: '<strong>{ first_name } { last_name }</strong> ({ role }) changed the Due Date to { date }'
+          durationUpdated: '<strong>{ first_name } { last_name }</strong> ({ role }) changed Duration to { duration }'
+          nameUpdated: '<strong>{ first_name } { last_name }</strong> ({ role }) changed the name of this Action from { from_name } to { to_name }'
+          roleAssigned: '<strong>{ first_name } { last_name }</strong> ({ role }) changed the Owner to { to_role }'
+          stateUpdated: '<strong>{ first_name } { last_name }</strong> ({ role }) changed State to { to_state }'
     view:
       viewViews:
         layoutView:
--- a/src/js/views/patients/sidebar/action-sidebar_views.js
+++ b/src/js/views/patients/sidebar/action-sidebar_views.js
@@ -1,7 +1,7 @@
 import _ from 'underscore';
 import Backbone from 'backbone';
 import hbs from 'handlebars-inline-precompile';
-import { View, CollectionView } from 'marionette';
+import { View } from 'marionette';
 
 import Optionlist from 'js/components/optionlist';
 
@@ -18,19 +18,6 @@ const SaveView = View.extend({
   triggers: {
     'click .js-cancel': 'cancel',
     'click .js-save': 'save',
-  },
-});
-
-const ActivityView = View.extend({
-  template: hbs`{{ type }}`,
-});
-
-const ActivitiesView = CollectionView.extend({
-  childView(model) {
-    const type = model.get('type');
-    switch (type) {
-      default: return ActivityView;
-    }
   },
 });
 
@@ -98,5 +85,4 @@ const LayoutView = View.extend({
 
 export {
   LayoutView,
-  ActivitiesView,
 };
--- a/test/support/api/events.js
+++ b/test/support/api/events.js
@@ -3,7 +3,7 @@ import { getResource } from 'helpers/json-api';
 
 Cypress.Commands.add('routeActionActivity', (mutator = _.identity) => {
   cy
-    .fixture('collections/events').as('fxEvents');
+    .fixture('test/events').as('fxEvents');
 
   cy.route({
     url: '/api/actions/**/activity*',
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -3,7 +3,8 @@ import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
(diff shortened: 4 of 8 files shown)


src/js/views/patients/sidebar/action-sidebar_activity-views.js
```diff
@@ -0,0 +1,105 @@
...
+      ActionStateUpdated: StateUpdatedView,
+    };
+
+    return ChildViews[model.get('type')];
+  },
+});
+
+export {
+  ActivitiesView,
+};
```
luketlancaster: I'm not really jazzed about any of this tbh...maybe it's the best that can be done/doesn't matter all that much...but suggestions are welcome

src/js/views/patients/sidebar/action-sidebar_views.js
```diff
@@ -1,10 +1,12 @@
 import _ from 'underscore';
 import Backbone from 'backbone';
 import hbs from 'handlebars-inline-precompile';
-import { View, CollectionView } from 'marionette';
+import { View } from 'marionette';
 
 import Optionlist from 'js/components/optionlist';
 
+import { ActivitiesView } from './action-sidebar_activity-views';
```
luketlancaster: is this right? Should this be imported here only to be exported? Feels not right.

src/js/views/patients/sidebar/action-sidebar_activity-views.js
```diff
@@ -0,0 +1,105 @@
...
+    {{#with metadata.editor}}
+    {{formatMessage (intlGet "patients.sidebar.actionSidebar.activityViews.dueDateUpdated") first_name = first_name last_name = last_name role = role.name}}
+    {{/with}} {{formatMoment metadata.value.to "LONG"}}
+    <div>{{formatMoment date "LONG"}}</div>
+  `,
+});
+
+const DurationUpdatedView = View.extend({
+  template: hbs`
+    {{formatMessage (intlGet "patients.sidebar.actionSidebar.activityViews.durationUpdated") first_name = first_name last_name = last_name role = role duration = duration}}
```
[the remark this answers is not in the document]
paulfalgout: Alternatively we could add this to `js/base/helpers`
```js
Handlebars.registerHelper({
  formatDuration(dur, key, { hash = {} }) {
    if (!dur) return new Handlebars.SafeString(hash.defaultHtml || '');

    return duration(dur, hash.inputKey).get(key);
  },
});
```
Then it'd be `duration = (formatDuration value.to "minutes")`
Which is kind of consistent with `formatMoment`

Then at that point every childview is just a template..

At which point I think this file could change to 
```js
const ActionAddedTemplate = hbs``;

const ActivityView = View.extend({
  getTemplate() {
    const Templates = [
      ActionAdded: ActionAddedTemplate,
      ...
     ];
     return Templates[this.model.get('type')];
  },
});

const ActivitiesView = CollectionView.extend({
  childView: ActivityView,
});
```
_or_ these two views could move back into the other views file and just add
`import ActionAddedTemplate from './action-added.hbs';` etc at the top

test/integration/patients/sidebar.js
```diff
@@ -0,0 +1,42 @@
+import _ from 'underscore';
```
paulfalgout: let's call this file `action-sidebar` maybe even nest it in a `sidebar/` directory.. we're likely to end up with a lot of these.

test/integration/patients/sidebar.js
```diff
@@ -0,0 +1,42 @@
...
+import moment from 'moment';
+
+context('patient sidebar', function() {
```
paulfalgout: action sidebar

test/integration/patients/sidebar.js
```diff
@@ -0,0 +1,42 @@
...
+  specify('display patient data', function() {
+    const dob = moment().subtract(10, 'years');
+
+    cy
+      .server()
+      .routePatientActions(_.identity, '1')
+      .routeActionActivity()
+      .routePatient(fx => {
```
paulfalgout: I don't see any reason to change the patient for these tests

test/integration/patients/sidebar.js
```diff
@@ -0,0 +1,42 @@
...
+          first_name: 'First',
+          last_name: 'Last',
+          birth_date: dob.format('YYYY-MM-DD'),
+          mrn: 'a12345',
+          sex: 'f',
+        };
+
+        return fx;
+      })
+      .visit('/patient/dashboard/1')
```
paulfalgout: rather than hitting the table item we can just visit `/patient/1/action/1`

test/support/api/events.js
```diff
@@ -3,13 +3,13 @@ import { getResource } from 'helpers/json-api';
...
   cy
-    .fixture('collections/events').as('fxEvents');
+    .fixture('test/events').as('fxEvents');
 
   cy.route({
     url: '/api/actions/**/activity*',
     response() {
       return mutator({
-        data: getResource(_.sample(this.fxEvents, 5), 'events'),
+        data: getResource(this.fxEvents, 'events'),
```
paulfalgout: I think I'd prefer this stay sampled for the generic use, so that every other action test isn't overly long and then when we want to test them all we can do
```js
.routeActionActivity(fx => {
  fx.data = getResource(this.fxEvents, 'events');
  return fx;
})
```

src/js/views/patients/sidebar/action-sidebar_activity-views.js
```diff
@@ -0,0 +1,105 @@
...
+    {{formatMessage (intlGet "patients.sidebar.actionSidebar.activityViews.stateUpdated") first_name = editor.first_name last_name = editor.last_name role = editor.role.name to_state = value.to.name}}
+    {{/with}}
+    <div>{{formatMoment date "LONG"}}</div>
+  `,
+});
+
+const ActivitiesView = CollectionView.extend({
+  childView(model) {
+    const ChildViews = {
+      ActionAdded: AddedView,
```
paulfalgout: I believe this is changing to `ActionCreated` https://app.clubhouse.io/roundingwell/story/20723/modify-action-search-filters#activity-20725

src/js/i18n/en-US.yml
```diff
@@ -60,6 +60,17 @@ careOptsFrontend:
               other {}
             }
           sexLabel: Sex
+    sidebar:
+      actionSidebar:
+        activityViews:
+          addedHTML: "<strong>{ first_name } { last_name }</strong> ({ role }) added this Action"
```
paulfalgout: I know it was a convention before to add `HTML` to i18n keys, but I no longer think it's useful.  It provides protection to use the html suffix in `templateContext` so that we're clear we're escaping user data, but in the case of i18n it's only in-house html.. the data still gets escaped.  (I think we didn't know for sure when we were first doing the i18n and making stylistic decisions)

PR #42, opened 2019-07-17
Adds hardcoded MVP action views (lists)

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20732]

### DIFF
--- a/src/js/apps/patients/view/view_app.js
+++ b/src/js/apps/patients/view/view_app.js
@@ -1,3 +1,6 @@
+import _ from 'underscore';
+import moment from 'moment';
+
 import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
@@ -10,8 +13,8 @@ export default App.extend({
   },
   onBeforeStart({ viewId }) {
     if (this.isRestarting()) return;
-    const currentClinician = Radio.request('auth', 'currentUser');
-    this.groups = currentClinician.getGroups();
+    this.currentClinician = Radio.request('auth', 'currentUser');
+    this.groups = this.currentClinician.getGroups();
 
     this.showView(new LayoutView({ viewId }));
     this.showFilterView();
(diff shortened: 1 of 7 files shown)


src/js/apps/patients/view/view_app.js
```diff
@@ -54,4 +59,20 @@ export default App.extend({
...
+  _filtersById(viewId, currentClinician) {
+    const clinician = currentClinician.id;
+
+    const filters = {
+      'owned-by-me': { clinician },
+      'actions-for-coordinators': { role: 'coordinator' },
+      'new-actions': { created: moment().subtract(1, 'day').format() },
+      'updated-past-three-days': { updated: moment().subtract(3, 'days').format() },
+      'done-last-thirty-days': {
+        updated: moment().subtract(30, 'days').format(),
```
luketlancaster: wasn't sure one the time situation here...does this need to be `.utc()`'d?

paulfalgout:
@BenCoffeed The preview environment here doesn't appear to contain the build from this PR

src/js/apps/patients/view/view_app.js
```diff
@@ -10,8 +13,8 @@ export default App.extend({
   },
   onBeforeStart({ viewId }) {
     if (this.isRestarting()) return;
-    const currentClinician = Radio.request('auth', 'currentUser');
-    this.groups = currentClinician.getGroups();
+    this.currentClinician = Radio.request('auth', 'currentUser');
```
paulfalgout: it doesn't look like you need to attach `currentClinician`

src/js/apps/patients/view/view_app.js
```diff
@@ -54,4 +59,20 @@ export default App.extend({
 
     return groups;
   },
+  _filtersById(viewId, currentClinician) {
+    const clinician = currentClinician.id;
+
+    const filters = {
+      'owned-by-me': { clinician },
+      'actions-for-coordinators': { role: 'coordinator' },
+      'new-actions': { created: moment().subtract(1, 'day').format() },
```
paulfalgout: I think we should specifically do `subtract(24, 'hours')` here.
Then in the other spots we want `moment().startOf('day').subtract(n, 'days')`

src/js/apps/patients/view/view_app.js
paulfalgout: nope.  The time is relative to the user so as long as it has a timestamp (which it will here), we're good.

src/js/apps/patients/view/view_app.js
luketlancaster: so should I just do another Radio.request to get the `currentClinician` down in `beforeStart`?

paulfalgout: ah ha.  I see.  github collapsed 18-22 for me.

PR #45, opened 2019-07-18
Uses role id for CO role to filter actions

luketlancaster:

### DIFF
--- a/src/js/apps/patients/view/view_app.js
+++ b/src/js/apps/patients/view/view_app.js
@@ -61,10 +61,12 @@ export default App.extend({
   },
   _filtersById(viewId, currentClinician) {
     const clinician = currentClinician.id;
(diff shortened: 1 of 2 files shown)


src/js/apps/patients/view/view_app.js
```diff
@@ -61,10 +61,12 @@ export default App.extend({
   },
   _filtersById(viewId, currentClinician) {
     const clinician = currentClinician.id;
+    const roles = Radio.request('entities', 'roles:collection');
+    const role = roles.find({ short_name: 'CO' });
```
paulfalgout: did this work? seems like you still need `.id` on this line

luketlancaster: whoopsies. I thought it worked because it didn't blow up...but that's just because it wasn't passing the role filter at all...

PR #48, opened 2019-07-22
Add state/owner/due components to actions

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch20742]

This is essentially phase 1 of this work, so I'll break up the card into two.
What's left is duration and handling when the sidebar can be saved along with empty state text for action creation.

There's also a ton of styling and some i18n work todo as well as tests that'll need to come on the end of this.
I'm still pretty unsure as to how these will _actually_ work in the end as widths on the lists of the components with variables names seems pretty impossible, but for now it's doneish.

Also so far `PATCH` looks correct locally and the endpoint responds with a 204, but nothing ever updates, so hopefully running this with a public URL will make a difference or at least shine light on the issue.

### DIFF
--- a/src/js/apps/patients/patient/dashboard/dashboard_app.js
+++ b/src/js/apps/patients/patient/dashboard/dashboard_app.js
@@ -1,4 +1,3 @@
-import _ from 'underscore';
 import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
@@ -10,6 +9,7 @@ export default App.extend({
     'click:add': 'click:add',
   },
   onBeforeStart({ patient }) {
+    this.patient = patient;
     this.showView(new LayoutView({ model: patient }));
     this.getRegion('content').startPreloader();
   },
@@ -17,7 +17,6 @@ export default App.extend({
     return Radio.request('entities', 'fetch:patientActions:collection', { patient });
   },
   onStart({ patient }, actions) {
-    this.patient = patient;
     this.actions = actions;
 
     this.showChildView('content', new ListView({ collection: actions }));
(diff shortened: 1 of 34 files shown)


shadowhand:
> Also so far PATCH looks correct locally and the endpoint responds with a 204, but nothing ever updates, so hopefully running this with a public URL will make a difference or at least shine light on the issue.

I'm still confused about this. I assume that changes to models are stored locally, then the diff is sent as a `PATCH` request? If that is the case, then the `204` semantics are correct.

Or are you saying that the `PATCH` is sent with the diff, but upon reloading the same model, the data does not match the local (browser) copy?

paulfalgout:
@shadowhand so if you goto https://fe-view-action-update.preview.roundingwell.com and pick the view that brings up the two available actions (right now that'll be the "< 3 days" one) you can pull up the network tab and see the requests made if you update the due date or state from the list.  To me the data looks correct and I get a 204 back, but upon hard refresh, it goes back to the fixture data

shadowhand:
@paulfalgout that's not what happens for me. I get this error:

Image 2019-07-22 at 9 57 12 AM

paulfalgout:
Not sure how you're getting it to `PUT` at all:

Screen Shot 2019-07-23 at 12 00 01 AM

shadowhand:
@paulfalgout click on action, use sidebar to change owner/state, click "save".

paulfalgout:
Yeah the sidebar doesn't work yet and isn't ready.  Goto the a list.. change the state or due date on the list.. don't click on the action

shadowhand:
@paulfalgout however, if I do it from the list UI, I see that no data is submitted with the PATCH:

Image 2019-07-22 at 10 03 44 AM

paulfalgout:
There's data submitted in your screenshot

shadowhand:
Ah yes, the clinician change is not `attributes` but `relationships`. My mistake, it looks like everything is going through correctly... trying to figure out how to check the database.

paulfalgout:
Once the current deploy goes out setting ajax to `json` `contentType` makes devtools a much nicer experience, but at least locally it's functioning the same

shadowhand:
And yet our BE tests prove that it works just fine. I feel like this has to be a problem with the `/api` proxy routing swallowing errors or something. Direct HTTP requests have no issue. 😕

paulfalgout:
ping @BenCoffeed

shadowhand:
@paulfalgout this should be fixed once the build completes:

https://circleci.com/gh/RoundingWell/care-ops-backend/1565

PR #52, opened 2019-08-01
Show which nav item is currently selected

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20766]

### DIFF
--- a/src/js/apps/globals/app-frame_app.js
+++ b/src/js/apps/globals/app-frame_app.js
@@ -1,3 +1,5 @@
+import _ from 'underscore';
+
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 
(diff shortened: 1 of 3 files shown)


src/js/apps/globals/app-frame_app.js
```diff
@@ -9,10 +11,56 @@ import PatientsMainApp from 'js/apps/patients/patients-main_app';
...
   radioRequests: {
-    'select': 'select',
+    'select': 'onSelect',
+  },
+  onSelect(appName, event, eventArgs) {
+    const navMatch = viewsNav.find(model => this._navMatch(model, event, eventArgs))
+      || patientsNav.find(model => this._navMatch(model, event, eventArgs));
+
+    if (navMatch) {
+      this.getChildView('nav').$('.is-selected').removeClass('is-selected');
```
paulfalgout: as much as it seems unnecessary I'd like to move this to a function on the nav view so that  DOM querying isn't in the apps

src/js/apps/globals/app-frame_app.js
```diff
@@ -9,10 +11,56 @@ import PatientsMainApp from 'js/apps/patients/patients-main_app';
...
+]);
+
 export default App.extend({
   channelName: 'nav',
   radioRequests: {
-    'select': 'select',
+    'select': 'onSelect',
+  },
+  onSelect(appName, event, eventArgs) {
+    const navMatch = viewsNav.find(model => this._navMatch(model, event, eventArgs))
```
paulfalgout: I think this might be cleaner if it was `this._navMatch(viewsNav, event, eventArgs) || ...`  then `_navMatch` can just be a `find(model => { return...`  makes `_navMatch` a little bit more complex, but this line is simpler.

src/js/apps/globals/app-frame_app.js
paulfalgout: Ah there is a `removeSelected()` just not being used

PR #53, opened 2019-08-02
Preloaders/tooltips/empty views/sidebar animation

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20748]

### DIFF
--- a/src/js/views/patients/patient/dashboard/dashboard_views.js
+++ b/src/js/views/patients/patient/dashboard/dashboard_views.js
@@ -1,4 +1,5 @@
 import Radio from 'backbone.radio';
+import hbs from 'handlebars-inline-precompile';
 import { View, CollectionView, Region } from 'marionette';
 
 import 'sass/modules/buttons.scss';
@@ -12,6 +13,15 @@ import LayoutTemplate from './layout.hbs';
 
 import '../patient.scss';
 import './dashboard.scss';
+
+const EmptyView = View.extend({
+  tagName: 'tr',
+  template: hbs`
+    <td class="patient-empty-list">
+      <h2>{{ @intl.patients.patient.dataEvents.dataEventsViews.emptyView }}</h2>
+    </td>
+  `,
+});
 
 const ItemView = View.extend({
   modelEvents: {
@@ -81,6 +91,7 @@ const ListView = CollectionView.extend({
   className: 'table-list',
   tagName: 'table',
   childView: ItemView,
+  emptyView: EmptyView,
   viewFilter({ model }) {
     return !model.isDone();
   },
--- a/src/js/views/patients/patient/patient.scss
+++ b/src/js/views/patients/patient/patient.scss
@@ -97,3 +97,13 @@
   border-bottom: solid 1px $black-90;
   padding: 16px;
 }
+
+.patient-empty-list {
+  border-bottom: 1px solid $black-90;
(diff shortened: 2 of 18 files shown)


src/js/views/patients/patient/dashboard/dashboard_views.js
```diff
@@ -13,6 +14,15 @@ import LayoutTemplate from './layout.hbs';
...
 
+const EmptyView = View.extend({
+  tagName: 'tr',
+  template: hbs`
+    <td class="patient-empty-list__body">
+      <h2 class="patient-empty-list__sub-heading">{{ @intl.patients.patient.dataEvents.dataEventsViews.emptyView }}</h2>
+    </td>
+  `,
+});
+
```
luketlancaster: this one doesn't actually show up because the 'new action' template is always there (and I think should only be there after clicking 'add action' or whatever and should go away if you close the sidebar w/o adding an action?)

paulfalgout: I don't follow.  The new action template only shows up for me after pushing add action, and does go away when I close the new action sidebar (if it's open)

src/sass/modules/table-list.scss
```diff
@@ -35,3 +35,17 @@
   font-size: 12px;
   padding: $pad-small;
 }
+
+.table-empty-list__body {
```
paulfalgout: I think this can probably be done with a single class

src/js/views/patients/patient/dashboard/dashboard_views.js
luketlancaster: 🤔 maybe it was just the testing env I had setup that was wonky

src/sass/modules/table-list.scss
luketlancaster: were you thinking that I should move the class up out of `table-list` and the patient one as well, and just have some 'generic' `.empty-list` class both can use?

paulfalgout: I think where they're defined is fine.  I just think it can be done with one class rather than two

src/js/views/patients/patient/patient.scss
```diff
@@ -97,3 +97,14 @@
   border-bottom: solid 1px $black-90;
   padding: 16px;
 }
+
+.patient-empty-list {
+  border-bottom: 1px solid $black-90;
+  color: $black-80;
+  font-size: 16px;
+  font-weight: 600;
+  margin: 40px auto;
```
paulfalgout: This can be written `margin: 40px auto 20px;`

src/js/views/patients/sidebar/action/action-sidebar_views.js
```diff
@@ -128,6 +134,14 @@ const LayoutView = View.extend({
   initialize({ action }) {
     this.action = action;
   },
+  onAttach() {
+    anime({
+      targets: this.el,
+      translateX: [{ value: 20, duration: 0 }, { value: 0, duration: 200 }],
+      opacity: [{ value: 0, duration: 0 }, { value: 1, duration: 300 }],
+      easing: 'easeInOutQuad()',
```
paulfalgout: in other places we're not using the `()` so I think it's safe to remove it.. I _think_ that might be deprecated if I'm remembering correctly

PR #54, opened 2019-08-06
Update styles and markup to MVP

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch20777]


paulfalgout:
There's a new issue on add new where it's just adding a bunch of them :-/

PR #63, opened 2019-08-13
Redirect to initially requested url after auth success

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20804]

this what you're thinking?


paulfalgout:
:tada:

PR #65, opened 2019-08-15
Remove organization

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch20812]


src/js/entities-service/entities/organizations.js
```diff
@@ -1,31 +1,24 @@
...
-  urlRoot: '/api/organizations',
-  getGroups() {
-    return Radio.request('entities', 'groups:collection', this.get('_groups'));
+  getStates() {
+    return this.get('states').clone();
   },
   getRoles() {
-    return Radio.request('entities', 'roles:collection', this.get('_roles'));
+    return this.get('roles').clone();
   },
```
luketlancaster: do we still want to access the states/roles through the organization? That's what's happening here right?

src/js/services/auth.js
```diff
@@ -108,17 +108,18 @@ export default App.extend({
...
-        return Radio.request('entities', 'fetch:clinicians:byGroup', group);
-      });
-
-      $.when(...groupRequests).done(() => {
-        d.resolve(currentUser);
+      this.currentOrg = Radio.request('entities', 'organizations:model', {
+        states: states[0],
+        roles: roles[0],
       });
+      console.log(this.currentOrg);
```
luketlancaster: forgotten log

shadowhand: Debug left behind.

src/js/services/auth.js
```diff
@@ -108,17 +108,18 @@ export default App.extend({
   },
   initBootstrap() {
     const d = $.Deferred();
-    $.when(Radio.request('entities', 'fetch:clinicians:current')).then(currentUser => {
+    const fetchCurrentUser = Radio.request('entities', 'fetch:clinicians:current');
+    const fetchGroups = Radio.request('entities', 'fetch:groups:collection');
+    const fetchRoles = Radio.request('entities', 'fetch:roles:collection');
+    const fetchStates = Radio.request('entities', 'fetch:states:collection');
+    $.when(fetchCurrentUser, fetchRoles, fetchStates, fetchGroups).done((currentUser, roles, states) => {
```
shadowhand: 👀 does this mean we can get rid of compound documents?!

src/js/services/auth.js
```diff
@@ -108,17 +108,18 @@ export default App.extend({
...
-
-      const groupRequests = this.currentOrg.getGroups().map(group => {
-        return Radio.request('entities', 'fetch:clinicians:byGroup', group);
-      });
-
-      $.when(...groupRequests).done(() => {
-        d.resolve(currentUser);
+      this.currentOrg = Radio.request('entities', 'organizations:model', {
+        states: states[0],
+        roles: roles[0],
```
luketlancaster: can you explain this for me? This is getting the cached organization (which is the default, for now) based on the roles/states that come back from the api?

src/js/services/auth.js
paulfalgout: requests that can be made in parallel are far different than requests the must be made serially.   This is a trivial request.  Making an additional and separate patient request for each action that is returned on an action list is quite a bit more complicated

src/js/services/auth.js
paulfalgout: hmm.. CI should fail and didn't :-/

src/js/services/auth.js
paulfalgout: This is just creating an organization from nowhere. it's not cached, but it has a default value set for the name.  Then I'm attaching states and roles mainly because it's convenient, requires the fewest changes, and largely still makes sense from the FE architecture's perspective.  We could attach states and roles here like we're attaching `currentUser` and `currentOrg` and add 2 additional auth requests.  I was on the fence.  The roles, states, and groups can still be considered the org ones..  though at the moment we only need to request groups related to the currentUser (though we need all groups cached for various display)  thoughts?

luketlancaster: 🤔 it makes sense as these three models are owned by the org (and, in the case of roles/states will eventually be customized by them). I think the difficulty lies in just changing the way I think about "Orgs" in the context of the application, and less in how it's architected if that makes sense. 

On the other hand, the frontend application will only ever get back roles/states for the 'current' org because of sso/auth magic right? So does accessing them through the org make sense there? Or should they be their own, standalone, thing? I dunno, I'm fine either way.

paulfalgout: the roles, states and groups will always be the same for the subdomain.  The organization is essentially non-existent, but for the frontend it's a name we throw on a model to display on the top left menu.. so since it's an entity for FE.. it is effectively equivalent to the subdomain which owns the other entities.  organization for FE is to a certain extent "global"

PR #68, opened 2019-08-19
Switch to 3 states

luketlancaster:
* queued - To Do
* started - In Progress
* done - Done

- [x] Add the Clubhouse Story ID: [ch20823]

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
   ],
   "fontawesome": {
     "fas": [
-      "faAdjust",
       "faCaretLeft",
       "faCaretRight",
       "faCheckCircle",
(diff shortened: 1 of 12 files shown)


src/js/views/patients/actions/actions_views.js
```diff
@@ -19,9 +19,8 @@ import 'sass/domain/action-state.scss';
 import './actions.scss';
 
 const StatusIcons = {
-  needs_attention: 'exclamation-circle',
-  open: 'dot-circle',
-  pending: 'adjust',
```
paulfalgout: this font icon needs to be removed from package.json

package.json
```diff
@@ -33,7 +33,6 @@
       "faChevronLeft",
       "faChevronRight",
       "faDotCircle",
-      "faExclamationCircle",
```
paulfalgout: wrong one.  This is "To Do" / "Queued"

luketlancaster: yeah just saw that :/

PR #69, opened 2019-08-22
Replace selectlist with droplist and input

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch20858]


paulfalgout:
The coverage change here is related to the `contains` method for detecting if a click is within a component.  selectlist needed a state we don't currently need, but that line is useful for future components, but probably shouldn't be ignored either.

PR #74, opened 2019-09-05
Move auth to a separate chunk

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch20925]
This also improves the auth experience.


.eslintrc
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "eslint:recommended",
   "parserOptions": {
-    "ecmaVersion": 6,
+    "ecmaVersion": 2020,
```
luketlancaster: Nothing like bumping it up 2014 versions :trollface:

PR #80, opened 2019-09-11
Update VersionSuffix to match backend practice

BenCoffeed:
- [x] Add the Clubhouse Story ID: [ch20979]
- [ ] Add QA tasks


BenCoffeed:
I added in some logging because it's adding an extra `-` at the end of the VersionSuffix

BenCoffeed:
Updated. There was something weird going on with the `-c` or complementary flag that was adding a trailing `-`. Switching to a just using `[:punct:][:blank:]` to strip out any non alphanumeric characters seems to have resolved.

PR #81, opened 2019-09-16
Reshow component after model is updated

paulfalgout:
We were reshowing on the change event, but the handler on the component would happen prior to whatever else was listening updated the model.

- [x] Add the Clubhouse Story ID: [ch20996]
- [x] Add QA tasks


sofiaroundingwell:
Well, nevermind on the merge. I Do not have permission to merge. So... someone merge? please? thanks

PR #82, opened 2019-09-16
Fixes action activity to use new data format

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch20959]

felt like there was a lot of "hand wavy" and "this'll do" going on while working on this...so...you know...

### DIFF
--- a/src/js/entities-service/entities/events.js
+++ b/src/js/entities-service/entities/events.js
@@ -1,3 +1,4 @@
+import Radio from 'backbone.radio';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
@@ -6,6 +7,26 @@ const TYPE = 'events';
 
 const _Model = BaseModel.extend({
   type: TYPE,
+
+  getClinician() {
+    return Radio.request('entities', 'clinicians:model', this.get('_clinician'));
+  },
+  getEditor() {
+    if (!this.get('_editor')) {
+      return Radio.request('entities', 'clinicians:model', {
+        first_name: 'RoundingWell',
+        last_name: '',
+      });
+    }
+
+    return Radio.request('entities', 'clinicians:model', this.get('_editor'));
+  },
+  getRole() {
+    return Radio.request('entities', 'roles:model', this.get('_role'));
+  },
+  getState() {
+    return Radio.request('entities', 'states:model', this.get('_state'));
+  },
 });
 
 const Model = Store(_Model, TYPE);
--- a/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
+++ b/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
@@ -1,61 +1,47 @@
+import _ from 'underscore';
+
 import hbs from 'handlebars-inline-precompile';
 import { View, CollectionView } from 'marionette';
 
 import './action-sidebar.scss';
 
 const CreatedTemplate = hbs`
-  {{#with metadata.editor}}
-  {{formatHTMLMessage (intlGet "patients.sidebar.action.activityViews.created") name = name role = role.name}}
-  {{/with}}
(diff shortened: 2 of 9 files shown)


test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -1,6 +1,6 @@
 import moment from 'moment';
 
-import { getResource } from 'helpers/json-api';
+// import { getResource } from 'helpers/json-api';
```
paulfalgout: can remove

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -219,7 +226,8 @@ context('action sidebar', function() {
       .wait('@routePatientActions')
       .wait('@routeAction')
       .wait('@routeActionActivity')
-      .wait('@routePatient');
+      .wait('@routePatient')
+      .wait('@routeGroups');
```
paulfalgout: I doubt this wait is necessary?  None of the routes above will run until this one occurs because they all wait on the bootstrap to be successful.  It's kind of implicitly `wait`'d

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -194,6 +194,13 @@ context('action sidebar', function() {
 
     cy
       .server()
+      .routeGroups(fx => {
+        fx.data[2].relationships.clinicians.data[1] = { id: '22222', type: 'clinicians' };
+        fx.included[0].id = '22222';
```
paulfalgout: probably should set the `type` here too to future proof any other includes being added to this groups call.  Might be safer to just do `fx.included[0] = { ... }` or to `_.find(fx.included, { type: 'clinicians' })`

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -75,6 +60,22 @@ const ActivityView = View.extend({
 
     return Templates[this.model.get('type')];
   },
+  templateContext() {
+    const editor = this.model.getEditor();
+    const clinician = this.model.getClinician();
+    const to_clinician = this.model.getClinician() ? `${ clinician.get('first_name') } ${ clinician.get('last_name') }` : null;
```
paulfalgout: I'm surprised this didn't fail the linter.. while we can send snake cased to templates, locally they should be camel cased and translated at the object `to_clinician: toClinician`

src/js/entities-service/entities/events.js
```diff
@@ -6,6 +7,20 @@ const TYPE = 'events';
 
 const _Model = BaseModel.extend({
   type: TYPE,
+
+  getClinician() {
+    return Radio.request('entities', 'clinicians:model', this.get('_clinician'));
+  },
+  getEditor() {
```
paulfalgout: I think this might be a little cleaner if we do something like:
```js
  if (!this.get('_editor')) {
    return Radio.request('entities', 'clinician:model', {
      first_name: 'RoundingWell',
      last_name: '',
     });
  }
```
with a similar assumption on `clinician.getRole` this way if/when we display events elsewhere, this special case is in one place.  However this'll mean that we'll need to `_.trim` the name as `first_name + ' ' + '';` will be one too many spaces.

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
paulfalgout: should also probably be `clinician ?`  However it looks to me like it will always be true?  `getClinician` `Radio.request('entities', 'clinicians:model', this.get('_clinician'));` will always return at least an empty clinician model

paulfalgout: you could also just always do this instead of `: null` and `_.trim` it which for empty clinician models will result in an empty string which'd be nullish

test/integration/patients/sidebar/action-sidebar.js
luketlancaster: whoops!

PR #84, opened 2019-09-18
Add formio POC

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21006]


paulfalgout:
Something strange going on with istanbul and the handlebars update https://github.com/wycats/handlebars.js/issues/1553  I'm going to revert that update shortly.  I think the test are ok, but I need to test further.

paulfalgout:
Going to rebuild the commits, but here's the form example: https://fe-feature-ch21006-formio-poc.preview.roundingwell.com/forms/12345

PR #89, opened 2019-09-30
Switch to auth0 access token

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21061]

Should be merged with https://github.com/RoundingWell/care-ops-backend/pull/279

### DIFF
--- a/src/js/auth.js
+++ b/src/js/auth.js
@@ -1,18 +1,7 @@
-import auth0 from 'auth0-js';
+import $ from 'jquery';
+import createAuth0Client from '@auth0/auth0-spa-js';
 
-const AUTHD_PATH = '/authenticated';
-
-let webAuth;
-
-function authorize(connection, prompt = 'none') {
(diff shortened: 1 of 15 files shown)


luketlancaster:
@paulfalgout you want this to go into the release?

src/js/auth.js
```diff
@@ -1,39 +1,19 @@
...
-function authorize(connection, prompt = 'none') {
-  webAuth.authorize({
-    responseType: 'id_token',
-    responseMode: 'fragment',
-    redirectUri: window.location.origin + AUTHD_PATH,
-    connection,
-    prompt,
-  });
-}
+let auth0;
```
shadowhand: Shouldn't this be in a separate import? Such as:

```js
# src/js/utils/auth.js
import createAuth0Client from '@auth0/auth0-spa-js'

const auth = await createAuth0Client({
  /* ??? */
});

export default auth;
```

?
export default

PR #91, opened 2019-10-09
Use audience for auth0

paulfalgout:
Without setting the audience you get an opaque access token


paulfalgout:
Ignore codecov here.  It's a false report.

PR #93, opened 2019-10-14
Add FE release script

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21119]

An example release made from the script https://github.com/RoundingWell/care-ops-frontend/tree/release/20191014

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,3 @@
+{
+  "name": "roundingwell/care-ops-frontend"
+}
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -0,0 +1,3 @@
+{
+  "name": "roundingwell/care-ops-frontend"
```
luketlancaster: should there be anything more in here? Or is the bare minimum fine?

paulfalgout: the other fields we could put are like author, license, keyword etc.  we don't want version here..  couldn't find anything else useful.

PR #94, opened 2019-10-15
Update cypress e2e setup for access token

paulfalgout:
- [ ] Add QA tasks


paulfalgout:
this is a pretty small change entirely behind the `_DEVELOP_` flag so I'm merging it.

PR #95, opened 2019-10-15
Add program list page

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21120]

### DIFF
--- a/src/js/apps/programs/list/programs-all_app.js
+++ b/src/js/apps/programs/list/programs-all_app.js
@@ -1,0 +1,17 @@
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+import { ListView, LayoutView } from 'js/views/programs/list/programs-all_views';
+
+export default App.extend({
+  onBeforeStart() {
+    this.showView(new LayoutView());
+    this.getRegion('list').startPreloader();
+  },
+  beforeStart() {
+    return Radio.request('entities', 'fetch:programs:collection');
+  },
+  onStart(options, collection) {
+    this.showChildView('list', new ListView({ collection }));
+  },
+});
--- a/src/js/apps/programs/programs-main_app.js
+++ b/src/js/apps/programs/programs-main_app.js
@@ -1,0 +1,22 @@
+import RouterApp from 'js/base/routerapp';
+
+import ProgramsAllApp from 'js/apps/programs/list/programs-all_app';
+
+export default RouterApp.extend({
+  routerAppName: 'ProgramsApp',
+
+  childApps: {
+    programsAll: ProgramsAllApp,
+  },
+
+  eventRoutes: {
+    'programs:all': {
+      action: 'showProgramsAll',
+      route: 'programs',
+      isList: true,
+    },
+  },
+  showProgramsAll() {
+    this.startCurrent('programsAll');
+  },
+});
--- a/src/js/entities-service/programs.js
+++ b/src/js/entities-service/programs.js
@@ -1,0 +1,13 @@
+import BaseEntity from 'js/base/entity-service';
+import { _Model, Model, Collection } from './entities/programs';
+
+const Entity = BaseEntity.extend({
+  Entity: { _Model, Model, Collection },
+  radioRequests: {
+    'programs:model': 'getModel',
+    'programs:collection': 'getCollection',
+    'fetch:programs:collection': 'fetchCollection',
+  },
+});
+
+export default new Entity();
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -155,6 +155,15 @@ careOptsFrontend:
             updated_past_three_days {Actions updated in the last 3 days by any clinician in your Group(s). States included are To Do and In Progress.}
             done_last_thirty_days {Actions set to Done by any clinician in your Group(s) in the last 30 days.}
           }
+  programs:
+    list:
+      programsAllViews:
+        emptyView: No Programs
+        itemView:
+          off: Off
+          on: On
+        layoutView:
+          title: Programs
   regions:
     preload:
       loading: Loading...
--- a/src/js/views/programs/list/programs-all_views.js
+++ b/src/js/views/programs/list/programs-all_views.js
@@ -1,0 +1,63 @@
+import moment from 'moment';
+
+import hbs from 'handlebars-inline-precompile';
+import { View, CollectionView } from 'marionette';
+
+import PreloadRegion from 'js/regions/preload_region';
+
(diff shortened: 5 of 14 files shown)


luketlancaster:
It's all there...though some of it is not pretty. Wanted to get it in before lunch for review though

src/js/apps/programs/list/programs-all_app.js
```diff
@@ -0,0 +1,21 @@
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+import { ListView, LayoutView } from 'js/views/programs/list/programs-all_views';
+
+export default App.extend({
+  stateEvents: {
```
paulfalgout: I don't see a need for this `stateEvents`

src/js/apps/programs/list/programs-all_app.js
```diff
@@ -0,0 +1,21 @@
...
+    'change': 'restart',
+  },
+  onBeforeStart() {
+    if (this.isRestarting()) return;
```
paulfalgout: can probably leave this out unless we need the restart

src/js/apps/programs/programs-main_app.js
```diff
@@ -0,0 +1,22 @@
...
+  },
+
+  eventRoutes: {
+    'programs:all': {
+      action: 'showProgramsAll',
+      route: 'programs',
+      isList: true,
+    },
+  },
+  showProgramsAll(formId) {
```
paulfalgout:
```diff
+
- formId
```

src/js/entities-service/programs.js
```diff
@@ -0,0 +1,16 @@
+import BaseEntity from 'js/base/entity-service';
+import { _Model, Model, Collection } from './entities/programs';
+
+const Entity = BaseEntity.extend({
+  Entity: { _Model, Model, Collection },
+  radioRequests: {
+    'programs:model': 'getModel',
+    'programs:collection': 'getCollection',
+    'fetch:programs:collection': 'fetchPrograms',
```
paulfalgout: this can just be `fetchCollection` unless something extra needs to happen

src/js/i18n/en-US.yml
```diff
@@ -155,6 +156,15 @@ careOptsFrontend:
             updated_past_three_days {Actions updated in the last 3 days by any clinician in your Group(s). States included are To Do and In Progress.}
             done_last_thirty_days {Actions set to Done by any clinician in your Group(s) in the last 30 days.}
           }
+  programs:
+    list:
+      programsAllViews:
+        emptyView: No Programs
+        itemView:
+          'off': "Off"
```
paulfalgout: what's with the quotes on the left side?  If they're needed let's change the text to `offText` or something so they're not

src/js/views/programs/list/programs-list.scss
```diff
@@ -0,0 +1,7 @@
+.program--on {
+  color: color(green);
+}
+
+.program--words {
+  margin-left: 6px;
```
paulfalgout: I'm nearly sure we should only be using 4px increments, so 4 or 8.  That said there's a little more styling to do, so I wouldn't worry too much about it.. can get caught on a second pass.

test/integration/programs/list/programs-all.js
```diff
@@ -0,0 +1,49 @@
...
+            name: 'First in List',
+            published: true,
+            updated_at: moment().format(),
+          },
+        };
+
+        fx.data[1].attributes = {
+          name: 'Last in List',
+          published: true,
+          updated_at: moment().subtract(2, 'days').format(),
```
paulfalgout: all the dates within a cypress test that are added to data should be `moment.utc()` and all the results should be `moment()`

test/integration/programs/list/programs-all.js
```diff
@@ -0,0 +1,49 @@
...
+        };
+
+        fx.data[2].attributes = {
+          name: 'Secomd in List, Not Published',
```
paulfalgout: it's the same in both places, so doesn't matter, but "Secomd"

test/integration/programs/list/programs-all.js
```diff
@@ -0,0 +1,49 @@
...
+        return fx;
+      })
+      .visit('/programs');
+
+    cy
+      .get('.table-list__item')
+      .first()
+      .should('contain', 'First in List')
+      .should('contain', 'On');
+
```
paulfalgout: you _could_ just `.next()` here and I think it's be ok.. the context doesn't change.

src/js/i18n/en-US.yml
luketlancaster: syntax highlighter made it look like they were reserved words, so I just did this to make it happy

src/js/views/programs/list/programs-all_views.js
```diff
@@ -0,0 +1,66 @@
...
+
+const ItemView = View.extend({
+  className: 'table-list__item',
+  tagName: 'tr',
+  template: hbs`
+    <td class="table-list__cell w-60">{{ name }}</td>
+    <td class="table-list__cell w-20 program--published">
+    {{#if published}}<span class="program--on">{{fas "toggle-on"}}<span class="program--words">{{ @intl.programs.list.programsAllViews.itemView.on }}</span></span>
+    {{else}}{{far "toggle-off"}}<span class="program--words">{{ @intl.programs.list.programsAllViews.itemView.off }}</span>{{/if}}
+    </td>
```
luketlancaster: thoughts on this mess? Is it fine? Seems like there might be a better way....

src/js/i18n/en-US.yml
paulfalgout: image

src/js/views/programs/list/programs-all_views.js
paulfalgout: it's bordering on needing it's own template, but fine for now.

test/integration/programs/list/programs-all.js
luketlancaster: Ah, was wondering why the test failed first time around 🙃

src/js/views/programs/list/programs-all_views.js
```diff
@@ -0,0 +1,66 @@
...
+    </div>
+    <div class="flex-region list-page__list js-list" data-list-region></div>
+  `,
+  regions: {
+    list: {
+      el: '[data-list-region]',
+      regionClass: PreloadRegion,
+    },
+  },
+  ui: {
```
paulfalgout: just caught this.  doesn't seem like `ui` is used or `.js-list`

PR #96, opened 2019-10-16
Remove FE deploy

paulfalgout:
So we won't need to build step, but the deploy step will need to move over to BE somehow.

Otherwise I think the clean up here is easy.  @BenCoffeed certainly don't have to deal with this until the FE dependency deploying is a thing. I just wanted to investigate what the ci would look like post-FE deploy.


paulfalgout:
Ah ha, and whatever github mechanism that's waiting on the automated preview will need to be cleaned up too.  cool.

PR #97, opened 2019-10-16
Adds admin link to the top nav menu

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21111]

left some comments in, as I'm sure there'll be requested changes and I can remove them on the next push

### DIFF
--- a/src/js/apps/globals/app-frame_app.js
+++ b/src/js/apps/globals/app-frame_app.js
@@ -13,6 +13,36 @@ import ProgramsMainApp from 'js/apps/programs/programs-main_app';
 
 import { AppNavView, AppNavCollectionView } from 'js/views/globals/app-nav/app-nav_views';
 
+const topNavMenu = new Backbone.Collection([
+  {
+    onSelect() {
+      Radio.trigger('event-router', 'patients:all');
+    },
+    id: 'your-workspace',
+    isFas: false,
+    icon: 'window',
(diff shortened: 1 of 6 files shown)


src/js/apps/globals/app-frame_app.js
```diff
@@ -84,9 +103,13 @@ export default App.extend({
     });
   },
   showAppNav() {
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    if (!currentUser.can('program-admin')) topNavMenu.remove(topNavMenu.at(0)); // UGLY
```
paulfalgout: add an id that can be a string so you can do `topNavMenu.remove('program-admin');`

src/js/apps/globals/app-frame_app.js
```diff
@@ -13,6 +13,36 @@ import ProgramsMainApp from 'js/apps/programs/programs-main_app';
...
+    isFas: true,
+    icon: 'tools',
+    text: 'globals.appNav.topMenu.admin',
+  },
+  {
+    onSelect() {
+      Radio.request('auth', 'logout');
+    },
+    id: 'sign-out',
+    isFas: true,
```
luketlancaster: couldn't get the `itemTemplateContext` in the OptionList to work for whatever reason, so did this instead

paulfalgout: works for now :+1:

PR #101, opened 2019-10-21
Add navigation switching

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21150]


paulfalgout:
well at least the rule on the main menu is only show stuff if there's more than 1..  since we don't have views yet, it probably doesn't make sense to add just program

paulfalgout:
whelp mark decided he wanted the single menu item after all

PR #103, opened 2019-10-21
Add program sidebar

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21121]


src/js/apps/admin/programs/sidebar/program-sidebar_app.js
```diff
@@ -0,0 +1,39 @@
...
+        // TODO: display errors in form
+        // const errors = this.program.parseErrors(responseJSON);
+      });
+  },
+  onDelete() {
+    this.program.destroy();
+    this.stop();
+  },
+  onStop() {
+    this.program.trigger('editing', false);
```
paulfalgout: we won't need this trigger.  This is what we use to highlight the action being edited in the list, but you can'd edit a program while seeing the list so we won't be using it.

luketlancaster:
So some of the misses were related to deletion/displaying existing programs which isn't a part of this PR. Given that I removed that code to increase the coverage, and when it gets added back in it should get tested in the normal course of development

PR #104, opened 2019-10-22
Add program errors

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21149]
Also added style updates, moving more naming from programs to admin and including [ch21139]
Also moved animation the sidebar animation to a separate shared import


src/js/i18n/en-US.yml
```diff
@@ -1,5 +1,40 @@
 ---
 careOptsFrontend:
+  admin:
```
luketlancaster: good, was toying around with doing this myself. I'll have conflicts, but it needed to happen

PR #105, opened 2019-10-22
Adds program layout + context trail

luketlancaster:
throws in a half done sidebar for free because I misread the cards 🙍

- [x] Add the Clubhouse Story ID: [ch21122]


src/js/views/admin/programs/program/sidebar/sidebar-layout.hbs
```diff
@@ -0,0 +1,18 @@
...
+    <div>
+      <button class="button--icon u-margin--r-8 js-menu">{{far "ellipsis-h"}}</button>
+    </div>
+  </div>
+  <div class="u-margin--t-16">
+    <h1>{{ name }}</h1>
+    Description:
+    <h2>{{ details }}</h2>
+    Program State
+    {{#if published}}
```
paulfalgout: don't have to worry about this here, and I realize this was half-done, but I modified the list stuff from this mark up, so check that out for the next work.  also just for clarity we shouldn't reuse these styles.  I should have renamed them in the previous PR, but the ones in the `programs-list.scss` will be `programs-list__` prefixed.  and any similar style here should be duplicated for use in this file.  if we ever reuse styles they should be in `sass/modules` and never across view scss files

PR #109, opened 2019-10-24
Add program actions

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21171]

A lot of this is copy-pasting... so there's a ton of added lines..


src/js/views/admin/actions/actions.scss
```diff
@@ -0,0 +1,9 @@
+.programactions__role {
```
luketlancaster: `.program-actions__role`?

paulfalgout: good catch

PR #110, opened 2019-10-24
Update dependencies

paulfalgout:
My last PR made one test a bit brittle due to timestamp issue.  I commented on it so that it doesn't happen again.  I also tried upgrading cypress, but we can't until this issue is resolved:
https://github.com/cypress-io/cypress/issues/5431

But I went ahead and updated the other dependencies.


test/integration/patients/patient/dashboard.js
```diff
@@ -130,7 +130,8 @@ context('patient dashboard page', function() {
       .wait('@routePatchAction')
       .its('request.body')
       .should(({ data }) => {
-        expect(data.attributes.due_date).to.equal(moment.utc().format('YYYY-MM-DD'));
+        // Datepicker doesn't user timestamp so due_date is local.
```
luketlancaster:
```suggestion
        // Datepicker doesn't use timestamp so due_date is local.
```
?

PR #112, opened 2019-10-25
Adds empty/null states for clearing duration/due action events

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21155]


src/js/i18n/en-US.yml
```diff
@@ -234,11 +234,13 @@ careOptsFrontend:
             cancelBtn: Cancel
             saveBtn: Save
         activityViews:
+          clinicianAssigned: '<strong>{ name }</strong> ({ role }) changed the Owner to { to_name }'
```
luketlancaster: did a sort on these, that's why this one moved

PR #113, opened 2019-10-29
Handle error codes

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21246]

### DIFF
--- a/src/js/app.js
+++ b/src/js/app.js
@@ -46,6 +46,8 @@ const Application = App.extend({
     this.configComponents();
     this.startServices();
     this.setListeners();
+    // Ensure Error is the first app initialized
+    new ErrorApp({ region: this.getRegion('error') });
   },
 
   configComponents() {
@@ -97,9 +99,6 @@ const Application = App.extend({
   },
 
   onStart() {
-    // Ensure Error is the first app initialized
-    new ErrorApp({ region: this.getRegion('error') });
-
     this.addChildApp('appFrame', AppFrameApp);
     this.startChildApp('appFrame', { view: this.getView().appView });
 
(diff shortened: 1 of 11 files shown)


src/js/app.js
```diff
@@ -46,6 +46,8 @@ const Application = App.extend({
     this.configComponents();
     this.startServices();
     this.setListeners();
+    // Ensure Error is the first app initialized
+    new ErrorApp({ region: this.getRegion('error') });
```
paulfalgout: moving this to `onBeforeStart` handles 403/500s for the bootstrap endpoints

src/js/entities-service/entities/clinicians.js
```diff
@@ -24,7 +24,11 @@ const _Model = BaseModel.extend({
     return Radio.request('entities', 'roles:model', this.get('_role'));
   },
   can(prop) {
-    return this.get('access') === 'manager';
+    /* istanbul ignore next */
+    return (_DEVELOP_ && !sessionStorage.getItem('cypress'))
```
paulfalgout: this'll make it so that local development always has access to everything all the time, but prod builds won't be affected.  _so_ we can use the e2e library and/or cypress to test different user accesses, or temporarily comment out this line during local development.  Figure this is better than constantly overriding this for testing.

src/js/entities-service/entities/clinicians.js
```diff
@@ -24,7 +24,11 @@ const _Model = BaseModel.extend({
...
+      || this.get('access') === 'manager'
```
paulfalgout: also this is super simplistic and not very good.  Left it in this state as a clear "we're definitely going to refactor this at some point once we have more use cases" kind of thing.

PR #121, opened 2019-11-05
Refactor/simplify fetch patterns

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21319]

OK.. This PR does 2 things:

**Instance Fetch**
I'm not thrilled with modifying the documented behavior of Bb's `fetch` however as you'll see it greatly reduces the complexity of using `fetch` in every instance we use it.  Also on success/fail it is still returning the jqxhr data, but on success it's inserting the instance of the model or collection which in general is what we want.   We may get to a place where we really need the jqxhr to be returned by the fetch and I think in that instance we can add an `options.useXhr` or something that would return that instead of the instance promise.

**`beforeStart` accepts an array of promises**
This one should likely remove our need for really ever using `$.Deferred` outside of the bootstrap service.  See the effects here https://github.com/RoundingWell/care-ops-frontend/compare/refactor/simplify-fetch-patterns?expand=1#diff-d6e4c48d0be5ae617f4aed9e9f2dc7a2R10
Two things to note with this change.
When one thing is resolved it'll be `onStart(options, instance, jqData, etc)`  when an array is used it'll be `onStart(options, [instance, jqData, etc], [instance, jqData, etc], ...)`
so
```js
beforeStart() {
  return Radio.request('entities', 'fetch:foo:model', id);
},
onStart(options, foo)
```
and
```js
beforeStart() {
  return [
    Radio.request('entities', 'fetch:foo:model', id),
    Radio.request('entities', 'fetch:bar:model', id),
  ];
},
onStart(options, [foo], [bar])
```

Also with promises.. `$.when`ing something that is not a promise would resolve it's value immediately.. so now arrays will act funny.  here's what I mean:
```js
// Simple value
beforeStart() {
  return 1;
},
onStart(options, one)

// Array before this PR
beforeStart() {
  return [1,2,3];
},
onStart(options, array)

// Array after this PR
beforeStart() {
  return [1,2,3];
},
onStart(options, one, two, three)
```

we rarely if ever resolve a value.. and if we do it'd probably be a model/collection instance.. we'll probably never ever return an array of values.. but something to keep in mind.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1520,6 +1520,12 @@
       "integrity": "sha512-ce5d3q03Ex0sy4R14722Rmt6MT07Ua+k4FwDfdcToYJcMKNtRVQvJ6JCAPdAmAnbRb6CsX6aYb9m96NGod9uTw==",
       "dev": true
     },
+    "@types/sizzle": {
+      "version": "2.3.2",
+      "resolved": "https://registry.npmjs.org/@types/sizzle/-/sizzle-2.3.2.tgz",
+      "integrity": "sha512-7EJYyKTL7tFR8+gDbB6Wwz/arpGa0Mywk1TJbNzKzHtzbwVmY4HR9WqS5VV7dsBUKQmPNr192jHr/VpBluj/hg==",
+      "dev": true
+    },
     "@types/tapable": {
       "version": "1.0.4",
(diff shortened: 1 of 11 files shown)


paulfalgout:
For future viewers this did get merged.  github was having a sad and so the merge for whatever reason wasn't reflected on the PR

PR #122, opened 2019-11-05
Feature/register auth0 callback url

BenCoffeed:
- [x] Add the Clubhouse Story ID: [ch21300]
- [ ] Add QA tasks

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,22 @@ jobs:
           name: Set GitHub status to failure
           command: .circleci/bin/update-gh-status failure
           when: on_fail
+
+  addPreviewSubdomainAuth0:
+    docker:
+      - image: lambci/lambda:build-python3.7
+    steps:
+      - attach_workspace:
+          at: ~/
+      - run:
+          name: Add Preview Subdomain to Auth0 configs
+          command: |
+            source .circleci/bin/auth0_tools
+            add_auth0_subdomain "https://fe-"`echo "$CIRCLE_BRANCH" | tr '[:punct:][:blank:]' '-' | tr '[:upper:]' '[:lower:]'`".preview.roundingwell.com"
+      - run:
+          name: Set GitHub status to failure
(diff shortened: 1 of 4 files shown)


src/assets/config.json
```diff
@@ -1,6 +1,6 @@
 {
-  "domain": "roundingwell-care-team.auth0.com",
-  "client_id": "zmfrKoBuMKsm5H4ZekNHeA4FRzIS44cz",
+  "domain": "rwell-careops-dev.auth0.com",
```
paulfalgout: we can't change this.  It's the config for local development.  We'll need a different `config.json` for the `*.preview.`s

.circleci/config.yml
```diff
@@ -52,6 +52,27 @@ jobs:
...
+      - run:
+          name: install curl
+          command: apk --no-cache add curl
+      - run:
+          name: Update GitHub status
+          command: |
+            echo "export VERSION_SUFFIX=$(echo "$CIRCLE_BRANCH" | tr '[:punct:][:blank:]' '-' | tr '[:upper:]' '[:lower:]')" >> $BASH_ENV
+            source $BASH_ENV  
+            source .circleci/bin/auth0_tools
+            add_auth0_callback "https://fe-${VersionSuffix}.preview.roundingwell.com/authenticated"
```
paulfalgout: This will also need to happen for Allowed Web Origins and Logout URLs

src/assets/config.json
BenCoffeed: 👍

.circleci/config.yml
BenCoffeed: Taken care of.

src/assets/config.json
BenCoffeed: Fixed.

BenCoffeed:
This is now operating as expected. Still want to hold off until we get the config.json redirector deployed.

PR #123, opened 2019-11-06
Chore/parallel requests

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21286]

In addition to changing group clinicians in the bootstrap and patient fields to parallel requests, I looked at entity requests as a whole and tried to standardize naming and functionality.  In some places we were passing the entity to get the id and some places we were passing the id, and in those cases we didn't have the entity, so the id everywhere is probably best.

Also I found a minor bug in the model/collection fetch change we just merged.  It was aborting calls based on the root url of the entity.. so when we fetched clinicians in the bootstrap it was aborting the `clinicians/me` request due to the root url of the clinician model being the same.  So now the xhr control uses the `url` when we override it and it was modified along with recent changes which no longer use a `success` callback.

Finally the patient-actions request wasn't setting up the filter quite right, so this'll be a bug in that epic until this gets in, but I also modified the style such that changing that on its own would be conflicting.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1520,12 +1520,6 @@
       "integrity": "sha512-ce5d3q03Ex0sy4R14722Rmt6MT07Ua+k4FwDfdcToYJcMKNtRVQvJ6JCAPdAmAnbRb6CsX6aYb9m96NGod9uTw==",
       "dev": true
     },
-    "@types/sizzle": {
-      "version": "2.3.2",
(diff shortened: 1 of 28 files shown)


paulfalgout:
tests are timing out in different places on cypress... going to see if I can determine why it's stalling... :-/

paulfalgout:
Well good news/bad news  Good news is reverting cypress consistently fixes the test hanging issue which is also the bad news.

paulfalgout:
I'm watching this and might try and debug it some if there's time https://github.com/cypress-io/cypress/issues/5581

PR #124, opened 2019-11-06
Uses program action status enum instead of published bool

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21321]

I _think_ this covers it...though I have to admit I stayed up late doing most of this so...you know...

### DIFF
--- a/src/js/apps/admin/program/action/action_app.js
+++ b/src/js/apps/admin/program/action/action_app.js
@@ -11,7 +11,7 @@ export default App.extend({
         _program: programId,
         _role: null,
         days_until_due: null,
-        published: false,
+        status: 'draft',
       });
     }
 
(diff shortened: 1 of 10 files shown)


test/support/api/program-actions.js
```diff
@@ -53,7 +53,7 @@ Cypress.Commands.add('routeAllProgramActions', (mutator = _.identity, programIds
     .fixture('test/roles').as('fxRoles');
 
   cy.route({
-    url: '/api/program-actions?*',
```
paulfalgout: this is the only change that makes me nervous as it essentially matches the same route as a single program-action.  Is it not matching with the `?` it seems like any use of this would call `?filter[status]`   Otherwise maybe no `*`?

luketlancaster: looks like it works fine if I switch it back....must have initially switched it when something else was wrong?

PR #129, opened 2019-11-07
Filter out initial action created event if action created from program

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21350]

### DIFF
--- a/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
+++ b/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
@@ -96,6 +96,10 @@ const ActivityView = View.extend({
 
 const ActivitiesView = CollectionView.extend({
   childView: ActivityView,
+  viewFilter({ model }) {
+    if (model.get('type') === 'ActionCreated' && this.model.get('_program')) return false;
+    return true;
+  },
 });
 
 const TimestampsView = View.extend({
--- a/test/integration/patients/sidebar/action-sidebar.js
+++ b/test/integration/patients/sidebar/action-sidebar.js
@@ -246,7 +246,6 @@ context('action sidebar', function() {
       })
       .routePatientActions(fx => {
         fx.data[0] = actionData;
-
         fx.data[0].relationships.clinician.data = { id: '11111' };
 
         return fx;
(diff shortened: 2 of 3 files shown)


test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -511,7 +512,7 @@ context('action sidebar', function() {
 
     cy
       .get('[data-activity-region]')
-      .should('contain', 'RoundingWell (System) added this Action')
+      .should('not.contain', 'RoundingWell (System) added this Action')
       .should('contain', 'Clinician McTester (Nurse) changed the Owner to Another Clinician')
       .should('contain', 'Clinician McTester (Nurse) changed the details of this Action')
       .should('contain', 'Clinician McTester (Nurse) changed the Due Date to ')
```
luketlancaster: you think we need a whole 'nother set of tests for a non-program action?

paulfalgout: was that the only test for `RoundingWell (System)` cause I don't want to lose that.  But yeah we probably should have another set of tests for the other option.. 1 for code coverage, but 2 `.should('not.contain', 'Whatever')` would also pass.. really we could probably add a test for an action with a program and have 2 events.. one create and the other the pathway one and then check for length of 1.

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -96,6 +96,10 @@ const ActivityView = View.extend({
 
 const ActivitiesView = CollectionView.extend({
   childView: ActivityView,
+  viewFilter({ model }) {
+    if (model.get('type') === 'ActionCreated' && this.model.getProgram()) return false;
```
paulfalgout: for some reason I assumed we already had a `getProgram` when slacking earlier.  We probably won't need it for anything.. we can just use `get('_program')` here and add less code.

test/integration/patients/sidebar/action-sidebar.js
luketlancaster: alrighty, I added another round for 'created by program action'. take a look and lemme know

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -202,7 +202,7 @@ context('action sidebar', function() {
       .should('not.exist');
   });
 
-  specify('display action sidebar', function() {
+  specify.only('display action sidebar', function() {
```
LisaManresa: remove .only

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -246,7 +246,7 @@ context('action sidebar', function() {
       })
       .routePatientActions(fx => {
         fx.data[0] = actionData;
-
+        console.log(fx.data);
```
LisaManresa: rogue console.log

test/integration/patients/sidebar/action-sidebar.js
luketlancaster: It is one way to make the tests run faster 🙃 (fixed in latest push)

LisaManresa: "QA engineers hate him. Find out this dev's brilliant secret to speeding up integration tests!"

PR #130, opened 2019-11-08
IE11 Support

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21311]

Slowly working my way through the webpack set up to get the app running in IE11. This fixes a couple of major errors that were preventing the app from even starting up. Unfortunately, testing IE11 with a local dev environment is problematic. SauceLabs has proved to work quite well, but I've run into an issue with Oauth where it gets stuck in a loop asking me to log in. I can't discern whether it's an IE11 issue or an issue with SauceLabs' tunneling. So I'd like to throw this up on quality-assurance and test in my Windows VM.

### DIFF
--- a/config/fontawesome-plugin.js
+++ b/config/fontawesome-plugin.js
@@ -15,7 +15,7 @@ function faLoader({ fas, far, fal }) {
   });
 
   return () => {
-    return `const { library } = require(\'@fortawesome/fontawesome-svg-core\');
+    return `const library = require(\'@fortawesome/fontawesome-svg-core\').library;
       library.add(
         ${ [...fasIcons, ...farIcons, ...falIcons].join(',') }
       );`;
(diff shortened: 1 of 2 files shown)


luketlancaster (approved):
looks good to me, but I'm guessing there could be more incoming once it gets tested out on a qa box?

LisaManresa:
Yeah. I'm sure there are other issues, but I need to verify whether or not this Oauth issue is real or just an environmental thing.

paulfalgout:
Yep so I think the auth0 thing is a saucelabs proxy issue.  Here's the error on auth0 `Callback URL mismatch. http://127.0.0.1:8081/authenticated is not in the list of allowed callback URLs`  I think we can add that proxy though so that it is supported.

paulfalgout:
There's no reason this can't go in now.  There's more to do, but we can get it in subsequent PRs

PR #134, opened 2019-11-12
Login prompt view

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21334]
- [ ] Add QA tasks


paulfalgout (approved):
~~This seems good to me, but can we check `npm run analyze` and see if the css file stays built into it's own file?~~

:+1: this works

ah but it moves the file out of assets anyhow..

src/js/auth.js
```diff
@@ -46,10 +48,16 @@ function login(success, config) {
       return;
     }
 
-    auth0.loginWithRedirect({
-      appState: location.pathname,
-      prompt: 'none',
-    });
+    if (location.pathname === '/login') {
+      auth0.loginWithRedirect({
+        appState: '/',
```
paulfalgout: actually this is breaking..  we need to be able to pass the appState from the original url.. and this'll make users login each time.  the card instructions were wrong.  hmm we also don't want to go to the login screen at all unless it's needed...

paulfalgout (dismissed):

src/assets/css/login-prompt.css
```diff
@@ -0,0 +1,94 @@
...
+
+
+.svg-inline--fa {
+  display: inline-block;
+  height: 1em;
+  vertical-align: middle;
+}
+
+.login-prompt__roundingwell-logo {
+  margin-bottom: 50px;
```
paulfalgout: Ah so now all px settings everywhere other than fonts should be multiples of 4px and ideally multiples of 8px

paulfalgout:
Ok the login logic was a lot more complicated than I thought it'd be.

Also webpack was separating the styles and loading them nicely, but it didn't need the css to be in assets after all.

Then I noticed that the styles should not overlap at all with the main app, so we can't style `body` or `.svg-inline--fa`  I don't _think_ it was a huge problem, but it's avoidable so 🎉

luketlancaster (approved):

PR #136, opened 2019-11-13
Only show groups filter if current clinician has > 1 group

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21377]

### DIFF
--- a/src/js/apps/patients/list/patients-all_app.js
+++ b/src/js/apps/patients/list/patients-all_app.js
@@ -25,6 +25,8 @@ export default App.extend({
     const groupId = this.getState('groupId') || groups.at(0).id;
     this.setState({ groupId });
 
+    if (groups.length === 1) return;
+
     const groupsSelect = new GroupsDropList({
       collection: groups,
       state: { selected: groups.get(groupId) },
--- a/src/js/apps/patients/view/view_app.js
+++ b/src/js/apps/patients/view/view_app.js
@@ -44,6 +44,8 @@ export default App.extend({
     const groupId = this.getState('groupId') || groups.at(0).id;
     this.setState({ groupId });
 
+    if (this.groups.length === 1) return;
+
     const groupsSelect = new GroupsDropList({
       collection: groups,
       state: { selected: groups.get(groupId) },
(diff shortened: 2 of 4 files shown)


src/js/apps/patients/view/view_app.js
```diff
@@ -44,6 +44,8 @@ export default App.extend({
     const groupId = this.getState('groupId') || groups.at(0).id;
     this.setState({ groupId });
 
+    if (this.groups.length === 1) return;
```
paulfalgout, line 47: ditto above

test/integration/patients/view.js
```diff
@@ -250,7 +250,9 @@ context('view page', function() {
     cy
       .get('.list-page__filters')
       .contains('Another Group');
+  });
 
+  specify('clinician in only one group', function() {
```
paulfalgout: this looks like it should be specify `group filtering - new actions` or something like that.  and there needs to be _another_ specify with one group down above 282.

src/js/apps/patients/list/patients-all_app.js
```diff
@@ -25,6 +25,8 @@ export default App.extend({
     const groupId = this.getState('groupId') || groups.at(0).id;
     this.setState({ groupId });
 
+    if (groups.length === 1) return;
```
paulfalgout, line 28: can we add a bail like this for `0` above 25 as a cheap work around no groups?  I'm not sure what the endpoint does when no group filter is sent, but I imagine it'll either be 0 results or something that just leaves the spinner vs a console error.

paulfalgout: actually, let's try it and if it's no results, this is probably worth doing.. but if it's a spinner, no groups should just be handled differently.

paulfalgout (changes requested):

src/js/apps/patients/list/patients-all_app.js
luketlancaster: the endpoint 400's, so we'll probably need to handle it differently right?

paulfalgout: yep ok

paulfalgout (approved):

PR #141, opened 2019-11-14
Use only assigned roles for actions

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21408]

### DIFF
--- a/test/integration/patients/sidebar/action-sidebar.js
+++ b/test/integration/patients/sidebar/action-sidebar.js
@@ -221,6 +221,17 @@ context('action sidebar', function() {
 
     cy
       .server()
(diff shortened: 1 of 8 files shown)


test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -369,6 +380,7 @@ context('action sidebar', function() {
 
     cy
       .get('.picklist')
+      .should('not.contain', 'Not Included')
```
paulfalgout, line 386: FYI: I tested this test with `getRoles` in the `patient/action_views` and the test fails. :tada:

luketlancaster (approved):
looks good to me

PR #142, opened 2019-11-15
ie11 support: Part 2

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21311]

In which Lisa takes an embarrassingly long time debugging IE11 only to realize we just needed additional polyfills.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -31,6 +31,8 @@ module.exports = {
   entry: [
     'core-js/modules/es.promise',
     'core-js/modules/es.array.iterator',
+    'core-js/modules/es.array.from',
+    'core-js/modules/es.string.includes',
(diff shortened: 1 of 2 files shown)


webpack.config.js
```diff
@@ -40,6 +42,7 @@ module.exports = {
     chunkFilename: '[name].[contenthash].bundle.js',
   },
   devServer: {
+    disableHostCheck: true,
```
LisaManresa, line 45: This prevents an `Invalid Host/Origin header` error in IE11.

luketlancaster (commented):
:shipit:

luketlancaster (approved):
whoopsies wrong radio button selected on the last one

PR #143, opened 2019-11-15
Display attachment/form related events on action sidebar

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21416]


test/support/defaults.js
```diff
@@ -25,6 +25,11 @@ beforeEach(function() {
     .routePatients() // Setup default patients/all route
     .routeStates()
     .routeRoles()
-    .routeForms()
+    .routeForms(fx => {
+      fx.data[0].id = '11111';
+      fx.data[0].attributes.name = 'Test Form';
+
+      return fx;
+    })
```
luketlancaster: thoughts? Instead of needing to change it in a bunch of places (potentially)?

paulfalgout: I think you should modify the actual `routeForms` and do something like:
```js
      return mutator({
        data: getResource(_.union(this.fxTestForms, _.sample(this.fxForms, 5)), 'forms'),
        included: [],
      });
```

paulfalgout (changes requested):

PR #147, opened 2019-11-18
Feature/render forms responses

paulfalgout:
We'll still need to test a form and response directly, but this ought to be enough to roll a release and for the feature to start getting QA'd


paulfalgout:
Surprisingly the only reason coverage went down was because we're not currently using `go:back` anymore.. fine with leaving that hanging.. _not_ sure why we're not seeing complaints about formapp.js coverage...

PR #149, opened 2019-11-18
Store entire submission and use only readOnly for forms

paulfalgout:
html mode for whatever reason doesn't show radio button data


luketlancaster (approved):
I _for sure_ understand what this is doing

wweaver4th:
It all checks out.

wweaver4th (approved):
Everything seems to work.

PR #151, opened 2019-11-18
Uses new created/updated_since filters instead of old deprecated ones

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21440]

### DIFF
--- a/src/js/apps/patients/view/view_app.js
+++ b/src/js/apps/patients/view/view_app.js
@@ -75,13 +75,13 @@ export default App.extend({
     const filters = {
       'owned-by-me': { clinician, status },
       'actions-for-my-role': { role: role.id, status },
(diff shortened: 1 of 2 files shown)


luketlancaster:
turns out I didn't pay attention to the actual fields when creating the faker script for forms and used `created_at` and `updated_at`...so those are already correct 😱

paulfalgout (approved):

shadowhand:
@luketlancaster do we not need to worry about the usage of `created` and `updated` timestamp fields?

luketlancaster:
@shadowhand we're not doing anything with those attrs on form/form-responses yet, and I actually did the [faker](https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/fixtures/config/forms.js#L10-L14) generator for forms 'incorrect', using `created_at` and `updated_at` initially (because I just copied over the attrs from other entities)

PR #152, opened 2019-11-19
Give root container div height of 100%

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21460]

Also piggybacking on this to adjust positioning for the login page, per Mark's request

### DIFF
--- a/src/sass/formapp.scss
+++ b/src/sass/formapp.scss
@@ -1,5 +1,3 @@
-html, body, #root { height: 100% }
-
 body {
   padding: 24px;
 }
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -5,7 +5,10 @@ import 'js/utils/formatting';
 
 import PreloadRegion from 'js/regions/preload_region';
 
+import 'sass/app-root.scss';
+
 const configVersion = '3';
+
 
 function startPreloader() {
   const preloadRegion = new PreloadRegion({ el: '#root' });
--- a/src/js/views/globals/login/login-prompt.css
+++ b/src/js/views/globals/login/login-prompt.css
@@ -8,7 +8,7 @@
 }
 
 .login-prompt__roundingwell-logo {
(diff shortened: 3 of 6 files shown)


src/sass/root.scss
```diff
@@ -0,0 +1 @@
+html, body, .app-root { height: 100% }
```
paulfalgout: can we add this class to the `formapp.scss` too?

paulfalgout (changes requested):

src/sass/root.scss
paulfalgout: also I think html/body are covered elsewhere for the rest of the app.  only formapp has it's own css because it's based on bootstrap

paulfalgout: for that matter this can just be `.app-root` and we can import this scss over in formapp.js and remove the #root style from that stylesheet.

LisaManresa: Both are set by _base.scss, but that is only included by `provider-core` which isn't imported until the app starts up. So we also need to set them in this app-root file. Should I import this into `provider-core.scss` as well and remove the height: 100%; declarations from that one?

paulfalgout: ah yeah ok that makes some sense.. dynamic loading is so much fun.  Ok well maybe we pull it out of the `_base` then...  this means that it _doesn't_ need to be in `formapp.scss` at all then correct?

LisaManresa: Ok, I'll pull it out of `_base.scss` and confirm the formapp styles.

src/sass/formapp.scss
```diff
@@ -1,4 +1,4 @@
-html, body, #root { height: 100% }
+@import './app-root.scss';
```
paulfalgout: I think this isn't needed?  since formapp is loaded from index.  These styles would already be in the DOM

paulfalgout (approved):

PR #153, opened 2019-11-20
Test display of new and complete form

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21449]

### DIFF
--- a/test/integration/forms/formapp.js
+++ b/test/integration/forms/formapp.js
@@ -1,0 +1,84 @@
+context('Formapp', function() {
+  specify('load basic form', function() {
+    cy
+      .server()
+      .routeAction()
+      .routeFormDefinition()
+      .visit('/formapp/1/a/1')
+      .wait('@routeAction')
+      .wait('@routeFormDefinition');
+
+    let reloadStub;
+
+    cy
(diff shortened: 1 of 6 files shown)


test/integration/forms/formapp.js
```diff
@@ -0,0 +1,56 @@
...
+      .get('body')
+      .should('contain', 'Family Medical History');
+
+    cy
+      .get('textarea')
+      .type('Here is some typing');
+
+    cy
+      .get('button')
+      .click();
```
paulfalgout: we should be able to test that it POSTs to the correct URL.

paulfalgout (changes requested):
on small test on the POST, otherwise :+1:

Actually run `npm run dev:coverage` locally and check the local coverage report after running the form tests to see if everything in formapp.js gets covered.  I don't think it is.  We can probably istanbul comment out the error handler, but ideally we're 100% on all 4 columns on all of the formapp files.

paulfalgout:
There are 2 misses we should take care of... I believe one is the error handler and the other is the radio request we're stubbing.  Both can have an istanbul ignore comment

PR #155, opened 2019-11-21
Update modal styles

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21482]

### DIFF
--- a/src/js/views/globals/modal/modal.hbs
+++ b/src/js/views/globals/modal/modal.hbs
@@ -11,7 +11,7 @@
 
 <div class="modal-footer" data-footer-region>
   <div class="u-float--right">
-    <a href="#" class="modal-cancel js-close">{{ cancelText }}</a>
+    <button class="button--text js-close">{{ cancelText }}</button>
     <button type="submit" class="{{ buttonClass }} js-submit">{{ submitText }}</button>
   </div>
 </div>
--- a/src/sass/modules/modals.scss
+++ b/src/sass/modules/modals.scss
@@ -1,4 +1,6 @@
 .modal {
+  background-color: $white;
+  border-radius: 4px;
   position: absolute;
   width: 600px;
 }
@@ -14,15 +16,12 @@
 }
 
 .modal-header {
-  background-color: $black-95;
-  border-bottom: 1px solid $black-85;
-  border-radius: 6px 6px 0 0;
-  padding: 16px 24px;
+  padding: 16px 16px 16px 24px;
   position: relative;
 }
 
 .modal-close {
-  color: $black-40;
+  color: $black-60;
   display: block;
   font-size: 16px;
   position: absolute;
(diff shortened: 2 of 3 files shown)


src/js/views/globals/modal/modal.hbs
```diff
@@ -12,6 +12,6 @@
 <div class="modal-footer" data-footer-region>
   <div class="u-float--right">
     <a href="#" class="modal-cancel js-close">{{ cancelText }}</a>
-    <button type="submit" class="{{ buttonClass }} js-submit">{{ submitText }}</button>
+    <button type="submit" class="button--primary {{ buttonClass }} js-submit">{{ submitText }}</button>
```
paulfalgout: I don't think `.button--primary` would go here.  If it's the better button than `button--blue` it should be the default, but `button--primary` is a specific style

src/sass/modules/modals.scss
```diff
@@ -1,4 +1,6 @@
 .modal {
+  background-color: $white;
+  border-radius: 6px;
```
paulfalgout: we should maybe check with mark on this, but I think 6px is a copy over from mainapp.  this should likely be 4 or 8

src/sass/modules/modals.scss
```diff
@@ -37,13 +36,7 @@
 }
 
 .modal-cancel {
```
paulfalgout: if this is all that's on this style I think we can use `u-margin--r-8`, but is there no hover/focus style at all for these cancels?

paulfalgout (changes requested):

src/sass/modules/modals.scss
LisaManresa: I'll check with Mark. Seems like there should be.

src/js/views/globals/modal/modal.hbs
```diff
@@ -11,7 +11,7 @@
 
 <div class="modal-footer" data-footer-region>
   <div class="u-float--right">
-    <a href="#" class="modal-cancel js-close">{{ cancelText }}</a>
-    <button type="submit" class="{{ buttonClass }} js-submit">{{ submitText }}</button>
+    <button class="modal-cancel button--text js-close">{{ cancelText }}</button>
+    <button type="submit" class="button--blue {{ buttonClass }} js-submit">{{ submitText }}</button>
```
paulfalgout: Ah so I mean that the value of `{{ buttonClass }}` should default to `blue--button` and no class should be added here.

paulfalgout (changes requested):

src/js/views/globals/modal/modal.hbs
```diff
@@ -11,7 +11,7 @@
 
 <div class="modal-footer" data-footer-region>
   <div class="u-float--right">
-    <a href="#" class="modal-cancel js-close">{{ cancelText }}</a>
-    <button type="submit" class="{{ buttonClass }} js-submit">{{ submitText }}</button>
+    <button class="modal-cancel button--text js-close">{{ cancelText }}</button>
```
paulfalgout: :+1:

src/js/views/globals/modal/modal.hbs
LisaManresa: Ah, that makes way more sense.

src/js/views/globals/modal/modal.hbs
paulfalgout: we can actually remove `.modal-cancel` here now.  I assume it's not used anywhere

paulfalgout (approved):

paulfalgout:
The CI failures here aren't related to this PR.  See: https://github.com/RoundingWell/care-ops-frontend/pull/159

PR #157, opened 2019-11-26
Adds program action droplist

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21491]


src/js/views/admin/program/workflows/workflows_views.js
```diff
@@ -6,6 +6,10 @@ import { View, CollectionView } from 'marionette';
 
 import 'sass/modules/buttons.scss';
 
+import Droplist from 'js/components/droplist';
+
+// import intl from 'js/i18n';
```
paulfalgout: can remove

src/js/apps/admin/program/workflows/workflows_app.js
```diff
@@ -20,6 +23,27 @@ export default App.extend({
     this.actions = actions;
 
     this.showChildView('content', new ListView({ collection: actions }));
+
+    if (_DEVELOP_) {
+      const lists = [];
+
+      lists.unshift({
```
paulfalgout: I don't think we'll need to `unshift` here I think we'll just be able to define the list with new action and new flow, and we can add the new flow option in this PR.    Really there's only one list here so we don't need to use lists at all.  This could be a collection like the publish state droplist which specifies test and icon.  You could also probably handle the click event like the `MainNavDroplist` where the event is handled on the component definition and calls `model.onSelect()`

paulfalgout (changes requested):

src/js/views/admin/program/workflows/workflows_views.js
luketlancaster: whoopsies

src/js/apps/admin/program/workflows/workflows_app.js
```diff
@@ -20,6 +20,16 @@ export default App.extend({
     this.actions = actions;
 
     this.showChildView('content', new ListView({ collection: actions }));
+
+    if (_DEVELOP_) {
+      this.showChildView('add', new AddActionDroplist({
+        picklistEvents: {
+          'picklist:item:select': ({ model }) => {
+            model.attributes.onSelect(this.program);
```
paulfalgout: we should use `model.get('onSelect')(this.program)`  While it's perfectly ok here, backbone cautions against using `attributes` in general, so we've stuck with that stylistically. 
I do think this event would be better on the view like:
```js
  picklistEvents: {
    'picklist:item:select': 'onSelect',
  },
  onSelect({ model }) {
    model.get('onSelect')();
  },
```

And I think the collection and `onSelect => () {`s using `this.program.id` can be here in the app.

src/js/views/admin/program/workflows/workflows_views.js
```diff
@@ -82,14 +87,58 @@ const LayoutView = View.extend({
...
+      },
+    ]);
+  },
+  popWidth: 248,
+  picklistOptions() {
+    return {
+      attr: 'name',
+      className: 'picklist workflows__picklist',
+      headingText: 'Add Workflow',
+      lists: [{
```
paulfalgout: so because there's one list this can be:
```
attr: 'name',
className: 'picklist workflows__picklist',
headingText: 'Add Workflow'
itemTemplate: hbs``
```
And droplist.collection is automatically handled as the droplist's primary list collection, so not needed to be attached.

src/js/views/admin/program/workflows/workflows_views.js
```diff
@@ -82,14 +87,58 @@ const LayoutView = View.extend({
...
     'click .js-add': 'click:add',
   },
 });
 
+const AddActionDroplist = Droplist.extend({
+  initialize({ model }) {
+    this.collection = new Backbone.Collection([
+      {
+        onSelect(program) {
+          Radio.trigger('event-router', 'program:action:new', program.id); // probably not gonna work?
```
paulfalgout: should work / seems to work?

paulfalgout (changes requested):

src/js/apps/admin/program/workflows/workflows_app.js
```diff
@@ -20,6 +23,32 @@ export default App.extend({
     this.actions = actions;
 
     this.showChildView('content', new ListView({ collection: actions }));
+
+    if (_DEVELOP_) {
+      const actrionDroplistMenu = new Backbone.Collection([
```
paulfalgout: `actionDroplistMenu`

src/js/apps/admin/program/workflows/workflows_app.js
```diff
@@ -20,6 +23,32 @@ export default App.extend({
     this.actions = actions;
 
     this.showChildView('content', new ListView({ collection: actions }));
+
+    if (_DEVELOP_) {
```
paulfalgout: since we can, let's do `if (!_DEVELOP) return;` here so later when we pull the flag out, the change is just the one line

paulfalgout (changes requested):
One little nit and a spell-check and this is :shipit:

src/js/apps/admin/program/workflows/workflows_app.js
```diff
@@ -20,6 +23,32 @@ export default App.extend({
...
     this.showChildView('content', new ListView({ collection: actions }));
+
+    if (!_DEVELOP_) return;
+
+    const actionDroplistMenu = new Backbone.Collection([
+      {
+        onSelect: () => {
+          Radio.trigger('event-router', 'program:action:new', this.program.id);
+        },
+        id: 'new-action',
```
paulfalgout: bah I just saw one other thing worth doing.  We're not using these `id`s so might as well remove them.  Also instead of using `name` let's just use `text` and remove `attr: 'name'` from the component definition.

paulfalgout (changes requested):
sorry... one more time "one more thing" probably the last one maybe. probably

src/js/apps/admin/program/workflows/workflows_app.js
luketlancaster: no problem

paulfalgout (approved):

PR #159, opened 2019-12-02
Fix formapp test

paulfalgout:
This test is flaky.
I _think_ it's because the stub gets called after the request which we're waiting on after the test.

### DIFF
--- a/test/integration/forms/formapp.js
+++ b/test/integration/forms/formapp.js
@@ -35,10 +35,7 @@ context('Formapp', function() {
 
     cy
       .get('button')
-      .click()
-      .then(() => {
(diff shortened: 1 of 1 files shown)


luketlancaster (approved):

paulfalgout:
didn't fix it :-/

paulfalgout:
The stub is getting called.. it's just _after_ the test fails...

paulfalgout:
From Cypress docs:
```
Note: Prefer .should() with callback over .then() for assertions as they are automatically rerun until no assertions throw within it but be aware of differences.
```

luketlancaster (approved):
good to know

PR #160, opened 2019-12-04
Create compact/icon only versions of published, owner, due components

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21519]

### DIFF
--- a/src/js/views/admin/actions/actions_views.js
+++ b/src/js/views/admin/actions/actions_views.js
@@ -15,10 +15,15 @@ import './actions.scss';
 
 const i18n = intl.admin.actions.actionsViews;
 
-const PublishedTemplate = hbs`<span class="{{ className }}">{{far icon}}{{ name }}</span>`;
-
 const PublishedComponent = Droplist.extend({
   isCompact: false,
+  getTemplate() {
+    if (this.getOption('isCompact')) {
+      return hbs`<span class="{{ className }}">{{far icon}}</span>`;
+    }
+
+    return hbs`<span class="{{ className }}">{{far icon}}{{ name }}</span>`;
+  },
   initialize({ model }) {
     this.collection = new Backbone.Collection([
       {
@@ -49,14 +54,15 @@ const PublishedComponent = Droplist.extend({
     const isCompact = this.getOption('isCompact');
 
     return {
-      className: isCompact ? 'button-secondary--compact' : 'button-secondary w-100',
-      template: PublishedTemplate,
+      className: isCompact ? 'button-secondary--compact is-icon-only' : 'button-secondary w-100',
+      template: this.getTemplate(),
     };
   },
   picklistOptions: {
     headingText: i18n.publishedComponent.headingText,
     getItemFormat({ attributes }) {
-      return new Handlebars.SafeString(PublishedTemplate(attributes));
+      const template = hbs`<span class="{{ className }}">{{far icon}}{{ name }}</span>`;
+      return new Handlebars.SafeString(template(attributes));
     },
   },
 });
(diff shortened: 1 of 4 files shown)


src/js/views/admin/actions/actions_views.js
```diff
@@ -49,35 +54,67 @@ const PublishedComponent = Droplist.extend({
...
-      'change:_role': 'render',
-    },
-    className: 'button-secondary w-100',
-    template: hbs`{{far "user-circle"}}{{ name }}{{#unless name}}{{ @intl.admin.actions.actionsViews.ownerComponent.defaultText }}{{/unless}}`,
+  viewOptions() {
+    const isCompact = this.getOption('isCompact');
+    const name = this.getState('selected');
+    return {
+      model: this.model,
+      tagName: 'button',
```
paulfalgout: this `tagName` and `attributes` is already this on the viewclass of a droplist.  I don't think we want to set the model here either.  It should get automatically set to the selected model... so `className` should be able to use `!this.model` instead of `!name`

src/js/views/admin/actions/actions_views.js
```diff
@@ -104,19 +141,43 @@ const days = _.map(_.range(100), function(day) {
...
+
     return hbs`{{far "stopwatch"}}{{ day }} {{formatMessage  (intlGet "admin.actions.actionsViews.dueDayComponent.unitLabel") day=day}}`;
   },
   viewOptions() {
+    const isCompact = this.getOption('isCompact');
+    const day = this.getState('selected').get('day');
+
     return {
-      className: 'button-secondary w-100',
+      model: this.model,
```
paulfalgout: like above, shouldn't need to set `model`, `tagName`, or `attributes` on a droplist

src/js/views/admin/actions/actions_views.js
```diff
@@ -49,35 +54,67 @@ const PublishedComponent = Droplist.extend({
...
+          return 'button-secondary--compact';
+        }
+
+        return 'button-secondary w-100';
+      },
+      template: this.getTemplate(),
+    };
   },
   initialize({ model }) {
+    this.model = model;
```
paulfalgout: I don't think we need to attach the model

src/js/views/admin/actions/actions_views.js
```diff
@@ -137,6 +198,7 @@ const DueDayComponent = Droplist.extend({
     },
   },
   initialize({ model }) {
+    this.model = model;
```
paulfalgout: don't think we need to attach the model

paulfalgout (changes requested):

src/js/views/admin/actions/actions_views.js
luketlancaster: so I could either use `name` or keep model attached below...went with `name`

src/js/views/admin/actions/actions_views.js
```diff
@@ -49,33 +54,57 @@ const PublishedComponent = Droplist.extend({
     const isCompact = this.getOption('isCompact');
 
     return {
-      className: isCompact ? 'button-secondary--compact' : 'button-secondary w-100',
-      template: PublishedTemplate,
+      tagName: 'button',
```
paulfalgout: doesn't need tagName

src/js/views/admin/actions/actions_views.js
```diff
@@ -49,33 +54,57 @@ const PublishedComponent = Droplist.extend({
...
-    template: hbs`{{far "user-circle"}}{{ name }}{{#unless name}}{{ @intl.admin.actions.actionsViews.ownerComponent.defaultText }}{{/unless}}`,
+  viewOptions() {
+    const isCompact = this.getOption('isCompact');
+    const name = this.getState('selected');
+    return {
+      modelEvents: {
+        'change:_role': 'render',
+      },
+      className() {
+        if (!name && isCompact) {
```
paulfalgout: it's not really `name`  it's `selected`  but:
```
const selected = this.getState('selected');
className() {
  this.model === selected // true 
```
because:
https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/components/droplist/index.js#L47

attaching the `model` should not make a difference.

paulfalgout (changes requested):

src/js/views/admin/actions/actions_views.js
luketlancaster: so when I log out `this.model` inside of `className()` it's `undefined` no matter the value of the owner

paulfalgout: ah.. right because `className` happens prior to the `initialize`.. hmm so it'd have to be `this.getOption('state').get('selected')`  😐 ok well maybe just call it `selected`

paulfalgout (approved):

PR #161, opened 2019-12-05
Add program flow item view to workflows

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21517]

Screen Shot 2019-12-05 at 6 35 57 AM

### DIFF
--- a/src/js/apps/admin/program/workflows/workflows_app.js
+++ b/src/js/apps/admin/program/workflows/workflows_app.js
@@ -23,10 +23,8 @@ export default App.extend({
     ];
   },
   onStart({ program }, [actions], [flows]) {
-    this.actions = actions;
-
-    const collection = new Backbone.Collection([...actions.models, ...flows.models]);
-    this.showChildView('content', new ListView({ collection }));
+    this.actions = new Backbone.Collection([...actions.models, ...flows.models]);
+    this.showChildView('content', new ListView({ collection: this.actions }));
 
     if (!_DEVELOP_) return;
 
--- a/src/js/apps/admin/program/workflows/workflows_app.js
+++ b/src/js/apps/admin/program/workflows/workflows_app.js
@@ -17,12 +17,16 @@ export default App.extend({
     this.getRegion('content').startPreloader();
   },
   beforeStart({ program }) {
-    return Radio.request('entities', 'fetch:programActions:collection:byProgram', { programId: program.id });
+    return [
+      Radio.request('entities', 'fetch:programActions:collection:byProgram', { programId: program.id }),
+      Radio.request('entities', 'fetch:programFlows:collection', { program: program.id }),
+    ];
   },
-  onStart({ program }, actions) {
+  onStart({ program }, [actions], [flows]) {
     this.actions = actions;
 
-    this.showChildView('content', new ListView({ collection: actions }));
+    const collection = new Backbone.Collection([...actions.models, ...flows.models]);
+    this.showChildView('content', new ListView({ collection }));
 
     if (!_DEVELOP_) return;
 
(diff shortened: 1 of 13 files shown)


src/js/views/admin/program/workflows/workflows.scss
```diff
@@ -43,21 +43,21 @@
...
 
-.workflows__action-icon {
+.workflows__icon {
   position: relative;
   top: -1px;
   vertical-align: 1px;
 
   .icon {
-    color: color(blue);
+    color: $black-60;
```
paulfalgout, line 61: The action icon is blue.  Only the flow folder is grey
https://marvelapp.com/68ec5id/screen/63607908

src/js/views/admin/program/workflows/workflows_views.js
```diff
@@ -77,6 +73,17 @@ const ItemView = View.extend({
...
   },
+});
+
+const ActionItemView = ItemView.extend({
+  template: ActionItemTemplate,
+  onClick() {
+    Radio.trigger('event-router', 'program:action', this.model.get('_program'), this.model.id);
+  },
+  onRender() {
+    ItemView.prototype.onRender.apply(this);
```
paulfalgout: I'd rather duplicate the `onRender`s than call the prototype

src/js/apps/admin/program/workflows/workflows_app.js
```diff
@@ -17,12 +17,16 @@ export default App.extend({
...
+      Radio.request('entities', 'fetch:programActions:collection:byProgram', { programId: program.id }),
+      Radio.request('entities', 'fetch:programFlows:collection', { program: program.id }),
+    ];
   },
-  onStart({ program }, actions) {
+  onStart({ program }, [actions], [flows]) {
     this.actions = actions;
 
-    this.showChildView('content', new ListView({ collection: actions }));
+    const collection = new Backbone.Collection([...actions, ...flows]);
```
paulfalgout: I think both of these should be `...actions.models` and `..flows.models`  But does this work?  I can't figure out why..

paulfalgout (changes requested):

src/js/apps/admin/program/workflows/workflows_app.js
LisaManresa: Huh, I'm not sure why it works either. I'll change it to models.

LisaManresa:
I need to address the cypress tests, so I'm going to tag this as do not merge for now.

test/integration/admin/programs/program/workflows.js
```diff
@@ -44,9 +43,18 @@ context('program workflows page', function() {
         fx.data = testAction;
         return fx;
       })
+      .routeProgramFlows(fx => {
```
paulfalgout, line 46: I think this may be flaky if the random timestamp isn't after the `updated_at`s above

paulfalgout (changes requested):

paulfalgout:
failing tests seem legit

test/support/api/program-flows.js
```diff
@@ -0,0 +1,33 @@
+import _ from 'underscore';
+import { getResource, getRelationship } from 'helpers/json-api';
+
+Cypress.Commands.add('routeProgramFlows', (mutator = _.identity, programId) => {
+  cy
+    .fixture('collections/program-flows').as('fxProgramFlows')
```
luketlancaster, line 6: don't know if you've caught this yet, but you'll need to add a generator for this collection here: https://github.com/RoundingWell/care-ops-frontend/tree/develop/test/fixtures/config

LisaManresa:
Tests should be passing now. This should also take care of [ch21521].

PR #164, opened 2019-12-09
Changes default route to worklist/owned-by-me

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21555]

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -13,7 +13,7 @@ const i18n = intl.globals.nav;
 const appNavMenu = new Backbone.Collection([
   {
     onSelect() {
-      Radio.trigger('event-router', 'patients:all');
+      Radio.trigger('event-router', 'default');
     },
     id: 'PatientsApp',
     isFas: false,
--- a/src/js/apps/patients/patients-main_app.js
+++ b/src/js/apps/patients/patients-main_app.js
@@ -1,5 +1,7 @@
 import _ from 'underscore';
 
+import Radio from 'backbone.radio';
+
 import RouterApp from 'js/base/routerapp';
 
 import PatientApp from 'js/apps/patients/patient/patient_app';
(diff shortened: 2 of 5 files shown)


src/js/apps/globals/nav_app.js
```diff
@@ -13,7 +13,7 @@ const i18n = intl.globals.nav;
 const appNavMenu = new Backbone.Collection([
   {
     onSelect() {
-      Radio.trigger('event-router', 'patients:all');
+      Radio.trigger('event-router', 'worklist', 'owned-by-me');
```
luketlancaster: tried going to `default`...but that didn't work, so I did this instead.

paulfalgout: Why because the test failed or because it didn't go to the owned-by-me list?

luketlancaster: it didn't go to the owned-by-me list

src/js/apps/patients/patients-main_app.js
```diff
@@ -24,11 +24,12 @@ export default RouterApp.extend({
...
 
   defaultRoute() {
-    const defaultRoute = 'patients:all';
+    const defaultRoute = 'worklist';
+    const defaultWorklist = 'owned-by-me';
 
     this.routeAction(defaultRoute, () => {
-      this.replaceRoute(defaultRoute);
-      this.showPatientsAll();
+      this.replaceRoute(defaultRoute, defaultWorklist);
```
paulfalgout: This will fix the `'event-router', 'default'` issue:
```js
    this.routeAction(defaultRoute, () => {
      _.defer(()=> {
        this.replaceRoute(defaultRoute, defaultWorklist);
        Radio.request('nav', 'select', this.routerAppName, defaultRoute, [defaultWorklist]);
        this.showPatientsWorklist(defaultWorklist);
      });
    });
```

paulfalgout (changes requested):
We can also remove the `initialize` from patients-main_app

luketlancaster:
alrighty, updated with requests. Had to leave the `initialize` otherwise you'd get a 404 page on first load

paulfalgout:
does it work to remove `initialize` if you change `route` in the `default` `eventRoutes` from `'/'` to `''`?  Those should be doing the same thing.

luketlancaster:
Yup that got it

PR #166, opened 2019-12-12
Add Picklist/Droplist info text

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21585]
- [ ] Add QA tasks


paulfalgout:
sooo. cypress does not like sticky.  Was a bad solution anyhow.

LisaManresa (approved):

PR #169, opened 2019-12-16
Disable flow publishing when no actions are present

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21600]

This prevents a flow from being published if there are no published actions on it.

Only thing left is to sort out styles on the droplist. Supporting a disabled item in the picklist is complicated and requires more discussion.

### DIFF
--- a/src/js/views/admin/program/flow/flow_views.js
+++ b/src/js/views/admin/program/flow/flow_views.js
@@ -2,6 +2,8 @@ import Radio from 'backbone.radio';
 import hbs from 'handlebars-inline-precompile';
 import { View, CollectionView } from 'marionette';
 
+import intl from 'js/i18n';
+
 import PreloadRegion from 'js/regions/preload_region';
 
 import { DueDayComponent, OwnerComponent, PublishedComponent } from 'js/views/admin/actions/actions_views';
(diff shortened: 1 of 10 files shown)


src/js/views/admin/program/flow/flow_views.js
```diff
@@ -72,7 +74,12 @@ const HeaderView = View.extend({
     this.showOwner();
   },
   showPublished() {
-    const publishedComponent = new PublishedComponent({ model: this.model, isCompact: true });
+    const publishedComponent = new PublishedComponent({
+      model: this.model,
+      isCompact: true,
+      infoText: intl.admin.program.flowViews.flowStatusInfoText,
```
paulfalgout: I'm not 100% sure, but I assumed the infoText would only be there if publish was disabled.  We'll also need to do this same treatment to the flow sidebar

paulfalgout (changes requested):

src/js/views/admin/program/flow/flow_views.js
LisaManresa: Good call.

LisaManresa:
Also, just a note on coverage. It's dropped a little since we need to implement the actions logic before we can fully test the publish disabling/enabling. But once the actions are implemented I can quickly write tests to cover the other test cases.

LisaManresa:
[ch21600]

luketlancaster (approved):
Everything looks good to me, for whatever that's worth 🙃

PR #172, opened 2019-12-17
Add sort to worklists

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21586]

### DIFF
--- a/src/js/views/patients/worklist/worklist_views.js
+++ b/src/js/views/patients/worklist/worklist_views.js
@@ -4,7 +4,7 @@ import Radio from 'backbone.radio';
 import hbs from 'handlebars-inline-precompile';
 import { View, CollectionView } from 'marionette';
 
-import { renderTemplate } from 'js/i18n';
+import intl, { renderTemplate } from 'js/i18n';
 
 import 'sass/modules/list-pages.scss';
 import 'sass/modules/table-list.scss';
@@ -99,7 +99,10 @@ const LayoutView = View.extend({
   template: hbs`
     <div class="list-page__header">
       <div class="list-page__title">{{formatMessage (intlGet "patients.worklist.worklistViews.listTitles") title=worklistId role=role}}<span class="list-page__header-icon js-title-info">{{fas "info-circle"}}</span></div>
-      <div class="list-page__filters" data-filters-region></div>
+      <div class="list-page__filters flex">
+        <div data-filters-region></div>
+        <div class="worklist-list__filter" data-sort-region></div>
+      </div>
       <table class="w-100 js-list-header"><tr>
         <td class="table-list__header w-15">{{ @intl.patients.worklist.worklistViews.layoutView.patientHeader }}</td>
         <td class="table-list__header w-40">{{ @intl.patients.worklist.worklistViews.layoutView.actionHeader }}</td>
@@ -119,6 +122,7 @@ const LayoutView = View.extend({
   },
   regions: {
     filters: '[data-filters-region]',
+    sort: '[data-sort-region]',
     list: {
       el: '[data-list-region]',
       regionClass: PreloadRegion,
@@ -163,9 +167,6 @@ const ListView = CollectionView.extend({
   tagName: 'table',
   childView: ItemView,
   emptyView: EmptyView,
-  viewComparator({ model }) {
-    return - model.moment('updated_at').format('X');
-  },
   onAttach() {
     this.triggerMethod('update:listDom', this);
   },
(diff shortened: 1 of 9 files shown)


src/js/views/patients/worklist/worklist_views.js
```diff
@@ -186,8 +187,25 @@ const GroupsDropList = Droplist.extend({
   },
 });
 
+const SortDropList = Droplist.extend({
+  popWidth: 248,
+  picklistOptions() {
+    return {
+      headingText: intl.patients.worklist.worklistViews.sortDropList.headingText,
+      itemTemplate: hbs`
```
paulfalgout: I think this is essentially the default `itemTemplate`.  Is it needed?

test/integration/patients/worklist.js
```diff
@@ -45,12 +47,14 @@ context('worklist page', function() {
         fx.data[1].relationships.state = { data: { id: '55555' } };
         fx.data[1].attributes.name = 'Last In List';
         fx.data[1].attributes.updated_at = moment().subtract(2, 'days').format();
+        fx.data[1].attributes.due_date = moment().add(5, 'days').format();
```
paulfalgout: I think these'll be `moment(local)....format('YYYY-MM-DD')`  I don't think we should format these with a timezone.

paulfalgout (changes requested):

src/js/views/patients/worklist/worklist_views.js
luketlancaster: nope, sure isn't

test/integration/patients/worklist.js
```diff
@@ -44,13 +46,15 @@ context('worklist page', function() {
...
-        fx.data[1].attributes.updated_at = moment().subtract(2, 'days').format();
+        fx.data[1].attributes.updated_at = moment(local).subtract(2, 'days').format('YYYY-MM-DD');
+        fx.data[1].attributes.due_date = moment(local).add(5, 'days').format('YYYY-MM-DD');
 
         fx.data[2].relationships.state = { data: { id: '55555' } };
         fx.data[2].relationships.patient = { data: { id: '1' } };
         fx.data[2].id = '2';
         fx.data[2].attributes.name = 'Second In List';
-        fx.data[2].attributes.updated_at = moment().subtract(1, 'days').format();
+        fx.data[2].attributes.updated_at = moment(local).subtract(1, 'days').format('YYYY-MM-DD');
```
paulfalgout: update_at is a timestamp :-)  So these will be `.format()` and really for test data should probably be `moment.utc()`  On day-dates are localized to the user's date since they store no timezone.

paulfalgout (changes requested):
I think the coverage issue here is the unit test not being included in the report.  Can you confirm locally that the unit test covers the file?

test/integration/patients/worklist.js
luketlancaster: man one day I'll actually internalize what to do with mocking date/time...until then, 1000 reviews it is

luketlancaster (commented):
ran the unit tests and they all passed locally, so I think we're good there?

paulfalgout:
What about the local coverage reports though.

test/integration/patients/worklist.js
```diff
@@ -32,7 +34,7 @@ context('worklist page', function() {
             details: null,
             duration: 0,
             due_date: null,
-            updated_at: moment().format(),
+            updated_at: moment.utc().format('YYYY-MM-DD'),
```
paulfalgout: no format arg here :-D

paulfalgout (changes requested):

test/integration/patients/worklist.js
luketlancaster: ok so...why this comment from earlier? 
https://github.com/RoundingWell/care-ops-frontend/pull/172#discussion_r358941041

i think this is where some of my confusion is coming from...

luketlancaster: oh whoops, I keep looking at different parts of this test (because someone put updated_at and due_date in different orders here and below)

(that someone is me)

test/integration/patients/worklist.js
```diff
@@ -32,7 +34,7 @@ context('worklist page', function() {
             details: null,
             duration: 0,
             due_date: null,
-            updated_at: moment().format(),
+            updated_at: moment.utc(),
```
paulfalgout: still needs a format.. just no arguments

paulfalgout (changes requested):

luketlancaster:
iSuck

PR #175, opened 2019-12-18
Remove _DEVELOP_ flag and old workflow add button

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21606]

### DIFF
--- a/src/js/views/admin/program/workflows/layout.hbs
+++ b/src/js/views/admin/program/workflows/layout.hbs
@@ -3,7 +3,5 @@
     {{~ @intl.admin.program.workflows.layoutTemplate.workflowsLabel ~}}
   </span>
 </div>
-<div class="workflows__actions" data-add-region>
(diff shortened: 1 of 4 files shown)


src/js/views/admin/program/workflows/layout.hbs
```diff
@@ -3,7 +3,5 @@
     {{~ @intl.admin.program.workflows.layoutTemplate.workflowsLabel ~}}
   </span>
 </div>
-<div class="workflows__actions" data-add-region>
-  <button class="button-primary js-add">{{far "plus-circle"}}{{ @intl.admin.program.workflows.layoutTemplate.addActionBtn }}</button>
```
paulfalgout, line 7: i18n should be removed.. also whatever was using `.js-add` etc

paulfalgout (changes requested):

luketlancaster:
dangit I should've known it wasn't that easy

paulfalgout (approved):

PR #177, opened 2019-12-18
Change get of flows to go through program endpoint

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21626]

tested this locally and it worked like a charm


LisaManresa (approved):

luketlancaster:
`fetch:programFlows:collection` wasn't being used anymore, so I took that out (should make coverage happy). Then this can go in and we can move on to find the next 🐛 🙃

PR #179, opened 2019-12-19
Various fire-drill fixes for program-flow

paulfalgout:
- [ ] Add QA tasks


LisaManresa (approved):

luketlancaster (approved):

paulfalgout:
The failing test is a flaky test and not a code issue.  @LisaManresa the test in the screenshot is failing because the `program-flow/1` page's endpoints aren't stubbed so even though it does go there correctly, the failure causes an additional redirect so a slower test run might fail.

PR #187, opened 2020-01-09
Add patient flow page

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21610]

I imagine this will be conflicting with other test files, but hopefully not too much else.

This is the base additions so that we can break up the other work.  It sets up the cypress scenario, but does no actual testing.  I wouldn't trust yet the functionality of the list or components to be accurate as much of the was copy-paste-modify but with this out of the way I can make the rest of the cards

### DIFF
--- a/src/js/apps/admin/program/flow/flow_app.js
+++ b/src/js/apps/admin/program/flow/flow_app.js
@@ -1,8 +1,6 @@
 import _ from 'underscore';
 import Radio from 'backbone.radio';
 
-import intl from 'js/i18n';
-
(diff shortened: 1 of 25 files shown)


paulfalgout:
Ok the coverage reports are expected.  I think this is ok.

luketlancaster (approved):
nothing jumping out as crazy to me. :shipit:

LisaManresa (approved):

LisaManresa:
Merging this so Luke and I can resolve conflicts and get PRs up.

PR #188, opened 2020-01-09
Display patient flows

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21609]

### DIFF
--- a/src/js/views/patients/patient/data-events/action-item.hbs
+++ b/src/js/views/patients/patient/data-events/action-item.hbs
@@ -1,0 +1,9 @@
+    <td class="table-list__cell w-40">
+      <span class="patient__action-icon">{{far "file-alt"}}</span>{{ name }}
+    </td>
+    <td class="table-list__cell w-60">
+      <span class="table-list__meta" data-state-region></span>
+      <span class="table-list__meta" data-owner-region></span>
+      <span class="table-list__meta" data-due-region></span>
+      <span class="patient__action-ts">{{formatMoment updated_at "TIME_OR_DAY"}}</span>
+    </td>
--- a/src/js/views/patients/patient/data-events/action-item.hbs
+++ b/src/js/views/patients/patient/data-events/action-item.hbs
@@ -2,8 +2,8 @@
       <span class="patient__action-icon">{{far "file-alt"}}</span>{{ name }}
     </td>
     <td class="table-list__cell w-60">
-      <span class="table-list__meta" data-state-region></span>
-      <span class="table-list__meta" data-owner-region></span>
-      <span class="table-list__meta" data-due-region></span>
-      <span class="patient__action-ts">{{formatMoment updated_at "TIME_OR_DAY"}}</span>
+      <span class="table-list__meta" data-state-region></span>{{~ remove_whitespace ~}}
+      <span class="table-list__meta" data-owner-region></span>{{~ remove_whitespace ~}}
+      <span class="table-list__meta" data-due-region></span>{{~ remove_whitespace ~}}
+      <span class="patient__action-ts">{{formatMoment updated_at "TIME_OR_DAY"}}</span>{{~ remove_whitespace ~}}
     </td>
--- a/test/integration/admin/programs/sidebar/flow-sidebar.js
+++ b/test/integration/admin/programs/sidebar/flow-sidebar.js
@@ -162,7 +162,13 @@ context('flow sidebar', function() {
 
         return fx;
       })
-      .routeProgramFlowActions(_.identity, '1')
+      .routeProgramFlowActions(fx => {
+        _.each(fx.included, action => {
+          action.attributes.status = 'published';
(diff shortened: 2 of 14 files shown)


src/js/views/patients/patient/dashboard/add-action.scss
```diff
@@ -7,6 +7,10 @@
     color: color(blue);
   }
 
+  .picklist__flow-item .icon {
```
paulfalgout: I think this should be `add-action` prefixed.  `picklist` makes it look like it'd be found in the picklist module scss.

src/js/views/patients/patient/dashboard/add-workflow_views.js
```diff
@@ -10,7 +10,7 @@ import intl from 'js/i18n';
 
 import './add-action.scss';
 
-const ItemTemplate = hbs`{{#if isFar}}{{far iconType}}{{else}}{{fas iconType}}{{/if}} <a>{{ name }}</a>`;
+const ItemTemplate = hbs`<span class="{{#if isFar}}picklist__action-item{{else}}picklist__flow-item{{/if}}">{{#if isFar}}{{far iconType}}{{else}}{{fas iconType}}{{/if}} <a>{{ name }}</a></span>`;
```
paulfalgout: I don't like reusing `isFar` here it means "is fontawesome regular" we might want a different check for the type.  Is the __action-item class used?

paulfalgout: actually should change even be in this PR?

test/fixtures/config/flows.js
```diff
@@ -0,0 +1,24 @@
+const faker = require('faker');
+const moment = require('moment');
+
+module.exports = {
+  generate() {
+    return {
+      id: faker.random.uuid(),
+      name: `${ faker.company.bsBuzz() } ${ faker.company.catchPhraseNoun() }`,
+      details: faker.lorem.sentences(),
+      due_date: moment(faker.date.between(
```
paulfalgout: I don't think flows have `due_date` or `duration` but will have a `created_at`

paulfalgout (changes requested):

src/js/views/patients/patient/dashboard/add-workflow_views.js
LisaManresa: I can remove it. Just figured it was easy pickings, but I can leave it for @luketlancaster's PR if that's the more logical place for it

test/support/api/flows.js
```diff
@@ -0,0 +1,94 @@
...
+
+      return mutator({
+        data: action,
+        included: [],
+      });
+    },
+  })
+    .as('routeFlow');
+
+  // cy.routeProgramByAction();
```
luketlancaster: this needed?

src/js/views/patients/patient/data-events/action-item.hbs
```diff
@@ -0,0 +1,9 @@
+    <td class="table-list__cell w-40">
+      <span class="patient__action-icon">{{far "file-alt"}}</span>{{ name }}
+    </td>
+    <td class="table-list__cell w-60">
+      <span class="table-list__meta" data-state-region></span>
+      <span class="table-list__meta" data-owner-region></span>
+      <span class="table-list__meta" data-due-region></span>
+      <span class="patient__action-ts">{{formatMoment updated_at "TIME_OR_DAY"}}</span>
+    </td>
```
luketlancaster, line 9: couplea nits to pick: does this need to be indented? And also should we throw in `{{~ remove_whitespace ~}}` like what we have in the `flow-item` template?

paulfalgout:
The tests are failing consistently.. seems like a legit issue.

LisaManresa:
Yup, I think I got them resolved now. Running coverage locally and checking the coverage report now.

test/integration/admin/programs/sidebar/flow-sidebar.js
```diff
@@ -162,7 +162,13 @@ context('flow sidebar', function() {
 
         return fx;
       })
-      .routeProgramFlowActions(_.identity, '1')
+      .routeProgramFlowActions(fx => {
```
paulfalgout, line 165: 👍 ah flake.  good catch

test/support/api/flows.js
```diff
@@ -1,5 +1,55 @@
 import _ from 'underscore';
-import { getResource, getRelationship } from 'helpers/json-api';
+import { getResource, getIncluded, getRelationship } from 'helpers/json-api';
+
+Cypress.Commands.add('routePatientFlows', (mutator = _.identity, patientId) => {
```
paulfalgout, line 4: it probably does make sense to create some sort of data generator function that these two can share at some point as you were moving towards earlier.. but we can circle back around.

paulfalgout (approved):

PR #189, opened 2020-01-10
Add flow to patient from workflow dropdown

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21726]

this _technically_ works...but boy oh boy are there still broken things floating around...

### DIFF
--- a/src/js/apps/patients/patient/dashboard/dashboard_app.js
+++ b/src/js/apps/patients/patient/dashboard/dashboard_app.js
@@ -58,6 +58,12 @@ export default App.extend({
   },
 
   onAddProgramFlow(programFlow) {
+    const flow = programFlow.getFlow(this.patient.id);
+
+    flow.saveAll().done(() => {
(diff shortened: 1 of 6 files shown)


src/js/apps/patients/patient/dashboard/dashboard_app.js
```diff
@@ -58,6 +58,14 @@ export default App.extend({
   },
 
   onAddProgramFlow(programFlow) {
+    const flow = programFlow.getFlow(this.patient.id);
+
+    flow.saveAll().done(() => {
+      this.collection.unshift(flow);
```
paulfalgout: I don't think there's reason to add this to the collection if we're going to immediately route off the page

paulfalgout (changes requested):

src/js/apps/patients/patient/dashboard/dashboard_app.js
luketlancaster: was wondering about that. sounds fair

PR #190, opened 2020-01-10
Add patient flow sidebar

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21766]

In the interest of being non-blocking I'd like to get this in along with all of the other stuff I haven't written tests for 👿 and come back to those once we have all the tickets underway.  It's possibly a very bad idea for sure, but I'm trying to come close to hitting the date on the epic.


luketlancaster (approved):
lookin fine to me

PR #192, opened 2020-01-10
Adds all currently available flow event types/activity views

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21803]

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -53,7 +53,7 @@ export default App.extend({
   },
   onDelete() {
     // Fake a destroy for patient-flow-actions
-    if (!this.action.get('_flow')) {
+    if (this.action.get('_flow')) {
       this.action.stopListening();
       this.action.trigger('destroy', this.action, this.action.collection);
       this.stop();
--- a/src/js/apps/patients/sidebar/flow-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/flow-sidebar_app.js
@@ -4,7 +4,8 @@ import App from 'js/base/app';
 
 import intl from 'js/i18n';
 
-import { LayoutView, ActivitiesView } from 'js/views/patients/sidebar/flow/flow-sidebar_views';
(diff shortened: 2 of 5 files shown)


luketlancaster:
That last commit was just me poking around/looking at test failures...I don’t actually run the tests again :grimace:

src/js/apps/admin/sidebar/action-sidebar_app.js
```diff
@@ -32,7 +32,7 @@ export default App.extend({
   },
   onDelete() {
     // Fake a destroy for program-flow-actions
-    if (!this.action.get('_program')) {
```
paulfalgout: I don't think this is correct.  Pretty sure a `program-action` is completely unaware of what `program-flow` it is related to which is ultimately the reason we have to do this in the first place (despite what the documentation may suggest).. if a program action does not have a program relationship then it is part of a flow.  That's the only way for us to know not to delete it directly.

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -53,7 +53,7 @@ export default App.extend({
   },
   onDelete() {
     // Fake a destroy for patient-flow-actions
-    if (!this.action.get('_flow')) {
```
paulfalgout, line 56: this however is clearly a fix and we _need_ the flow relationship somehow, but I don't yet know how we're going to get it.

paulfalgout (changes requested):

src/js/apps/admin/sidebar/action-sidebar_app.js
luketlancaster: So I think I was just going off the comment here...if it's a program-flow-action, it should have the `program-flow` relationship and we can fake destroy it. If it's a program action, it won't have that and we continue on? Of course I did it at EOD on friday so...

paulfalgout: Yeah I suppose what it means is that we're fake destroying the action so that the program-flow-action can do the destroy which auto-destroys the action on the BE.  If we destroy the action here everything goes :boom:  But if we don't fake it, the action doesn't get removed from the list.

PR #193, opened 2020-01-13
Make flow state read-only on lists

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21819]

This also includes a couple of copy-paste fixes on patient flows and it moves the view sharing to behaviors.  Taking out the state was just one more thing that wasn't common between the views and we were ending up with things in those views that weren't needed, but were only there to maintain inheritance.

This also makes a static file for the state icon translation that we now need outside of the component.

### DIFF
--- a/src/js/views/patients/actions/actions_views.js
+++ b/src/js/views/patients/actions/actions_views.js
@@ -20,11 +20,7 @@ import './actions.scss';
 
 const i18n = intl.patients.actions.actionsViews;
 
-const StatusIcons = {
-  queued: 'exclamation-circle',
-  started: 'dot-circle',
-  done: 'check-circle',
-};
+import { PatientStatusIcons } from 'js/static';
 
 const StateTemplate = hbs`<span class="action--{{ statusClass }}">{{fas statusIcon}}{{#unless isCompact}}{{ name }}{{/unless}}</span>`;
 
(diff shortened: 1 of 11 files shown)


src/js/views/patients/actions/actions_views.js
```diff
@@ -20,11 +20,7 @@ import './actions.scss';
...
 const i18n = intl.patients.actions.actionsViews;
 
-const StatusIcons = {
-  queued: 'exclamation-circle',
-  started: 'dot-circle',
-  done: 'check-circle',
-};
+import { PatientStatusIcons } from 'js/static';
 
 const StateTemplate = hbs`<span class="action--{{ statusClass }}">{{fas statusIcon}}{{#unless isCompact}}{{ name }}{{/unless}}</span>`;
```
LisaManresa, line 25: Is there any risk of a `far` icon being used at some point? One thing to note, I've noticed that when you get the font awesome type wrong, the view fails completely silently. I've gotten stuck troubleshooting that a few times. It might be worth adding a dev card to see if we can get those errors to surface.

LisaManresa (approved):

luketlancaster (approved):

src/js/views/patients/actions/actions_views.js
paulfalgout: it's possible but unlikely here.  Yeah.. I ran into that too today.. no error.. just nothing.. bizarre.. we probably need to catch the case in our handlebars helper.. but I'm not sure how we're going to accomplish that..  I suppose we could check against the keys in package.json... perhaps only when on a dev instance.. otherwise that seems like a potentially costly check at render.

PR #195, opened 2020-01-13
Support flow actions in worklists

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21791]

### DIFF
--- a/src/js/views/patients/worklist/worklist_views.js
+++ b/src/js/views/patients/worklist/worklist_views.js
@@ -32,7 +32,13 @@ const ItemView = View.extend({
   tagName: 'tr',
   template: hbs`
     <td class="table-list__cell w-15 worklist-list__patient-name js-patient">{{ patient.first_name }} {{ patient.last_name }}</td>
-    <td class="table-list__cell w-40"><span class="worklist-list__name-icon">{{far "file-alt"}}</span>{{ name }}</td>
+    <td class="table-list__cell w-40">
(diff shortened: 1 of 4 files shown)


src/js/views/patients/worklist/worklist_views.js
```diff
@@ -32,7 +32,13 @@ const ItemView = View.extend({
   tagName: 'tr',
   template: hbs`
     <td class="table-list__cell w-15 worklist-list__patient-name js-patient">{{ patient.first_name }} {{ patient.last_name }}</td>
-    <td class="table-list__cell w-40"><span class="worklist-list__name-icon">{{far "file-alt"}}</span>{{ name }}</td>
+    <td class="table-list__cell w-40">
+      <span class="worklist-list__name-icon">{{far "file-alt"}}</span>{{~ remove_whitespace ~}}
+      <span class="worklist-list__name">
+        {{#if flowName}}<span class="worklist-list__flow-name">{{ flowName }}</span>{{/if}}{{~ remove_whitespace ~}}
```
paulfalgout: I think rather than making a span `display:block` this could be a `div`

paulfalgout (changes requested):
~~A flow will only have an owner component and the state of the flow needs to be read-only~~ actually never mind 🤦‍♂ there's no flows on the list

paulfalgout (approved):
I think this is just the one nitpick on the `display: block` `span`

PR #197, opened 2020-01-14
Print form work

wweaver4th:
- [x] Add the Clubhouse Story ID: [ch21619]
- [] Add QA tasks


luketlancaster:
this seems...not good...
Screen Shot 2020-01-14 at 9 42 45 AM

my guess is it's the redraw on render 🤔

src/js/formapp.js
```diff
@@ -46,6 +46,9 @@ function renderForm({ formId, actionId, patientId }) {
         .then(form => {
           form.nosubmit = true;
           form.submission = { data: fields.data.attributes };
+          form.on('render', () => {
+            form.redraw();
+          });
```
luketlancaster: so I _think_ this is what's triggering the infinite loop. But it's only doing it in Firefox from what I can tell (and it's only a warning so...). Chrome is fine, and performance is fine as well, so it might be ok? Looking at the failed test now

wweaver4th: Yeah, sorry for my crappy code. Originally this was an on change, but this would redraw the form everytime anything was changed and causing issues with the user experience. This redraw is to convert code tags in the HTML blocks. What's a better way to ensure it only does that once after the form is loaded?

paulfalgout: I don't think an event is needed here.  I think it can just be:
```
form.submission = 
form.redraw();
```

src/js/formapp.js
```diff
@@ -55,7 +58,7 @@ function renderForm({ formId, actionId, patientId }) {
             }).then(res => {
               form.emit('submitDone', res);
             }).fail(errors => {
-              /* istanbul ignore next: Don't need to test error handler */
+            /* istanbul ignore next: Don't need to test error handler */
```
paulfalgout: can this change be reverted?

src/js/formapp.js
```diff
@@ -75,10 +78,26 @@ Radio.reply('forms', 'reload', () => {
...
+        {% } %}
+        </div>
+        `,
+      };
       Formio.createForm(document.getElementById('root'), formDef, {
         readOnly: true,
+        renderMode: 'html',
       }).then(form => {
         form.submission = response;
+        form.on('change', () => {
```
paulfalgout, line 94: same here:
```js
form.submission = response;
form.redraw();
```

src/sass/formapp.scss
```diff
@@ -3,12 +3,67 @@ body {
...
+  font-size: 18px;
+  font-weight: bold;
+}
 .form-control:disabled, .form-control[readonly] {
   background-color: inherit;
   border: none !important;
 }
-.btn:disabled {
-  opacity: .5;
+.block {
```
paulfalgout: This seems like a bad idea.  `block` looks to be a utility class to make something into a block.

src/sass/formapp.scss
```diff
@@ -3,12 +3,67 @@ body {
...
+    .bg-light {
+      background-color: #fff !important;
+    }
+    .card {
+      border: none;
+    }
+  }
+}
+@media print {
+  .body {
```
paulfalgout: I don't think this class is doing anything and should be removed

src/sass/formapp.scss
```diff
@@ -3,12 +3,67 @@ body {
...
   background-color: inherit;
   border: none !important;
 }
-.btn:disabled {
-  opacity: .5;
+.block {
+  padding: 10px 20px;
+}
+@media screen {
+  .just-values th {
```
paulfalgout: are these being used in the forms?  Might be worth adding a comment to any custom classes that are not overriding formio or bootstrap.

paulfalgout (changes requested):
`package-lock.json` shouldn't be in the PR

src/js/formapp.js
wweaver4th: will do

src/sass/formapp.scss
paulfalgout: actually I'm not sure where `.block` is coming from, is this a custom class unrelated to form.io / bootstrap?  If so a name other than `block` might be useful due to `display: block`

src/js/formapp.js
wweaver4th: Both @luketlancaster and I tried that before and it didn't seem to work. But now it is ¯\_(ツ)_/¯

paulfalgout: I blame @luketlancaster

luketlancaster: https://gph.is/2C3S5eh

PR #199, opened 2020-01-15
Update flow actions when save is done

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21821]

### DIFF
--- a/src/js/apps/admin/program/flow/flow_app.js
+++ b/src/js/apps/admin/program/flow/flow_app.js
@@ -66,8 +66,8 @@ export default SubRouterApp.extend({
         flowAction.saveAll({ _program_action: action.id })
           .done(() => {
             this.setState('preventSort', false);
+            this.flow.set('_program_flow_actions', _.union(flowActionRelations, [{ id: flowAction.id }]));
           });
-        this.flow.set('_program_flow_actions', _.union(flowActionRelations, [{ id: flowAction.id }]));
         Radio.trigger('event-router', 'programFlow:action', this.flow.id, action.id);
       },
       'destroy'(action) {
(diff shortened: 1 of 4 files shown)


paulfalgout (dismissed):
a couple of things.. for whatever reason it doesn't look like the new line is getting covered.. it seems like it should with the new flow action, but it isn't.. in any case, the only way to test for this bug is to make sure that all of the actions in the list are published so that the published option is available on the flow.. and then adding a new action should make it unpublishable.

luketlancaster:
🤔 so are you saying that a flow is only publishable if all the actions are published? I thought it was publishable if one of the actions is published?

paulfalgout:
oh yeah you're right.. so.. well.. hmm.. the issue above is true.. the id wasn't available until the `done`, but there must be something else going on as adding an action can never change what the publishable state should be..

paulfalgout:
OK.. so this was a thing.. for one the flow action endpoints weren't stubbed so that's why the `done` was never occurring.. but it almost doesn't matter because we can't align the FE generated uuids from FE POSTs with Cypress responses... so we're not getting the correct events and we're so the new action never finishes its lifecycle..  so _yes_ we can test that the POST worked correctly, but we can't test specifically for this scenario in the bug :-/  It'd take the resolution of https://github.com/cypress-io/cypress/issues/521 I believe to make this work.. or alternatively we need to be able to control the result of the `uuid` utility from Cypress somehow.

Additionally we were storing and passing data we weren't using, so this also added a bit of clean up.

luketlancaster (commented):
those changes look good to me. Merge away, unless there's more work that I need to do...

PR #200, opened 2020-01-16
Add attachment link button to data & events actions

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch21835]

### DIFF
--- a/test/integration/patients/patient/data-and-events.js
+++ b/test/integration/patients/patient/data-and-events.js
@@ -26,6 +26,7 @@ context('patient data and events page', function() {
             clinician: { data: { id: '11111' } },
             role: { data: { id: null } },
             state: { data: { id: '55555' } },
+            forms: { data: [{ id: '1' }] },
           },
         };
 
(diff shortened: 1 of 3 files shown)


test/integration/patients/patient/data-and-events.js
```diff
@@ -208,9 +211,22 @@ context('patient data and events page', function() {
       .find('tr')
       .should('have.lengthOf', 3);
 
-    // TODO: Uncomment when patient flow app is implemented
-    // cy
-    //   .url()
-    //   .should('contain', 'flows/2');
```
luketlancaster, line 214: thoughts on the best way to handle needing to click through to new links on multiple items? A new specify block just for that?

paulfalgout: I think we can use `cy.go('back')` in some cases

paulfalgout (approved):

test/integration/patients/patient/data-and-events.js
luketlancaster: I tried that initially and it looked like it wasn't going to work...lemme play around with it more

PR #205, opened 2020-01-21
Add a couple of event displays

luketlancaster:
Also adds a check for events that exist but don't have templates, and
moved some fixtures around

- [x] Add the Clubhouse Story ID: [ch21867]

### DIFF
--- a/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
+++ b/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
@@ -32,6 +32,15 @@ const DueDateUpdatedTemplate = hbs`
   <div>{{formatMoment date "AT_TIME"}}</div>
 `;
 
+const DueTimeUpdatedTemplate = hbs`
+  {{#unless value}}
+  {{formatHTMLMessage (intlGet "patients.sidebar.action.activityViews.dueTimeCleared") name = name role = role }}
+  {{else}}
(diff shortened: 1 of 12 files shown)


src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -89,7 +101,9 @@ const ActivityView = View.extend({
       ActionStateUpdated: StateUpdatedTemplate,
     };
 
-    return Templates[this.model.get('type')];
+    if (!Object.keys(Templates).includes(type)) return hbs``;
```
paulfalgout: I think `!Templates[type]` works here.  Also `.includes` doesn't work in IE11 without a polyfill.  I think we have it, but I don't want to rely on it.

paulfalgout (changes requested):

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -32,6 +32,15 @@ const DueDateUpdatedTemplate = hbs`
   <div>{{formatMoment date "AT_TIME"}}</div>
 `;
 
+const DueTimeUpdatedTemplate = hbs`
+  {{#unless value}}
+  {{formatHTMLMessage (intlGet "patients.sidebar.action.activityViews.dueTimeCleared") name = name role = role }}
+  {{else}}
+  {{formatHTMLMessage (intlGet "patients.sidebar.action.activityViews.dueTimeUpdated") name = name role = role date = (formatMoment value "TIME")}}
```
luketlancaster, line 39: oh yeah I just guessed on this one...is this `"TIME"` correct here?

PR #208, opened 2020-01-22
Feature/quick search modal

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21859]

This is still a WIP, but I wanted to get eyes on it.

I still need to sort out some issues with an empty results collection and I need to write tests.

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -7,6 +7,7 @@ import App from 'js/base/app';
 import intl from 'js/i18n';
 
 import { AppNavView, AppNavCollectionView, MainNavDroplist, PatientsAppNav, AdminAppNav } from 'js/views/globals/app-nav/app-nav_views';
+import { PatientSearchModal } from 'js/views/globals/search/patient-search_views';
 
 const i18n = intl.globals.nav;
 
@@ -159,6 +160,10 @@ export default App.extend({
     navView.showChildView('patients', patientsCollectionView);
     navView.showChildView('worklists', workflowsCollectionView);
 
+    this.listenTo(navView, 'search', () => {
+      this.showSearchModal(navView);
+    });
+
     return navView;
   },
   getAdminAppNav() {
@@ -169,5 +174,34 @@ export default App.extend({
     navView.showChildView('admin', adminCollectionView);
 
     return navView;
+  },
+  showSearchModal(navView) {
+    const resultsCollection = Radio.request('entities', 'searchPatients:collection');
+
+    const patientSearchModal = new PatientSearchModal({
+      resultsCollection,
+    });
+
+    this.listenTo(patientSearchModal, {
+      'update:query': state => {
+        const query = state.get('query');
+
+        Radio.request('entities', 'fetch:searchPatients:collection', query).then(results => {
+          if (!results) return;
+          resultsCollection.set(results.models);
+        });
+      },
+      'item:select': ({ model }) => {
+        Radio.trigger('event-router', 'patient:dashboard', model.get('_patient'));
+        patientSearchModal.destroy();
+      },
+      'destroy': () => {
+        navView.triggerMethod('search:active', false);
+      },
+    });
+
+    Radio.request('modal', 'show:custom', patientSearchModal);
+
+    navView.triggerMethod('search:active', true);
   },
 });
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -163,7 +163,13 @@ export default App.extend({
     this.listenTo(navView, 'search', () => {
       this.showSearchModal(navView);
     });
-
+    
+    const hotkeyCh = Radio.channel('hotkey');
+    navView.listenTo(hotkeyCh, 'search', evt => {
+      evt.preventDefault();
+      this.showSearchModal(navView);
+    });
+    
     return navView;
   },
   getAdminAppNav() {
@@ -176,21 +182,11 @@ export default App.extend({
     return navView;
   },
   showSearchModal(navView) {
-    const resultsCollection = Radio.request('entities', 'searchPatients:collection');
-
     const patientSearchModal = new PatientSearchModal({
-      resultsCollection,
+      collection: Radio.request('entities', 'searchPatients:collection'),
     });
 
     this.listenTo(patientSearchModal, {
-      'update:query': state => {
-        const query = state.get('query');
-
-        Radio.request('entities', 'fetch:searchPatients:collection', query).then(results => {
-          if (!results) return;
-          resultsCollection.set(results.models);
-        });
-      },
       'item:select': ({ model }) => {
         Radio.trigger('event-router', 'patient:dashboard', model.get('_patient'));
         patientSearchModal.destroy();
--- a/src/js/components/picklist/index.js
+++ b/src/js/components/picklist/index.js
@@ -158,7 +158,7 @@ const Picklists = CollectionView.extend({
     return !!childView.children.length;
   },
   childViewEvents: {
-    'filter'() {
+    'before:render:children'() {
       return this.debouncedFilter();
     },
   },
--- a/src/js/components/picklist/index.js
+++ b/src/js/components/picklist/index.js
@@ -24,6 +24,8 @@ const CLASS_OPTIONS = [
   'noResultsText',
   'PicklistItem',
   'placeholderText',
+  'emptyView',
+  'emptyViewOptions',
 ];
 
 const CLASS_OPTIONS_ITEM = [
--- a/src/js/entities-service/entities/patient-search-results.js
+++ b/src/js/entities-service/entities/patient-search-results.js
@@ -1,3 +1,4 @@
+import _ from 'underscore';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
 
@@ -10,8 +11,30 @@ const Model = BaseModel.extend({
 const Collection = BaseCollection.extend({
   url: '/api/patients',
   model: Model,
+  initialize() {
+    this._debouncedSearch = _.debounce(this._debouncedSearch, 150);
+  },
+  search(search = '') {
+    if (search.length < 3) {
+      if (!search.length) this.reset();
+      this.isSearching = false;
+      return;
+    }
+    
+    this.isSearching = true;
+    this._debouncedSearch(search);
+  },
+  _debouncedSearch(search) {
+    const filter = { search };
+    
+    this.fetch({ data: { filter } }).then(() => {
+      this.isSearching = false;
+      this.trigger('search', this);
+    });
+  },
 });
 
 export {
   Collection,
+  Model,
 };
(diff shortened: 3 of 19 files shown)


src/js/entities-service/entities/patient-search-results.js
```diff
@@ -0,0 +1,22 @@
+import Store from 'backbone.store';
+import BaseCollection from 'js/base/collection';
+import BaseModel from 'js/base/model';
+
+const TYPE = 'patients-search-results';
+
+const _Model = BaseModel.extend({
+  type: TYPE,
+  urlRoot: '/api/patients',
```
paulfalgout: I don't think we need to set the `urlRoot`

src/js/entities-service/entities/patient-search-results.js
```diff
@@ -0,0 +1,22 @@
...
+});
+
+const Model = Store(_Model, TYPE);
```
paulfalgout: these also probably don't need to be stored

src/js/views/globals/search/patient-search_views.js
```diff
@@ -0,0 +1,78 @@
...
+  className: 'picklist patient-search__picklist',
+  isSelectlist: true,
+  placeholderText: intl.globals.search.patientSearchViews.patientSearchPicklist.placeholderText,
+  getItemFormat(item) {
+    return `${ item.get('first_name') } ${ item.get('last_name') }`;
+  },
+  itemTemplate: hbs`
+    <a{{#if isSelected}} class="is-selected"{{/if}}>
+      {{matchText text query}}{{~ remove_whitespace ~}}
+      <span class="picklist-item__birthdate">{{ birthdate }}</span>
```
paulfalgout: should be able to use `formatMoment` here instead of the context

src/js/views/globals/search/patient-search_views.js
```diff
@@ -0,0 +1,78 @@
...
+    this.listenTo(picklistView.getState(), 'change:query', this.onChangeQuery);
+    this.showChildView('picklist', picklistView);
+  },
+  onChangeQuery(state) {
+    this.triggerMethod('update:query', state);
+  },
+});
+
+export {
+  PatientSearchPicklist,
```
paulfalgout: don't need to export it

src/js/entities-service/patient-search-results.js
```diff
@@ -0,0 +1,18 @@
...
+import BaseEntity from 'js/base/entity-service';
+import { _Model, Model, Collection } from './entities/patients';
+
+const Entity = BaseEntity.extend({
+  Entity: { _Model, Model, Collection },
+  radioRequests: {
+    'fetch:search:patients': 'fetchSearchPatients',
+  },
+  fetchSearchPatients(query) {
+    const filter = { search: _.searchSanitize(query) };
```
paulfalgout: I don't think we need to `searchSanitize` here

src/js/views/globals/search/patient-search_views.js
```diff
@@ -0,0 +1,78 @@
...
+
+import intl from 'js/i18n';
+
+import Picklist from 'js/components/picklist';
+
+import './patient-search.scss';
+
+const PatientSearchModalHeader = hbs`{{far "search"}}`;
+
+const PicklistEmpty = View.extend({
```
paulfalgout: Maybe we could handle the searchTip as part of the modal view logic if there's no query and just use the noResultsText for the picklist itself.  That'd be less customizing

src/js/apps/globals/nav_app.js
```diff
@@ -170,4 +175,46 @@ export default App.extend({
 
     return navView;
   },
+  onSearch(navView) {
+    const resultsCollection = new Backbone.Collection([]);
+
+    const modalBody = new PatientSearchModalBody({
```
paulfalgout: what if we just passed the collection to the modal view here and let the modal view pass the collection to the picklist, and we may not need the event prefix?

src/js/apps/globals/nav_app.js
```diff
@@ -170,4 +175,46 @@ export default App.extend({
 
     return navView;
   },
+  onSearch(navView) {
+    const resultsCollection = new Backbone.Collection([]);
```
paulfalgout: this should probably be `Radio.request('entities', 'search:patients');`

src/js/entities-service/patient-search-results.js
```diff
@@ -0,0 +1,18 @@
+import _ from 'underscore';
+
+import BaseEntity from 'js/base/entity-service';
+import { _Model, Model, Collection } from './entities/patients';
+
+const Entity = BaseEntity.extend({
+  Entity: { _Model, Model, Collection },
+  radioRequests: {
+    'fetch:search:patients': 'fetchSearchPatients',
```
paulfalgout: maybe `fetch:searchPatients` since this isn't returning a patient model

src/js/views/globals/app-nav/app-nav.scss
```diff
@@ -53,3 +65,22 @@ $menu-hover-color: #375467;
     font-weight: 700;
   }
 }
+
+
+.modal.quick-search-modal {
```
paulfalgout: unused

src/js/views/globals/search/patient-search.scss
```diff
@@ -0,0 +1,90 @@
+.modal.patient-search__modal {
```
paulfalgout: I don't think we should chain or nest classes unless it cannot be avoided

paulfalgout (changes requested):
I think the regular modal is getting in the way.. what if we added a `show:custom` to the modal service that was:
```js
showCustom(view) {
  this.modalRegion.show(view);
  return view;
}
```
Then we don't need to override the generic modal stuff.

src/js/apps/globals/nav_app.js
```diff
@@ -170,4 +175,40 @@ export default App.extend({
...
+
+        Radio.request('entities', 'fetch:searchPatients', query).then(results => {
+          resultsCollection.reset(results.models);
+        });
+      },
+      'item:select': ({ model }) => {
+        Radio.trigger('event-router', 'patient:dashboard', model.get('_patient'));
+        patientSearchModal.destroy();
+      },
+      'close': () => {
```
paulfalgout: Isn't `destroy` always called?  Is `close` really needed?  If so that seems like a bug?

src/js/entities-service/patient-search-results.js
```diff
@@ -0,0 +1,17 @@
+import BaseEntity from 'js/base/entity-service';
+import { Model, Collection } from './entities/patient-search-results';
+
+const Entity = BaseEntity.extend({
+  Entity: { Model, Collection },
+  radioRequests: {
+    'searchPatients:collection': 'getCollection',
+    'fetch:searchPatients': 'fetchSearchPatients',
```
paulfalgout: ah this should probably be `:collection`  missed that before but we also postfix those right now.

src/js/views/globals/search/patient-search.scss
```diff
@@ -0,0 +1,78 @@
...
+
+.patient-search__picklist {
+  background-color: transparent;
+  box-shadow: none;
+  height: 100%;
+  max-height: 100%;
+  position: relative;
+  width: auto;
+
+  .picklist__fixed-heading {
```
paulfalgout: does this need to be nested?

src/js/views/globals/search/patient-search_views.js
```diff
@@ -0,0 +1,86 @@
...
+      {{matchText text query}}{{~ remove_whitespace ~}}
+      <span class="picklist-item__birthdate">{{ birthdate }}</span>
+    </a>
+  `,
+  itemTemplateContext() {
+    return {
+      birthdate: moment(this.model.get('birth_date')).format('MM/DD/YYYY'),
+    };
+  },
+  emptyView() {
```
paulfalgout: I think you can wrap this in `viewOptions` so we don't need to the picklist component.  Also we could probably add a `template` to the `viewOptions` to avoid needing to override the styles of `input-primary--small`

src/js/components/picklist/index.js
```diff
@@ -24,6 +24,8 @@ const CLASS_OPTIONS = [
   'noResultsText',
   'PicklistItem',
   'placeholderText',
+  'emptyView',
```
paulfalgout, line 27: if we can avoid this, that'd be best

src/js/apps/globals/nav_app.js
```diff
@@ -159,6 +160,10 @@ export default App.extend({
     navView.showChildView('patients', patientsCollectionView);
     navView.showChildView('worklists', workflowsCollectionView);
 
+    this.listenTo(navView, 'search', () => {
+      this.onSearch(navView);
```
paulfalgout: I think maybe this should be `showSearchModal`.  It seems awkward to call a `on___` method directly.

src/js/views/globals/search/patient-search_views.js
paulfalgout: I think this'd be `{{formatMoment birth_date "MM/DD/YYYY" inputFormat="YYYY-MM-DD"}}`

src/js/views/globals/search/patient-search_views.js
```diff
@@ -0,0 +1,80 @@
...
+    return `${ item.get('first_name') } ${ item.get('last_name') }`;
+  },
+  itemTemplate: hbs`
+    <a class="patient-search__picklist-item{{#if isSelected}} is-selected{{/if}}">
+      {{matchText text query}}{{~ remove_whitespace ~}}
+      <span class="picklist-item__birthdate">{{formatMoment birth_date "MM/DD/YYYY"}}</span>
+    </a>
+  `,
+  emptyView() {
+    if (this.model.get('query') && this.model.get('query').length > 0) return EmptyResultsViews;
```
paulfalgout: I'm not sure the `&&` will ever happen here?  or was length supposed to be `> 2`?

paulfalgout (changes requested):

src/js/views/globals/search/patient-search.scss
LisaManresa: Right now, yes in order to avoid affecting other picklists outside this modal. But if I can figure out the template thing you mentioned [here](https://github.com/RoundingWell/care-ops-frontend/pull/208/files#r370335140), we may not need it.

src/js/views/globals/search/patient-search_views.js
LisaManresa: `query` starts out as `undefined`, but after the user types and then clears it, it will be an empty string. My thought is that if the user has completely cleared their query, we should show the tip view again.

paulfalgout: an empty string is falsy though so I don't think the `length > 0` is doing anything

src/js/components/picklist/index.js
paulfalgout: You're not using these now

paulfalgout:
@LisaManresa @luketlancaster 👀 🙏

LisaManresa:
I can't approve since I opened the PR. But :shipit:

PR #210, opened 2020-01-24
due time component for actions

luketlancaster:
Wanted to get this up and reviewed before the weekend

- [x] Add the Clubhouse Story ID: [ch21864]

### DIFF
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -234,8 +234,13 @@ careOptsFrontend:
           bodyText: There are actions not done on this flow. Are you sure you want to set the flow to done?
           headingText: Set Flow to Done?
           submitText: Set to Done
-        dueComponent:
+        dueDayComponent:
           defaultText: Select Date...
+        dueTimeComponent:
+          defaultText: Set Time...
+          clear: Clear Time
+          placeholderText: Time
+          headingText: Set Time
         durationComponent:
           clear: Clear Duration
           defaultText: Select Duration...
--- a/src/js/views/patients/actions/actions.scss
+++ b/src/js/views/patients/actions/actions.scss
@@ -16,3 +16,16 @@
 .actions__due .is-overdue {
   color: color(red);
 }
+
+.actions__due {
+  border-radius: 4px 0 0 4px;
+}
+
+.actions__time {
+  border-left: 1px solid transparent;
+  border-radius: 0 4px 4px 0;
+
+  &:hover {
+    border-left: 1px solid $black-90;
+  }
+}
--- a/src/js/views/patients/actions/actions_views.js
+++ b/src/js/views/patients/actions/actions_views.js
@@ -185,7 +185,7 @@ const OwnerComponent = Droplist.extend({
   },
 });
 
-const DueComponent = Component.extend({
+const DueDayComponent = Component.extend({
   isCompact: false,
   viewOptions() {
     const isCompact = this.getOption('isCompact');
@@ -219,7 +219,7 @@ const DueComponent = Component.extend({
         </span>
       `,
       templateContext: {
-        defaultText: isCompact ? '' : i18n.dueComponent.defaultText,
+        defaultText: isCompact ? '' : i18n.dueDayComponent.defaultText,
         dateFormat: isCompact ? 'SHORT' : 'LONG',
         isOverdue() {
           if (!this.due_date) return;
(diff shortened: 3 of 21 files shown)


luketlancaster:
I'm on 6% battery...so wanted to get this up before the laptop dies/reviewed so I have something to do this afternoon. No tests yet, because ugh

src/js/i18n/en-US.yml
```diff
@@ -234,8 +234,13 @@ careOptsFrontend:
...
           headingText: Set Flow to Done?
           submitText: Set to Done
-        dueComponent:
+        dueDayComponent:
           defaultText: Select Date...
+        dueTimeComponent:
+          defaultText: Set Time...
+          clear: Clear Time
+          placeholderText: Time
+          headingText: Set Time (TZ)
```
paulfalgout: I don't think we need to add the timezone label at all now.

src/js/views/patients/actions/actions.scss
```diff
@@ -16,3 +16,17 @@
 .actions__due .is-overdue {
   color: color(red);
 }
+
+.program-action-sidebar__due {
```
paulfalgout: ideally we don't need to nest classes or `div`s

src/js/views/patients/actions/actions_views.js
```diff
@@ -254,6 +254,80 @@ const DueComponent = Component.extend({
...
+    }
+
+    return hbs`{{far "clock"}} {{time}}`;
+  },
+  viewOptions() {
+    const isCompact = this.getOption('isCompact');
+    const time = (this.getState('selected')) ? this.getState('selected').get('time') : null; // not great bob
+
+    return {
+      className() {
```
paulfalgout: can we do a `this.getClassName(time)` like the template one?

src/js/views/patients/actions/actions_views.js
```diff
@@ -254,6 +254,80 @@ const DueComponent = Component.extend({
   },
 });
 
+// Every 15 mins for 24 hours starting at 7am
+const start = moment({ hour: 6, minute: 45, second: 0 });
+
+const times = _.times(96, function() {
+  return { time: start.add(15, 'minutes').format('h:mm A') };
```
paulfalgout: I think this format should actually be `LT` for i18n consideration.. here and below where it reads the format.

paulfalgout (changes requested):

luketlancaster:
@paulfalgout wanna give this another look while I start down the testing path?

src/js/views/patients/actions/actions_views.js
```diff
@@ -254,6 +254,82 @@ const DueComponent = Component.extend({
...
+    placeholderText: i18n.dueTimeComponent.placeholderText,
+    isSelectlist: true,
+    getItemFormat(item) {
+      const time = item.get('time');
+
+      if (!time) {
+        return i18n.dueTimeComponent.clear;
+      }
+
+      return time;
```
paulfalgout: actually second look I think this is the only place that needs `.format('LT')` everything else can be `HH:mm:ss` which should remove the need for reformatting the time in a few places.

paulfalgout (changes requested):
need to review more, but wanted to get one comment in while I had it..

src/js/views/patients/actions/actions_views.js
luketlancaster: so if I don't have `format` up on [line 261](https://github.com/RoundingWell/care-ops-frontend/pull/210/files#diff-0a759ec2641c5a9b600cafd635ba96faR261) it generates 95 of the same time (6:45am), which is why I've ended up needing to format everywhere

src/js/views/patients/actions/actions.scss
```diff
@@ -16,3 +16,16 @@
 .actions__due .is-overdue {
   color: color(red);
 }
+
+.actions__due {
+  border-radius: 4px 0 0 4px;
+}
+
+.actions__time {
+  border-left: 1px solid white;
```
luketlancaster: this is kind of a lazy hack, and is noticeable if the screen gets big enough. Done so the button doesn't visibly shift on hover, when the left border goes from 0px to 1px

src/js/views/patients/actions/actions_views.js
```diff
@@ -254,6 +254,82 @@ const DueComponent = Component.extend({
...
+    }
+
+    if (!time) {
+      return hbs`{{far "clock"}}{{@intl.patients.actions.actionsViews.dueTimeComponent.placeholderText}}`;
+    }
+
+    return hbs`{{far "clock"}} {{time}}`;
+  },
+  viewOptions() {
+    const time = (this.getState('selected')) ? this.getState('selected').get('time') : null; // not great bob
```
luketlancaster: done because if `'selected'` is null everything blows up

src/js/views/patients/actions/actions_views.js
paulfalgout: Right so `.add` overwrites the moment, but there it should probably be `.format('HH:mm:ss')` so that everywhere the data is the same and the only place we're using `LT` is directly before displaying the data.

src/js/views/patients/actions/actions.scss
paulfalgout: why is it visible on a big screen?  What if the color was `transparent`?

src/js/views/patients/actions/actions_views.js
paulfalgout: when/why is selected null?

paulfalgout (changes requested):

src/js/views/patients/actions/actions.scss
luketlancaster: because there's 1px on the top and bottom of the border that "isn't there." And it'll need to be transparent anyway (if this is the way to go), because of the highlighting on the row if it's selected (click on the image and zoom way in)
Screen Shot 2020-01-28 at 9 33 57 AM

src/js/views/patients/actions/actions_views.js
luketlancaster: well apparently never...don't know why I had to do this previously, but don't need to anymore...

src/js/views/patients/actions/actions.scss
```diff
@@ -16,3 +16,16 @@
 .actions__due .is-overdue {
   color: color(red);
 }
+
+.actions__due {
+  border-radius: 4px 0 0 4px;
+}
+
+.actions__time {
+  border-left: 1px solid transparent;
```
luketlancaster, line 25: this is what mark liked the most

PR #213, opened 2020-01-28
Support full page form view in form app

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21687]

A first pass at toggling the full width view for forms. Need to review behaviors with Mark to confirm they are correct.

This uses 2 different properties on the state to keep track:
1. `sidebar`: The view/app that is currently active in the sidebar, regardless of whether it is collapsed or not.
2. `fullPageForm`: Keeps track of whether the form is expanded or not

As of now, if the action sidebar button is clicked while the form is expanded, we compress the form and display the action sidebar (Even if the previously displayed sidebar was the patient sidebar). This is the behavior I need to review with Mark.


src/js/apps/forms/form/form_app.js
```diff
@@ -25,6 +25,11 @@ export default App.extend({
     this.patient = patient;
     this.action = action;
 
+    this.setState({
+      'sidebar': false,
```
paulfalgout: don't need to be `'`'d

src/js/apps/forms/form/form_app.js
```diff
@@ -64,6 +72,30 @@ export default App.extend({
 
     this.listenTo(sidebar, 'stop', this.showPatientSidebar);
 
-    this.setState('actionSidebar', sidebar);
+    this.setState({
+      'sidebar': sidebar,
```
paulfalgout: same here with the `'`

src/js/views/forms/form/form_views.js
```diff
@@ -85,6 +95,9 @@ const LayoutView = View.extend({
   onClickPrintButton() {
     this.ui.iframe[0].contentWindow.print();
   },
+  onClickExpandButton() {
```
paulfalgout: can be removed?

src/js/apps/forms/form/form_app.js
```diff
@@ -64,6 +72,30 @@ export default App.extend({
...
+      'fullPageForm': false,
+    });
+  },
+  onClickExpandButton() {
+    const sidebar = this.getState('sidebar');
+
+    this.setState('fullPageForm', !this.getState('fullPageForm'));
```
paulfalgout: I think there's a `this.toggleState('fullPageForm')`

paulfalgout (changes requested):
I think the previous version of this was just using the state as a flag, but I think we might be better served here by reacting to the state change where the user interactions all just call a single `setState` function (or `toggleState`) and then based on state change events we re-render or something like that which run through the same group of logic based on the state.  I also think it may be better to store a flag (or flags) in the state rather than an app.  If we do go in the direction of being able to open the form app in various states, it'll be easier to pass a set of flags than it will be to determine which sidebar app should be stored in the state.

paulfalgout (approved):
Is this mark approved?

LisaManresa:
Yup. Screen shared with Mark yesterday and went over the functionality. Added in the tooltips at his request.

LisaManresa:
[ch21920]

PR #221, opened 2020-02-05
Change app widths

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch21928]

Makes all sidebars absolutely positioned and increased the site minimum width to 1200px


luketlancaster (approved):
👍🏻

PR #223, opened 2020-02-06
Migrate actions and flows to use owner instead of clinician/role

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21918]

### DIFF
--- a/src/js/entities-service/entities/actions.js
+++ b/src/js/entities-service/entities/actions.js
@@ -3,8 +3,16 @@ import Radio from 'backbone.radio';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
+import JsonApiMixin from 'js/base/jsonapi-mixin';
 
 const TYPE = 'patient-actions';
+const { parseRelationship } = JsonApiMixin;
+
+const _parseRelationship = function(relationship, key) {
+  if (!relationship || key === 'owner') return relationship;
+
+  return parseRelationship(relationship, key);
+};
 
 const _Model = BaseModel.extend({
   urlRoot() {
@@ -22,21 +30,13 @@ const _Model = BaseModel.extend({
 
     return forms.at(0);
   },
-  getClinician() {
-    const clinicianId = this.get('_clinician');
-    if (!clinicianId) return;
-    return Radio.request('entities', 'clinicians:model', clinicianId);
-  },
   getPatient() {
     return Radio.request('entities', 'patients:model', this.get('_patient'));
   },
-  getRole() {
-    const roleId = this.get('_role');
-    if (!roleId) return;
-    return Radio.request('entities', 'roles:model', roleId);
-  },
   getOwner() {
-    return this.getClinician() || this.getRole();
+    const owner = this.get('_owner');
+    if (!owner) return;
+    return Radio.request('entities', `${ owner.type }:model`, owner.id);
   },
   getFlow() {
     if (!this.get('_flow')) return;
@@ -66,45 +66,36 @@ const _Model = BaseModel.extend({
       },
     });
   },
-  saveClinician(clinician) {
-    return this.save({ _clinician: clinician.id, _role: null }, {
-      relationships: {
-        clinician: this.toRelation(clinician),
-      },
-    });
-  },
-  saveRole(role) {
-    return this.save({ _role: role.id, _clinician: null }, {
-      relationships: {
-        role: this.toRelation(role),
-      },
-    });
-  },
   saveOwner(owner) {
-    if (owner.type === 'clinicians') {
-      return this.saveClinician(owner);
-    }
-
-    return this.saveRole(owner);
+    return this.save({
+      _owner: owner,
+    },
+    {
+      relationships: {
+        owner: this.toRelation(owner),
+      },
+    });
   },
   saveAll(attrs) {
     attrs = _.extend({}, this.attributes, attrs);
+
     const relationships = {
       forms: this.toRelation(attrs._forms, 'forms'),
-      role: this.toRelation(attrs._role, 'roles'),
-      clinician: this.toRelation(attrs._clinician, 'clinicians'),
+      owner: this.toRelation(attrs._owner),
       state: this.toRelation(attrs._state, 'states'),
       program: this.toRelation(attrs._program, 'programs'),
     };
 
     return this.save(attrs, { relationships }, { wait: true });
   },
+  parseRelationship: _parseRelationship,
 });
 
 const Model = Store(_Model, TYPE);
 const Collection = BaseCollection.extend({
   url: '/api/actions',
   model: Model,
+  parseRelationship: _parseRelationship,
 });
 
 export {
--- a/src/js/entities-service/entities/flows.js
+++ b/src/js/entities-service/entities/flows.js
@@ -3,8 +3,16 @@ import Radio from 'backbone.radio';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
+import JsonApiMixin from 'js/base/jsonapi-mixin';
 
 const TYPE = 'flows';
+const { parseRelationship } = JsonApiMixin;
+
+const _parseRelationship = function(relationship, key) {
+  if (!relationship || key === 'owner') return relationship;
+
+  return parseRelationship(relationship, key);
+};
 
 const _Model = BaseModel.extend({
   urlRoot() {
@@ -17,17 +25,9 @@ const _Model = BaseModel.extend({
     return Radio.request('entities', 'patients:model', this.get('_patient'));
   },
   getOwner() {
-    return this.getClinician() || this.getRole();
-  },
-  getClinician() {
-    const clinicianId = this.get('_clinician');
-    if (!clinicianId) return;
-    return Radio.request('entities', 'clinicians:model', clinicianId);
-  },
-  getRole() {
-    const roleId = this.get('_role');
-    if (!roleId) return;
-    return Radio.request('entities', 'roles:model', roleId);
+    const owner = this.get('_owner');
+    if (!owner) return;
+    return Radio.request('entities', `${ owner.type }:model`, owner.id);
   },
   isDone() {
     const state = Radio.request('entities', 'states:model', this.get('_state'));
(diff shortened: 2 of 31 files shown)


src/js/views/patients/actions/actions_views.js
```diff
@@ -141,7 +140,7 @@ const OwnerComponent = Droplist.extend({
     }
     return {
       modelEvents: {
-        'change:_role change:_clinician': 'render',
+        'change:_role change:_clinician change:_owner': 'render',
```
LisaManresa: Just a note, we need to retain `change:_role change:_clinician` until the flows models are also migrated to use owner.

src/js/entities-service/entities/actions.js
```diff
@@ -22,21 +30,21 @@ const _Model = BaseModel.extend({
...
+    // modify the OwnerComponent to call getOwner() rather than getRole()
+    const owner = this.get('_owner');
+    if (!owner || owner.type !== 'roles') return;
+    return Radio.request('entities', 'roles:model', owner.id);
   },
   getOwner() {
-    return this.getClinician() || this.getRole();
+    const owner = this.get('_owner');
+    if (!owner) return;
+    if (owner.type === 'clinicians') return Radio.request('entities', 'clinicians:model', owner.id);
```
paulfalgout: I'd be ok with 
```
`${ owner.type }:model`
```
That really _should_ be true every time.

src/js/entities-service/entities/actions.js
```diff
@@ -66,45 +74,36 @@ const _Model = BaseModel.extend({
...
-      relationships: {
-        clinician: this.toRelation(clinician),
-      },
-    });
-  },
-  saveRole(role) {
-    return this.save({ _role: role.id, _clinician: null }, {
+  saveOwner(owner) {
+    return this.save({
+      _owner: owner,
```
paulfalgout, line 71: This is making me wonder if we should bite the bullet and make all single relationships `{ id, type }` in the near future

src/js/entities-service/entities/actions.js
```diff
@@ -66,45 +74,36 @@ const _Model = BaseModel.extend({
...
-    return this.saveRole(owner);
-  },
   saveAll(attrs) {
     attrs = _.extend({}, this.attributes, attrs);
+
     const relationships = {
       forms: this.toRelation(attrs._forms, 'forms'),
-      role: this.toRelation(attrs._role, 'roles'),
-      clinician: this.toRelation(attrs._clinician, 'clinicians'),
+      owner: attrs._owner ? this.toRelation(attrs._owner.id, attrs._owner.type) : null,
```
paulfalgout: wouldn't `owner: this.toRelation(attrs._owner)` work?

src/js/entities-service/entities/program-actions.js
```diff
@@ -22,29 +30,44 @@ const _Model = BaseModel.extend({
...
+  getOwner() {
+    const role = this.get('_owner');
+    if (!role || role.type !== 'roles') return;
+    return Radio.request('entities', 'roles:model', role.id);
+  },
+  saveOwner(owner) {
+    const type = 'roles';
+    return this.save({
+      _owner: {
+        id: owner.id,
```
paulfalgout: I think we need to handle if owner is null too?

paulfalgout (changes requested):

test/integration/admin/programs/sidebar/flow-sidebar.js
```diff
@@ -134,7 +134,7 @@ context('flow sidebar', function() {
       .wait('@routePostFlow')
       .its('request.body')
       .should(({ data }) => {
-        expect(data.relationships.role.data).to.be.null;
+        expect(data.relationships.owner).to.be.null;
```
paulfalgout: This needs to be `owner.data).to.be.null`

src/js/entities-service/entities/actions.js
```diff
@@ -66,45 +66,36 @@ const _Model = BaseModel.extend({
...
-  saveRole(role) {
-    return this.save({ _role: role.id, _clinician: null }, {
+  saveOwner(owner) {
+    return this.save({
+      _owner: owner,
+    },
+    {
       relationships: {
-        role: this.toRelation(role),
+        owner: owner ? this.toRelation(owner) : null,
```
paulfalgout: This should be able to be `owner: this.toRelation(owner)` if `owner` is `null` this will set owner to `{ data: null }` which is correct `owner: null` is invalid

src/js/entities-service/entities/actions.js
```diff
@@ -66,45 +66,36 @@ const _Model = BaseModel.extend({
...
-    return this.saveRole(owner);
-  },
   saveAll(attrs) {
     attrs = _.extend({}, this.attributes, attrs);
+
     const relationships = {
       forms: this.toRelation(attrs._forms, 'forms'),
-      role: this.toRelation(attrs._role, 'roles'),
-      clinician: this.toRelation(attrs._clinician, 'clinicians'),
+      owner: attrs._owner ? this.toRelation(attrs._owner) : null,
```
paulfalgout: Like above `owner: this.toRelation(attrs._owner)` should work.

src/js/entities-service/entities/actions.js
```diff
@@ -22,21 +30,13 @@ const _Model = BaseModel.extend({
...
   },
-  getRole() {
-    const roleId = this.get('_role');
-    if (!roleId) return;
-    return Radio.request('entities', 'roles:model', roleId);
-  },
   getOwner() {
-    return this.getClinician() || this.getRole();
+    const owner = this.get('_owner');
+    if (!owner || !owner.id) return;
```
paulfalgout: are both of these cases possible?  I think only `!owner` should be an issue... but I'm not 100% sure what happens with new actions.  It could be this is just supporting the `toRelation` below and won't be needed if that's fixed.  In fact I don't think actions can ever have a null owner.  program-actions can, but actions should 100% of the time have an `_owner`

src/js/entities-service/entities/flows.js
```diff
@@ -17,17 +25,10 @@ const _Model = BaseModel.extend({
...
-    const clinicianId = this.get('_clinician');
-    if (!clinicianId) return;
-    return Radio.request('entities', 'clinicians:model', clinicianId);
-  },
-  getRole() {
-    const roleId = this.get('_role');
-    if (!roleId) return;
-    return Radio.request('entities', 'roles:model', roleId);
+    const owner = this.get('_owner');
+    if (!owner || !owner.id) return;
```
paulfalgout: This should be able to match the action version

src/js/entities-service/entities/program-actions.js
```diff
@@ -22,29 +30,43 @@ const _Model = BaseModel.extend({
...
-    const action = this.pick('name', 'details', '_role', '_program', '_forms');
+    const action = this.pick('name', 'details', '_owner', '_program', '_forms');
     const dueDay = this.get('days_until_due');
     const dueDate = (dueDay === null) ? null : moment().add(dueDay, 'days').format('YYYY-MM-DD');
 
     _.extend(action, {
       _patient: patientId,
       _state: states.at(0).id,
-      _clinician: action._role ? null : currentUser.id,
+      _owner: action._owner ? action._owner : {
```
paulfalgout: could be `action._owner || {...`  though either is ok

src/js/entities-service/entities/program-actions.js
```diff
@@ -22,29 +30,43 @@ const _Model = BaseModel.extend({
...
 
     return Radio.request('entities', 'actions:model', action);
   },
-  getRole() {
-    const roleId = this.get('_role');
-    if (!roleId) return;
-    return Radio.request('entities', 'roles:model', roleId);
+  getOwner() {
+    const owner = this.get('_owner');
+    if (!owner || owner.type !== 'roles') return;
```
paulfalgout: This check seems odd to me

src/js/entities-service/entities/program-flows.js
```diff
@@ -16,10 +24,10 @@ const _Model = BaseModel.extend({
   validate({ name }) {
     if (!_.trim(name)) return 'Flow name required';
   },
-  getRole() {
-    const roleId = this.get('_role');
-    if (!roleId) return;
-    return Radio.request('entities', 'roles:model', roleId);
+  getOwner() {
+    const owner = this.get('_owner');
+    if (!owner || owner.type !== 'roles') return;
```
paulfalgout: is the type check necessary?

src/js/entities-service/entities/program-actions.js
```diff
@@ -22,29 +30,43 @@ const _Model = BaseModel.extend({
...
+    return Radio.request('entities', 'roles:model', owner.id);
   },
-  saveRole(role) {
-    return this.save({ _role: role.id }, {
+  saveOwner(owner) {
+    if (!owner) {
+      return this.save({ _owner: null }, { relationships: { owner: null } });
```
paulfalgout: I think this and program flow could be
```js
owner = this.toRelation(owner);
return this.save({ _owner:  owner.data }, {
  relationships: { owner },
});

```

src/js/entities-service/entities/program-flows.js
```diff
@@ -33,17 +41,29 @@ const _Model = BaseModel.extend({
...
+        owner: this.toRelation(_owner),
       },
     });
   },
   saveAll(attrs) {
     attrs = _.extend({}, this.attributes, attrs);
+
     const relationships = {
-      role: this.toRelation(attrs._role, 'roles'),
+      owner: attrs.owner ? this.toRelation(attrs._owner, 'roles') : null,
```
paulfalgout: null won't work here.. `toRelation` should be all that's needed.

paulfalgout (changes requested):

src/js/entities-service/entities/actions.js
LisaManresa: Oh, this explains the `expect(data.relationships.owner).to.be.null;` above. Thanks.

src/js/entities-service/entities/actions.js
LisaManresa: This was just an overly paranoid check. But yeah, there's no way for id to be empty. It's always just `owner` that is null.

src/js/entities-service/entities/program-flows.js
LisaManresa: No, it'll never be anything other than 'roles'

src/js/views/patients/patient/patient.scss
```diff
@@ -150,3 +150,7 @@
   vertical-align: middle;
   width: 34px;
 }
+
+.patient__flow-status-icon {
```
paulfalgout: Can we look at the two other icons with no button on the lists and come up with a solution that works for all of them? `.workflows__flow-status` and `.flow-action--attachment`  Also it'd be great to stick to 8px or 4px increments, so this'd need to be `padding: 0 8px;` with a 1px border that's transparent if this is the best solution for all the places.

paulfalgout (changes requested):
one small request for the flow icon related issues.

PR #224, opened 2020-02-06
Fix flow status icon on patient dashboard

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21985]

### DIFF
--- a/src/js/views/patients/patient/patient.scss
+++ b/src/js/views/patients/patient/patient.scss
@@ -150,3 +150,7 @@
   vertical-align: middle;
   width: 34px;
 }
+
+.patient__flow-status-icon {
+  padding: 0 9px;
+}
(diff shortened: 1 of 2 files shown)


sofiaroundingwell (approved):

src/js/views/patients/patient/patient.scss
```diff
@@ -150,3 +150,7 @@
   vertical-align: middle;
   width: 34px;
 }
+
+.patient__flow-status-icon {
+  padding: 0 9px;
```
paulfalgout, line 155: this should probably be `padding: 8px` and `border: 1px solid transparent;`

paulfalgout (changes requested):
Might be ok to get this in for the release, but we should readdress this.  We handle this same situation differently for workflows__flow-status and we're not handling flow-action--attachment which I imagine has the same issue.

paulfalgout:
Since this is in the other PR and already in release we can handle it over there.

PR #229, opened 2020-02-13
Formapp update button

paulfalgout:
This work includes the update button and a refactor to the forms work

- [x] Add the Clubhouse Story ID: [ch21902]

~~This is probably better in two commits, but I'm getting it up quick.  Still need to finish out the tests, but I _think_ these should pass just without coverage~~


LisaManresa (approved):

luketlancaster:
💃

PR #236, opened 2020-02-20
Move similar sidebar styles into modules/sidebar

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch22069]

trying to split this up a bit so it's easier to review


luketlancaster:
[ch22069]

paulfalgout (approved):
This seems ok, but I'll wait on merging to see if we're doing a quick release today

PR #238, opened 2020-02-21
Tweaking style for textareas

wweaver4th:
- [ ] Add QA tasks

### DIFF
--- a/src/sass/formapp.scss
+++ b/src/sass/formapp.scss
@@ -25,7 +25,7 @@ body {
 }
 .card-body {
   border: none;
-  padding: .375rem .75rem;
+  padding: 6px 12px;
 }
 .print-body {
   background-color: #fff;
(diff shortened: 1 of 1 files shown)


src/sass/formapp.scss
```diff
@@ -20,6 +20,13 @@ body {
     font-size: 18px;
   }
 }
+.bg-light {
+  background-color: #fff !important;
+}
+.card-body {
+  border: none;
+  padding: .375rem .75rem;
```
paulfalgout: why are these `rem`s instead of `6px 12px`?  Are there `rem`s elsewhere?

paulfalgout (changes requested):

src/sass/formapp.scss
wweaver4th: probably because that was what was used elesewhere. I'll change.

PR #240, opened 2020-02-26
Move similar action-item templates, to DRY things up

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch22052]

had this one sitting around for a bit, so I'll keep pushing them up in chunks instead of letting it hang out locally forever.


src/js/views/admin/program/workflows/workflows_views.js
```diff
@@ -60,6 +59,7 @@ const ActionItemView = View.extend({
   templateContext() {
     return {
       hasAttachment: this.model.getForm(),
+      newActionText: intl.admin.program.workflows.actionItemTemplate.newProgramAction,
```
paulfalgout: These keys are in the view now so they should move.. alternatively.. could set an `isFlowAction` flag and keep these in the template.  I don't know how often we set text from i18n keys in templateContext.. it's a bit more abstract.. good if we do a 3rd one I suppose.

paulfalgout (changes requested):

PR #241, opened 2020-02-27
Program flow actions drag and drop sort

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch21625]

Demoed this for Mark and he has approved.

### DIFF
--- a/src/js/views/admin/program/flow/action-item.hbs
+++ b/src/js/views/admin/program/flow/action-item.hbs
@@ -6,13 +6,16 @@
   </span>
 </td>
 <td class="table-list__cell w-60">
-  <span class="table-list__meta" data-published-region></span>{{~ remove_whitespace ~}}
-  <span class="table-list__meta" data-owner-region></span>{{~ remove_whitespace ~}}
-  <span class="table-list__meta" data-due-region></span>{{~ remove_whitespace ~}}
-  <span class="table-list__meta">
-    {{#if hasAttachment}}
-      <span class="program-flow__action-attachment">{{far "link"}}</span>
-    {{/if }}
-  </span>{{~ remove_whitespace ~}}
-  <span class="program-flow__action-ts">{{formatMoment updated_at "TIME_OR_DAY"}}</span>
+  <div class="pos--relative">
+    <span class="table-list__meta" data-published-region></span>{{~ remove_whitespace ~}}
+    <span class="table-list__meta" data-owner-region></span>{{~ remove_whitespace ~}}
+    <span class="table-list__meta" data-due-region></span>{{~ remove_whitespace ~}}
+    <span class="table-list__meta">
+      {{#if hasAttachment}}
+        <span class="program-flow__action-attachment">{{far "link"}}</span>
+      {{/if }}
+    </span>{{~ remove_whitespace ~}}
+    <span class="program-flow__action-ts">{{formatMoment updated_at "TIME_OR_DAY"}}</span>
+    <span class="program-flow__sort-handle js-sort">{{fas "grip-vertical"}}</span>{{~ remove_whitespace ~}}
+  </div>
 </td>
--- a/src/js/views/admin/program/flow/flow_views.js
+++ b/src/js/views/admin/program/flow/flow_views.js
@@ -5,9 +5,10 @@ import { View, CollectionView } from 'marionette';
 import PreloadRegion from 'js/regions/preload_region';
 
 import { DueDayComponent, OwnerComponent, PublishedComponent } from 'js/views/admin/actions/actions_views';
+import SortableList from 'js/behaviors/sortable-list';
 
-import HeaderTemplate from './header.hbs';
 import ActionItemTemplate from './action-item.hbs';
+import HeaderTemplate from './header.hbs';
 
 import './program-flow.scss';
 
@@ -109,9 +110,10 @@ const ActionItemView = View.extend({
     'change:id': 'render',
   },
   className() {
-    if (this.model.isNew()) return 'table-list__item is-selected';
+    const className = 'table-list__item program-flow__action-item js-draggable';
+    if (this.model.isNew()) return `${ className } is-selected`;
 
-    return 'table-list__item';
+    return className;
   },
   initialize() {
     this.action = this.model.getAction();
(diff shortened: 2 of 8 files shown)


src/js/views/admin/program/flow/program-flow.scss
```diff
@@ -98,6 +98,95 @@
...
+  .program-flow__action-sort {
+    display: inline-block;
+  }
+
+  .program-flow__action-ts {
+    display: none;
+  }
+}
+
+.program-flow__list.is-dragging {
```
LisaManresa: I'm not thrilled about this, there's a lot of overriding going on. But SortableJS doesn't play nice with list items that have hover states. The hover styles will remain applied to whatever item fills the space where the dragged item used to be, creating a confusing UI that makes it difficult to discern where the dragged item will be dropped.

The solution was to add a class to the entire list when an item is being dragged and override the hover styles.

src/js/views/admin/program/flow/action-item.hbs
```diff
@@ -0,0 +1,21 @@
+<td class="table-list__cell w-40">
```
LisaManresa: This might not be ideal, given we are trying to dedupe our codebase. But this template contains a wrapping div within the meta `td` so that we can absolutely position the added `program-flow__action-sort` element. We could conditionally add both of those things to the generic `action-item.hbs` using an `isSortable` flag. @paulfalgout do you have a preference?

src/js/views/admin/program/flow/flow_views.js
```diff
@@ -1,17 +1,43 @@
...
-import { DueDayComponent, OwnerComponent, PublishedComponent, ActionItemTemplate } from 'js/views/admin/actions/actions_views';
+import { DueDayComponent, OwnerComponent, PublishedComponent } from 'js/views/admin/actions/actions_views';
 
+import ActionItemTemplate from './action-item.hbs';
 import HeaderTemplate from './header.hbs';
 
 import './program-flow.scss';
 
+
+const SortableCollection = Behavior.extend({
```
paulfalgout: This behavior could be generic in the behaviors directory outside of the `this.view.collection.updateSequences();` line.  That line could be added to the view's `onDragEnd`

src/js/views/admin/program/flow/flow_views.js
```diff
@@ -1,17 +1,43 @@
...
+  onRender() {
```
paulfalgout: also what if we added
```
options: {
  handle: '.js-sort',
  item: '.js-draggable',
}
```
to the behavior and then if we need to override them we can, but they're sensible defaults.

src/js/views/admin/program/flow/flow_views.js
```diff
@@ -221,6 +255,7 @@ const LayoutView = View.extend({
       regionClass: PreloadRegion,
       replaceElement: true,
     },
+    testList: '[data-test-list-region]',
```
paulfalgout: leftover

src/js/views/admin/program/flow/action-item.hbs
paulfalgout: hmm bummer.  I think we should revert and/or readdress https://github.com/RoundingWell/care-ops-frontend/pull/240  @luketlancaster   There might be some stuff to keep from it, but looking at it again I let some over-generic things through as well.  `action-icon` is too generic since we have program and patient actions.. it it's something that _could_ be that generic it's a `/domain` style.  There may be some compromise in general such that this would work though.. more in other comments

src/js/views/admin/program/flow/program-flow.scss
```diff
@@ -98,6 +98,60 @@
   border-top: $black-90 1px solid;
 }
 
+.program-flow__action-meta {
```
paulfalgout: not sure what this is

src/js/views/admin/program/flow/program-flow.scss
```diff
@@ -98,6 +98,60 @@
...
+    color: $black-60;
+    font-size: 20px;
+    vertical-align: middle;
+    width: 20px;
+  }
+}
+
+.table-list__item:hover {
+  .program-flow__action-sort {
+    align-items: center;
```
paulfalgout: does this class need more than display?  what happens if these two centers move to the non-hover class?

src/js/views/admin/program/flow/flow_views.js
```diff
@@ -110,9 +136,10 @@ const ActionItemView = View.extend({
     'change:id': 'render',
   },
   className() {
-    if (this.model.isNew()) return 'table-list__item is-selected';
+    const className = 'table-list__item table-list__item--draggable js-draggable';
```
paulfalgout: `.table-list__item--draggable` doesn't seem to be in use

src/js/views/admin/program/flow/program-flow.scss
paulfalgout: `table-list__` classes shouldn't show up in this file at all.  Likely the only thing that should be in this file is some sort of `program-flow__list-item` which would handle hiding the `program-action_ts` on hover.
Then maybe all of the sort stuff can be made generic and we can move it to the `table-list` styles.  Hopefully any other table list _would_ do this the same way, so it seems like it would make sense to put it there.

paulfalgout (changes requested):

src/js/views/admin/program/flow/action-item.hbs
luketlancaster: alrighty, put up #242 to revert #240, in the interest of getting this finished quicker. We can readdress that down the line once this has settled

src/js/views/admin/program/flow/program-flow.scss
LisaManresa: Extraneous from the original design. Meant to remove it.

src/js/views/admin/program/flow/program-flow.scss
```diff
@@ -133,6 +133,37 @@
...
+  }
+
+  .program-flow__action-ts {
+    display: none;
+  }
+}
+
+.is-dragging {
+  .program-flow__action-item:not(.sortable-ghost) {
+    .program-flow__sort-handle {
```
LisaManresa, line 161: I moved most of the dragging styles into table-list.scss. Opted to leave the handle display styles in the program-flow in case other lists don't have the handle displayed on hover. But it should be doable to move all the handle styles into table-list.scss and use a generic `sort--handle` classname if that's appropriate.

LisaManresa:
Apologies for all the noise on this PR. I realized merging the rollback didn't go as smoothly as I thought and keep finding classnames here and there that are incorrect.

PR #246, opened 2020-03-04
Worklist toggle

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22088]

### DIFF
--- a/src/js/apps/patients/patients-main_app.js
+++ b/src/js/apps/patients/patients-main_app.js
@@ -22,7 +22,7 @@ export default RouterApp.extend({
   },
 
   defaultRoute() {
-    const defaultRoute = 'worklist';
+    const defaultRoute = 'worklist:flows';
     const defaultWorklist = 'owned-by-me';
 
     this.routeAction(defaultRoute, () => {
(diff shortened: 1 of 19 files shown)


src/js/views/patients/worklist/worklist_views.js
```diff
@@ -124,19 +125,84 @@ const ItemView = View.extend({
...
+            {{ @intl.patients.worklist.worklistViews.layoutView.actionHeader }}
+          {{/if}}
+        </td>
+        <td class="table-list__header w-40">
+          {{#if isFlowList}}
+            {{ @intl.patients.worklist.worklistViews.layoutView.flowAttrHeader }}
+          {{ else }}
+            {{ @intl.patients.worklist.worklistViews.layoutView.actionAttrHeader }}
+          {{/if}}
+        </td>
```
luketlancaster: wonder if this template is getting long/complicated enough to pull out into it's own file?

luketlancaster (approved):
looks fine w/me. I'm sure @paulfalgout will have thoughts though 😁

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -63,10 +63,13 @@ export default App.extend({
     'change': 'onChangeState',
   },
   onChangeState(state) {
-    if (state.hasChanged('groupId')) {
+    if (state.hasChanged('groupId') || state.hasChanged('worklistType')) {
```
paulfalgout: I don't think we should store the type on the state.  we're going to want to store the state model in the future for recalling previous sort/filter and so the type really only needs to come from the url and can probably just be an option like the id

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -76,9 +79,14 @@ export default App.extend({
 
     this.currentClinician = Radio.request('bootstrap', 'currentUser');
     this.groups = this.currentClinician.getGroups();
-    this.setState({ 'sortId': 'sortUpdateDesc' });
+    this.setState({
+      'sortId': 'sortUpdateDesc',
```
paulfalgout: doesn't need `'`s

src/js/apps/patients/patients-main_app.js
```diff
@@ -40,9 +41,13 @@ export default RouterApp.extend({
       action: 'defaultRoute',
       route: '',
     },
-    'worklist': {
+    'worklistNotFound': {
```
paulfalgout: Is this needed?

src/js/apps/patients/patients-main_app.js
```diff
@@ -84,9 +89,14 @@ export default RouterApp.extend({
       'done-last-thirty-days': 'lastThirty',
     };
 
-    if (!worklistsById[worklistId]) Radio.trigger('event-router', 'notFound');
+    if (!worklistsById[worklistId] || !worklistType) Radio.trigger('event-router', 'notFound');
```
paulfalgout: should there be a route that doesn't return !worklistType?  I could see the type being wrong.. which maybe that should 404.

src/js/views/patients/worklist/worklist_views.js
LisaManresa: I'm kind of leaning that way as well.

paulfalgout (changes requested):

src/js/apps/patients/patients-main_app.js
LisaManresa: This is paired with the `worklistNotFound` event above. Basically, it's accounting for a url like `worklist/owned-by-me`, but now I'm thinking that's overkill. I could remove that logic and add a check for `worklistType` equaling either `flows` or `actions`.

paulfalgout: Yeah I just think a route like `worklistNotFound` is never needed.

src/js/apps/patients/patients-main_app.js
```diff
@@ -75,18 +76,19 @@ export default RouterApp.extend({
...
       'owned-by-me': 'ownedByMe',
       'for-my-role': 'forMyRole',
       'new-past-day': 'newPastDay',
       'updated-past-three-days': 'pastThree',
       'done-last-thirty-days': 'lastThirty',
     };
+    const types = ['actions', 'flows'];
 
-    if (!worklistsById[worklistId]) Radio.trigger('event-router', 'notFound');
+    if (!worklistsById[worklistId] || !types.includes(worklistType)) Radio.trigger('event-router', 'notFound');
```
paulfalgout: do we have a polyfill for includes for IE11... just thought of another thing.. above it could be `route: ['worklist/:id/actions', 'worklist/:id/flows']`  and then there's no need to test it.

paulfalgout (changes requested):

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: @LisaManresa

src/js/views/patients/worklist/worklist_views.js
paulfalgout: @LisaManresa this is a good idea

LisaManresa: Sorry, I missed this change. It's been pushed up now.

src/js/views/patients/worklist/layout.hbs
```diff
@@ -0,0 +1,30 @@
+<div class="list-page__header">
+  <div class="list-page__title">{{formatMessage (intlGet "patients.worklist.worklistViews.listTitles") title=worklistId role=role}}<span class="list-page__header-icon js-title-info">{{fas "info-circle"}}</span></div>
```
paulfalgout: Need to update these to `patients.worklist.layoutTemplate.X`

paulfalgout (changes requested):

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -30,7 +32,7 @@ const EmptyView = View.extend({
   `,
 });
```
luketlancaster, line 42: we'll need to update the `EmptyView` too to display the proper text (No Actions vs No Flows)

LisaManresa: Ah thank you. That was on my radar then slipped my mind

PR #248, opened 2020-03-09
Save worklist filter/sort state in localstorage

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22091]

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -1,5 +1,6 @@
 import _ from 'underscore';
 import moment from 'moment';
+import store from 'store';
 
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
@@ -12,6 +13,10 @@ import App from 'js/base/app';
 import { ListView, LayoutView, GroupsDropList, SortDropList } from 'js/views/patients/worklist/worklist_views';
 
 const i18n = intl.patients.worklist.worklistApp;
+const DEFAULT_STATE = {
+  sortId: 'sortUpdateDesc',
+  groupId: null,
+};
 
 const sortOptions = new Backbone.Collection([
   {
@@ -63,6 +68,7 @@ export default App.extend({
     'change': 'onChangeState',
   },
   onChangeState(state) {
+    store.set(this.stateId, state);
     if (state.hasChanged('groupId')) {
       this.restart();
     }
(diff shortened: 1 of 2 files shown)


src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -63,27 +68,42 @@ export default App.extend({
...
+    }
+  },
+  onBeforeStart(options) {
     if (this.isRestarting()) {
       this.getRegion('list').startPreloader();
       return;
     }
-    this.worklistId = worklistId;
-    this.worklistType = worklistType;
+
```
paulfalgout, line 94: I think what was here was fine without options unless there's a reason `initListState` shouldn't use `this.worklistId`?

paulfalgout (changes requested):
small nit/question.. otherwise :shipit:

src/js/apps/patients/worklist/worklist_app.js
LisaManresa: Nope, no reason.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -63,27 +68,42 @@ export default App.extend({
...
+    store.set(this.stateId, state);
     if (state.hasChanged('groupId')) {
       this.restart();
     }
   },
   viewEvents: {
     'toggle:listType': 'onToggleListType',
   },
+  initListState() {
+    if (!this.stateId) {
```
paulfalgout: what's the use case for the else here?  when is `stateId` defined by this point?

paulfalgout (changes requested):
had one other thought

src/js/apps/patients/worklist/worklist_app.js
LisaManresa: When the app is restarted, `stateId` remains defined, but `worklistId` and `worklistType` end up undefined in `onBeforeStart` because options is undefined. So if we don't have that check, `stateId` gets overwritten with undefined values.

paulfalgout: if the app is restarted I don't believe this function is called at all.

LisaManresa: Oh geez. Sorry I see what happened. Originally that `stateId` logic was in `onBeforeStart` above the `isRestarting` check. You're right, we don't need that anymore

paulfalgout (approved):

PR #252, opened 2020-03-12
Add flow progress bar to worklists and flow page

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch22044]
- [ ] Add QA tasks

### DIFF
--- a/src/js/apps/patients/patient/flow/flow_app.js
+++ b/src/js/apps/patients/patient/flow/flow_app.js
@@ -20,7 +20,6 @@ export default SubRouterApp.extend({
   },
   beforeStart({ flowId }) {
     return [
-      Radio.request('entities', 'fetch:patients:model:byFlow', flowId),
       Radio.request('entities', 'fetch:flows:model', flowId),
       Radio.request('entities', 'fetch:actions:collection:byFlow', flowId),
     ];
@@ -29,14 +28,12 @@ export default SubRouterApp.extend({
     Radio.trigger('event-router', 'notFound');
     this.stop();
   },
-  onStart({ currentRoute }, [patient], [flow], [actions]) {
-    this.patient = patient;
+  onStart({ currentRoute }, [flow], [actions]) {
     this.flow = flow;
     this.actions = actions;
 
     this.showChildView('contextTrail', new ContextTrailView({
       model: this.flow,
-      patient: this.patient,
     }));
 
     this.showHeader();
@@ -92,7 +89,7 @@ export default SubRouterApp.extend({
 
   showSidebar() {
     this.showChildView('sidebar', new SidebarView({
-      model: this.patient,
+      model: this.flow.getPatient(),
     }));
   },
 
--- a/src/js/apps/patients/patient/flow/flow_app.js
+++ b/src/js/apps/patients/patient/flow/flow_app.js
@@ -44,35 +44,33 @@ export default SubRouterApp.extend({
     this.showSidebar();
 
     this.listenTo(this.actions, {
-      'change:_state': this.onActionsChangeState,
-      'destroy': this.onActionsDestroy,
+      'change:_state': this.onActionChangeState,
+      'destroy': this.onActionDestroy,
     });
   },
-  onActionsChangeState(action) {
+  onActionChangeState(action) {
     const { complete, total } = this.flow.get('_progress');
-    const currentOrg = Radio.request('bootstrap', 'currentOrg');
-    const states = currentOrg.getStates();
-    const current = action.get('_state');
-    const previous = action.previous('_state');
-    const doneStateId = states.find({ status: 'done' }).id;
+    const isDone = action.isDone();
+
+    const prevState = Radio.request('entities', 'states:model', action.previous('_state'));
+    const isPrevDone = prevState.get('status') === 'done';
 
     // No change in completion
-    if (current !== doneStateId && previous !== doneStateId) return;
+    if (!isPrevDone && !isDone) return;
 
-    const isNewComplete = current === doneStateId;
-    this.flow.set({ _progress: { complete: complete + (isNewComplete ? 1 : -1), total } });
+    this.flow.set({ _progress: {
+      complete: complete + (isDone ? 1 : -1),
+      total,
+    } });
   },
-  onActionsDestroy(action) {
-    const progress = this.flow.get('_progress');
-    let complete = progress.complete;
-    const currentOrg = Radio.request('bootstrap', 'currentOrg');
-    const states = currentOrg.getStates();
-    const current = action.get('_state');
-    const doneStateId = states.find({ status: 'done' }).id;
+  onActionDestroy(action) {
+    const { complete, total } = this.flow.get('_progress');
+    const isDone = action.isDone();
 
-    if (current === doneStateId) complete -= 1;
-
-    this.flow.set({ _progress: { complete: complete, total: progress.total - 1 } });
+    this.flow.set({ _progress: {
+      complete: complete - (isDone ? 1 : 0),
+      total: total - 1,
+    } });
   },
   showHeader() {
     const headerView = new HeaderView({
--- a/src/js/apps/patients/patient/flow/flow_app.js
+++ b/src/js/apps/patients/patient/flow/flow_app.js
@@ -42,8 +42,38 @@ export default SubRouterApp.extend({
     this.showHeader();
     this.showActionList();
     this.showSidebar();
-  },
 
+    this.listenTo(this.actions, {
+      'change:_state': this.onActionsChangeState,
+      'destroy': this.onActionsDestroy,
+    });
+  },
+  onActionsChangeState(action) {
+    const { complete, total } = this.flow.get('_progress');
+    const currentOrg = Radio.request('bootstrap', 'currentOrg');
+    const states = currentOrg.getStates();
+    const current = action.get('_state');
+    const previous = action.previous('_state');
+    const doneStateId = states.find({ status: 'done' }).id;
+
+    // No change in completion
+    if (current !== doneStateId && previous !== doneStateId) return;
+
+    const isNewComplete = current === doneStateId;
+    this.flow.set({ _progress: { complete: complete + (isNewComplete ? 1 : -1), total } });
+  },
+  onActionsDestroy(action) {
+    const progress = this.flow.get('_progress');
+    let complete = progress.complete;
+    const currentOrg = Radio.request('bootstrap', 'currentOrg');
+    const states = currentOrg.getStates();
+    const current = action.get('_state');
+    const doneStateId = states.find({ status: 'done' }).id;
+
+    if (current === doneStateId) complete -= 1;
+
+    this.flow.set({ _progress: { complete: complete, total: progress.total - 1 } });
+  },
   showHeader() {
     const headerView = new HeaderView({
       model: this.flow,
--- a/src/js/base/jsonapi-mixin.js
+++ b/src/js/base/jsonapi-mixin.js
@@ -45,6 +45,10 @@ export default {
 
     modelData.__cached_ts = moment.utc().format();
 
+    _.each(data.meta, (value, key) => {
+      modelData[`_${ _.underscored(key) }`] = value;
+    });
+
     return this.parseRelationships(modelData, data.relationships);
   },
   toRelation(entity, entityType) {
--- a/src/js/base/jsonapi-mixin.js
+++ b/src/js/base/jsonapi-mixin.js
@@ -27,7 +27,11 @@ export default {
     return _.map(relationship, item => {
       const itemRelationship = { id: item.id };
 
-      if (item.meta) itemRelationship._meta = item.meta;
+      if (item.meta) {
+        _.each(item.meta, (value, key) => {
+          itemRelationship[`_${ _.underscored(key) }`] = value;
+        });
+      }
 
       return itemRelationship;
     });
--- a/src/js/views/patients/patient/dashboard/dashboard_views.js
+++ b/src/js/views/patients/patient/dashboard/dashboard_views.js
@@ -156,9 +156,7 @@ const FlowItemView = View.extend({
   },
   template: FlowItemTemplate,
   templateContext() {
-    const currentOrg = Radio.request('bootstrap', 'currentOrg');
-    const states = currentOrg.getStates();
-    const status = states.get(this.model.get('_state')).get('status');
+    const status = this.model.getState().get('status');
 
     return {
       statusClass: _.dasherize(status),
--- a/src/js/views/patients/patient/dashboard/dashboard_views.js
+++ b/src/js/views/patients/patient/dashboard/dashboard_views.js
@@ -9,6 +9,7 @@ import { View, CollectionView, Behavior } from 'marionette';
 import { PatientStatusIcons } from 'js/static';
 
 import 'sass/modules/buttons.scss';
+import 'sass/modules/progress-bar.scss';
 import 'sass/modules/table-list.scss';
 
 import PreloadRegion from 'js/regions/preload_region';
--- a/src/js/views/patients/patient/dashboard/flow-item.hbs
+++ b/src/js/views/patients/patient/dashboard/flow-item.hbs
@@ -7,5 +7,6 @@
     <span class="patient__flow-status action--{{ statusClass }}">{{fas statusIcon}}</span>{{~ remove_whitespace ~}}
   </span>{{~ remove_whitespace ~}}
   <span class="table-list__meta" data-owner-region></span>{{~ remove_whitespace ~}}
+  <progress class="progress-bar patient-flow__progress" value={{ _progress.complete }} max={{ _progress.total }}></progress>{{~ remove_whitespace ~}}
   <span class="patient__action-ts">{{formatMoment updated_at "TIME_OR_DAY"}}</span>
 </td>
--- a/src/js/views/patients/patient/data-events/data-events_views.js
+++ b/src/js/views/patients/patient/data-events/data-events_views.js
@@ -4,6 +4,7 @@ import Radio from 'backbone.radio';
 import hbs from 'handlebars-inline-precompile';
 import { View, CollectionView, Behavior } from 'marionette';
 
+import 'sass/modules/progress-bar.scss';
 import 'sass/modules/table-list.scss';
 
 import PreloadRegion from 'js/regions/preload_region';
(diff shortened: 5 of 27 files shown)


src/js/views/patients/patient/dashboard/flow-item.hbs
```diff
@@ -7,5 +7,8 @@
     <span class="patient__flow-status action--{{ statusClass }}">{{fas statusIcon}}</span>{{~ remove_whitespace ~}}
   </span>{{~ remove_whitespace ~}}
   <span class="table-list__meta" data-owner-region></span>{{~ remove_whitespace ~}}
+  <span class="patient-flow__progress">
+    <div class="patient-flow__progress-outer"><div class="patient-flow__progress-inside"></div></div>
+  </span>{{~ remove_whitespace ~}}
```
luketlancaster: whoops forgot to fix these guys

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -53,6 +53,35 @@ export default SubRouterApp.extend({
       'edit': this.onEditFlow,
     });
 
+    const actions = this.flow.getActions();
```
paulfalgout: there's a `this.actions` already

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -53,6 +53,35 @@ export default SubRouterApp.extend({
...
+    const currentOrg = Radio.request('bootstrap', 'currentOrg');
+    const states = currentOrg.getStates();
+
+    this.listenTo(actions, {
+      'change:_state'(action) {
+        const progress = this.flow.get('_progress');
+        const previous = states.get(action._previousAttributes._state);
```
paulfalgout: I think `action.previous('_state')` would work here.

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -53,6 +53,35 @@ export default SubRouterApp.extend({
...
+        const current = states.get(action.get('_state'));
+
+        if (current.get('status') === 'done') {
```
paulfalgout: what might be easier is getting the `doneStateId` and comparing that rather than `states.get` and `get('status')` everywhere.

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -53,6 +53,35 @@ export default SubRouterApp.extend({
...
+          this.flow.set({ _progress: { complete: progress.complete + 1, total: progress.total } });
+        }
+
+        if (current.get('status') !== 'done' && previous.get('status') === 'done') {
```
paulfalgout: assuming we're using ids we can probably do `if (currentId === previousId) return;` at the top.  then complete is add if done and decremented otherwise

Could even do something like `const completeChange = (currentId === doneStateId) ? 1 : -1;` with a single `set` and `+ completeChange'

src/js/apps/patients/patient/flow/flow_app.js
paulfalgout: Also I think this could move into a function in `onStart`  The header displays the results, but it's not part of showing the header really.  Also it won't be as clean with the states, but it's probably a good idea to break this up into
```
this.listenTo(this.actions, {
  'change:_state': this.onActionsChangeState,
  'destroy': this.onActionsDestroy,
}
```
in which case that `listenTo` may not need its own function.

src/js/base/jsonapi-mixin.js
```diff
@@ -45,6 +45,10 @@ export default {
 
     modelData.__cached_ts = moment.utc().format();
 
+    _.each(data.meta, (value, key) => {
```
paulfalgout, line 52: I'm fine with this, but we should handle all `meta` the same way.. so line 30 would need to change as well.. plus anywhere that used that `_meta` within the app.

src/js/views/patients/patient/dashboard/dashboard_views.js
```diff
@@ -183,6 +184,14 @@ const FlowItemView = View.extend({
 
     this.showChildView('owner', ownerComponent);
   },
+  setProgress() {
+    const prog = this.model.get('_progress');
+    const percentage = (prog.complete / prog.total) * 100;
+
+    this.$el.find('.worklist-flow__progress-inside').css({
```
paulfalgout: this should be `ui` element and `.js-` so that this is `this.ui.progress.css`

paulfalgout: although.. I don't think this is a thing..

src/js/views/patients/patient/flow/flow_views.js
```diff
@@ -91,6 +92,19 @@ const HeaderView = View.extend({
 
     this.showChildView('owner', ownerComponent);
   },
+  setProgress() {
+    const prog = this.model.get('_progress');
+    const bar = this.$el.find('.patient-flow__header__progress');
```
paulfalgout: should be a `ui` element and `.js-` selector  `this.ui.bar`

src/js/views/patients/patient/flow/flow_views.js
```diff
@@ -91,6 +92,19 @@ const HeaderView = View.extend({
...
+
+    bar.attr({ value: prog.complete, max: prog.total });
+
+    this.listenTo(this.model, {
```
paulfalgout: this `listenTo` shouldn't be here.  right now it's adding a listener for every view render.  Plus it's doing 100% of what's above.  this could safely be done in `initialize` but in this case all you need to do is add `'change:_progress': 'setProgress',` to the `modelEvents`

src/js/views/patients/patient/flow/patient-flow.scss
```diff
@@ -106,3 +106,33 @@
...
   text-align: center;
 }
+
+.patient-flow__header__progress {
+  background: #E6E6E6;
+  border: 0;
+  border-radius: 16px;
+  display: inline-block;
+  height: 16px;
+  margin-top: 0.75em;
```
paulfalgout: this was just for jsfiddle testing. margin should be 4px increments if needed at all.

src/js/views/patients/patient/flow/patient-flow.scss
```diff
@@ -106,3 +106,33 @@
   padding: 20px;
   text-align: center;
 }
+
+.patient-flow__header__progress {
+  background: #E6E6E6;
```
paulfalgout: $black-90

src/js/views/patients/patient/flow/patient-flow.scss
```diff
@@ -106,3 +106,33 @@
...
+  margin-top: 0.75em;
+  width: 144px;
+}
+
+.patient-flow__header__progress::-webkit-progress-bar {
+  background: transparent;
+}
+
+.patient-flow__header__progress::-moz-progress-bar {
+  background: #1dbe6c;
```
paulfalgout: this also should likely be one of our green colors in sass and not hard-coded here.

src/js/views/patients/patient/flow/patient-flow.scss
```diff
@@ -106,3 +106,33 @@
...
+  border-radius: 32px;
```
paulfalgout: these should probably be 16px

src/js/views/patients/patient/patient.scss
```diff
@@ -151,3 +151,33 @@
   padding: 0 8px;
   vertical-align: middle;
 }
+
+.patient-flow__progress {
```
paulfalgout, line 155: I think the progress bar styles can be moved into a `/module` with the exception of `width` `margin-top` or any other thing that is likely context specific.

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -182,6 +182,11 @@ const FlowItemView = View.extend({
 
     this.showChildView('owner', ownerComponent);
   },
+  setProgress() {
+    const progress = this.model.get('_progress');
+
+    this.$el.find('.worklist-flow__progress').attr({ value: progress.complete, max: progress.total });
```
paulfalgout: if these values aren't going to change we might as well put them in the template.. for that matter they should always be put in the template, and this function in the other views should just be called `onChangeFlowState`

paulfalgout (changes requested):

src/js/apps/patients/patient/flow/flow_app.js
luketlancaster: so `currentId` will never equal `previousId`, because you can't "change" a state that way...what I'm checking for here is the fact that you can change from To Do to Doing back to To Do, and that'll keep decrementing because current isn't done, and it isn't previous either. That make sense?

src/js/views/patients/patient/dashboard/dashboard_views.js
luketlancaster: oh i forgot to change this to look for `<progress>` instead of some nested `<div>`s

src/js/apps/patients/patient/flow/flow_app.js
paulfalgout: ah wouldn't that require both of these to be `!==`?

luketlancaster: don't think so? If the current state isn't done, we want to decrement. But we only want to decrement if we're coming from a done state.

paulfalgout: ah yeah I was thinking in terms of my original suggestion.. something like this:
```js
onChangeState(action) {
  const { complete, total } = this.flow.get('_progress');
  const current = action.get('_state');
  const previous = action.previous('_state');

  // No change in completion
  if(current !== doneStateId && previous !== doneStateId) return;

  const isNewComplete = current === doneStateId;
  this.flow.set({ _progress: { complete: complete + (isNewComplete ? 1  : -1), total } });
}
```

luketlancaster: ooh that's way better than what I have...

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -42,8 +42,38 @@ export default SubRouterApp.extend({
...
+      'destroy': this.onActionsDestroy,
+    });
   },
+  onActionsChangeState(action) {
+    const { complete, total } = this.flow.get('_progress');
+    const current = action.get('_state');
+    const previous = action.previous('_state');
+    const currentOrg = Radio.request('bootstrap', 'currentOrg');
+    const states = currentOrg.getStates();
+    const doneStateId = states.findWhere({ name: 'Done' }).id;
```
paulfalgout: `find({ status: 'done' })` I think would be better.  `findWhere` is essentially a less useful alias of `find` and `status` is better than `name` since that'd change with i18n

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -42,8 +42,38 @@ export default SubRouterApp.extend({
     this.showHeader();
     this.showActionList();
     this.showSidebar();
+
+    this.listenTo(this.actions, {
+      'change:_state': this.onActionsChangeState,
+      'destroy': this.onActionsDestroy,
+    });
   },
+  onActionsChangeState(action) {
```
paulfalgout: let's make this function and `onActionsDestroy` more similar by using the same var names and orders where possible.

src/js/views/patients/patient/dashboard/flow-item.hbs
```diff
@@ -7,5 +7,6 @@
     <span class="patient__flow-status action--{{ statusClass }}">{{fas statusIcon}}</span>{{~ remove_whitespace ~}}
   </span>{{~ remove_whitespace ~}}
   <span class="table-list__meta" data-owner-region></span>{{~ remove_whitespace ~}}
+  <progress class="js-progress progress-bar patient-flow__progress" value={{ progressValue }} max={{ progressMax }}></progress>{{~ remove_whitespace ~}}
```
paulfalgout: you don't need `templateContext` here.  `{{ _progress.complete }}` should work.

src/js/views/patients/patient/data-events/data-events_views.js
```diff
@@ -129,9 +130,19 @@ const FlowItemView = View.extend({
...
+    const progress = this.model.get('_progress');
+    return {
+      progressValue: progress.complete,
+      progressMax: progress.total,
+    };
+  },
   triggers: {
     'click': 'click',
   },
+  ui: {
```
paulfalgout: only need to add the ui and the `.js-` selector to the template if it is used

src/js/views/patients/patient/flow/flow_views.js
```diff
@@ -53,18 +54,29 @@ const HeaderView = View.extend({
   modelEvents: {
     'editing': 'onEditing',
     'change': 'render',
+    'change:_progress': 'onChangeFlowState',
```
paulfalgout: should probably be `onChangeFlowProgress`

src/sass/modules/progress-bar.scss
```diff
@@ -0,0 +1,28 @@
...
+}
+
+.progress-bar::-moz-progress-bar {
+  background: color(green);
+  border-radius: 16px;
+}
+
+.progress-bar::-webkit-progress-value {
+  background: color(green);
+  border-radius: 32px;
```
paulfalgout: can this one also be 16 I assume?

test/integration/patients/patient/patient-flow.js
```diff
@@ -361,4 +361,158 @@ context('patient flow page', function() {
...
       .contains('No Actions');
   });
+
+  specify('flow progress bar', function() {
+    cy
+      .server()
+      .routeFlow(fx => {
+        fx.data.id = '1';
+
+        fx.data.attributes.name = 'Test Flow';
```
paulfalgout: I don't think we need to change the flow for this test?  Let's only add test data if we're going to use and rely on whatever is in the fixtures otherwise.

test/integration/patients/patient/patient-flow.js
```diff
@@ -361,4 +361,158 @@ context('patient flow page', function() {
...
+        fx.data[1].relationships.patient.data.id = '1';
+        fx.data[1].relationships.state.data.id = '22222';
+
+        fx.data[2].id = '3';
+        fx.data[2].relationships.patient.data.id = '1';
+        fx.data[2].relationships.state.data.id = '22222';
+
+        return fx;
+      }, '1')
+      .routePatientByFlow(fx => {
```
paulfalgout: I assume this could just be `routePatientByFlow()`?

paulfalgout (changes requested):

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -42,8 +42,38 @@ export default SubRouterApp.extend({
...
 
+    // No change in completion
+    if (current !== doneStateId && previous !== doneStateId) return;
+
+    const isNewComplete = current === doneStateId;
+    this.flow.set({ _progress: { complete: complete + (isNewComplete ? 1 : -1), total } });
+  },
+  onActionsDestroy(action) {
+    const progress = this.flow.get('_progress');
+    let complete = progress.complete;
```
paulfalgout: actually I'm rethinking this.  since we're not concerned with previous this'd be better with `const isDone = action.isDone();`  with a `complete + (isDone ? -1 : 1);` like above

test/integration/patients/patient/patient-flow.js
```diff
@@ -361,4 +361,138 @@ context('patient flow page', function() {
...
+      .routeProgramByAction()
+      .visit('/flow/1')
+      .wait('@routeFlow')
+      .wait('@routeFlowActions')
+      .wait('@routePatientByFlow');
+
+    cy
+      .get('.patient-flow__header__progress')
+      .should('have.value', 0)
+      .should($progress => {
```
paulfalgout: I _think_ this can be something like `should('have.attr', 'max', 3)`

test/integration/patients/patient/patient-flow.js
```diff
@@ -361,4 +361,138 @@ context('patient flow page', function() {
...
+
+    cy
+      .get('.picklist')
+      .contains('Delete Action')
+      .click();
+
+    cy
+      .get('.patient-flow__header__progress')
+      .should('have.value', 1)
+      .should($progress => {
```
paulfalgout: same has above..`have.attr` should work.

test/integration/patients/worklist.js
```diff
@@ -133,6 +145,18 @@ context('worklist page', function() {
...
 
+    cy
+      .get('@firstRow')
+      .find('.worklist-flow__progress')
+      .should('have.value', 0);
+
+    cy
+      .get('@firstRow')
+      .find('.worklist-flow__progress')
+      .should($progress => {
```
paulfalgout: same above

paulfalgout (changes requested):

paulfalgout (changes requested):
One more thing.. pull in `develop`
Let's change ~~these two lines~~ to use
~~`this.isDoneList = flow.isAllDone();`~~
`const isAllComplete = flow.isAllDone();`
and `isAllDone` can test if both `_progress` values are equal.
Then we shouldn't need `flow.getActions();`

paulfalgout:
didn't realize the PTO.  I'll pick this up

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -322,10 +298,6 @@ const ListView = CollectionView.extend({
     if (!this.isAttached()) return;
     this.triggerMethod('update:listDom', this);
   },
-  viewFilter({ model }) {
```
paulfalgout, line 325: @LisaManresa I'm walking this back.  We don't current filter the worklists after changes.. only doing that on patient dashboards.  Any reason to introduce it here to only flows?

LisaManresa: This was in place to ensure that when an item's state is changed from `done` it is visually removed from the list (on the "done < 30 days" list only). But perhaps the better approach would be to remove them from the collection in the list view on `change:visible`.

paulfalgout: we don't currently visually remove anything from any worklist if ~~it's status~~ whatever is putting it on that particular list changes

LisaManresa: Ah ok. Then yeah, this can go.

PR #253, opened 2020-03-12
Make flow state read only on worklists

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22136]

### DIFF
--- a/src/js/views/patients/worklist/worklist_views.js
+++ b/src/js/views/patients/worklist/worklist_views.js
@@ -1,3 +1,4 @@
+import anime from 'animejs';
 import _ from 'underscore';
 
 import Radio from 'backbone.radio';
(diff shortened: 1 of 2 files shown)


src/js/views/patients/worklist/worklist_views.js
```diff
@@ -163,7 +185,27 @@ const FlowItemView = View.extend({
     this.showState();
     this.showOwner();
   },
+  onChangeState() {
+    if (!this.model.isDone()) {
```
paulfalgout: I don't think the `if` is necessary?  won't it always be true

paulfalgout (changes requested):

paulfalgout (approved):

PR #255, opened 2020-03-17
Clinicians list

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22140]

There is some redundancy between the RoleComponent and the OwnerComponent. Perhaps we can share the same component, but that may not be the right approach.

Screen Shot 2020-03-17 at 12 17 59 PM

### DIFF
--- a/src/js/apps/admin/admin-main_app.js
+++ b/src/js/apps/admin/admin-main_app.js
@@ -1,6 +1,9 @@
+import Radio from 'backbone.radio';
+
 import RouterApp from 'js/base/routerapp';
 
 import ProgramsAllApp from 'js/apps/admin/list/programs-all_app';
+import CliniciansAllApp from 'js/apps/admin/list/clinicians-all_app';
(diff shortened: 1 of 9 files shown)


src/js/views/admin/actions/actions_views.js
```diff
@@ -265,9 +265,70 @@ const AttachmentComponent = Droplist.extend({
   },
 });
 
+const RoleItemTemplate = hbs`<a{{#if isSelected}} class="is-selected"{{/if}}>{{matchText name query}} <span class="program-actions__role">{{matchText short query}}</span></a>`;
```
paulfalgout: I know we don't have a great place to put these common elements right now, but this really isn't action related.  Could probably just go into the clinician views for now

paulfalgout (changes requested):
Should be able to add istanbul ignore if/else comments to avoid the drop in coverage related to the `_DEVELOP_` flags

PR #256, opened 2020-03-17
Remove program-flow-action

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch22151]

~~There's undoubtedly still some tests to look into.~~ For whatever reason the intermediate object was a bit more forgiving in the fixtures..
~~I also probably need to build commits a bit better.~~

This takes care of removing program-flow-action, but also address disabling the sorting when sorting can't happen.


src/js/behaviors/sortable-list.js
```diff
@@ -6,13 +6,36 @@ export default Behavior.extend({
...
+  /* istanbul ignore next: sensible but unused default */
+  shouldDisable() {
+    return this.view.collection.length < 2;
+  },
+  collectionEvents: {
+    'update change:id': 'updateDisabled',
+  },
+  updateDisabled() {
+    const shouldDisabled = this.shouldDisable();
+    this.view.$el.toggleClass('is-dragable', !shouldDisabled);
```
LisaManresa: Not a huge deal since the classname in the scss is the same, but typo.`is-draggable`

LisaManresa (approved):

luketlancaster (approved):

PR #259, opened 2020-03-23
Break out owned-by worklist and add clinician filter

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch22141]

### DIFF
--- a/src/js/apps/patients/worklist/filters_app.js
+++ b/src/js/apps/patients/worklist/filters_app.js
@@ -1,0 +1,78 @@
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+
+import { FiltersView, GroupsDropList, ClinicianDropList, ClinicianClearButton } from 'js/views/patients/worklist/worklist_views';
+
+export default App.extend({
+  stateEvents: {
+    'change:clinicianId': 'showFilters',
+  },
+  onStart(options) {
+    this.currentClinician = options.currentClinician;
+    this.groups = this.currentClinician.getGroups();
+
+    this.showView(new FiltersView({
+      model: this.currentClinician,
+    }));
+    this.showFilters();
+  },
+  showFilters() {
+    this.showGroupsFilterView();
+    this.showCliniciansFilterView();
+    this.showResetButton();
+  },
+  showGroupsFilterView() {
+    const groups = this._getGroups();
+    if (groups.length <= 2) return;
+
+    const groupsSelect = new GroupsDropList({
+      collection: groups,
+      state: { selected: groups.get(this.getState('groupId')) },
+    });
+
+    this.listenTo(groupsSelect.getState(), 'change:selected', (state, { id }) => {
+      this.setState('groupId', id);
+    });
+
+    this.showChildView('group', groupsSelect);
+  },
+  showCliniciansFilterView() {
+    const selected = Radio.request('entities', 'clinicians:model', this.getState('clinicianId'));
+
+    const clinicianFilter = new ClinicianDropList({
+      model: this.currentClinician,
+      state: { selected },
+    });
+
+    this.listenTo(clinicianFilter, 'change:selected', ({ id }) => {
+      this.setState('clinicianId', id);
+    });
+
+    this.showChildView('clinician', clinicianFilter);
+  },
+  showResetButton() {
+    if (this.currentClinician.id === this.getState('clinicianId')) {
+      this.getRegion('reset').empty();
+      return;
+    }
+
+    const clearButton = new ClinicianClearButton();
+
+    this.listenTo(clearButton, 'click', () => {
+      this.setState('clinicianId', this.currentClinician.id);
+    });
+
+    this.showChildView('reset', clearButton);
+  },
+  _getGroups() {
+    const groups = this.groups.clone();
+
+    groups.unshift({
+      id: 'all-groups',
+      name: 'All Groups',
+    });
+
+    return groups;
+  },
+});
--- a/src/js/apps/patients/worklist/worklist-owned-by_app.js
+++ b/src/js/apps/patients/worklist/worklist-owned-by_app.js
@@ -1,25 +1,28 @@
-import _ from 'underscore';
-import moment from 'moment';
 import store from 'store';
 
 import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
-import { ListView, LayoutView, GroupsDropList, SortDropList, sortOptions } from 'js/views/patients/worklist/worklist_views';
+import FiltersApp from './filters_app';
 
-const DEFAULT_STATE = {
-  sortId: 'sortUpdateDesc',
-  groupId: null,
-};
+import { ListView, LayoutView, SortDropList, sortOptions } from 'js/views/patients/worklist/worklist_views';
 
 export default App.extend({
+  childApps: {
+    filters: {
+      AppClass: FiltersApp,
+      restartWithParent: false,
+      regionName: 'filters',
+      getOptions: ['currentClinician'],
+    },
+  },
   stateEvents: {
     'change': 'onChangeState',
   },
   onChangeState(state) {
     store.set(this.stateId, state);
-    if (state.hasChanged('groupId')) {
+    if (state.hasChanged('groupId') || state.hasChanged('clinicianId')) {
       this.restart();
     }
   },
@@ -29,9 +32,15 @@ export default App.extend({
   initListState() {
     this.stateId = `worklist-${ this.worklistId }-${ this.worklistType }`;
 
+    const default_state = {
+      sortId: 'sortUpdateDesc',
+      groupId: 'all-groups',
+      clinicianId: this.currentClinician.id,
+    };
+
     if (!store.get(this.stateId)) {
-      store.set(this.stateId, DEFAULT_STATE);
-      this.setState(DEFAULT_STATE);
+      store.set(this.stateId, default_state);
+      this.setState(default_state);
     } else {
       this.setState(store.get(this.stateId));
     }
@@ -44,19 +53,19 @@ export default App.extend({
 
     this.worklistId = worklistId;
     this.worklistType = worklistType;
-
-    this.initListState();
-
     this.currentClinician = Radio.request('bootstrap', 'currentUser');
     this.groups = this.currentClinician.getGroups();
+
+    this.initListState();
 
     this.showView(new LayoutView({
       worklistId: this.worklistId,
       isFlowList: this.worklistType === 'flows',
     }));
+
     this.getRegion('list').startPreloader();
-    this.showFilterView();
     this.showSortView();
+    this.startFiltersApp();
   },
   beforeStart() {
     let groupId = this.getState('groupId');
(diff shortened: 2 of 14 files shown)


luketlancaster:
there's like a 100% chance I missed something, but it was time to get another set of eyes on this beast before I went totally crazy

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -300,16 +300,63 @@ const ListView = CollectionView.extend({
...
+  },
+  initialize({ model }) {
+    const groups = model.getGroups();
+
+    this.lists = groups.map(group => {
+      return {
+        collection: group.getClinicians(),
+        headingText: group.get('name'),
+        itemTemplate: hbs`<a{{#if isSelected}} class="is-selected"{{/if}}>{{matchText name query}}</a>`,
+        getItemSearchText() {
```
paulfalgout: I think we're doing this too often in the code base.. we need this if we're doing something with two data points like `Nurse (NUR)` or whatever.. but in this case `attr: 'name'` should be enough

src/js/apps/patients/worklist/worklist-owned-by_app.js
```diff
@@ -0,0 +1,194 @@
...
+    this.showSortView();
+  },
+  beforeStart() {
+    let groupId = this.getState('groupId');
+
+    if (groupId === 'all-groups') {
+      groupId = this.groups.pluck('id').join(',');
+    }
+
+    const filter = _.extend({ group: groupId }, this._filtersById(this.worklistId, this.getState('clinicianId')));
```
paulfalgout: let's make this line `const filter = this._filtersBy(this.getState('groupId'), this.getState('clinicianId'));`

src/js/apps/patients/worklist/worklist-owned-by_app.js
```diff
@@ -0,0 +1,194 @@
...
+      this.setState('sortId', selected.id);
+      this.getChildView('list').setComparator(selected.get('comparator'));
+    });
+
+    this.showChildView('sort', sortSelect);
+  },
+  onToggleListType(type) {
+    Radio.trigger('event-router', `worklist:${ type }`, this.worklistId);
+  },
+  _getGroups() {
```
paulfalgout: unused

src/js/apps/patients/worklist/worklist-owned-by_app.js
```diff
@@ -0,0 +1,194 @@
...
+    if (!store.get(this.stateId)) {
+      store.set(this.stateId, default_state);
+      this.setState(default_state);
+    } else {
+      this.setState(store.get(this.stateId));
+    }
+  },
+  onBeforeStart({ worklistId, worklistType }) {
+    if (this.isRestarting()) {
+      this.getRegion('list').startPreloader();
```
paulfalgout, line 50: this happens in the `if` and the `else`  let's just make it the first line

src/js/apps/patients/worklist/worklist-owned-by_app.js
```diff
@@ -0,0 +1,194 @@
...
+  beforeStart() {
+    let groupId = this.getState('groupId');
+
+    if (groupId === 'all-groups') {
+      groupId = this.groups.pluck('id').join(',');
+    }
+
+    const filter = _.extend({ group: groupId }, this._filtersById(this.worklistId, this.getState('clinicianId')));
+
+    const filtersApp = this.startChildApp('filters', {
```
paulfalgout: let's move this start into a function and call it in `onBeforeStart` along with the `listenTo`.  Also with changing the childApp above we won't need to set region or clinician here.  Then we should not need to set defaults with the `||`  if we do then the defaults aren't working above.

src/js/apps/patients/worklist/worklist-owned-by_app.js
```diff
@@ -0,0 +1,194 @@
...
+    text: i18n.sortOptions.sortUpdateDesc,
+    comparator(a, b) {
+      return dateSort('desc', a.model.get('updated_at'), b.model.get('updated_at'));
+    },
+  },
+]);
+
+export default App.extend({
+  childApps: {
+    filters: FiltersApp,
```
paulfalgout: let's change this to:
```
  filters: {
    AppClass: FiltersApp,
    regionName: 'filters',
    getOptions: ['currentClinician'],
  }
```

src/js/apps/patients/worklist/worklist-owned-by_app.js
```diff
@@ -0,0 +1,194 @@
...
+    const filter = _.extend({ group: groupId }, this._filtersById(this.worklistId, this.getState('clinicianId')));
+
+    const filtersApp = this.startChildApp('filters', {
+      region: this.getRegion('filters'),
+      clinician: this.currentClinician,
+      groupId: this.getState('groupId') || this.groups.at(0).id,
+      clinicianId: this.getState('clinicianId') || this.currentClinician.id,
+    });
+
+    this.listenTo(filtersApp, {
```
paulfalgout: you _could_ simplify this a bit I think.. I don't think it'd be too brittle, but this could be:
```js
this.listenTo(filtersApp, 'change:state', state = > {
  this.setState(state);
});
```
then in the change event's you'll be triggering `{ groupId: id }` etc

src/js/apps/patients/worklist/worklist-owned-by_app.js
```diff
@@ -0,0 +1,194 @@
...
+      'filter:group': this.filterGroup,
+      'filter:clinician': this.filterClinician,
+      'filter:currentClinician': this.filterClinician,
+    });
+
+    if (this.worklistType === 'actions') {
+      return Radio.request('entities', 'fetch:actions:collection', { filter });
+    }
+
+    return Radio.request('entities', 'fetch:flows:collection', { filter });
```
paulfalgout: we can update the other app too, but I don't see any reason we couldn't use `fetch:${ this.worklistType }:collection`

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -0,0 +1,74 @@
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+
+import { FiltersView, GroupsDropList, ClinicianDropList, ClinicianClearButton } from 'js/views/patients/worklist/worklist_views';
+
+export default App.extend({
+  onBeforeStart(options) {
```
paulfalgout: probably a decent case for `this.mergeOptions(options, ['currentClinician', 'groupId', 'clinicianId']);`

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -0,0 +1,74 @@
...
+
+    this.showFilters();
+  },
+  showFilters() {
+    this.showGroupsFilterView();
+    this.showCliniciansFilterView();
+    this.showResetButton();
+  },
+  showGroupsFilterView() {
+    if (this.groups.length === 1) return;
```
paulfalgout: let's change this to:
```
const groups = this.currentClinician.getGroups();
if(groups.length < 2) return;

const groupsSelect = new GroupsDropList({
  collection: this._getGroups(groups),
```

paulfalgout (changes requested):

src/js/apps/patients/worklist/worklist-owned-by_app.js
luketlancaster: so if I move this to `onBeforeStart` the filters only work one time, then they don't seem to be changing the state at all

paulfalgout: ah.. add `restartWithParent: false` to the childApp definition with `AppClass`

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -0,0 +1,73 @@
...
+export default App.extend({
+  onBeforeStart(options) {
+    this.mergeOptions(options, ['currentClinician', 'groupId', 'clinicianId']);
+    this.groups = this.currentClinician.getGroups();
+
+    this.showView(new FiltersView({
+      model: this.currentClinician,
+    }));
+  },
+  onStart() {
```
paulfalgout: since there's no async here all of the above can just go in `onStart`

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -300,16 +300,61 @@ const ListView = CollectionView.extend({
...
   picklistOptions: {
     attr: 'name',
   },
 });
 
+const ClinicianDropList = Droplist.extend({
+  picklistOptions: {
+    isSelectlist: true,
+  },
+  viewOptions() {
```
paulfalgout: this doesn't need to be a function

src/js/apps/patients/worklist/worklist-owned-by_app.js
```diff
@@ -0,0 +1,172 @@
...
+
+import App from 'js/base/app';
+
+import FiltersApp from './filters_app';
+
+import { ListView, LayoutView, SortDropList } from 'js/views/patients/worklist/worklist_views';
+
+const i18n = intl.patients.worklist.worklistApp;
+
+const sortOptions = new Backbone.Collection([
```
paulfalgout: ya know.. 100% of this is view related.    Why don't we move `sortOptions` to `worklist_views` for now.

paulfalgout (changes requested):

src/js/apps/patients/worklist/worklist-owned-by_app.js
luketlancaster: yeah I was running into an issue with this...lemme see if I can't reproduce it

luketlancaster: yup, so if I just make `this.getRegion('list')` the first line, it errors because the region is undefined. If keep only the bottom one, I get a "TypeError: pattern is undefined"

paulfalgout: ah right.. because it's based off the layout.. carry on

paulfalgout (changes requested):
Ok.. so.. after thinking about it.. one more iteration.. what if we moved calling `startFiltersApp` to after `showSortView` then in the method do `listenTo(filtersApp.getState(), 'change', state => { this.setState(state.attributes`  then make the app `restartWithParent: false`  and up in the filters app change the `triggerMethod`s to `setState`s  and instead of `mergeOptions` pass `state: this.getState().attributes }` which'll initiate the state model.  Then the only thing that needs to be added is:
```
stateEvents: {
  'change:clinicianId': 'showResetButton',
},
```

paulfalgout:
Looks like we lost some coverage https://codecov.io/gh/RoundingWell/care-ops-frontend/compare/92b270f0eab480ef0e792ee63f2e92eb32b7e9cc...f91f5dd0918e4ec339a3753e3204da00e6c1bac9/src/src/js/apps/patients/worklist/worklist_app.js  need to add a couple of tests that are now only done on the owned-by page

paulfalgout (approved):
Approved once line coverage is 100%

luketlancaster:
pushing to update coverage because it runs faster in the ☁️

I _think_ I still have one miss, but I'm not sure why because it looks like the tests should be going down that path

PR #260, opened 2020-03-24
Clinician sidebar

LisaManresa:
- [x] Add the Clubhouse Story ID: [c22168] [ch22167]

Still lots to do on this app, but it would be good to get eyes on this so far. This sets up the routing and views for the clinician sidebar (add and edit).

### DIFF
--- a/src/js/apps/admin/list/clinicians-all_app.js
+++ b/src/js/apps/admin/list/clinicians-all_app.js
@@ -1,18 +1,58 @@
 import Radio from 'backbone.radio';
 
-import App from 'js/base/app';
+import SubRouterApp from 'js/base/subrouterapp';
+
+import intl from 'js/i18n';
+
 import { ListView, LayoutView } from 'js/views/admin/list/clinicians-all_views';
 
-export default App.extend({
+export default SubRouterApp.extend({
+  routerAppName: 'CliniciansApp',
+  eventRoutes: {
+    'clinician': 'showClinicianSidebar',
+    'clinician:new': 'showClinicianSidebar',
+  },
+  viewEvents: {
+    'click:addClinician': 'onClickAddClinician',
+  },
   onBeforeStart() {
     this.showView(new LayoutView());
     this.getRegion('list').startPreloader();
+
+    this.clinicians = Radio.request('entities', 'clinicians:collection');
   },
   beforeStart() {
     return Radio.request('entities', 'fetch:clinicians:collection');
   },
-  onStart(options, collection) {
-    this.clinicians = collection;
-    this.showChildView('list', new ListView({ collection }));
+  onStart({ currentRoute }, collection) {
+    this.clinicians.add(collection.models);
+
+    this.showChildView('list', new ListView({ collection: this.clinicians }));
+
+    this.startRoute(currentRoute);
+  },
+  showClinicianSidebar(clinicianId) {
+    const clinician = clinicianId ? this.clinicians.get(clinicianId) : Radio.request('entities', 'clinicians:model', {});
+
+    if (!clinician) {
+      Radio.request('alert', 'show:error', intl.admin.list.cliniciansAllApp.notFound);
+      Radio.trigger('event-router', 'clinicians:all');
+      return;
+    }
+
+    Radio.request('sidebar', 'start', 'clinician', { clinician });
+
+    this.editClinician(clinician);
+  },
+  editClinician(clinician) {
+    if (clinician.isNew()) {
+      this.clinicians.unshift(clinician);
+      return;
+    }
+
+    clinician.trigger('editing', true);
+  },
+  onClickAddClinician() {
+    Radio.trigger('event-router', 'clinician:new');
   },
 });
--- a/src/js/apps/admin/sidebar/clinician-sidebar_app.js
+++ b/src/js/apps/admin/sidebar/clinician-sidebar_app.js
@@ -1,0 +1,37 @@
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+
+import { SidebarView } from 'js/views/admin/sidebar/clinician/clinician-sidebar_views';
+
+export default App.extend({
+  onBeforeStart({ clinician }) {
+    this.clinician = clinician;
+
+    this.showView(new SidebarView({
+      model: this.clinician,
+    }));
+  },
+  viewEvents: {
+    'save': 'onSave',
+    'close': 'stop',
+    'delete': 'onDelete',
+  },
+  onSave({ model }) {
+    this.clinician.save(model.attributes).done(() => {
+      Radio.trigger('event-router', 'clinician', this.clinician.id);
+    });
+  },
+  // onDelete() {
+  //   this.clinician.trigger('editing', false);
+  //   this.clinician.destroy();
+  //   this.stop();
+  // },
+  onStop() {
+    this.clinician.trigger('editing', false);
+    if (this.clinician && this.clinician.isNew()) this.clinician.destroy();
+
+    Radio.request('sidebar', 'close');
+    Radio.trigger('event-router', 'clinicians:all');
+  },
+});
--- a/src/js/entities-service/entities/clinicians.js
+++ b/src/js/entities-service/entities/clinicians.js
@@ -14,6 +14,15 @@ const _Model = BaseModel.extend({
   preinitialize() {
     this.on('change:_role', this.onChangeRole);
   },
+  validate(attrs) {
+    if (!attrs.name) {
+      return 'A clinician name is required';
+    }
+
+    if (!attrs.email) {
+      return 'A clinician email address is required';
+    }
+  },
   onChangeRole() {
     const previousRole = Radio.request('entities', 'roles:model', this.previous('_role'));
     previousRole.set('_clinicians', _.reject(previousRole.get('_clinicians'), { id: this.id }));
--- a/src/js/views/admin/list/clinicians-all_views.js
+++ b/src/js/views/admin/list/clinicians-all_views.js
@@ -1,74 +1,27 @@
 import _ from 'underscore';
 import hbs from 'handlebars-inline-precompile';
 import Radio from 'backbone.radio';
-import { View, CollectionView } from 'marionette';
+import { View, CollectionView, Behavior } from 'marionette';
 
 import PreloadRegion from 'js/regions/preload_region';
-import Droplist from 'js/components/droplist';
-
-import intl from 'js/i18n';
+import { RoleComponent } from 'js/views/admin/shared/clinicians_components';
 
 import 'sass/modules/list-pages.scss';
 import 'sass/modules/table-list.scss';
 
-
-const RoleItemTemplate = hbs`<a{{#if isSelected}} class="is-selected"{{/if}}>{{matchText name query}} <span class="program-actions__role">{{matchText short query}}</span></a>`;
-
-const RoleComponent = Droplist.extend({
-  isCompact: false,
-  getTemplate() {
-    if (this.getOption('isCompact')) {
-      return hbs`{{far "user-circle"}}{{ short }}`;
-    }
-
-    return hbs`{{far "user-circle"}}{{ name }}{{#unless name}}{{ @intl.admin.list.cliniciansAllViews.roleComponent.roleDefaultText.defaultText }}{{/unless}}`;
+const RowBehavior = Behavior.extend({
+  modelEvents: {
+    'editing': 'onEditing',
+    'change': 'onChange',
   },
-  popWidth() {
-    const isCompact = this.getOption('isCompact');
-
-    return isCompact ? null : this.getView().$el.outerWidth();
+  onChange() {
+    this.view.render();
   },
-  picklistOptions: {
-    isSelectlist: true,
+  onEditing(isEditing) {
+    this.$el.toggleClass('is-selected', isEditing);
   },
-  viewOptions() {
-    const isCompact = this.getOption('isCompact');
-    const selected = this.getState('selected');
-    return {
-      modelEvents: {
-        'change:_role': 'render',
-      },
-      className() {
-        if (!selected && isCompact) {
-          return 'button-secondary--compact is-icon-only';
-        }
-
-        if (isCompact) {
-          return 'button-secondary--compact';
-        }
-
-        return 'button-secondary w-100';
-      },
-      template: this.getTemplate(),
-    };
-  },
-  initialize({ model }) {
-    const currentOrg = Radio.request('bootstrap', 'currentOrg');
-    const roles = currentOrg.getActiveRoles();
-
-    this.lists = [{
-      collection: roles,
-      itemTemplate: RoleItemTemplate,
-      headingText: intl.admin.list.cliniciansAllViews.roleComponent.rolesHeadingText,
-      getItemSearchText() {
-        return this.$el.text();
-      },
-    }];
-
-    this.setState({ selected: model.getRole() });
-  },
-  onChangeSelected(selected) {
-    this.triggerMethod('change:role', selected);
+  onInitialize() {
+    if (this.view.model.isNew()) this.$el.addClass('is-selected');
   },
 });
 
@@ -83,12 +36,16 @@ const EmptyView = View.extend({
 
 const ItemView = View.extend({
   className: 'table-list__item',
+  behaviors: [RowBehavior],
   tagName: 'tr',
   regions: {
     role: '[data-role-region]',
   },
+  triggers: {
+    'click': 'click',
+  },
   template: hbs`
-    <td class="table-list__cell w-20">{{ name }}</td>
+    <td class="table-list__cell w-20">{{#unless name}}{{ @intl.admin.list.cliniciansAllViews.itemView.newClinician }}{{/unless}}{{ name }}</td>
     <td class="table-list__cell w-40">{{#each groups}}{{#unless @first}}, {{/unless}}{{ this.name }}{{/each}}</td>
     <td class="table-list__cell w-30" data-role-region></td>
   `,
@@ -99,6 +56,13 @@ const ItemView = View.extend({
   },
   onRender() {
     this.showRole();
+  },
+  onClick() {
+    if (this.model.isNew()) {
+      return;
+    }
+
+    Radio.trigger('event-router', 'clinician', this.model.id);
   },
   showRole() {
     const roleComponent = new RoleComponent({ model: this.model, isCompact: true });
@@ -116,6 +80,7 @@ const LayoutView = View.extend({
   template: hbs`
     <div class="list-page__header">
       <div class="list-page__title">{{ @intl.admin.list.cliniciansAllViews.layoutView.title }}</div>
+      <button class="button-primary js-add-clinician">{{far "plus-circle"}}{{ @intl.admin.list.cliniciansAllViews.layoutView.addClinicianButton }}</button>
     </div>
     <div class="flex-region list-page__list">
       <table class="w-100"><tr>
@@ -131,9 +96,14 @@ const LayoutView = View.extend({
       el: '[data-list-region]',
       regionClass: PreloadRegion,
     },
+    sidebar: '[data-sidebar-region]',
+    addClinician: {
+      el: '[data-add-region]',
+      replaceElement: true,
+    },
   },
   triggers: {
-    'click .js-add': 'click:add',
+    'click .js-add-clinician': 'click:addClinician',
   },
 });
 
--- a/src/js/views/admin/shared/clinicians_components.js
+++ b/src/js/views/admin/shared/clinicians_components.js
@@ -1,0 +1,68 @@
+import hbs from 'handlebars-inline-precompile';
+import Radio from 'backbone.radio';
+
+import Droplist from 'js/components/droplist';
+
+import intl from 'js/i18n';
+
+const RoleItemTemplate = hbs`<a{{#if isSelected}} class="is-selected"{{/if}}>{{matchText name query}} <span class="clinician__role">{{matchText short query}}</span></a>`;
+
+const RoleComponent = Droplist.extend({
+  isCompact: false,
+  getTemplate() {
+    if (this.getOption('isCompact')) {
+      return hbs`{{far "user-circle"}}{{ short }}`;
+    }
+
+    return hbs`{{far "user-circle"}}{{ name }}{{#unless name}}{{ @intl.admin.shared.cliniciansComponents.roleComponent.roleDefaultText }}{{/unless}}`;
+  },
+  popWidth() {
+    const isCompact = this.getOption('isCompact');
+
+    return isCompact ? null : this.getView().$el.outerWidth();
+  },
+  picklistOptions: {
+    isSelectlist: true,
+  },
+  viewOptions() {
+    const isCompact = this.getOption('isCompact');
+    const selected = this.getState('selected');
+    return {
+      modelEvents: {
+        'change:_role': 'render',
+      },
+      className() {
+        if (!selected && isCompact) {
+          return 'button-secondary--compact is-icon-only';
+        }
+
+        if (isCompact) {
+          return 'button-secondary--compact';
+        }
+
+        return 'button-secondary w-100';
+      },
+      template: this.getTemplate(),
+    };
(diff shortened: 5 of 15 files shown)


src/js/entities-service/entities/clinicians.js
```diff
@@ -25,7 +25,7 @@ const _Model = BaseModel.extend({
     return Radio.request('entities', 'groups:collection', this.get('_groups'));
   },
   getRole() {
-    if (!this.get('_role')) {
+    if (!this.get('_role') && !this.isNew()) {
```
LisaManresa: Added this to allow for an empty role on a new clinician model so that we can display the droplist default text of "Select Role..."

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,226 @@
...
+import 'sass/modules/forms.scss';
+import 'sass/modules/sidebar.scss';
+
+import ClinicianSidebarTemplate from './clinician-sidebar.hbs';
+import ClinicianNameTemplate from './clinician-name.hbs';
+import ClinicianEmailTemplate from './clinician-email.hbs';
+
+const RoleItemTemplate = hbs`<a{{#if isSelected}} class="is-selected"{{/if}}>{{matchText name query}} <span class="clinician__role">{{matchText short query}}</span></a>`;
+
+const RoleComponent = Droplist.extend({
```
LisaManresa: This is identical to the RoleComponent in clinicians-all_views.js. So we'll probably want to find a common place for this. Would something like `clinicians-actions_views.js` make sense?

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,226 @@
...
+    const roleComponent = new RoleComponent({ model: this.model, state: { isDisabled } });
+
+    this.listenTo(roleComponent, 'change:role', role => {
+      this.model.saveRole(role);
+    });
+
+    this.showChildView('role', roleComponent);
+  },
+  showSave() {
+    if (!this.model.isValid()) return this.showDisabledSave();
```
LisaManresa: We're probably going to want an email validator for this check, so that a model with an invalid email cannot be saved.

test/integration/admin/list/clinicians-all.js
```diff
@@ -2,8 +2,13 @@ import _ from 'underscore';
 
 context('clinicians list', function() {
   specify('display clinicians list', function() {
+    // routeClinicians gets called twice in this app because of the clinicians bootstrap call.
```
LisaManresa: I think this was the cause of the 404 issue I was having in cypress tests. There may be a better way to handle this, but starting out on another view then navigating to the clinicians list was a reliable workaround. Added this comment anywhere I was doing that for context.

src/js/entities-service/entities/clinicians.js
```diff
@@ -25,7 +25,7 @@ const _Model = BaseModel.extend({
     return Radio.request('entities', 'groups:collection', this.get('_groups'));
   },
   getRole() {
-    if (!this.get('_role')) {
+    if (!this.get('_role') && this.get('name') === 'RoundingWell') {
```
LisaManresa: This is pretty hacky, but added this to allow for an empty role on a new clinician model so that we can display the droplist default text of "Select Role..."

src/js/apps/admin/clinician/clinician_app.js
```diff
@@ -0,0 +1,24 @@
+import Radio from 'backbone.radio';
+
+import intl from 'js/i18n';
+
+import App from 'js/base/app';
+
+export default App.extend({
+  beforeStart({ clinicianId, programId }) {
```
paulfalgout: programId?

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,226 @@
...
+    this.lists = [{
+      collection: roles,
+      itemTemplate: RoleItemTemplate,
+      headingText: intl.admin.sidebar.clinician.clinicianSidebarViews.roleComponent.rolesHeadingText,
+      getItemSearchText() {
+        return this.$el.text();
+      },
+    }];
+
+    this.setState({ selected: model.getRole() });
```
paulfalgout: what if we handled `model.get('_role')` with `roles` here for now to get around needing to change `getRole`

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,226 @@
...
+  },
+  initialize({ model }) {
+    const currentOrg = Radio.request('bootstrap', 'currentOrg');
+    const roles = currentOrg.getActiveRoles();
+
+    this.lists = [{
+      collection: roles,
+      itemTemplate: RoleItemTemplate,
+      headingText: intl.admin.sidebar.clinician.clinicianSidebarViews.roleComponent.rolesHeadingText,
+      getItemSearchText() {
```
paulfalgout: I think this can just be `attr: 'name',` instead of overriding `getItemSearchText`

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,226 @@
...
+    }];
+
+    this.setState({ selected: model.getRole() });
+  },
+  onChangeSelected(selected) {
+    this.triggerMethod('change:role', selected);
+  },
+});
+
+const NameView = View.extend({
```
paulfalgout, line 12: we can simplify this I think.  it can just be a regular input and doesn't need to account for multiple lines.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,226 @@
...
+    };
+  },
+  onDomRefresh() {
+    if (this.model.isNew()) {
+      this.ui.input.focus();
+    }
+  },
+});
+
+const EmailView = View.extend({
```
paulfalgout, line 31: similarly doesn't need to account for multiple lines.  just an `<input`

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,226 @@
...
+    save: '[data-save-region]',
+    // access: '[data-access-region]',
+    role: '[data-role-region]',
+    // groups: '[data-groups-region]',
+  },
+  onRender() {
+    this.showName();
+    this.showEmail();
+    this.showRole();
+    this.showSave();
```
paulfalgout: we only need to have save if the model is new.  we can't edit names or emails of existing clinicians

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
paulfalgout: I think the BE will have to make this error anyhow for testing for other duplicate possibilities or other email restrictions so we can probably just rely on that check

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,226 @@
...
+    this.showChildView('save', new SaveView({ model: this.model }));
+  },
+  showDisabledSave() {
+    this.showChildView('save', new DisabledSaveView());
+  },
+  onSave() {
+    this.getRegion('save').empty();
+  },
+  onCancel() {
+    if (this.model.isNew()) {
```
paulfalgout: I think cancel will likely only ever be available if the model is new.

paulfalgout (changes requested):
I realize this'll be a large edit, but I think we can potentially simplify this list app by making the sidebar more like the flow or program sidebar where the list app would just start the sidebar via the service with the model rather than using the id and having a separate app load the model.  Actions are a bit more complicated because the list can be dependent on the action (completed vs incomplete) but right now the clinician will always be loaded in the list.  So we shouldn't need `GET clinician/id` right now.

I'm also looking into an alternative solution for the testing bootstrap problem.

src/js/apps/admin/list/clinicians-all_app.js
```diff
@@ -1,18 +1,54 @@
...
   },
-  onStart(options, collection) {
+  onStart({ currentRoute }, collection) {
     this.clinicians = collection;
     this.showChildView('list', new ListView({ collection }));
   },
+  showClinicianSidebar(clinicianId) {
+    const clinician = clinicianId ? this.clinicians.get(clinicianId) : Radio.request('entities', 'clinicians:model', {});
+
+    if (clinicianId && !clinician) {
```
paulfalgout: no need to check `clinicianId` here `!clinician` is enough

src/js/apps/admin/list/clinicians-all_app.js
```diff
@@ -1,18 +1,54 @@
...
+      Radio.request('alert', 'show:error', intl.admin.list.cliniciansAllViews.notFound);
```
paulfalgout: should be `cliniciansAllApp.notFound`

src/js/apps/admin/list/clinicians-all_app.js
```diff
@@ -1,18 +1,54 @@
...
+      Radio.trigger('event-router', 'clinicians:all');
```
paulfalgout, line 39: need to add a `return;` to this `if`

src/js/apps/admin/sidebar/clinician-sidebar_app.js
```diff
@@ -0,0 +1,41 @@
...
+  },
+  onSave({ model }) {
+    if (model.isNew()) {
+      this.clinician.save(model.attributes).done(() => {
+        Radio.trigger('event-router', 'clinician', this.clinician.id);
+      });
+      return;
+    }
+
+    this.clinician.save(model.pick('name', 'email'));
```
paulfalgout: We can probably deal with this later, but there's a real change that save will only ever be an option on new clinicians.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,216 @@
...
+    this.setState({ selected: role });
+  },
+  onChangeSelected(selected) {
+    this.triggerMethod('change:role', selected);
+  },
+});
+
+const NameView = View.extend({
+  className: 'pos--relative',
+  template: hbs`<input class="input-primary w-100 js-input"{{#if isNew}} placeholder="{{ @intl.admin.sidebar.clinician.clinicianSidebarViews.nameView.placeholder }}"{{/if}} value="{{ name }}" />`,
```
paulfalgout: I think the placeholder can always be available

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,216 @@
...
+  behaviors: [InputWatcherBehavior],
```
paulfalgout: No need to key watch.  Can just do `'input .js-input` in events

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,216 @@
...
+    const roles = currentOrg.getActiveRoles();
+
+    this.lists = [{
+      collection: roles,
+      itemTemplate: RoleItemTemplate,
+      headingText: intl.admin.sidebar.clinician.clinicianSidebarViews.roleComponent.rolesHeadingText,
+      attr: 'name',
+    }];
+
+    const role = Radio.request('entities', 'roles:model', model.get('_role'));
```
paulfalgout: could just do `roles.get(model.get('_role'));`

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,216 @@
...
+    if (!this.model.isValid()) return this.showDisabledSave();
+
+    this.showChildView('save', new SaveView({ model: this.model }));
+  },
+  showDisabledSave() {
+    this.showChildView('save', new DisabledSaveView());
+  },
+  showForm() {
+    this.stopListening(this.model);
+    this.model = this.model.clone();
```
paulfalgout: I don't think this is going to work.  I don't see how the original model will understand if something changes on the clone.

paulfalgout (changes requested):

src/js/apps/admin/list/clinicians-all_app.js
LisaManresa: showClinicianSidebar can be called without an ID to create a new clinician. So there will still be a clinician model in that case. This is to catch when `this.clinicians.get(clinicianId)` is empty. Although, to be honest, I'm not even sure that will work as I haven't sorted out how to test it yet.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
LisaManresa: Don't we need the watcher to determine when to enable/disable the save view?

src/js/apps/admin/list/clinicians-all_app.js
paulfalgout: `if(!clinician)` catches the case for `this.clinicians.get(clinicianId)` being empty without needing to check the `clinicianId`  clinician will never be falsey when clinicianId is falsey, so the extra check isn't needed.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,149 @@
...
+    if (!this.model.isNew()) return;
+    if (!this.model.isValid()) return this.showDisabledSave();
+
+    this.showChildView('save', new SaveView({ model: this.model }));
+  },
+  showDisabledSave() {
+    this.showChildView('save', new DisabledSaveView());
+  },
+  showForm() {
+    this.listenTo(this.model, 'change:name change:email', this.showSave);
```
paulfalgout: rather than bailing out of `showSave` if `!isNew` this listener could just be in the `isNew` below.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
paulfalgout: `onWatchKeydown` I don't think should be needed?

So the behavior then I believe is equivalent to:
```js
events: {
  'input @ui.input`: 'onInput',
},
onInput() {
  this.model.set('name', this.ui.input.val());
},
```
I suppose we could keep it for consistency?  But I don't think the keydown watch is necessary for an input type=text?

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,149 @@
...
+
+    if (this.model.isNew()) this.showDisabledSave();
+    else this.getRegion('save').empty();
```
paulfalgout: I don't think this empty will ever be necessary.  `showForm` is only currently called in `onRender` so it is emptied during that render.

src/js/apps/admin/list/clinicians-all_app.js
```diff
@@ -1,18 +1,55 @@
...
+  },
   onBeforeStart() {
     this.showView(new LayoutView());
     this.getRegion('list').startPreloader();
   },
   beforeStart() {
     return Radio.request('entities', 'fetch:clinicians:collection');
   },
-  onStart(options, collection) {
+  onStart({ currentRoute }, collection) {
```
paulfalgout, line 27: it seems like the `currentRoute` needs to be handled otherwise these subroutes will only work if the app is already started

src/js/apps/admin/list/clinicians-all_app.js
```diff
@@ -1,18 +1,55 @@
...
+      return;
+    }
+
+    Radio.request('sidebar', 'start', 'clinician', { clinician });
+
+    this.editClinician(clinician);
+  },
+  editClinician(clinician) {
+    if (clinician.isNew()) {
+      this.clinicians.unshift(clinician);
```
paulfalgout, line 49: So I think if you were to push the add new button while the list is loading this'd fail because `this.clinicians` isn't available.  So maybe the new button should show `onStart`

src/js/apps/admin/list/clinicians-all_app.js
```diff
@@ -1,18 +1,55 @@
...
+
+import intl from 'js/i18n';
+
 import { ListView, LayoutView } from 'js/views/admin/list/clinicians-all_views';
 
-export default App.extend({
+export default SubRouterApp.extend({
+  routerAppName: 'CliniciansApp',
+  eventRoutes() {
+    return {
```
paulfalgout: this could just be `eventRoutes: { ... }` instead of a function.

src/js/views/admin/list/clinicians-all_views.js
```diff
@@ -100,6 +57,13 @@ const ItemView = View.extend({
   onRender() {
     this.showRole();
   },
+  onClick() {
+    if (this.model.isNew()) {
+      Radio.trigger('event-router', 'clinician:new');
```
paulfalgout: I think this will cancel anything not saved in the sidebar.  Best probably to just `return;` here?

src/js/views/admin/shared/clinicians_components.js
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        return 'button-secondary w-100';
+      },
+      template: this.getTemplate(),
+    };
+  },
+  initialize({ model }) {
+    const currentOrg = Radio.request('bootstrap', 'currentOrg');
+    const roles = currentOrg.getActiveRoles();
```
paulfalgout: I think this should be `getRoles`  Clinicians can be set to a role that doesn't have other clinicians.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,149 @@
...
+  onWatchKeydown(evt) {
+    if (evt.which === _.ENTER_KEY) {
+      evt.preventDefault();
+      return;
+    }
+  },
+  onWatchChange(text) {
+    this.model.set('email', text);
+  },
+  templateContext() {
```
paulfalgout: this doesn't seem to be in use

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -0,0 +1,149 @@
...
+  onWatchKeydown(evt) {
+    if (evt.which === _.ENTER_KEY) {
+      evt.preventDefault();
+      return;
+    }
+  },
+  onWatchChange(text) {
+    this.model.set('name', text);
+  },
+  templateContext() {
```
paulfalgout: doesn't seem to be in use

paulfalgout (changes requested):
I'm still looking into the cypress testing issue.. but in the very least I'd like to move it to a central location so there's one location that requires such a big exception.

src/js/apps/admin/list/clinicians-all_app.js
```diff
@@ -1,18 +1,63 @@
...
 
-export default App.extend({
+import intl from 'js/i18n';
+
+import { AddClinicianButtonView, ListView, LayoutView } from 'js/views/admin/list/clinicians-all_views';
+
+export default SubRouterApp.extend({
+  routerAppName: 'CliniciansApp',
+  eventRoutes: {
+    'clinicians:all': 'showCliniciansAll',
```
paulfalgout: I don't think this should be a route in the subapp.  it happens for any case

src/js/apps/admin/list/clinicians-all_app.js
```diff
@@ -1,18 +1,63 @@
...
     return Radio.request('entities', 'fetch:clinicians:collection');
   },
-  onStart(options, collection) {
+  onStart({ currentRoute }, collection) {
     this.clinicians = collection;
-    this.showChildView('list', new ListView({ collection }));
+
+    this.startRoute(currentRoute);
+  },
+  showCliniciansAll() {
```
paulfalgout: Seems like this needs to happen regardless of the route

src/js/apps/admin/list/clinicians-all_app.js
paulfalgout: Had a thought if it looks better to have the add button available during the load..  

`onBeforeStart` could have `this.collection = Radio.request('entities', 'clinicians:collection');`
Then in `onStart` it'd be `this.collection.add(collection.models);`

src/js/views/admin/list/clinicians-all_views.js
```diff
@@ -111,11 +75,19 @@ const ItemView = View.extend({
   },
 });
 
+const AddClinicianButtonView = View.extend({
+  template: hbs`<button class="button-primary js-add-clinician">{{far "plus-circle"}}{{ @intl.admin.list.cliniciansAllViews.layoutView.addClinician }}</button>`,
```
paulfalgout:
```diff
-layoutView
+addClinicianButtonView
```

test/integration/admin/programs/sidebar/clinician-sidebar.js
```diff
@@ -0,0 +1,323 @@
...
+            },
+          ],
+        };
+
+        return fx;
+      })
+      .navigate('/clinicians')
+      .wait('@routeClinicians');
+
+    cy
```
paulfalgout, line 64: In the sidebar tests let's navigate to the sidebar instead of using the list.  In the list test file let's test that when you click the list item or the add button that it routes to the right place, and stop there.

test/integration/admin/programs/sidebar/clinician-sidebar.js
```diff
@@ -0,0 +1,323 @@
...
+        url: '/api/clinicians',
+        response: {
+          data: {
+            id: '1',
+          },
+        },
+      })
+      .as('routePostClinician');
+
+    cy
```
paulfalgout, line 232: same here let's navigate to the new sidebar in this test file

paulfalgout (changes requested):

src/js/apps/admin/list/clinicians-all_app.js
LisaManresa: If some mapping for `clinicians:all` doesn't exist, then `startRoute()` in `onStart` fails. But we need `startRoute` to handle starting the clinician sidebar. Is there another step I'm missing?

src/js/apps/admin/list/clinicians-all_app.js
```diff
@@ -1,18 +1,61 @@
...
+  viewEvents: {
+    'click:addClinician': 'onClickAddClinician',
+  },
   onBeforeStart() {
     this.showView(new LayoutView());
     this.getRegion('list').startPreloader();
+
+    this.clinicians = Radio.request('entities', 'clinicians:collection');
+
+    const addClinicianView = this.showChildView('addClinician', new AddClinicianButtonView());
```
paulfalgout: so this'll be real annoying since I suggested breaking this out.. _but_ now it could be back in the layout safely... thoughts?

paulfalgout (changes requested):

src/js/apps/admin/list/clinicians-all_app.js
LisaManresa: Good call. I'm all for moving it back to layout.

src/js/views/admin/list/clinicians-all_views.js
```diff
@@ -116,6 +80,7 @@ const LayoutView = View.extend({
   template: hbs`
     <div class="list-page__header">
       <div class="list-page__title">{{ @intl.admin.list.cliniciansAllViews.layoutView.title }}</div>
+      <button class="button-primary js-add-clinician">{{far "plus-circle"}}{{ @intl.admin.list.cliniciansAllViews.addClinicianButtonView }}</button>
```
paulfalgout: this i18n should be under `layoutView`

paulfalgout (changes requested):
one last nitpick and this is :shipit:

PR #261, opened 2020-03-26
Only show due sorts on flow sort droplist

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch22236]


paulfalgout (changes requested):
So let's do this a hair different.  Rather than `sortOptions` being a `Backbone.Collection` in the views let's just make it two arrays.  `sortUpdateOptions` and `sortDueOptions`  then in the worklist_apps it'd be:
```js
const sortOptions = new Backbone.Collection(sortUpdateOptions);
if (this.worklistType === 'actions') sortOptions.add(sortDueOptions, { at: 0 });
```

PR #262, opened 2020-03-30
Feature/worklist type state

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch22236]
- [ ] Add QA tasks


src/js/apps/patients/worklist/filters_app.js
```diff
@@ -27,7 +27,7 @@ export default App.extend({
     if (this.groups.length < 2) return;
 
     const groups = this._getGroups();
-    const selected = groups.get(this.getState('groupId'));
+    const selected = groups.get(this.getState('groupId')) || groups.at(0);
```
paulfalgout, line 33: We should really be making sure `groupId` is correct before this point rather than picking `at(0)`.. what is this use case accounting for?  _although_ I can see a edge case for when a user is removed from a group.. but otherwise this seems like there's a bug somewhere.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -134,7 +133,6 @@ export default App.extend({
     this.showSortDroplist();
     this.showTypeToggle();
     this.showTableHeaders();
-    this.startFiltersApp();
```
paulfalgout: why was this change necessary?  Work was done to make sure the filters app was updated as data changed.. so that'd need to come out.. but ideally it could be in onBeforeStart

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -148,17 +146,18 @@ export default App.extend({
...
+    this.startFiltersApp();
   },
   getComparator() {
     const sortId = this.getState().getSort();
     return this.sortOptions.get(sortId).get('comparator');
   },
   getSortOptions() {
     if (this.getState().isFlowType()) {
-      return _.union(sortDueOptions, sortUpdateOptions);
+      return sortUpdateOptions;
```
paulfalgout, line 155: :+1: oops

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: ah... this is because the `filters.type` is changing outside of the filters app.  Ok.. well.. bad news... but I think it makes sense..  we need to move the toggle into the filters app.  Then all of the filters state is controlled by the same app.

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -108,6 +111,28 @@ const ListView = CollectionView.extend({
   },
 });
 
+const TypeToggleView = View.extend({
+  template: hbs`
+    <button class="button-secondary {{#unless isFlowList}}button--blue{{/unless}} worklist-list__toggle-actions js-toggle-actions">{{far "file-alt"}}{{ @intl.patients.worklist.layoutTemplate.actionsButton }}</button>{{~ remove_whitespace ~}}
```
paulfalgout: i18n need updating

paulfalgout (changes requested):
So I also just realized we'll have to extract the tooltip somehow from the layout.  We can't pass `isFlowList` to the layout as it would have to be re-rendered when the state changes.

src/js/apps/patients/worklist/filters_app.js
luketlancaster: if `groupId` is `null`, `groups.get` doesn't return anything. So I guess make the id for "All Groups" `all-groups` again?

src/js/apps/patients/worklist/worklist_app.js
luketlancaster: I changed this back because the first click on a filter worked, but any subsequent click didn't fire off an event

src/js/apps/patients/worklist/filters_app.js
paulfalgout: yeah :-/    just check the places where `groupId` is falsey checked.. or is id `0` hacky?  kind of..

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: I think I get it.. so.. we should remove stateEvents from the filters app.. since those'll never happen.. and let's put this back in `onBeforeStart` and add this to the `if(isRestarting`  I _think_ that'll work and handle state changes during loading

luketlancaster: nope, no dice. Same behavior as before: first interaction with a filter component changes does what you think it would: sets the state in the store, new requests are fired off, list updates (though the 'reset button' doesn't show up until a page refresh). Then every other interaction with a filter only updates the droplist, but the store isn't being updated and so requests aren't being fired/the list doesn't get updated

src/js/apps/patients/worklist/filters_app.js
luketlancaster: so `0` is kinda hacky...though the `groups` right here is private to this app, and so we know that "All Groups" is the first thing (unless for some reason we get a request in the future from business to make "All Groups" the last thing in the droplist?)

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: back to the drawing board I guess

paulfalgout: Ok.. I'm pretty sure everything works as it was in my commit with the exception of this line which needs to change:
```js
  this.listenTo(filtersApp.getState(), 'change', ({ attributes }) => {
    this.setState({ filters: _.clone(attributes) });
  });
```
The issue is that we're changing an object for the same object so even though the data on the object changed the object referenced did not so no change event was occurring..  Generally why we don't nest state data (and want flat data from endpoints) but I dunno.. seemed cleaner this time :-/

luketlancaster:
alrighty, take another look @paulfalgout. Still need to do the tooltips, but I _think_ this should be closer?

test/integration/services/patient-quick-search.js
```diff
@@ -17,7 +17,9 @@ context('Patient Quick Search', function() {
 
         return fx;
       })
-      .visit();
+      .visit()
+      .navigate('/worklist/owned-by')
```
paulfalgout: why does this need to navigate or wait at all?

test/integration/patients/worklist.js
```diff
@@ -925,6 +880,32 @@ context('worklist page', function() {
...
+      .get('.app-frame__content')
+      .find('.table-list__item')
+      .last()
+      .should('contain', 'Updated Least Recent');
+
+    cy
+      .get('.worklist-list__filter-sort')
+      .click()
+      .get('.picklist')
+      .should('not.contain', 'Due Date: Oldest - Newest')
```
paulfalgout: could just have one `should` with `not.contain`, 'Due Date`

src/js/views/patients/worklist/filters_views.js
```diff
@@ -0,0 +1,93 @@
...
+      isFlowList: this.getOption('isFlowList'),
+    };
+  },
+  triggers: {
+    'click .js-toggle-actions': 'click:toggleActions',
+    'click .js-toggle-flows': 'click:toggleFlows',
+  },
+  onClickToggleActions() {
+    this.triggerMethod('toggle:listType', 'actions');
+    this.$el.find('button').toggleClass('button--blue');
```
paulfalgout:
```
ui: {
  buttons: 'button',
},
...
this.ui.toggleClass('button--blue')
```

src/js/views/patients/worklist/filters_views.js
```diff
@@ -0,0 +1,93 @@
...
+const ClinicianClearButton = View.extend({
+  template: hbs`<button class="button-secondary worklist-list__clear-filter">{{ @intl.patients.worklist.filtersViews.clearClinicianFilter }}</button>`,
+  triggers: {
+    'click': 'click',
+  },
+});
+
+const TypeToggleView = View.extend({
+  template: hbs`
+    <button class="button-secondary {{#unless isFlowList}}button--blue{{/unless}} worklist-list__toggle-actions js-toggle-actions">{{far "file-alt"}}{{ @intl.patients.worklist.typeToggle.actionsButton }}</button>{{~ remove_whitespace ~}}
```
paulfalgout: worklist.filtersViews.typeToggleView.actionsButton

paulfalgout (changes requested):
cool.. so pretty much just the tooltip

test/integration/services/patient-quick-search.js
luketlancaster: it sure doesn't

paulfalgout (changes requested):
looks like a hair of coverage missing

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -69,7 +84,7 @@ export default App.extend({
     const groups = this.groups.clone();
 
     groups.unshift({
-      id: 'all-groups',
+      id: null,
       name: 'All Groups',
```
LisaManresa: Should this be i18n-ified?

LisaManresa (approved):
One small thing, but probably not a blocker since it's already in develop 😬

src/js/apps/patients/worklist/filters_app.js
luketlancaster: yeah you're right it should 100% be i18n'd

PR #263, opened 2020-03-30
Introduce cypress navigate

paulfalgout:
This PR does 3 things.
- Updates dependencies which may fail.. 🤞 for Cypress.  I have some trouble with "run all" locally in the gui, but I think it's a memory thing.  Also updates form.io @wweaver4th 
- Small modification to the component command to be more in line with cypress's recommended styles
- Add a navigate command for navigating within the app after the test visit.. something we constantly want to do for one reason or another.. and it _should_ handle the test issue in the clinician work.  This is surely to conflict with that PR, but it should be an easy conflict to resolve.


LisaManresa (approved):
🎉

test/integration/admin/list/clinicians-all.js
```diff
@@ -35,7 +36,7 @@ context('clinicians list', function() {
 
         return fx;
       })
-      .visit('/clinicians')
+      .navigate('/clinicians')
```
paulfalgout, line 39: @luketlancaster for your mid-specify `visit` desires

test/support/defaults.js
```diff
@@ -22,9 +22,11 @@ beforeEach(function() {
...
-    .routeFlows() // Setup default worklist route
     .routeStates()
     .routeRoles()
     .routeForms()
-    .routeGroupsBootstrap();
+    .routeGroupsBootstrap()
+    .routeFlows(); // Setup default route
 });
+
+Cypress.env('defaultRoute', '/worklist/owned-by/flows');
```
paulfalgout, line 32: @luketlancaster 👀 bout to merge this and it's going to 💥 the worklist branch

test/integration/admin/list/clinicians-all.js
luketlancaster: i knew if I made that mistake enough you'd figure out a way to make it so I wouldn't bug you with it anymore 🙃

PR #264, opened 2020-03-31
Repo cleanup

paulfalgout:
Removed unused files and suggestions from LGTM

Made a PR on the auth0 lib today and they're using LGTM and synk which are both free for open source and look like useful, quick, and cheap audits.


luketlancaster (approved):
💃

PR #265, opened 2020-04-01
Update sidebar timestamps w/new design

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch22172]

### DIFF
--- a/src/sass/modules/sidebar.scss
+++ b/src/sass/modules/sidebar.scss
@@ -21,12 +21,16 @@
   color: $white;
 }
 
-.sidebar__timestamps {
+.sidebar__footer {
   border-top: 1px solid $black-90;
   font-size: 12px;
   font-weight: 600;
   margin-top: 16px;
   padding-top: 16px;
+}
+
+.sidebar__footer-left {
+  margin-right: 32px;
 }
 
 .sidebar__label {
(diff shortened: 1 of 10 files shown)


src/sass/modules/sidebar.scss
```diff
@@ -29,6 +29,10 @@
   padding-top: 16px;
 }
 
+.sidebar__timestamps-created {
```
paulfalgout: Not much about this style is "created" related.  If we added two other timestamps below the two we have now what would be a better reusable class name?
Also "timestamps" and "created" are really pushing domain instead of generic.  I think `sidebar__timestamps` was a domain class renamed..  really this one and the one above should talk about more what this style is doing for the sidebar and not the content that it contains.

paulfalgout (changes requested):

src/sass/modules/sidebar.scss
luketlancaster: `.sidebar__footer` and `.sidebar__footer-spacing`?

paulfalgout: or `sidebar__footer--left` ?

luketlancaster: 🤘

paulfalgout (approved):

PR #267, opened 2020-04-07
Add/Remove clinician groups

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22179]

Still have to write tests for this, but wanted to get eyes on it to verify that this is the correct approach.


package-lock.json
```diff
@@ -6098,8 +6098,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
```
paulfalgout: don't think this should be checked in

src/js/services/bootstrap.js
```diff
@@ -20,6 +21,9 @@ export default App.extend({
   getCurrentOrg() {
     return this.currentOrg;
   },
+  getGroups() {
```
paulfalgout: probably don't need to add this right now.  it won't get coverage

src/js/views/admin/shared/clinician-groups.scss
```diff
@@ -0,0 +1,55 @@
+.clinician-groups {
+  .clinician-groups__item:first-child {
```
paulfalgout: does this BEM'd class need to be nested?

src/js/views/admin/shared/clinician-groups.scss
```diff
@@ -0,0 +1,55 @@
...
+  overflow: hidden;
+  padding: 8px 16px 8px 40px;
+  position: relative;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+.clinician-groups__item-icon .icon {
+  font-size: 16px;
+  left: 14px;
```
paulfalgout: `14px` seems problematic here and in the droplist.  it'd be one of the only cases in the app where we're not using a multiple of 4

src/js/views/admin/shared/clinicians_components.js
```diff
@@ -63,6 +66,126 @@ const RoleComponent = Droplist.extend({
...
+    this.listenTo(droplist, 'add:clinician', this.addClinicianGroup);
+  },
+  addClinicianGroup(group) {
+    this.clinicianGroups.add(group);
+  },
+  removeClinicianGroup(group) {
+    this.clinicianGroups.remove(group);
+  },
+  onAddClinicianGroup(group) {
+    Radio.request('entities', 'groups:assignClinician', group.id, this.getOption('clinician'));
```
paulfalgout: I think this should be `group.addClinician(clinician);`  because it needs to update both `group.get('_clinicians')` and `clinician.get('_groups')` or the next time `clinician.getGroups()` or `group.getClinicians()` is called it'll be wrong.

src/js/views/admin/shared/clinicians_components.js
```diff
@@ -63,6 +66,122 @@ const RoleComponent = Droplist.extend({
...
+  },
+});
+
+const ClinicianGroupsView = CollectionView.extend({
+  tagName: 'ul',
+  className: 'clinician-groups',
+  childView: GroupsItemView,
+  childViewOptions() {
+    return {
+      disabled: this.getOption('state').isDisabled,
```
paulfalgout: doesn't seem to need disabled

src/js/views/admin/shared/clinicians_components.js
```diff
@@ -63,6 +66,122 @@ const RoleComponent = Droplist.extend({
...
+    'remove': 'remove',
+  },
+  onRemove(child) {
+    this.triggerMethod('remove:clinician', child.model);
+  },
+});
+
+const GroupsDropList = Droplist.extend({
+  initialize() {
+    this.listenTo(this.getOption('clinicianGroups'), 'add remove', this.show);
```
paulfalgout: can be `update` instead of `add remove`

src/js/views/admin/shared/clinicians_components.js
```diff
@@ -63,6 +66,122 @@ const RoleComponent = Droplist.extend({
...
+          text: clinicianGroups.length > 0 ? 'Add Group...' : 'Select Group...',
+        };
+      },
+    };
+  },
+  picklistOptions: {
+    attr: 'name',
+    itemTemplate: hbs`<a>{{matchText text query}}</a>`,
+  },
+  onChangeSelected(selected) {
```
paulfalgout: elsewhere we're overriding `onPicklistSelect` to prevent state selection.. which should also potentially mean you don't need to override `itemTemplate` above.  Also is there a `!selected` scenario?

src/js/views/admin/shared/clinicians_components.js
```diff
@@ -63,6 +66,122 @@ const RoleComponent = Droplist.extend({
...
+    });
+  },
+  onRender() {
+    this.showGroups();
+    this.showDroplist();
+  },
+  showGroups() {
+    const groupsView = this.showChildView('groups', new ClinicianGroupsView({
+      collection: this.clinicianGroups,
+      state: this.getOption('state') || {},
```
paulfalgout: may not need the state

src/js/views/admin/shared/clinicians_components.js
```diff
@@ -63,6 +66,122 @@ const RoleComponent = Droplist.extend({
...
+  template: hbs`
+    <div data-groups-region></div>
+    <div data-droplist-region></div>
+  `,
+  regions: {
+    groups: '[data-groups-region]',
+    droplist: '[data-droplist-region]',
+  },
+  initialize() {
+    this.currentClinician = Radio.request('bootstrap', 'currentUser');
```
paulfalgout, line 124: might be worth a `this.clinician = this.getOption('clinician');`

paulfalgout (changes requested):
:-D you're implementing some of these suggestions already.

src/js/views/admin/shared/clinicians_components.js
paulfalgout: ok so just cleared with mark that we don't need to switch the text.  It can always say "Add Group..." so we should no longer need to pass `clinicianGroups` to this droplist

paulfalgout: _however_ we probably need to listen to `this.collection` `update` like here.. and onShow disable the component if `!groups.length`..

src/js/views/admin/shared/clinician-groups.scss
LisaManresa: This is because the Fontawesome "users" icon is 4px wider than the "user-circle" icon. So the button text does in the Role and Groups components does not align properly without that 14px definition. Is there a better way to account for this? I was basing this off of the Marvel https://marvelapp.com/7ihgbgb/screen/63561658/handoff

paulfalgout: elsewhere we've done
```css
  // Handles odd width icon spacing
  margin-right: -2px;
```
At least there should be a comments as to why it's not 4x, but if we can handle it the same way that's probably ideal.

LisaManresa: Ah perfect, thanks.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -98,6 +100,21 @@ const SidebarView = View.extend({
...
+    const isDisabled = this.model.isNew();
+
+    const groups = this.showChildView('groups', new GroupsComponent({
+      clinician: this.model,
+      state: {
+        isDisabled,
+      },
+    }));
+
+    this.listenTo(groups, {
```
paulfalgout: can remove?

src/js/entities-service/entities/groups.js
```diff
@@ -11,6 +13,45 @@ const _Model = BaseModel.extend({
   getClinicians() {
     return Radio.request('entities', 'clinicians:collection', this.get('_clinicians'));
   },
+  addClinician(clinician) {
+    const url = `/api/groups/${ this.id }/relationships/clinicians`;
+    const group = this.toRelation(this);
+
+    clinician.set({ _groups: clinician.get('_groups').concat({ id: this.id }) }, {
+      relationships: { group },
```
paulfalgout: I don't think there's a need to update relationships on a set.  not sure that really does anything.  might not need `toRelation` in this method

src/js/entities-service/entities/groups.js
```diff
@@ -11,6 +13,45 @@ const _Model = BaseModel.extend({
...
+    clinician = this.toRelation(clinician);
+
+    this.set({ _clinicians: this.get('_clinicians').concat(clinician.data) }, {
+      relationships: { clinician },
+    });
+
+    $.ajax({
+      type: 'POST',
+      url,
+      data: JSON.stringify({ data: [clinician.data] }),
```
paulfalgout: does it expect the data to be in an array here and in the delete?  I would expect it's just looking for `{ id, type }` ?

src/js/views/admin/shared/clinician-groups.scss
```diff
@@ -0,0 +1,56 @@
...
+  .icon {
+    font-size: 12px;
+  }
+}
+
+.clinician-groups__droplist {
+  border-radius: 0;
+  width: 100%;
+
+  .icon {
```
paulfalgout, line 53: same `14px` issue here

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -98,6 +100,21 @@ const SidebarView = View.extend({
 
     this.showChildView('role', roleComponent);
   },
+  showGroups() {
+    const isDisabled = this.model.isNew();
```
paulfalgout: it looks like the group component itself is handling this

src/js/views/admin/shared/clinicians_components.js
```diff
@@ -63,6 +66,107 @@ const RoleComponent = Droplist.extend({
...
+  },
+  viewOptions: {
+    className: 'button-secondary clinician-groups__droplist',
+    template: hbs`{{far "users"}}Add Group...`,
+  },
+  picklistOptions: {
+    attr: 'name',
+  },
+  onShow() {
+    if (this.getOption('clinician').isNew()) {
```
paulfalgout, line 101: this seems ok, but after looking at it.. I think maybe the `listenTo` and the state setting should happen up in the parent component and just redo the droplist component.. otherwise modifying the state on the parent would have no effect.  ie: `myGroupsComponent.setState('disabled', true)`  _or_ we just don't pass the state down at all.  But right now we're only supporting the initial state which seem incomplete.

src/js/views/admin/shared/clinicians_components.js
```diff
@@ -63,6 +66,107 @@ const RoleComponent = Droplist.extend({
...
+      collection: this.clinicianGroups,
+    }));
+
+    this.listenTo(groupsView, 'remove:clinician', this.removeClinicianGroup);
+  },
+  showDroplist() {
+    const droplist = this.showChildView('droplist', new GroupsDropList({
+      collection: this.groups,
+      clinicianGroups: this.clinicianGroups,
+      state: this.getOption('state') || {},
```
paulfalgout: if we pass the state down this should likely be `state: this.getState().attributes,`

paulfalgout (changes requested):

src/js/entities-service/entities/groups.js
LisaManresa: The API docs seem to indicate that the endpoints accept an array. But is it possible that it can accept either a single clinician or an array?

src/js/views/admin/shared/clinicians_components.js
LisaManresa: I'd lean toward keeping the state self-contained within the Droplist component. Passing the state into it was extraneous anyway, it wasn't utilizing it at all.

src/js/entities-service/entities/groups.js
paulfalgout: yeah you're right.. that's surprising, but I guess it's easy enough to implement.

LisaManresa: I assumed it was to allow for bulk adding/removing.

src/js/entities-service/entities/groups.js
```diff
@@ -11,6 +13,51 @@ const _Model = BaseModel.extend({
   getClinicians() {
     return Radio.request('entities', 'clinicians:collection', this.get('_clinicians'));
   },
+  addClinician(clinician) {
+    const url = `/api/groups/${ this.id }/relationships/clinicians`;
+    const groups = clinician.get('_groups') || [];
+
+    clinician.set({ _groups: groups.concat({ id: this.id }) });
+
+    this.set({ _clinicians: this.get('_clinicians').concat({
```
paulfalgout: so had a thought here and above.. I _think_ that `_clinicians` isn't going to get  a change event because `concat` modifies the data on the attributes.  So maybe `_.union` would get around that along with matching the `_.reject` pattern below.

src/js/entities-service/entities/groups.js
```diff
@@ -11,6 +13,51 @@ const _Model = BaseModel.extend({
...
+      id: clinician.id,
+      type: clinician.type,
```
paulfalgout: only needs to be id on the `_foo` relationships.  I think adding type would actually end up adding type to the data.

src/js/entities-service/entities/groups.js
```diff
@@ -11,6 +13,51 @@ const _Model = BaseModel.extend({
...
+      }),
+    });
+  },
+  removeClinician(clinician) {
+    const url = `/api/groups/${ this.id }/relationships/clinicians`;
+
+    clinician.set({ _groups: _.reject(clinician.get('_groups'), { id: this.id }) });
+
+    this.set({
+      _clinicians: _.reject(this.get('_clinicians'), {
```
paulfalgout: I think like above `type` shouldn't be here.

src/js/views/admin/shared/clinicians_components.js
```diff
@@ -63,6 +66,106 @@ const RoleComponent = Droplist.extend({
...
+    const groupsView = this.showChildView('groups', new ClinicianGroupsView({
+      collection: this.clinicianGroups,
+    }));
+
+    this.listenTo(groupsView, 'remove:clinician', this.removeClinicianGroup);
+  },
+  showDroplist() {
+    const droplist = this.showChildView('droplist', new GroupsDropList({
+      collection: this.groups,
+      clinicianGroups: this.clinicianGroups,
```
paulfalgout: don't need to pass the `clinicianGroups` here.

src/js/views/admin/shared/clinicians_components.js
```diff
@@ -63,6 +66,106 @@ const RoleComponent = Droplist.extend({
...
+    <div data-groups-region></div>
+    <div data-droplist-region></div>
+  `,
+  regions: {
+    groups: '[data-groups-region]',
+    droplist: '[data-droplist-region]',
+  },
+  initialize() {
+    this.currentClinician = Radio.request('bootstrap', 'currentUser');
+    this.clinician = this.getOption('clinician');
```
paulfalgout: options gets passed to initialize so this could be
```
initialize({ clinician }) {
  this.clinician = clinician;
```

paulfalgout (changes requested):

src/js/entities-service/entities/groups.js
```diff
@@ -11,6 +13,45 @@ const _Model = BaseModel.extend({
   getClinicians() {
     return Radio.request('entities', 'clinicians:collection', this.get('_clinicians'));
   },
+  addClinician(clinician) {
+    const url = `/api/groups/${ this.id }/relationships/clinicians`;
+    const groups = clinician.get('_groups') || [];
+
+    clinician.set({ _groups: groups.concat({ id: this.id }) });
```
paulfalgout: same here.. I don't think and existing groups array will get a change event on this line

paulfalgout (changes requested):
one last little one

src/js/entities-service/entities/groups.js
LisaManresa: bah, forgot I was doing that in two places. Thanks

PR #268, opened 2020-04-09
Use program-action relationship for actions instead of program

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch22267]
- [ ] Add QA tasks

I _think_ this is it? Ran the sidebar and dashboard tests locally...we'll see what I missed!!!

### DIFF
--- a/src/js/entities-service/entities/program-actions.js
+++ b/src/js/entities-service/entities/program-actions.js
@@ -25,7 +25,7 @@ const _Model = BaseModel.extend({
     const currentUser = Radio.request('bootstrap', 'currentUser');
     const currentOrg = Radio.request('bootstrap', 'currentOrg');
     const states = currentOrg.getStates();
-    const action = this.pick('name', 'details', '_owner', '_program', '_form');
+    const action = this.pick('name', 'details', '_owner', '_form');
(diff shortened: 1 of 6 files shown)


src/js/entities-service/entities/program-actions.js
```diff
@@ -25,19 +25,20 @@ const _Model = BaseModel.extend({
...
-      _owner: action._owner || {
+      '_patient': patientId,
+      '_state': states.at(0).id,
+      '_owner': action._owner || {
         id: currentUser.id,
         type: 'clinicians',
       },
-      duration: 0,
-      due_date: dueDate,
+      '_program_action': this.id,
```
paulfalgout: no need to introduce the `'`s only if it needed to be `_program-action`  `_`s are ok

paulfalgout (changes requested):

src/js/entities-service/entities/program-actions.js
luketlancaster: ah yeah. leftover from when I was Doing It Wrong

paulfalgout (approved):
OK.  so now we need to make a release from this branch and get it into the BE branch.. After that branch is merged.  We can merge this one.

PR #271, opened 2020-04-14
Action sidebar comment form

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22280]

### DIFF
--- a/src/js/entities-service/entities/comments.js
+++ b/src/js/entities-service/entities/comments.js
@@ -1,5 +1,5 @@
-// import _ from 'underscore';
-// import Radio from 'backbone.radio';
+import _ from 'underscore';
+import Radio from 'backbone.radio';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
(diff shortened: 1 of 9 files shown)


src/js/views/patients/sidebar/action/action-sidebar.scss
```diff
@@ -1,3 +1,26 @@
 .action-sidebar__form {
   color: color(blue);
 }
+
+.sidebar__comment-form {
+  margin-top: 16px;
+}
+
+.sidebar__comment-label {
```
LisaManresa: @luketlancaster, we should sync up on this. Let me know if this solution works for you or if you already have another way of displaying the initials label.

paulfalgout: I could maybe see moving this to a comment module.  We don't use it elsewhere, but there's nothing about this comment form that is specifically sidebar.

src/js/entities-service/entities/comments.js
```diff
@@ -8,24 +8,26 @@ const TYPE = 'comments';
...
-//    };
-//
-//    return this.save(attrs, { relationships }, { wait: true });
-//  },
+  urlRoot() {
+    if (this.isNew()) return `/api/actions/${ this.get('_action') }/relationships/comments`;
+
+    return '/api/comments';
+  },
+  validate({ body }) {
```
paulfalgout: I believe the body of a comment is `message`

src/js/views/patients/sidebar/action/action-sidebar.hbs
```diff
@@ -35,4 +35,5 @@
   </div>
   {{/unless}}
 </div>
+<div data-comment-region></div>
```
paulfalgout, line 38: in the comp it looks to me like this is bottom aligned if there are no activities

src/js/views/patients/sidebar/action/action-sidebar.scss
```diff
@@ -1,3 +1,26 @@
 .action-sidebar__form {
   color: color(blue);
 }
+
+.sidebar__comment-form {
```
paulfalgout: could maybe just use `u-margin--t-16`

src/js/views/patients/sidebar/action/action-sidebar_views.js
```diff
@@ -307,6 +392,9 @@ const LayoutView = View.extend({
   showDisabledSave() {
     this.showChildView('save', new DisabledSaveView());
   },
+  showCommentForm() {
+    this.showChildView('comment', new CommentFormView({ actionId: this.model.id }));
```
paulfalgout: I think we'll probably want to modify this form in the future to handle not only new comments but edit them as well.  It may be best to pass in the comment model.   Really I think it might be best to pull showing the comment view out to the app level and have the app construct the data.  The app will also probably need to eventually add the successfully posted comment to the comment collection so should probably handle the `saveAll` as well.

paulfalgout (changes requested):

src/js/views/patients/sidebar/action/action-sidebar.scss
luketlancaster: this all works for me. I haven't done anything other than the most basic "I wonder what this classname should be" thinking, so I can just make whatever you have work

LisaManresa: That works for me. I can make adjustments to the classnames.

src/js/entities-service/entities/comments.js
LisaManresa: Ah thanks, meant to ask about that.

src/js/views/patients/sidebar/action/action-sidebar.hbs
LisaManresa: The activities container grows to fill the height regardless of how many activity events there are. So this stays bottom-aligned even if there is only the "created" event listed. At least in Chrome, I'll check other browsers.

paulfalgout (approved):

PR #272, opened 2020-04-15
Display comments in activity section of action sidebar

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch22279]

### DIFF
--- a/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
+++ b/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
@@ -5,6 +5,10 @@ import Radio from 'backbone.radio';
 import hbs from 'handlebars-inline-precompile';
 import { View, CollectionView } from 'marionette';
 
+import { renderTemplate } from 'js/i18n';
+
+import Tooltip from 'js/components/tooltip';
+
 import 'sass/modules/comments.scss';
 import 'sass/modules/sidebar.scss';
 
@@ -94,10 +98,13 @@ const CommentView = View.extend({
       <span class="comment__author-label">{{ initials }}</span>
       <span class="comment__author-name">{{ name }}</span>
       <span class="comment__timestamp">{{ formatMoment created_at "AT_TIME" }}</span>
-      {{#if canEdit}}<span class="comment__icon">{{far "pen"}} Edit</span>{{/if}}
+      {{#if canEdit}}<span class="js-edit comment__icon">{{far "pen"}} {{ @intl.patients.sidebar.action.activityViews.commentView.edit }}</span>{{/if}}
     </div>
-    <div class="comment__message">{{ message }}{{#if edited_at}}<span class="comment__edited"> (Edited) </span>{{/if}}</div>
+    <div class="comment__message">{{ message }}{{#if edited_at}}<span class="comment__edited"> {{ @intl.patients.sidebar.action.activityViews.commentView.edited }} </span>{{/if}}</div>
   `,
+  ui: {
+    edit: '.js-edit',
+  },
   templateContext() {
     const clinician = this.model.getClinician();
     const currentUser = Radio.request('bootstrap', 'currentUser');
@@ -107,6 +114,18 @@ const CommentView = View.extend({
       name: clinician.get('name'),
       initials: clinician.getInitials(),
     };
+  },
+  onRender() {
+    const edited = this.model.get('edited_at');
+    if (!edited) return;
+
+    const template = hbs`{{formatHTMLMessage (intlGet "patients.sidebar.action.activityViews.commentView.editTooltip")  edited = (formatMoment edited "TIME") }}`;
+
+    new Tooltip({
+      messageHtml: renderTemplate(template, { edited }),
+      uiView: this,
+      ui: this.ui.edit,
+    });
   },
 });
 
--- a/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
+++ b/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
@@ -1,6 +1,11 @@
+import moment from 'moment';
+
+import Radio from 'backbone.radio';
+
 import hbs from 'handlebars-inline-precompile';
 import { View, CollectionView } from 'marionette';
 
+import 'sass/modules/comments.scss';
 import 'sass/modules/sidebar.scss';
 
 import './action-sidebar.scss';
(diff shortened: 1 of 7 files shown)


src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -82,10 +87,22 @@ const StateUpdatedTemplate = hbs`
...
+    </div>
+    <div class="comment__message">{{ message }}{{#if edited_at}}<span class="comment__edited"> (Edited) </span>{{/if}}</div>
+  </div>
+`;
+
 const ActivityView = View.extend({
   className: 'u-margin--b-8',
   getTemplate() {
-    const type = this.model.get('type');
+    const type = this.model.get('type') || this.model.type;
```
paulfalgout: rather than use a template here can we make a `CommentView` and make `childView` a function on the collection view that checks the type

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -130,6 +159,15 @@ const ActivitiesView = CollectionView.extend({
...
     return true;
   },
+  viewComparator({ model }) {
+    const date = this._getSortDate(model);
+    return - moment(date).format('x');
+  },
+  _getSortDate(model) {
+    if (model.get('date')) return model.get('date');
+
+    return model.get('updated_at');
```
paulfalgout: created_at

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -82,10 +87,22 @@ const StateUpdatedTemplate = hbs`
   <div>{{formatMoment date "AT_TIME"}}</div>
 `;
 
+const CommentTemplate = hbs`
+  <div>
+    <div class="flex">
+      <span class="comment__author-label">{{ initials }}</span>
+      <span class="comment__author-name">{{ name }}</span>
+      <span class="comment__timestamp">{{ formatMoment updated_at "AT_TIME" }}</span>
```
paulfalgout: created_at

src/sass/modules/comments.scss
```diff
@@ -14,6 +13,35 @@
...
+  margin-left: auto;
+}
+
 .comment__input {
   min-height: 44px;
 }
+
+.comment__message {
+  color: $black-20;
+  margin-left: 34px;
```
paulfalgout: hmm yeah.. not divisible by 4.. this should probably be 32px and we should fix whatever is causing this.  Is it the author label?  That should also really be 24px.

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -673,6 +673,116 @@ context('action sidebar', function() {
...
 
+  specify('action comments', function() {
+    cy
+      .server()
+      .routePatient()
+      .routePatientActions()
+      .routePatientFlows()
+      .routeAction(fx => {
+        fx.data.id = '12345';
+        fx.data.relationships['program-action'] = { data: { id: '1' } };
```
paulfalgout: Do we need to test program-action or form here?  Really we may not even need to set the id

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -673,6 +673,116 @@ context('action sidebar', function() {
...
+        fx.data.relationships.form = { data: { id: '11111' } };
+        return fx;
+      })
+      .routeActionActivity(fx => {
```
paulfalgout, line 683: Is this the simplest activity we could use for this test?

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -673,6 +673,116 @@ context('action sidebar', function() {
...
+
+        fx.data[2].relationships.clinician.data = { id: '22222' };
+        fx.data[2].attributes.updated_at = moment(now).subtract(4, 'days').format();
+        fx.data[2].attributes.message = 'Message from Someone Else';
+        fx.data[2].attributes.edited_at = null;
+
+
+        return fx;
+      })
+      .routeProgramByAction(fx => {
```
paulfalgout: probably don't need a program

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -673,6 +673,116 @@ context('action sidebar', function() {
...
+      .wait('@routePatientActions')
+      .wait('@routeAction')
+      .wait('@routeActionActivity')
+      .wait('@routeActionComments')
+      .wait('@routeProgramByAction');
+
+    cy
+      .get('[data-activity-region]')
+      .find('.u-margin--b-8')
+      .first()
```
paulfalgout: `.eq(1)`

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -673,6 +673,116 @@ context('action sidebar', function() {
...
+      .should('contain', 'CM')
+      .should('contain', 'Clinician McTester')
+      .should('contain', 'Edit')
+      .should('contain', 'Most Recent Message from Clinician McTester')
+      .should('contain', '(Edited)');
+
+    cy
+      .get('[data-activity-region]')
+      .find('.u-margin--b-8')
+      .first()
```
paulfalgout: `.eq(2)`

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -82,10 +87,22 @@ const StateUpdatedTemplate = hbs`
   <div>{{formatMoment date "AT_TIME"}}</div>
 `;
 
+const CommentTemplate = hbs`
+  <div>
```
paulfalgout: does this need the outer wrapping `div`?

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -673,6 +673,116 @@ context('action sidebar', function() {
...
+      .wait('@routePatient')
+      .wait('@routePatientActions')
+      .wait('@routeAction')
+      .wait('@routeActionActivity')
+      .wait('@routeActionComments')
+      .wait('@routeProgramByAction');
+
+    cy
+      .get('[data-activity-region]')
+      .find('.u-margin--b-8')
```
paulfalgout: How about `td` instead of a utility class.  or add a `qa-activity-item` class.  I don't think we should test off of utility classes if it is avoidable.

paulfalgout (changes requested):

paulfalgout:
Let's add a tooltip on the "Editing" that shows the edited_at timestamp with `TIME` format.  Also the broken test looks legit.

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
luketlancaster: nope sure doesn't

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -82,8 +91,46 @@ const StateUpdatedTemplate = hbs`
...
+      edited: this.model.get('edited_at'),
+    };
+  },
+  onRender() {
+    if (!this.model.get('edited_at')) return;
+
+    const template = hbs`{{formatHTMLMessage (intlGet "patients.sidebar.action.activityViews.commentView.editTooltip")  edited = (formatMoment edited "TIME") }}`;
+
+    new Tooltip({
+      messageHtml: renderTemplate(template, this.templateContext()),
```
paulfalgout: I think this should be 
```
const edited = this.model.get('edited_at');
if (!edited) return;

...
renderTemplate(template, { edited });
```
no need to add this to `templateContext` for the view.

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -673,6 +673,108 @@ context('action sidebar', function() {
...
+      .find('.qa-activity-item')
+      .eq(1)
+      .should('contain', 'CM')
+      .should('contain', 'Clinician McTester')
+      .should('contain', 'Edit')
+      .should('contain', 'Most Recent Message from Clinician McTester')
+      .should('contain', '(Edited)');
+
+    cy
+      .get('.js-edit')
```
paulfalgout: can this be a find within a get of something specific to this area.. if we add a js-edit to the main menu or something in the future I don't want the test to break.

paulfalgout (changes requested):

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -82,8 +91,47 @@ const StateUpdatedTemplate = hbs`
...
+  },
+  templateContext() {
+    const clinician = this.model.getClinician();
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+
+    return {
+      canEdit: clinician.id === currentUser.id,
+      name: clinician.get('name'),
+      initials: clinician.getInitials(),
+      edited: this.model.get('edited_at'),
```
paulfalgout: this line isn't needed

paulfalgout (changes requested):

PR #275, opened 2020-04-17
edit comments in sidebar activity

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22306]

### DIFF
--- a/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
+++ b/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
@@ -9,6 +9,8 @@ import { renderTemplate } from 'js/i18n';
 
 import Tooltip from 'js/components/tooltip';
 
+import { CommentFormView } from 'js/views/patients/shared/comments_views';
+
 import 'sass/modules/comments.scss';
 import 'sass/modules/sidebar.scss';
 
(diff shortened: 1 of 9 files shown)


src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -129,6 +133,58 @@ const CommentView = View.extend({
...
+  },
+  template: hbs`
+    <div data-display-region></div>
+    <div class="sidebar_activity--comment-form" data-form-region></div>
+  `,
+  regions: {
+    display: '[data-display-region]',
+    form: '[data-form-region]',
+  },
+  onRender() {
```
paulfalgout: I think the `CommentDisplayView` could possibly just have a wrapping div that's a region.. then `onClickEdit` just show the comment form in that region.. and on save or cancel re-render that view which will clear the form rather than adding data to the model.

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -129,6 +133,58 @@ const CommentView = View.extend({
...
+  },
+  onClickEdit() {
+    this.model.set('editing', true);
+  },
+  onCancelEdit() {
+    this.model.set('editing', false);
+  },
+  onSaveComment({ model }) {
+    this.model.save({
+      message: model.attributes.message,
```
paulfalgout: probably `.get('message')`

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -129,6 +133,58 @@ const CommentView = View.extend({
...
+    });
+    this.model.set({
+      'editing': false,
```
paulfalgout: quotes aren't needed here.  Can we add the `edited_at` to the `save` so there's just a single change event?  If we ever do anything off the change event in the future, the `edited_at` won't be updated.  hopefully the server would just ignore it?  I believe that's true.  Also it needs to be `moment.utc().format()`

paulfalgout (changes requested):

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -93,18 +95,31 @@ const StateUpdatedTemplate = hbs`
 
 const CommentView = View.extend({
```
luketlancaster, line 96: would it make sense to move this over to `comments_views` and import it here?

LisaManresa: I've re-written this response 4 times now because I really can't decide where it should live. I was worried this might be too specific to the sidebar to live in a shared file.

paulfalgout: Yeah I think this is an activity view representing a comment, and the other view is very comment related.  _maybe_ activity at some point would be extracted from the sidebar, but for the moment I'd argue it's more related to activity than the comment

luketlancaster: makes sense to me. That was as much me wondering aloud than anything else, so I'm happy with it staying put

PR #278, opened 2020-04-20
Show info when clinician has no groups or role assigned

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22318]


paulfalgout (changes requested):
I'm not sure this should be part of the group component.  We're waiting on what it will look like if a role is not filled out, and that warning message may go in the same location.  Either way it seems like that message could be handled outside of this component.

PR #279, opened 2020-04-22
Add clinician admin access type

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch22315]


src/js/static.js
```diff
@@ -4,6 +4,22 @@ const PatientStatusIcons = {
...
+const ACCESS_TYPES = [
+  {
+    id: 'employee',
+    name: 'Employee',
+    details: 'can access Workspace but not Admin',
+  },
+  {
+    id: 'manager',
+    name: 'Manager',
+    details: 'can update Workplace and Admin',
```
LisaManresa: Are "Workplace" and "Workspace" two different things?

LisaManresa (approved):

src/js/static.js
paulfalgout: 🤦

PR #280, opened 2020-04-22
Display last active on clinicians list

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch###]

### DIFF
--- a/src/js/views/admin/list/clinicians-all_views.js
+++ b/src/js/views/admin/list/clinicians-all_views.js
@@ -47,8 +47,9 @@ const ItemView = View.extend({
   },
   template: hbs`
     <td class="table-list__cell w-20">{{#unless name}}{{ @intl.admin.list.cliniciansAllViews.itemView.newClinician }}{{/unless}}{{ name }}</td>
-    <td class="table-list__cell w-40">{{#each groups}}{{#unless @first}}, {{/unless}}{{ this.name }}{{/each}}</td>
+    <td class="table-list__cell w-30">{{#each groups}}{{#unless @first}}, {{/unless}}{{ this.name }}{{/each}}</td>
     <td class="table-list__cell w-30"><span class="u-margin--r-8" data-access-region></span><span data-role-region></span></td>
+    <td class="table-list__cell w-20 {{#unless last_active_at}}table-list__cell--empty{{/unless}}">{{formatMoment last_active_at "TIME_OR_DAY" defaultHtml=(intlGet "admin.list.cliniciansAllViews.itemView.noLastActive")}}</td>
   `,
   templateContext() {
     return {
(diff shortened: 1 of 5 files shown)


test/integration/admin/lists/clinicians-all.js
```diff
@@ -34,6 +38,7 @@ context('clinicians list', function() {
         fx.data[0].id = '1';
         fx.data[0].attributes.name = 'Test Clinician';
         fx.data[0].attributes.access = 'employee';
+        fx.data[0].attributes.last_active_at = local.format();
```
paulfalgout: a date coming from an endpoint should always be `utc`

test/integration/admin/lists/clinicians-all.js
```diff
@@ -1,4 +1,8 @@
 import _ from 'underscore';
+import formatDate from 'helpers/format-date';
+
+const moment = require('moment');
```
paulfalgout: `import`

paulfalgout (changes requested):
https://marvelapp.com/7ihgbgb/screen/63562970
Need to support a `null` value.

src/js/views/admin/list/clinicians-all_views.js
```diff
@@ -47,8 +47,15 @@ const ItemView = View.extend({
   },
   template: hbs`
     <td class="table-list__cell w-20">{{#unless name}}{{ @intl.admin.list.cliniciansAllViews.itemView.newClinician }}{{/unless}}{{ name }}</td>
-    <td class="table-list__cell w-40">{{#each groups}}{{#unless @first}}, {{/unless}}{{ this.name }}{{/each}}</td>
+    <td class="table-list__cell w-30">{{#each groups}}{{#unless @first}}, {{/unless}}{{ this.name }}{{/each}}</td>
     <td class="table-list__cell w-30"><span class="u-margin--r-8" data-access-region></span><span data-role-region></span></td>
+    <td class="table-list__cell w-20 {{#unless last_active_at}}table-list__cell--empty{{/unless}}">
+      {{#if last_active_at}}
```
paulfalgout: There is a way to set the value for null https://github.com/RoundingWell/RWell/blob/81f50ca2a61980ad000faaf20018d8658d92bba1/mainapp/assets/js/views/clinicians/clinicians/clinicians_views.js#L103
It's not _that_ much different, but I think I'd like to establish that pattern in case it is more useful in the future.

paulfalgout (changes requested):

src/js/views/admin/list/clinicians-all_views.js
LisaManresa: Oh nice, that's much cleaner.

paulfalgout (approved):

PR #282, opened 2020-04-23
Clinician edit

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22329]

### DIFF
--- a/src/js/apps/admin/sidebar/clinician-sidebar_app.js
+++ b/src/js/apps/admin/sidebar/clinician-sidebar_app.js
@@ -8,9 +8,7 @@ export default App.extend({
   onBeforeStart({ clinician }) {
     this.clinician = clinician;
 
-    this.showView(new SidebarView({
-      model: this.clinician,
-    }));
+    this.showView(new SidebarView({ clinician }));
   },
   viewEvents: {
     'save': 'onSave',
@@ -18,18 +16,20 @@ export default App.extend({
     'delete': 'onDelete',
   },
   onSave({ model }) {
-    this.clinician.save(model.attributes).done(() => {
+    this.clinician.save(model.attributes).then(() => {
       Radio.trigger('event-router', 'clinician', this.clinician.id);
+    }, ({ responseJSON }) => {
+      const errors = this.clinician.parseErrors(responseJSON);
+      this.getView().showErrors(errors);
     });
   },
-  // onDelete() {
-  //   this.clinician.trigger('editing', false);
-  //   this.clinician.destroy();
-  //   this.stop();
-  // },
+  onDelete() {
+    this.clinician.destroy();
+    this.stop();
+  },
   onStop() {
     this.clinician.trigger('editing', false);
-    if (this.clinician && this.clinician.isNew()) this.clinician.destroy();
+    if (this.clinician.isNew()) this.clinician.destroy();
 
     Radio.request('sidebar', 'close');
     Radio.trigger('event-router', 'clinicians:all');
--- a/src/js/views/admin/sidebar/clinician/clinician-sidebar.hbs
+++ b/src/js/views/admin/sidebar/clinician/clinician-sidebar.hbs
@@ -4,7 +4,9 @@
       <h3 class="sidebar__heading">Clinician</h3>
     </div>
     <div>
-      {{!-- <button class="button--icon u-margin--r-8 js-menu">{{far "ellipsis-h"}}</button> --}}
+      {{#if showDelete}}
+        <button class="button--icon u-margin--r-8 js-menu">{{far "ellipsis-h"}}</button>
+      {{/if}}
       <button class="button--icon js-close">{{fas "times"}}</button>
     </div>
   </div>
--- a/src/js/views/admin/sidebar/clinician/clinician-sidebar.scss
+++ b/src/js/views/admin/sidebar/clinician/clinician-sidebar.scss
@@ -7,3 +7,11 @@
     opacity: 0.5;
   }
 }
+
+.clinician-sidebar__locked-icon {
+  color: $black-60;
+  line-height: 24px;
+  position: absolute;
+  right: 16px;
+  top: 8px;
+}
--- a/src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
+++ b/src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
@@ -1,19 +1,39 @@
+import _ from 'underscore';
+import Backbone from 'backbone';
 import hbs from 'handlebars-inline-precompile';
+import Radio from 'backbone.radio';
 import { View } from 'marionette';
 
 import 'sass/modules/buttons.scss';
 import 'sass/modules/forms.scss';
 import 'sass/modules/sidebar.scss';
 
+import intl from 'js/i18n';
+
+import { animSidebar } from 'js/anim';
+
 import { GroupsComponent, RoleComponent, AccessComponent } from 'js/views/admin/shared/clinicians_components';
+import Optionlist from 'js/components/optionlist';
 
 import ClinicianSidebarTemplate from './clinician-sidebar.hbs';
 
 import './clinician-sidebar.scss';
 
+const i18n = intl.admin.sidebar.clinician.clinicianSidebarViews;
+
 const NameView = View.extend({
   className: 'pos--relative',
-  template: hbs`<input class="input-primary w-100 js-input" placeholder="{{ @intl.admin.sidebar.clinician.clinicianSidebarViews.nameView.placeholder }}" value="{{ name }}" />`,
+  template: hbs`
+    <input class="input-primary w-100 js-input{{#if error}} has-error{{/if}}" placeholder="{{ @intl.admin.sidebar.clinician.clinicianSidebarViews.nameView.placeholder }}" value="{{ name }}" {{#unless canEdit}}disabled{{/unless}} />{{~ remove_whitespace ~}}
+    {{#unless canEdit}}<span class="clinician-sidebar__locked-icon">{{far "lock"}}</span>{{/unless}}
+    {{#if error}}<div class="form-error u-margin--b-4">{{ error }}</div>{{/if}}
+  `,
+  templateContext() {
+    return {
+      canEdit: this.model.isEditable(),
+      error: this.getOption('error'),
+    };
+  },
   ui: {
     input: '.js-input',
   },
@@ -32,7 +52,17 @@ const NameView = View.extend({
 
 const EmailView = View.extend({
   className: 'pos--relative',
-  template: hbs`<input class="input-primary w-100 js-input" placeholder="{{ @intl.admin.sidebar.clinician.clinicianSidebarViews.emailView.placeholder }}" value="{{ email }}" />`,
+  template: hbs`
+    <input class="input-primary w-100 js-input{{#if error}} has-error{{/if}}" placeholder="{{ @intl.admin.sidebar.clinician.clinicianSidebarViews.emailView.placeholder }}" value="{{ email }}" {{#unless canEdit}}disabled{{/unless}} />{{~ remove_whitespace ~}}
+    {{#unless canEdit}}<span class="clinician-sidebar__locked-icon">{{far "lock"}}</span>{{/unless}}
+    {{#if error}}<div class="form-error u-margin--b-4">{{ error }}</div>{{/if}}
+  `,
+  templateContext() {
+    return {
+      canEdit: this.model.isEditable(),
+      error: this.getOption('error'),
+    };
+  },
   ui: {
     input: '.js-input',
   },
@@ -69,14 +99,19 @@ const InfoView = View.extend({
 });
 
 const SidebarView = View.extend({
-  modelEvents: {
-    'change:_groups change:_role': 'showInfo',
-  },
   className: 'sidebar flex-region',
   template: ClinicianSidebarTemplate,
+  templateContext() {
+    return {
+      showDelete: this.clinician.isEditable(),
+    };
+  },
   triggers: {
     'click .js-close': 'close',
-    // 'click .js-menu': 'click:menu',
+    'click @ui.menu': 'click:menu',
+  },
+  ui: {
+    menu: '.js-menu',
   },
   childViewTriggers: {
     'save': 'save',
@@ -91,6 +126,17 @@ const SidebarView = View.extend({
     info: '[data-info-region]',
     access: '[data-access-region]',
   },
+  initialize({ clinician }) {
+    this.clinician = clinician;
+
+    this.listenTo(this.clinician, {
+      'change:_role': this.showInfo,
+      'change:_groups': this.showInfo,
+    });
+  },
+  onAttach() {
+    animSidebar(this.el);
+  },
   onRender() {
     this.showForm();
     this.showAccess();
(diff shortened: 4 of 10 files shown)


src/js/views/admin/sidebar/clinician/clinician-sidebar.hbs
```diff
@@ -11,6 +11,9 @@
   <div class="u-margin--t-16">
     <div data-name-region></div>
     <div class="u-margin--t-8" data-email-region></div>
+    {{#if showDelete}}
```
paulfalgout: delete would only be available via the top left 3-dot menu

src/js/apps/admin/sidebar/clinician-sidebar_app.js
```diff
@@ -15,18 +15,18 @@ export default App.extend({
...
       Radio.trigger('event-router', 'clinician', this.clinician.id);
     });
   },
-  // onDelete() {
-  //   this.clinician.trigger('editing', false);
-  //   this.clinician.destroy();
-  //   this.stop();
-  // },
+  onDeleteClinician() {
+    this.clinician.trigger('editing', false);
```
paulfalgout: I don't think this trigger is needed

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -160,11 +171,23 @@ const SidebarView = View.extend({
...
   onSave() {
-    this.getRegion('save').empty();
+    this.render();
   },
   onCancel() {
     this.triggerMethod('close', this);
   },
+  onConfirmDelete() {
+    const modal = Radio.request('modal', 'show:small', {
+      bodyText: intl.admin.sidebar.clinician.clinicianSidebarViews.sidebarView.deleteModal.bodyText,
```
paulfalgout: could probably make a `const i18n = intl.admin.sidebar.clinician.clinicianSidebarViews;` at the top

src/js/apps/admin/sidebar/clinician-sidebar_app.js
```diff
@@ -15,18 +15,18 @@ export default App.extend({
   viewEvents: {
     'save': 'onSave',
     'close': 'stop',
-    'delete': 'onDelete',
+    'delete:clinician': 'onDeleteClinician',
```
paulfalgout: why add `:clinician` to this one?

paulfalgout (changes requested):

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -141,7 +152,7 @@ const SidebarView = View.extend({
     this.showChildView('save', new DisabledSaveView());
   },
   showForm() {
-    if (this.model.isNew()) {
+    if (this.model.isNew() || !this.model.get('last_active_at')) {
```
paulfalgout: are the fields read-only if they can't be edited?  They shouldn't look like they can be edited.

LisaManresa:
Tests should be passing, but we're waiting on design guidance for the read only name and email fields.

src/js/apps/admin/sidebar/clinician-sidebar_app.js
paulfalgout: I think this is still true.  the destroy should remove it from the collectionview that would listen to this trigger, so I'm not sure it is needed.

paulfalgout: I also noticed in the `onStop` there's a `this.clinician &&` that isn't needed.  If it is needed the `trigger` on the line before would error.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -98,11 +132,39 @@ const SidebarView = View.extend({
...
+      itemTemplate: hbs`<span class="sidebar__delete-icon">{{far "trash-alt"}}</span>{{ @intl.admin.sidebar.clinician.clinicianSidebarViews.sidebarView.menuOptions.delete }}`,
+      lists: [{ collection: menuOptions }],
+      align: 'right',
+      popWidth: 248,
+    });
+
+    optionlist.show();
+  },
+  canEdit() {
+    return this.model.isNew() || !this.model.get('last_active_at');
```
paulfalgout: I think `canEdit` can move to the clinician model and it only needs to be `return !this.model.get('last_active_at');`  A new model should never have a `last_active_at`

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -74,9 +98,19 @@ const SidebarView = View.extend({
   },
   className: 'sidebar flex-region',
   template: ClinicianSidebarTemplate,
+  templateContext() {
+    const showDelete = !this.model.isNew() && !this.model.get('last_active_at');
```
paulfalgout: I think this can also be `this.model.canEdit()` as below with no check to `isNew`  On all of the other sidebars you can delete a new sidebar model with the same method.  Maybe the method should be a name that indicated edit or delete?  `canModify`?

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -98,11 +132,39 @@ const SidebarView = View.extend({
...
+    optionlist.show();
+  },
+  canEdit() {
+    return this.model.isNew() || !this.model.get('last_active_at');
+  },
   showName() {
-    this.showChildView('name', new NameView({ model: this.model }));
+    this.showChildView('name', new NameView({
+      model: this.model,
+      canEdit: this.canEdit(),
```
paulfalgout: if we move `canEdit` to the model, we won't need to pass it here.  these views can just `this.model.canEdit`

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -1,19 +1,35 @@
...
 import ClinicianSidebarTemplate from './clinician-sidebar.hbs';
 
 import './clinician-sidebar.scss';
 
+const i18n = intl.admin.sidebar.clinician.clinicianSidebarViews;
+
 const NameView = View.extend({
-  className: 'pos--relative',
-  template: hbs`<input class="input-primary w-100 js-input" placeholder="{{ @intl.admin.sidebar.clinician.clinicianSidebarViews.nameView.placeholder }}" value="{{ name }}" />`,
+  className: 'pos--relative clinician-sidebar__input-container',
```
paulfalgout: I don't think `clinician-sidebar__input-container` is being used

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
paulfalgout: _although_ really `!this.model.get('last_active_at')` or `{{#unless last_active_at}}` is about the same.

paulfalgout (changes requested):

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -141,7 +203,7 @@ const SidebarView = View.extend({
     this.showChildView('save', new DisabledSaveView());
   },
   showForm() {
-    if (this.model.isNew()) {
+    if (this.canEdit()) {
```
paulfalgout: `onSave` now re-renders which will call showForm again duplicating the listener on `this.model`.   
Also we need to be cloning this model for editing.. so if you make edits to the clinician name/email and cancel, it  doesn't modify the clinician model in memory.  Currently if you make an edit to the name and cancel, it will remain edited until refetched.
Similarly cancel currently just closes the sidebar, which works for new clinicians, but not for existing clinicians

paulfalgout (changes requested):
This sidebar needs animSidebar.  Might as well get it in here.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
LisaManresa: Very true

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
LisaManresa: I'm thinking of using the method on the model in all scenarios here. It's not quite as clean in some cases, but it will at least be consistent if the criteria for editing a clinician ever changes.

paulfalgout (changes requested):
In the performance work I'm knee deep in I took a look at the action clone in the sidebar as there was some overlap and awkwardness.  Check out https://github.com/RoundingWell/care-ops-frontend/commit/5f33ad293685ce2ff13b8ad24fb42688f409ab88  I think this might be a more obvious pattern we should copy here.  And/or let's improve anything you see over there and get these in line.

src/js/views/admin/sidebar/clinician/clinician-sidebar.scss
```diff
@@ -7,3 +7,11 @@
     opacity: 0.5;
   }
 }
+
+.clinician-sidebar__locked-icon {
+  color: #999;
```
paulfalgout: $black-something

src/js/apps/admin/sidebar/clinician-sidebar_app.js
```diff
@@ -9,7 +9,8 @@ export default App.extend({
     this.clinician = clinician;
 
     this.showView(new SidebarView({
-      model: this.clinician,
+      clinician,
+      model: clinician,
```
paulfalgout: This isn't needed correct?

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -91,79 +122,131 @@ const SidebarView = View.extend({
...
   showDisabledSave() {
     this.showChildView('save', new DisabledSaveView());
   },
   showForm() {
-    if (this.model.isNew()) {
-      this.listenTo(this.model, 'change:name change:email', this.showSave);
+    this.cloneClinician();
+
+    if (this.clinician.isEditable()) {
+      this.listenTo(this.clonedClinician, 'change:name change:email', this.showSave);
```
paulfalgout: interesting this works..  I wonder if the issue with the one in init was that it was effectively this:  `this.listenTo(this.clonedClinician, { 'change:name change:email', this.showSave });`  and maybe making it a single `listenTo` would support both events.

paulfalgout (changes requested):
looks about right

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -1,27 +1,54 @@
...
   },
   events: {
     'input @ui.input': 'onInput',
   },
+  initialize({ error }) {
+    this.error = error;
+  },
   onInput(evt) {
     this.model.set('name', evt.target.value);
+    if (this.error) {
```
LisaManresa: No super thrilled with this, but it might be what's needed to achieve the clearing of the error. We don't want to call `showName` in the layout view because that will remove focus on the element every time the user types.

paulfalgout: you can check program, but I think we leave up the error until the next submission.  If that's the case we probably just need `getOption` in the `templateContext`

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -1,27 +1,54 @@
...
 
 import './clinician-sidebar.scss';
 
+const i18n = intl.admin.sidebar.clinician.clinicianSidebarViews;
+
 const NameView = View.extend({
   className: 'pos--relative',
-  template: hbs`<input class="input-primary w-100 js-input" placeholder="{{ @intl.admin.sidebar.clinician.clinicianSidebarViews.nameView.placeholder }}" value="{{ name }}" />`,
+  template: hbs`
+    <input class="input-primary w-100 js-input" placeholder="{{ @intl.admin.sidebar.clinician.clinicianSidebarViews.nameView.placeholder }}" value="{{ name }}" {{#unless canEdit}}disabled{{/unless}} />{{~ remove_whitespace ~}}
```
paulfalgout: I think this an the other input need a `{{#if error}} has-error{{/if}}`

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -91,79 +140,160 @@ const SidebarView = View.extend({
...
     this.showChildView('save', new DisabledSaveView());
   },
   showForm() {
-    if (this.model.isNew()) {
-      this.listenTo(this.model, 'change:name change:email', this.showSave);
+    if (_.isEmpty(this.state.get('errors'))) {
+      this.cloneClinician();
+    }
+
+    if (this.clinician.isEditable()) {
```
paulfalgout: I think it'll be ok if we leave the error state up until the following submission

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -91,79 +140,160 @@ const SidebarView = View.extend({
...
   },
+  showErrors(errors) {
+    this.state.set({ errors });
+    this.showName(errors.name);
+    this.showEmail(errors.email);
+    this.showSave();
+  },
   onSave() {
-    this.getRegion('save').empty();
+    this.state.set({ errors: null });
```
paulfalgout: would the save still empty?

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -91,79 +140,160 @@ const SidebarView = View.extend({
...
   },
+  initialize(options) {
+    this.mergeOptions(options, ['clinician', 'errors']);
+
+    this.listenTo(this.clinician, {
+      'change:_role': this.showInfo,
+      'change:_groups': this.showInfo,
+    });
+
+    this.state = new Backbone.Model();
```
paulfalgout: So there's nothing listening to this model, so I think perhaps this might be overcomplicating things.  if `showErrors(errors) {` just did `this.errors = errors` and inside `onSave` it can just be `this.errors = null` then there's no need for the `isEmpty` checks.  If we get an empty error we've got other problems.

paulfalgout (changes requested):

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
LisaManresa: Moved this to `showSave` to avoid having the button disappear and reappear when errors happen.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -91,79 +126,163 @@ const SidebarView = View.extend({
     info: '[data-info-region]',
     access: '[data-access-region]',
   },
+  initialize(options) {
+    this.mergeOptions(options, ['clinician', 'errors']);
```
paulfalgout: I don't think errors is an option

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -91,79 +126,163 @@ const SidebarView = View.extend({
...
   },
+  showErrors(errors) {
+    this.errors = errors;
+    this.showName(errors.name);
+    this.showEmail(errors.email);
+    this.showSave();
+  },
   onSave() {
-    this.getRegion('save').empty();
+    this.errors = null;
```
paulfalgout: what happens if you save and no error occurs?  Also can you push save over and over again while waiting to see if an error happens? If this hides save, `showSave` doesn't need to be in `showErrors`.  The button will appear when the user goes to fix the text.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -91,79 +126,163 @@ const SidebarView = View.extend({
...
+    this.showChildView('save', new SaveView({ model: this.clonedClinician }));
   },
   showDisabledSave() {
     this.showChildView('save', new DisabledSaveView());
   },
   showForm() {
-    if (this.model.isNew()) {
-      this.listenTo(this.model, 'change:name change:email', this.showSave);
-      this.showDisabledSave();
+    if (_.isEmpty(this.errors)) {
```
paulfalgout: actually this is a problem because we are adding another listener to the same clone.  _but_ I actually don't see a need here to prevent the clone.. cancel _should_ clear errors and cancel is the use case for needing a new clone..  That seems like the only time this is re-called.

I _think_ this can be:
```js
this.cloneClinician();
this.listenTo(this.clonedClinician, 'change:name change:email', this.showSave);
if (this.model.isNew()) this.showDisabledSave();
else this.getRegion('save').empty();
```
Then below we could simplify it to:
```js
  showErrors(errors) {
    this.showName(errors.name);
    this.showEmail(errors.email);
    // this.showDisabledSave();   <-- maybe, but if so add it to the program sidebar
  },
  onSave() {	
    if (this.model.isNew()) this.showDisabledSave();
    else this.getRegion('save').empty();
  },
}
```
and `showSave` can go back to disabling if invalid otherwise showing.. no need to check errors or clear.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -91,79 +126,163 @@ const SidebarView = View.extend({
...
 
     this.showName();
     this.showEmail();
   },
   showInfo() {
-    if (this.model.isNew()) return;
+    if (this.clinician.isNew()) return;
 
-    if (!this.model.get('_role') || this.model.getGroups().length === 0) {
+    if (!this.clinician.get('_role') || this.clinician.getGroups().length === 0) {
```
paulfalgout, line 231: interestingly I think this could be `if (!this.clinician.isActive())` and maybe should be, but that function is in my other branch.

paulfalgout (changes requested):

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
LisaManresa: Nope, don't need that any more.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
LisaManresa: Want me to add it to this PR?

paulfalgout: it'll be easier to come back to it probably.

src/js/views/admin/sidebar/clinician/clinician-sidebar_views.js
```diff
@@ -91,79 +126,146 @@ const SidebarView = View.extend({
...
       this.showChildView('info', new InfoView());
       return;
     }
 
     this.getRegion('info').empty();
   },
+  showErrors(errors) {
+    this.errors = errors;
```
paulfalgout: don't really need to attach or clear the errors

paulfalgout (changes requested):

paulfalgout (approved):

PR #283, opened 2020-04-23
Remove comment flag

paulfalgout:
And fix some styles.
And removed the clinician relation from save (it is implied)

- [x] Add the Clubhouse Story ID: [ch22307]


paulfalgout:
there's a bit more to do here.

LisaManresa (approved):

PR #285, opened 2020-04-27
Use PUT endpoint for clinician role

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22336]

### DIFF
--- a/src/js/entities-service/entities/clinicians.js
+++ b/src/js/entities-service/entities/clinicians.js
@@ -1,3 +1,4 @@
+import $ from 'jquery';
 import _ from 'underscore';
 import Radio from 'backbone.radio';
 import Store from 'backbone.store';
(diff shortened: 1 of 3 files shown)


src/js/entities-service/entities/clinicians.js
```diff
@@ -49,9 +50,15 @@ const _Model = BaseModel.extend({
       || this.get('access') === 'admin';
   },
   saveRole(role) {
+    const url = `/api/clinicians/${ this.id }/relationships/role`;
     role = this.toRelation(role);
-    return this.save({ _role: role.data }, {
-      relationships: { role },
+
+    this.set({ _role: role.data.id });
```
paulfalgout: can we do this set prior to the `toRelation` so it can just be `role.id` .. maybe the `toRelation` can be inside the `stringify` so we're not modifying `role`

paulfalgout (changes requested):
tiny request

PR #289, opened 2020-05-04
Feature/login preloader

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22317] [ch22322][ch22321]


src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -36,7 +36,6 @@ export default App.extend({
     this.showNewCommentForm();
   },
   showActivity() {
-    if (this.action.isNew()) return;
```
LisaManresa, line 39: Snuck this in just to clean up coverage reports. This is a redundant check.

test/support/api/clinicians.js
```diff
@@ -1,16 +1,28 @@
...
+    .fixture('collections/groups').as('fxGroups')
     .fixture('test/clinicians').as('fxTestClinicians');
 
   cy.route({
     url: '/api/clinicians/me',
     response() {
+      const groups = _.sample(this.fxGroups, 2);
+      const clinician = getResource(this.fxTestClinicians[0], 'clinicians');
+
+      clinician.attributes.last_active_at = moment.utc();
```
paulfalgout: `.format()`

test/support/api/clinicians.js
```diff
@@ -1,16 +1,28 @@
...
 
 Cypress.Commands.add('routeCurrentClinician', (mutator = _.identity) => {
   cy
+    .fixture('collections/groups').as('fxGroups')
     .fixture('test/clinicians').as('fxTestClinicians');
 
   cy.route({
     url: '/api/clinicians/me',
     response() {
+      const groups = _.sample(this.fxGroups, 2);
```
paulfalgout: this seems moderately problematic... maybe not, but the current clinicians is creating 2 groups, ensuring that the current clinician is in both groups.  We _probably_ need to hard set the group ids here and there to both be `11111` and `22222` and _not_ include the group data here.

src/js/views/globals/root_views.js
```diff
@@ -55,6 +57,17 @@ const TopRegionView = View.extend({
   },
 });
 
+const PreloaderRegionView = TopRegionView.extend({
+  // NOTE: ensures preload can't close on user click
+  contains() {
+    return this.getRegion('region').hasView();
+  },
+  className: 'foo',
```
paulfalgout: `foo`

src/js/views/admin/shared/clinician-state.scss
```diff
@@ -0,0 +1,18 @@
+.clinician-status {
```
paulfalgout: all the `status`s should probably be `state`s

src/js/entities-service/entities/clinicians.js
```diff
@@ -70,6 +70,15 @@ const _Model = BaseModel.extend({
   isEditable() {
     return !this.get('last_active_at');
   },
+  isActive() {
+    /* istanbul ignore next */
+    if (!_DEVELOP_) return true;
```
paulfalgout: might not need the flag now that `last_active_at` should be in BE `develop`

paulfalgout (changes requested):

test/support/api/clinicians.js
LisaManresa: The main issue I was trying to tackle here was guaranteeing that the currentClinician was active in all tests. I think the original issue was that, even though `routeGroupsBootstrap` sets the groups for all clinicians, if an individual test called `routeCurrentClinician`, it would override the data set by `routeGroupsBootstrap`, causing tests to get stuck on the "account not set up" view.

Hardcoding the group ids like you suggested works just as well, though.

test/support/api/clinicians.js
```diff
@@ -1,15 +1,22 @@
 import _ from 'underscore';
+import moment from 'moment';
 import { getResource, getIncluded, getRelationship } from 'helpers/json-api';
 
 Cypress.Commands.add('routeCurrentClinician', (mutator = _.identity) => {
   cy
+    .fixture('collections/groups').as('fxGroups')
```
paulfalgout, line 7: fixture is not used now

src/js/views/globals/prelogin/prelogin.scss
```diff
@@ -0,0 +1,106 @@
+.prelogin {
+  align-items: center;
+  background-color: #1F2B33;
+  display: flex;
+  flex-direction: column;
+  font-family: 'ProximaSoft', 'Droid Sans', 'helvetica neue', 'arial', sans-serif;
```
paulfalgout, line 6: I bet some of the styles in prelogin are redundant now that we're using scss.

src/js/apps/patients/sidebar/action-sidebar_app.js
paulfalgout: actually I think this whole app needs to bail if this is true.. currently it looks like we're showing a comment form on a new action.. can't post a comment to an action that has yet to exist... current bug

paulfalgout (approved):
There's a couple of small questions, but I'm going to get this in so I can start dealing with merge issues.

PR #290, opened 2020-05-05
Refactoring components

paulfalgout:
- [x] Add the Clubhouse Story ID: [c22416]

This PR admittedly does quite a lot.

It removes the dateSort entirely.  The date format we currently support should always be sortable alphabetically, so here's hoping we always keep this standard... it's faster.

https://github.com/RoundingWell/care-ops-frontend/commit/0d070aa3db11d3a214e9bbbec554c136db03ec8f
It adds a `canClear` option to droplists which adds a clear option outside of the collection to the droplist.  This work also standardized some things with components.. In particular it removed `getItemFormat` preferring always `itemTemplate`  this should always be workable and for the most part is what we've done anyhow.

While reviewing our usage of components I replaced the add component on the patient workflow page.  It was a droplist, but really a droplist should expect to store it's state.. if we're just picking an option and doing something with that action, we should be using an option list, so rather than update the incorrectly used droplist I update the component type. That change is here https://github.com/RoundingWell/care-ops-frontend/commit/c34fd0925e5f78bc59375852caeab9816d380874

Lastly it breaks up our `actions_views` type files and puts components in their own file under `{section}/shared/components` which is then added to `shared/actions_views` or `shared/flows_views` etc for those shared contexts.  Also notably I did my best to prevent creating new things within methods, particularly `initialize` on these components.  In list context, for instance, the time component was building the options at render time for every time component on the page (so 1 per result) and the setup time for that was significant.  This moves the collection creation outside of the component init and in many cases caches the results.

This needs lots of :eyes:


paulfalgout:
~~The codecov chart..  I think maybe it is incorrect.. I have full coverage locally, and the checks are showing full coverage as well.~~

Oh..  there's less covered lines, but there's less lines..  still 100% coverage.. that.. should not be red..

LisaManresa (approved):
Looks good. The new organization for components makes a lot of sense.

PR #291, opened 2020-05-05
Fix tab order on comment form

LisaManresa:
- [x ] Add the Clubhouse Story ID: [ch22423]


LisaManresa:
@paulfalgout cypress doesn't support tabbing in tests yet, correct?

PR #298, opened 2020-05-13
Feature/patient engagement sidebar

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22529]

### DIFF
--- a/src/js/apps/patients/patient/sidebar/sidebar_app.js
+++ b/src/js/apps/patients/patient/sidebar/sidebar_app.js
@@ -1,0 +1,23 @@
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+
+import { SidebarView, EngagementStatusPreloadView, EngagementStatusView } from 'js/views/patients/patient/sidebar/sidebar_views';
+
+export default App.extend({
+  onBeforeStart({ patient }) {
+    this.patient = patient;
+
+    this.showView(new SidebarView({ model: this.patient }));
+    this.showChildView('engagement', new EngagementStatusPreloadView());
+  },
+  beforeStart({ patient }) {
+    return Radio.request('entities', 'fetch:patient:engagementStatus', patient.id);
+  },
+  onStart() {
+    this.showChildView('engagement', new EngagementStatusView({ model: this.patient }));
+  },
+  onFail() {
+    this.showChildView('engagement', new EngagementStatusView({ model: this.patient }));
+  },
+});
--- a/src/js/entities-service/patients.js
+++ b/src/js/entities-service/patients.js
@@ -1,3 +1,4 @@
+import $ from 'jquery';
 import Radio from 'backbone.radio';
 
 import BaseEntity from 'js/base/entity-service';
@@ -9,12 +10,23 @@ const Entity = BaseEntity.extend({
     'patients:model': 'getModel',
     'fetch:patients:model': 'fetchModel',
     'fetch:patients:model:byAction': 'fetchPatientByAction',
+    'fetch:patient:engagementStatus': 'fetchPatientEngagementStatus',
+    // 'fetch:patient:engagementSettings': 'fetchPatientEngagementSettings',
   },
   fetchPatientByAction(actionId) {
     const patient = Radio.request('entities', 'patients:model');
 
     return patient.fetch({ url: `/api/actions/${ actionId }/patient` });
   },
+  fetchPatientEngagementStatus(patientId) {
+    const patient = Radio.request('entities', 'patients:model', patientId);
+
+    return $.ajax({
+      url: `/api/patients/${ patientId }/engagement-status`,
+    })
+      .then(response => patient.set(response.data));
+  },
+  // fetchPatientEngagementSettings(patientId) {},
 });
 
 export default new Entity();
--- a/src/js/views/patients/patient/sidebar/sidebar_views.js
+++ b/src/js/views/patients/patient/sidebar/sidebar_views.js
@@ -1,16 +1,18 @@
 import _ from 'underscore';
-import { View, CollectionView, Region } from 'marionette';
+import anime from 'animejs';
 import moment from 'moment';
 
 import hbs from 'handlebars-inline-precompile';
-import intl, { renderTemplate } from 'js/i18n';
+import { View, CollectionView, Region } from 'marionette';
 
+import intl, { renderTemplate } from 'js/i18n';
 import ModelRender from 'js/behaviors/model-render';
 
 import PatientSidebarTemplate from './patient-sidebar.hbs';
 import PatientSidebarGroupsTemplate from './patient-sidebar-groups.hbs';
 
 import './patient-sidebar.scss';
+import 'sass/domain/engagement-status.scss';
 
 const i18n = intl.patients.patient.sidebar.sidebarViews;
 
@@ -83,13 +85,38 @@ const GroupsView = View.extend({
   },
 });
 
+const EngagementStatusPreloadView = View.extend({
+  template: hbs`{{ @intl.patients.patient.sidebar.sidebarViews.engagementStatusPreloadView.loading }}`,
+  onRender() {
+    anime({
+      targets: this.el,
+      opacity: 0.5,
+      loop: true,
+      easing: 'easeInOutSine',
+      duration: 400,
+      direction: 'alternate',
+    });
+  },
+});
+
+const EngagementStatusView = View.extend({
+  template: hbs`
+    {{#if engagement.status}}
+      <span class="engagement-status__icon {{ engagement.status }} u-margin--r-4">{{fas "circle"}}</span>{{formatMessage (intlGet "patients.patient.sidebar.sidebarViews.engagementStatusView.status") status=engagement.status}}
+    {{else}}
+      <span class="patient-sidebar__no-engagement">{{ @intl.patients.patient.sidebar.sidebarViews.engagementStatusView.notAvailable }}</span>
+    {{/if}}
+  `,
+});
+
 const SidebarView = View.extend({
-  className: 'patient-sidebar',
+  className: 'patient-sidebar flex-region',
   template: PatientSidebarTemplate,
   regionClass: Region.extend({ replaceElement: true }),
   regions: {
     info: '[data-info-region]',
     groups: '[data-groups-region]',
+    engagement: '[data-engagement-region]',
   },
   onRender() {
     const model = this.model;
@@ -101,4 +128,6 @@ const SidebarView = View.extend({
 
 export {
   SidebarView,
+  EngagementStatusPreloadView,
+  EngagementStatusView,
 };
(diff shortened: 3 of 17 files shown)


src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -0,0 +1,19 @@
...
+  onBeforeStart({ patient }) {
+    this.patient = patient;
+
+    this.showView(new SidebarView({ model: this.patient }));
+  },
+  beforeStart({ patient }) {
+    return Radio.request('entities', 'fetch:patient:engagementStatus', patient.id);
+  },
+  onStart(options, status) {
+    this.patient.set({ status: status.get('status') });
```
LisaManresa: This is most certainly not the way we want to handle this. But I needed to unblock myself to get the rest of the PR done. See comment on the patient entity file for more...

src/js/entities-service/patients.js
```diff
@@ -9,12 +9,28 @@ const Entity = BaseEntity.extend({
...
   },
   fetchPatientByAction(actionId) {
     const patient = Radio.request('entities', 'patients:model');
 
     return patient.fetch({ url: `/api/actions/${ actionId }/patient` });
   },
+  fetchPatientEngagementStatus(patientId) {
+    const patient = Radio.request('entities', 'patients:model', patientId);
+
+    return patient.fetch({
```
LisaManresa: I'm pretty sure this is not how we intend to handle this data. Without the `success` function below, the status data does not merge with the rest of the patient entity correctly. It ends up setting the model `id` to undefined. This might be down to model parsing not working as we expected.

It also appears that the entities request above might not return the intended patient model. That might have to do with the patient model fetch not being complete by the time this method is called. But I'm not certain on that.

paulfalgout: Yeah we should probably do a `return $.ajax` call here and set the patient model in a `then` expecting the data
```
{
    "data": {
        "engagement": {
            "status": "active"
        }
    },
}
```
and setting `patient.set(data)` so that `patient.get('engagement').status` is the data.

test/integration/patients/patient/patient-flow.js
```diff
@@ -365,7 +365,7 @@ context('patient flow page', function() {
         fx.data.relationships.actions.data = flowActions;
         fx.data.relationships.state.data.id = '33333';
 
-        fx.data.meta.progress.complete = 0;
+        fx.data.meta.progress.complxete = 0;
```
paulfalgout: :boom:

paulfalgout (changes requested):

test/support/api/patients.js
```diff
@@ -92,3 +92,18 @@ Cypress.Commands.add('routePatientSearch', (mutator = _.identity) => {
   })
     .as('routePatientSearch');
 });
+
+Cypress.Commands.add('routePatientEngagementStatus', (mutator = _.identity) => {
```
paulfalgout: I think this can be `(status = _.sample(this.fxPatientEngagementStatuses)) =>`  then I think `cy.routePatientEngagementStatus(status)` can be added to `routePatient` ~~with `routePatient accepting `mutator, status`~~   actually.. just recalling `routePatientEngagementStatus` after `routePatient` should probably work for setting a test status

test/integration/patients/patient/sidebar.js
```diff
@@ -145,4 +148,263 @@ context('patient sidebar', function() {
...
+        .wait('@routePatientEngagementStatus');
+
+      cy
+        .get('.patient__sidebar')
+        .find('.active')
+        .parent()
+        .should('contain', 'Active');
+    });
+
+    specify('Not Scheduled', function() {
```
paulfalgout: I don't think we need to test all of the statuses.. if there was more logic than just picking a key out of the collection, but just testing a single engagement value for now should be enough

src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -0,0 +1,22 @@
...
+
+import App from 'js/base/app';
+
+import { SidebarView } from 'js/views/patients/patient/sidebar/sidebar_views';
+
+export default App.extend({
+  onBeforeStart({ patient }) {
+    this.patient = patient;
+
+    this.showView(new SidebarView({ model: this.patient }));
```
paulfalgout, line 11: I think this app should probably be 
```
onBeforeStart() {
...
  this.showView...
  this.showChildView('engagement', new Preloader());
},
onStart() {
  this.showChildView('engagement', new EngagementView({ model: this.patient });
},
onFail() {
  this.showChildView('engagement', new NotAvailableView());
},
```

paulfalgout (changes requested):

src/js/entities-service/patients.js
```diff
@@ -9,12 +11,28 @@ const Entity = BaseEntity.extend({
...
+  fetchPatientEngagementStatus(patientId) {
+    const patient = Radio.request('entities', 'patients:model', patientId);
+    const engagement = patient.getEngagement();
+
+    return $.ajax({
+      url: `/api/patients/${ patientId }/engagement-status`,
+    })
+      .then(response => {
+        return patient.set({
+          engagement: _.extend(engagement, response.data),
```
paulfalgout: extending the original engagement data means the set already happened.  I think `patient.set(response.data)` should be ok here.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -83,19 +85,64 @@ const GroupsView = View.extend({
...
+
+    return hbs`
+      {{#if statusUnavailable}}
+        <div class="patient-sidebar__no-engagement">{{ @intl.patients.patient.sidebar.sidebarViews.engagementStatusView.notAvailable }}</div>
+      {{else}}
+        <span class="engagement-status__icon {{ status }} u-margin--r-4">{{fas "circle"}}</span>{{formatMessage (intlGet "patients.patient.sidebar.sidebarViews.engagementStatusView.status") status=status}}
+      {{/if}}
+    `;
+  },
+  templateContext() {
```
paulfalgout: so if we move the logic into the app the view can just use `{{ engagement.status }}`

src/js/apps/patients/patient/sidebar/sidebar_app.js
paulfalgout: Actually I think onFail can also show EngagementView and we should show not available `{{#unless engagement.status}}`

test/support/api/patients.js
```diff
@@ -92,3 +92,18 @@ Cypress.Commands.add('routePatientSearch', (mutator = _.identity) => {
...
+  cy
+    .fixture('test/patient-engagement-statuses').as('fxPatientEngagementStatuses');
+
+  cy.route({
+    url: '/api/patients/**/engagement-status',
+    response() {
+      return mutator({
+        data: _.sample(this.fxPatientEngagementStatuses),
```
paulfalgout: with `(status = _.sample(this.fxPatientEngagementStatuses)) =>`
```
return {
  data: { engagement: { status } },
},
```

src/js/entities-service/entities/patients.js
```diff
@@ -14,6 +14,9 @@ const _Model = BaseModel.extend({
   getFields() {
     return Radio.request('entities', 'patientFields:collection', this.get('_patient_fields'));
   },
+  getEngagement() {
```
paulfalgout: probably won't need to do `|| {}` anywhere so likely won't need this `get`.. I think it's likely we'll only need to get the data in the template itself

test/integration/patients/patient/patient-flow.js
LisaManresa: 😳 How is that test passing?

test/support/api/patients.js
LisaManresa: Unfortunately the engagement status command needs to establish the fixture data on its own because there are a handful of scenarios where `routePatient` isn't called (flows and actions).

test/support/api/patients.js
```diff
@@ -92,3 +92,22 @@ Cypress.Commands.add('routePatientSearch', (mutator = _.identity) => {
   })
     .as('routePatientSearch');
 });
+
+Cypress.Commands.add('routePatientEngagementStatus', (mutator = _.identity, status) => {
```
paulfalgout: because the data is so simple I was suggesting not using a mutator..  Then no `_.identity` is needed when used.   If there's some reason to keep the mutator we can drop the 2nd param and just use the mutator.

paulfalgout (changes requested):

test/support/api/patients.js
paulfalgout: To this point though.. In _every_ instance where patient engagement status would be needed, patient _must_ be loaded somewhere..  _so_ it seems like either this could be added to `routePatient` and whatever the other route is, and then recalled very infrequently.. or we could just not add it _everywhere_ and let it fail unless we're testing it.   We're adding `routePatientEngagementStatus` 15 times, but probably only need it.. once or twice.

LisaManresa:
This also tackles [ch22532]

test/integration/patients/patient/sidebar.js
```diff
@@ -21,12 +22,14 @@ context('patient sidebar', function() {
 
         return fx;
       })
+      .routePatientEngagementStatus('active')
```
paulfalgout, line 25: what would happen if this was the only use of `routePatientEngagementStatus` in the tests?

LisaManresa: Oh good call. Now that we're handling failed calls correctly we don't need it everywhere there is a patient call.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -83,13 +85,43 @@ const GroupsView = View.extend({
...
+const EngagementStatusPreloadView = View.extend({
+  ui: {
+    loading: '.js-loading',
+  },
+  template: hbs`
+    <div class="js-loading">{{ @intl.patients.patient.sidebar.sidebarViews.engagementStatusPreloadView.loading }}</div>
+  `,
+  onRender() {
+    anime({
+      targets: this.ui.loading[0],
```
paulfalgout: sorry for the late comment.. _but_ would this work with `this.$el` not needing the `.js-loading` wrapper?

src/sass/domain/engagement-status.scss
```diff
@@ -0,0 +1,12 @@
+.engagement-status__icon {
+  .icon {
+    font-size: 14px;
+  }
+
+  &.active { color: color(green); }
```
paulfalgout, line 6: I _think_ we aren't going to end up needing to support all of these statuses.. but I mis-typed kim's email so waiting the hear back..  can probably just merge as is and clean up later if needed.  but dropping a note here.

paulfalgout (changes requested):
one small thing on the preloader and one note for later

PR #299, opened 2020-05-19
Display patient check-in event in Data & Events

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22463]

This also modifies the check-ins entity to match the path that BE expects us to hit.

### DIFF
--- a/src/js/apps/patients/patient/data-events/data-events_app.js
+++ b/src/js/apps/patients/patient/data-events/data-events_app.js
@@ -16,10 +16,11 @@ export default App.extend({
     return [
       Radio.request('entities', 'fetch:actions:collection:byPatient', { patientId: patient.id, filter }),
       Radio.request('entities', 'fetch:flows:collection:byPatient', { patientId: patient.id, filter }),
+      Radio.request('entities', 'fetch:patientEvents:collection', patient.id),
     ];
   },
-  onStart({ patient }, [actions], [flows]) {
-    this.collection = new Backbone.Collection([...actions.models, ...flows.models]);
+  onStart({ patient }, [actions], [flows], [events]) {
+    this.collection = new Backbone.Collection([...actions.models, ...flows.models, ...events.models]);
     this.showChildView('content', new ListView({ collection: this.collection }));
   },
   onEditAction(action) {
--- a/src/js/entities-service/check-ins.js
+++ b/src/js/entities-service/check-ins.js
@@ -1,3 +1,5 @@
+import Radio from 'backbone.radio';
+
 import BaseEntity from 'js/base/entity-service';
 import { _Model, Model, Collection } from './entities/check-ins';
 
@@ -5,7 +7,16 @@ const Entity = BaseEntity.extend({
   Entity: { _Model, Model, Collection },
   radioRequests: {
     'checkIns:model': 'getModel',
-    'fetch:checkIns:model': 'fetchModel',
+    'fetch:checkIns:model': 'fetchCheckInModel',
+  },
+  fetchCheckInModel({ checkInId, patientId }) {
+    // API needs patientId to verify clinician access
+    const checkIn = Radio.request('entities', 'checkIns:model');
+    const filter = { checkin: checkInId };
+    return checkIn.fetch({
+      url: `/api/patients/${ patientId }/checkins`,
+      data: { filter },
+    });
   },
 });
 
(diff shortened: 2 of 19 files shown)


src/js/entities-service/check-ins.js
```diff
@@ -1,11 +1,22 @@
...
 const Entity = BaseEntity.extend({
   Entity: { _Model, Model, Collection },
   radioRequests: {
     'checkIns:model': 'getModel',
-    'fetch:checkIns:model': 'fetchModel',
+    'fetch:checkIns:model': 'fetchCheckInModel',
+  },
+  fetchCheckInModel({ checkInId, patientId }) {
+    // API needs patientId to verify clinician access
+    const checkIn = Radio.request('entities', 'checkIns:model');
```
paulfalgout: `this.getModel()`

src/js/views/patients/patient/data-events/data-events_views.js
```diff
@@ -166,13 +166,38 @@ const FlowItemView = View.extend({
...
 });
 
+const PatientEventItemView = View.extend({
+  className: 'table-list__item',
+  tagName: 'tr',
+  behaviors: [RowBehavior],
+  template: hbs`
+    <td class="table-list__cell">
+      <span class="patient__event-icon">{{far "paper-plane"}}</span>{{~ remove_whitespace ~}}
+      <span class="patient__action-name">{{ @intl.patients.patient.dataEvents.dataEventsViews.patientEventItemView.checkInCompleted }}</span>
```
paulfalgout: `patient__action-name` likely needs to be renamed to something about how it is functionally and not "action"

src/js/views/patients/patient/data-events/data-events_views.js
```diff
@@ -181,10 +206,11 @@ const ListView = CollectionView.extend({
   },
   emptyView: EmptyView,
   viewComparator({ model }) {
-    return - model.moment('updated_at').format('x');
+    const comparatorAttr = model.get('finished_at') ? 'finished_at' : 'updated_at';
```
paulfalgout: I believe this can just be `return model.get('updated_at') || model.get('finished_at');`  _assuming_ that `finished_at` is the data format we've been using... if that's the case our default date-type sorts alphabetically and is way faster than using moment.

test/integration/check-ins/check-in.js
```diff
@@ -14,6 +14,7 @@ context('patient check-in', function() {
 
         return fx;
       })
+      .routePatientEngagementStatus()
```
paulfalgout: this commit seems unnecessary?

paulfalgout (changes requested):
Let's remove the `getPatient` on the patient-event completely until it's needed so that we're at coverage.  It's easy enough to add back when it's used.

src/js/views/patients/patient/data-events/data-events_views.js
LisaManresa: I think we're going to need a slightly more complex sort comparator since we need to display most recent date first.

src/js/views/patients/patient/data-events/data-events_views.js
```diff
@@ -166,25 +166,54 @@ const FlowItemView = View.extend({
...
 
     return ActionItemView;
   },
   emptyView: EmptyView,
-  viewComparator({ model }) {
-    return - model.moment('updated_at').format('x');
+  viewComparator({ model: modelA }, { model: modelB }) {
+    const modelAAttr = modelA.get('updated_at') || modelA.get('finished_at');
+    const modelBAttr = modelB.get('updated_at') || modelB.get('finished_at');
+    if (modelAAttr > modelBAttr) return -1;
```
paulfalgout: I think this could be `return modelAAttr.localeCompare(modelBAttr);`  If that's the wrong direction `localeCompare` returns a number so it'd just be `return -...`

paulfalgout (changes requested):

LisaManresa:
Hmm, my fix for the patient-flow test flakiness was apparently insufficient. Going to poke around with that some more.

src/js/apps/patients/patient/data-events/data-events_app.js
```diff
@@ -16,10 +16,11 @@ export default App.extend({
     return [
       Radio.request('entities', 'fetch:actions:collection:byPatient', { patientId: patient.id, filter }),
       Radio.request('entities', 'fetch:flows:collection:byPatient', { patientId: patient.id, filter }),
+      Radio.request('entities', 'fetch:patientEvents:collection', patient.id),
```
paulfalgout, line 19: If need be later we might have to come back and change this to `_DEVELOP_ ? { models: [] } " Radio.request....`   but let's see how things go.

PR #304, opened 2020-05-27
Check-in response view

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22562]

Note: This assumes the data will be formatted identically to https://github.com/RoundingWell/RWell/blob/develop/mainapp/assets/tests/fixtures/checkIn.json.

### DIFF
--- a/src/js/views/check-ins/check-in/check-in.scss
+++ b/src/js/views/check-ins/check-in/check-in.scss
@@ -54,7 +54,22 @@
   background-color: $white;
   border: 1px solid $black-90;
   border-radius: 4px 4px 0 0;
-  height: calc(100% - 16px);
+  height: calc(100% - 32px);
   margin-top: 16px;
-  padding: 8px;
+  overflow: auto;
+  padding: 8px 120px;
+}
+
+.check-in__item {
+  border-bottom: 1px solid $black-90;
+  padding: 16px 0;
+
+  &:last-of-type {
+    border-bottom: none;
+  }
+}
+
+.check-in__patient-answer {
+  background: color(yellow);
+  display: inline-block;
 }
--- a/src/js/views/check-ins/check-in/check-in_views.js
+++ b/src/js/views/check-ins/check-in/check-in_views.js
@@ -1,6 +1,6 @@
 import Radio from 'backbone.radio';
 import hbs from 'handlebars-inline-precompile';
-import { View } from 'marionette';
+import { View, CollectionView } from 'marionette';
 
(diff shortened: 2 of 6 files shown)


src/js/views/check-ins/check-in/check-in.scss
```diff
@@ -54,7 +54,27 @@
...
+.check-in__item {
+  border-bottom: 1px solid $black-90;
+  padding: 16px 0;
+
+  &:last-of-type {
+    border-bottom: none;
+  }
+}
+
+.check-in__item--message-label,
```
paulfalgout: I think these classes could use revision.  I don't think `--message-label` is a type of `check-in__item`

src/js/views/check-ins/check-in/check-in_views.js
```diff
@@ -36,8 +37,46 @@ const EngagementSidebarView = View.extend({
   template: hbs`Patient Engagement sidebar will go here`,
 });
 
-const CheckInView = View.extend({
-  template: hbs`Check-in details will go here`,
+const MessageItemView = View.extend({
+  className: 'check-in__item',
+  template: hbs`
```
paulfalgout, line 41: I think we could possibly do this as 
```
hbs`
  <strong>{{{ label_html }}}</strong>
  <div>{{{ body_html }}}</div>
`
```

src/js/views/check-ins/check-in/check-in_views.js
```diff
@@ -36,8 +37,46 @@ const EngagementSidebarView = View.extend({
...
+    },
+    body_html() {
+      return this.content.body;
+    },
+  },
+});
+
+const CheckInItemView = View.extend({
+  className: 'check-in__item',
+  template: hbs`
```
paulfalgout, line 57: same here I think
```
<div>{{{ question_html }}}</div>
<strong class="check-in__answer">{{{ answer_html }}}</strong>
```

src/js/views/check-ins/check-in/check-in_views.js
```diff
@@ -65,8 +104,9 @@ const LayoutView = View.extend({
     }));
   },
   showCheckInContent() {
+    const collection = new Backbone.Collection(this.model.get('content'));
```
paulfalgout: Can we move this to `this.model.getContent()` where the collection is made in the entity?

paulfalgout (changes requested):

paulfalgout (approved):

PR #306, opened 2020-05-28
Reports routing

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22563]

### DIFF
--- a/src/js/views/admin/report/report_views.js
+++ b/src/js/views/admin/report/report_views.js
@@ -1,0 +1,51 @@
+import Radio from 'backbone.radio';
+import hbs from 'handlebars-inline-precompile';
+
+import { View } from 'marionette';
+
+import PreloadRegion from 'js/regions/preload_region';
+
+import './report.scss';
+
+const ContextTrailView = View.extend({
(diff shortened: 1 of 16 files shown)


src/js/views/admin/report/report_views.js
```diff
@@ -0,0 +1,54 @@
+import Radio from 'backbone.radio';
+import hbs from 'handlebars-inline-precompile';
+
+import { View } from 'marionette';
+
+import PreloadRegion from 'js/regions/preload_region';
+
+import './report.scss';
+
+// const ReportView = View.extend({});
```
paulfalgout: what is `ReportView`?

paulfalgout (changes requested):

src/js/views/admin/report/report_views.js
LisaManresa: I figured it might be necessary to display the report. But that will probably behave the way forms do, where we just build the url for the report data and insert it into the iframe, right? I can remove that.

paulfalgout (approved):

PR #311, opened 2020-06-08
Adding kitchen sink fixture for forms tests

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22579]

### DIFF
--- a/test/integration/forms/formapp.js
+++ b/test/integration/forms/formapp.js
@@ -1,3 +1,7 @@
+import moment from 'moment';
+
+const now = moment.utc();
+
 context('Formapp', function() {
   specify('load basic form', function() {
     cy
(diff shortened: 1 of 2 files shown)


test/integration/forms/formapp.js
```diff
@@ -158,4 +162,83 @@ context('Formapp', function() {
...
+      .find('input[type=text]')
+      .first()
+      .type('hello')
+      .should('have.value', 'hello');
+
+    cy
+      .get('@formIOComponent')
+      .find('input[type=checkbox]')
+      .first()
+      .check()
```
paulfalgout: I'm not convinced this test wouldn't work if styles broke the checkbox/radio.. seems like we should try and click it?

paulfalgout (changes requested):

PR #312, opened 2020-06-08
Implement sidebar modal

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22601]

### DIFF
--- a/src/js/services/modal.js
+++ b/src/js/services/modal.js
@@ -11,10 +11,12 @@ export default App.extend({
     'show:small': 'showSmall',
     'show:tall': 'showTall',
     'show:custom': 'showCustom',
+    'show:sidebar': 'showSidebar',
   },
-  initialize({ modalRegion, modalSmallRegion }) {
+  initialize({ modalRegion, modalSmallRegion, modalSidebarRegion }) {
     this.modalRegion = modalRegion;
     this.modalSmallRegion = modalSmallRegion;
+    this.modalSidebarRegion = modalSidebarRegion;
   },
   showModal(options) {
     const ConfirmModal = ModalView.extend(options);
@@ -44,6 +46,17 @@ export default App.extend({
   },
   showCustom(view) {
     this.modalRegion.show(view);
+    return view;
+  },
+  showSidebar(options) {
+    const SidebarModal = ModalView.extend(_.extend({
+      className: 'modal--sidebar',
+      bodyClass: 'sidebar',
+    }, options));
+    const view = new SidebarModal();
+
+    this.modalSidebarRegion.show(view);
+
     return view;
   },
 });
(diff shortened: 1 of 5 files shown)


src/js/views/globals/root_views.js
```diff
@@ -107,6 +107,16 @@ const ModalRegionView = TopRegionView.extend({
   },
 });
 
+const ModalSidebarRegionView = ModalRegionView.extend({
+  setLocation() {
```
paulfalgout: `setLocation: _.noop,` should work here.

src/sass/modules/modals.scss
```diff
@@ -76,3 +76,24 @@
...
+  background: #fff;
+  display: flex;
+  flex-direction: column;
+  height: 100vh;
+  max-width: 400px;
+  min-height: 0;
+  position: absolute;
+  right: 0;
+  top: 0;
+  z-index: 1;
```
paulfalgout: I don't think adding a `z-index` here is correct.

src/sass/modules/modals.scss
```diff
@@ -76,3 +76,24 @@
 .modal-footer--empty {
   min-height: 8px;
 }
+
+.modal--sidebar {
+  background: #fff;
```
paulfalgout: $white

src/js/services/modal.js
```diff
@@ -46,4 +48,16 @@ export default App.extend({
     this.modalRegion.show(view);
     return view;
   },
+  showSidebar(options) {
+    const SidebarModal = ModalView.extend(_.extend({
+      className: 'modal--sidebar',
+      bodyClass: 'modal-content--sidebar',
+      isSidebar: true,
```
paulfalgout: I don't think this is needed

src/js/services/modal.js
```diff
@@ -46,4 +48,16 @@ export default App.extend({
     this.modalRegion.show(view);
     return view;
   },
+  showSidebar(options) {
+    const SidebarModal = ModalView.extend(_.extend({
+      className: 'modal--sidebar',
+      bodyClass: 'modal-content--sidebar',
```
paulfalgout: I think this is `sidebar`

src/js/views/globals/root_views.js
```diff
@@ -297,6 +307,7 @@ const RootView = CollectionView.extend({
     this.addRegionView('alert', new TopRegionView());
     this.addRegionView('modal', new ModalRegionView({ $body }));
     this.addRegionView('modalSmall', new ModalRegionView({ $body }));
+    this.addRegionView('modalSidebar', new ModalSidebarRegionView({ $body }));
```
paulfalgout: I believe the order here determines the stack placement.  Which is why we should not `z-index` these.  I think it's probably `modalSidebar` -> `modal` -> `modalSmall`

paulfalgout (changes requested):

paulfalgout (approved):

PR #313, opened 2020-06-10
Create worklist legacy app in preparation for multi-select feature

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22605]


paulfalgout (changes requested):
Bah.. so my bad..  dealing with coverage here is going to be too difficult keeping the views together.  better that we make legacy view files so that we can avoid dropping coverage during development.

PR #314, opened 2020-06-12
Add multiselect to worklists

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22608]

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -21,6 +21,8 @@ const StateModel = Backbone.Model.extend({
         groupId: null,
         clinicianId: this.currentClinician.id,
       },
+      selectedActions: {},
+      selectedFlows: {},
     };
   },
   preinitialize() {
@@ -78,6 +80,19 @@ const StateModel = Backbone.Model.extend({
     };
 
     return filters[this.id];
+  },
+  toggleSelected(model, isSelected) {
+    const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
+    const list = _.clone(this.get(listName));
+
+    this.set(listName, _.extend(list, {
+      [model.id]: isSelected,
+    }));
+  },
+  isSelected(model) {
+    const list = this.isFlowType() ? this.get('selectedFlows') : this.get('selectedActions');
+
+    return list[model.id];
   },
 });
 
@@ -138,9 +153,11 @@ export default App.extend({
     return Radio.request('entities', `fetch:${ this.getState().getType() }:collection`, { filter });
   },
   onStart(options, collection) {
+    this.collection = collection;
+
     const collectionView = new ListView({
-      collection,
-      isFlowList: this.getState().isFlowType(),
+      collection: this.collection,
+      state: this.getState(),
       viewComparator: this.getComparator(),
     });
 
--- a/src/js/views/patients/worklist/action_views.js
+++ b/src/js/views/patients/worklist/action_views.js
@@ -22,7 +22,16 @@ const ActionEmptyView = View.extend({
 });
 
 const ActionItemView = View.extend({
-  className: 'table-list__item',
+  className() {
+    const state = this.getOption('state');
+    const className = 'table-list__item';
+
+    if (state.isSelected(this.model)) {
+      return `${ className } is-selected`;
+    }
+
+    return className;
+  },
   tagName: 'tr',
   template: ActionItemTemplate,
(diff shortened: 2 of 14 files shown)


src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -138,12 +140,62 @@ export default App.extend({
...
+    const isFlowList = this.getState().isFlowType();
+    this.collection = collection;
+
+    this.mapSelectedItems(isFlowList ? this.getState('selectedFlows') : this.getState('selectedActions'));
+    this.initListView();
+  },
+  mapSelectedItems(ids) {
+    _.each(ids, id => {
+      const item = this.collection.get(id);
+      item.set('isSelected', true);
```
paulfalgout: rather than modifying the model, let's just pass the state model down through `childViewOptions` on the collection

src/js/views/patients/worklist/action_views.js
```diff
@@ -47,6 +55,10 @@ const ActionItemView = View.extend({
   triggers: {
     'click': 'click',
     'click .js-patient': 'click:patient',
+    'click .js-select': 'click:select',
+  },
+  modelEvents: {
+    'change:isSelected': 'render',
```
paulfalgout: this doesn't seem necessary

src/js/views/patients/worklist/action_views.js
```diff
@@ -59,6 +71,12 @@ const ActionItemView = View.extend({
   onClickPatient() {
     Radio.trigger('event-router', 'patient:dashboard', this.model.get('_patient'));
   },
+  onClickSelect() {
+    const isSelected = this.model.get('isSelected');
```
paulfalgout: if we pass the state down to the child we can possibly just do `listState.setSelected(this.model);` here... and I think we need to make sure we're not preventing the checkbox click from checking the box on the trigger definition `preventDefault: false`?

paulfalgout: Then `setSelected` on the state model can determine the type.. and handle the without/push stuff in the state model.  Then perhaps we could add a `listState.isSelected(this.model)`  It also might be cheaper to store these as `[{ 'id1': true, 'id2': false }]` so that each row doesn't have to loop over the array.. isSelected just becomes `selectedFoo[this.model.id]` and the values are either true or falsy (false/null)

paulfalgout (changes requested):

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -79,6 +81,26 @@ const StateModel = Backbone.Model.extend({
 
     return filters[this.id];
   },
+  setSelected(model) {
```
LisaManresa: This and `removeSelected` could be consolidated into 1 method if it also accepts an `isSelected` boolean. Thoughts?

paulfalgout: Yeah I think this could be a `toggleSelected` just like `toggleClass`.. and we can just support passing the boolean.. no need to auto-toggle for now.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -79,6 +81,26 @@ const StateModel = Backbone.Model.extend({
...
+
+    this.set(listName, _.extend(list, {
+      [model.id]: true,
+    }));
+  },
+  removeSelected(model) {
+    const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
+    const list = _.clone(this.get(listName));
+
+    delete list[model.id];
```
paulfalgout:
```js
    this.set(listName, _.extend(list, {
      [model.id]: false,
    }));
```
should be ok.

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -112,10 +112,25 @@ const ListView = CollectionView.extend({
...
   tagName: 'table',
   childView() {
-    return this.getOption('isFlowList') ? FlowItemView : ActionItemView;
+    return this.isFlowList ? FlowItemView : ActionItemView;
   },
   emptyView() {
-    return this.getOption('isFlowList') ? FlowEmptyView : ActionEmptyView;
+    return this.isFlowList ? FlowEmptyView : ActionEmptyView;
+  },
+  childViewTriggers: {
```
paulfalgout: can axe this

paulfalgout (changes requested):

paulfalgout (approved):

PR #316, opened 2020-06-17
Multi-select view for worklists

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22632]

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -9,7 +9,7 @@ import App from 'js/base/app';
 
 import FiltersApp from './filters_app';
 
-import { ListView, TooltipView, LayoutView, TableHeaderView, SortDroplist, sortDueOptions, sortUpdateOptions } from 'js/views/patients/worklist/worklist_views';
+import { ListView, TooltipView, LayoutView, TableHeaderView, SortDroplist, sortDueOptions, sortUpdateOptions, MultiEditButtonView } from 'js/views/patients/worklist/worklist_views';
 
 const StateModel = Backbone.Model.extend({
   defaults() {
@@ -21,6 +21,8 @@ const StateModel = Backbone.Model.extend({
         groupId: null,
         clinicianId: this.currentClinician.id,
       },
+      selectedActions: {},
+      selectedFlows: {},
     };
   },
   preinitialize() {
@@ -79,6 +81,56 @@ const StateModel = Backbone.Model.extend({
 
     return filters[this.id];
   },
+  getSelectedList() {
+    return this.isFlowType() ? this.get('selectedFlows') : this.get('selectedActions');
+  },
+  toggleSelected(model, isSelected) {
+    const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
+    const list = _.clone(this.get(listName));
+
+    this.set(listName, _.extend(list, {
+      [model.id]: isSelected,
+    }));
+  },
+  isSelected(model) {
+    const list = this.getSelectedList();
+
+    return list[model.id];
+  },
+  getSelected(collection) {
+    if (!collection) return;
+
+    const list = this.getSelectedList();
+    const collectionSelected = _.reduce(_.keys(list), (selected, item) => {
+      if (list[item] && collection.get(item)) {
+        selected.push({
+          id: item,
+        });
+      }
+
+      return selected;
+    }, []);
+
+    return Radio.request('entities', `${ this.getType() }:collection`, collectionSelected);
+  },
+  clearSelected() {
+    const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
+
+    this.set(listName, {});
+    this.trigger('select:none');
+  },
+  selectAll(collection) {
+    const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
+
+    const list = collection.reduce((selected, model) => {
+      selected[model.id] = true;
+      return selected;
+    }, {});
+
+    this.set(listName, list);
+
+    this.trigger('select:all');
+  },
 });
 
 export default App.extend({
@@ -95,9 +147,14 @@ export default App.extend({
     'change:filters': 'restart',
     'change:actionsSortId': 'onChangeStateSort',
     'change:flowsSortId': 'onChangeStateSort',
+    'change:selectedFlows': 'onChangeSelected',
+    'change:selectedActions': 'onChangeSelected',
   },
   onChangeStateSort() {
     this.getChildView('list').setComparator(this.getComparator());
+  },
+  onChangeSelected() {
+    this.toggleMultiSelect();
   },
   initListState() {
     const storedState = store.get(this.worklistId);
@@ -123,7 +180,26 @@ export default App.extend({
     this.showSortDroplist();
     this.showTableHeaders();
     this.showTooltip();
+    this.startFiltersApp();
+  },
+  beforeStart() {
+    const filter = this.getState().getEntityFilter();
+    return Radio.request('entities', `fetch:${ this.getState().getType() }:collection`, { filter });
+  },
+  onStart(options, collection) {
+    this.collection = collection;
 
+    const collectionView = new ListView({
+      collection: this.collection,
+      state: this.getState(),
+      viewComparator: this.getComparator(),
+    });
+
+    this.showChildView('list', collectionView);
+
+    this.toggleMultiSelect();
+  },
+  startFiltersApp() {
     const filtersApp = this.startChildApp('filters', {
       state: this.getState().getFilters(),
       shouldShowClinician: this.getState().id === 'owned-by',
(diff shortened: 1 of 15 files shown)


LisaManresa:
Still working out test failures, but wanted to get eyes on this.

src/sass/core/_reset.scss
```diff
@@ -130,7 +130,7 @@ input[type=checkbox] {
   border: 1px solid $black-20;
   border-radius: 2px;
   cursor: pointer;
-  display: block;
+  display: inline-block;
```
paulfalgout: So I wasn't paying attention on the last PR.  I think this should maybe move out of the reset.  The reset is just trying to make all elements the same between browsers, so we should add this to something else.. maybe either a core `_form` or just in the `modules/forms`

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -90,10 +93,31 @@ const StateModel = Backbone.Model.extend({
     }));
   },
   isSelected(model) {
-    const list = this.isFlowType() ? this.get('selectedFlows') : this.get('selectedActions');
+    const list = this.getSelectedList();
 
     return list[model.id];
   },
+  getSelected(collection) {
```
paulfalgout, line 100: I think this function instead of pick should use reduce that reduces to all available `collection.get(id)`s then this function should return `Radio.request('entities', `${ this.getType():collection', reducedModels);`  That collection can be passed to the button so that onClick it can pass it to the sidebar.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -90,10 +93,31 @@ const StateModel = Backbone.Model.extend({
...
+  },
+  clearSelected() {
+    const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
+
+    this.set(listName, {});
+  },
+  selectAll(collection) {
+    const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
+
+    this.set(listName, _.reduce(collection.models, (selected, model) => {
```
paulfalgout: I think this could be 
```
collection.map(model=> {
return [model.id] = true;
});
```

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -110,10 +134,23 @@ export default App.extend({
...
   },
+  onChangeSelected() {
+    const selected = _.keys(this.getState().getSelected(this.collection));
+
+    if (selected.length) {
+      this.showMultiEditButtonView();
+      return;
+    }
+
+    this.getRegion('filters').empty();
```
paulfalgout: why is the empty needed?

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -128,17 +165,40 @@ export default App.extend({
...
+    this.showChildView('list', this.collectionView);
+
+    if (_.keys(this.getState().getSelected(this.collection)).length) {
+      this.showMultiEditButtonView();
+      return;
+    }
+
+    this.startFiltersApp();
+  },
+  startFiltersApp() {
```
paulfalgout, line 202: hmm moving this to onStart is unfortunate.  it means that the filters will be reset every time they're changed..  So .. you pick a value in the filter.. and then realize you picked the wrong one.. so you start to pick the other one.. but during that the first request comes in and re-starts the filters and your droplist is now replaced with a new one.. 

I think what we can do is just be ok with the filters showing up during the `onBeforeStart` and the multi-select showing `onStart`  and in `onStart` it'll be ok if `startFiltersApp` is called again since it'll immediately bail if it is already running.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -110,10 +134,23 @@ export default App.extend({
...
     'change:actionsSortId': 'onChangeStateSort',
     'change:flowsSortId': 'onChangeStateSort',
+    'change:selectedFlows': 'onChangeSelected',
+    'change:selectedActions': 'onChangeSelected',
   },
   onChangeStateSort() {
     this.getChildView('list').setComparator(this.getComparator());
   },
+  onChangeSelected() {
+    const selected = _.keys(this.getState().getSelected(this.collection));
```
paulfalgout: I'd bet there could be one function that shows/clears depending on `selected.length` that could be called here or in onStart

src/js/views/patients/worklist/flow_views.js
```diff
@@ -81,6 +81,11 @@ const FlowItemView = View.extend({
     this.$el.toggleClass('is-selected', !isSelected);
     this.state.toggleSelected(this.model, !isSelected);
   },
+  // onAttach() {
```
paulfalgout: ?

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -148,20 +208,32 @@ export default App.extend({
...
     });
   },
-  beforeStart() {
-    const filter = this.getState().getEntityFilter();
-    return Radio.request('entities', `fetch:${ this.getState().getType() }:collection`, { filter });
-  },
-  onStart(options, collection) {
-    this.collection = collection;
+  showMultiEditButtonView() {
+    this.getChildApp('filters').stop();
```
paulfalgout: I think this stop could move to the function that switches between these two things.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -148,20 +208,32 @@ export default App.extend({
...
+  onClickMultiSelect() {
+    if (_.keys(this.getState().getSelected(this.collection)).length === this.collection.length) {
+      this.getState().clearSelected();
+      this.collectionView.triggerMethod('select:none');
+      return;
+    }
 
-    this.showChildView('list', collectionView);
+    this.getState().selectAll(this.collection);
+    this.collectionView.triggerMethod('select:all');
```
paulfalgout: I think these could move to triggers on the state itself at the end of `clearSelected` / `selectAll` which the collectionView can listen and react to.

paulfalgout (changes requested):

src/js/apps/patients/worklist/worklist_app.js
LisaManresa: Yup, I had mulled over that solution as well. I like that organization better as well.

src/js/apps/patients/worklist/worklist_app.js
LisaManresa: 👍 Although, I think we still need the `_.pick`. `list` is a misnomer, it's actually an object. And we need to check that the value isn't falsey before selecting it for the collection. So maybe `_.keys(_.pick(...))`

paulfalgout: `mapObject` I think

paulfalgout: err. that'll return if falsy.. hmm

src/js/apps/patients/worklist/worklist_app.js
LisaManresa: I think this is the same issue as above, the selected lists in the state aren't collections, they're objects:
```
selectedFlows: {
  '1': true,
  '2': false,
  '3': true
}
```
We could just do a forEach on the collection and then `list[model.id] = true`.

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: maybe...
`_.reduce(_.keys(list)`
then.. from that array of keys you can do `list[key]` to check the value and return an array of models, so all the `.get`s don't have to happen twice?

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: I think `reduce` would do it.  you can reduce an array to a single object.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -90,10 +93,42 @@ const StateModel = Backbone.Model.extend({
...
 
     return list[model.id];
   },
+  getSelected(collection) {
+    if (!collection) return;
+
+    const list = this.getSelectedList();
+    const collectionSelected = _.reduce(_.keys(list), (selected, item) => {
+      if (list[item] && collection.get(item)) {
+        return selected.concat([item]);
```
paulfalgout: `selected.push(collection.get(item))` I think.. does the entity work if it's just a collection of id strings?  I assumed it'd either need to be models or `{ id: 'idstring' }`

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -128,25 +168,17 @@ export default App.extend({
     }
 
     this.worklistId = worklistId;
-    this.initListState();
 
     this.showView(new LayoutView({
       worklistId: this.worklistId,
     }));
 
+    this.initListState();
```
paulfalgout: why did this move?

paulfalgout (changes requested):
seems good.. but one question on one of the reduces

src/js/apps/patients/worklist/worklist_app.js
LisaManresa: Ah thanks. I hadn't dug into the model attributes to verify it was building the collection correctly.

paulfalgout (approved):

PR #318, opened 2020-06-24
Trigger multi-edit sidebar, with (partial) delete functionality

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22692]


src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -282,4 +292,36 @@ export default App.extend({
 
     this.showChildView('tooltip', tooltipView);
   },
+  showEditActionsSidebar() {
+    const headerView = new BulkEditActionsHeaderView({
```
paulfalgout: is there a plan to change this here?  Seems like this needs to be the entire sidebar not just the header.  FWIW thought I think the rest of the sidebar can probably be handled within this view.. I just think maybe the naming is too narrow.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -282,4 +292,36 @@ export default App.extend({
...
+      actions: this.getState().getSelected(this.collection),
+    });
+
+    const modal = Radio.request('modal', 'show:sidebar', {
+      headerView: headerView,
+    });
+
+    this.listenTo(headerView, {
+      'delete'(actions) {
```
paulfalgout: I think we're going to need to do something like `actions.destroy();`  that function on the collection will need to do the batch delete and then on success... it'll probably have to simulate `model` `destroy` events so that any collection or whatever that's displaying that model anywhere handles it.  
_probably_ what we need to do is disable the delete button on click and not `destroy` the modal or clear the state until the patch delete succeeds. (which will remove the models from the collection without from the destroy event)

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -282,4 +292,36 @@ export default App.extend({
 
     this.showChildView('tooltip', tooltipView);
   },
+  showEditActionsSidebar() {
+    const headerView = new BulkEditActionsHeaderView({
+      actions: this.getState().getSelected(this.collection),
```
paulfalgout: We're doing this `state.getSelected` in a few places but they're all downstream of the first time.  I wonder if `toggleMultiSelect` should just store the selected collection on `this` and pass it to the multibutton so it doesn't have to get it again.
Similarly I think this is _slightly_ simpler if you just pass it here as `collection:` then in the view's template it's `items.length` or `this.collection.length` if needed in the view code.

paulfalgout (changes requested):
I'm not sure yet, but we may end up wanting to put the sidebar views in a separate file or directory.. maybe even making the sidebar itself a child app for org purposes.. a _little_ overkill probably, but all of these files are getting a bit big...and all of the components we're about to add aren't going to make that smaller

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: I just made this method into a separate card I'm going to look at real quick like.

src/js/apps/patients/worklist/worklist_app.js
LisaManresa: Yes, we'll also have a view for the sidebar body with the components. Since the sidebar is a modal in this case, we need a separate header, body and (potentially) footer view, don't we? Or would it make more sense to modify the sidebar modal to work more like the custom modal where we can pass a single view?

paulfalgout: ah no I see it now.. didn't realize this was being passed as `headerView` to the modal.. all good.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -282,4 +280,49 @@ export default App.extend({
...
+      collection: this.selected,
+    });
+
+    const modal = Radio.request('modal', 'show:sidebar', {
+      headerView: headerView,
+    });
+
+    this.listenTo(headerView, {
+      'delete'(flows) {
+        this.onBulkDelete(flows);
```
paulfalgout: I think if we want to make this an `on` Handler it should be `'delete': _.partial(this.onBulkDelete, modal)`  otherwise we can just call `flows.destroy();` `modal.destroy()` and clear the state here.  There should be no need to manually remove the models from `this.collection`

src/js/base/collection.js
```diff
@@ -34,4 +34,16 @@ export default Backbone.Collection.extend(_.extend({
...
   },
+  destroy(options) {
+    return this.sync('delete', this, {
+      url: _.result(this, 'url'),
+      data: JSON.stringify({
+        data: _.pluck(this.models, 'id', 'type'),
+      }),
+    }).always(() => {
+      this.models.each(action => {
+        action.trigger('destroy', action, action.collection, options);
```
paulfalgout: bah.  this says `action`  it should just be `model`  :-/  my bad.  I made this on actions and then realized it's completely generic.

paulfalgout (changes requested):

LisaManresa:
Still lots to do on this. But the basic functionality of the components (for actions only) is implemented. Still need to :
- apply this rendering logic to flows
- make the time component accommodate the values from the due date component
- Make the corresponding model changes/server requests on save

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
```diff
@@ -0,0 +1,436 @@
...
+  },
+  getBulkDuration() {
+    const duration = this.collection.at(0).get('duration');
+    const hasDifferentDuration = this.collection.some(item => {
+      return item.get('duration') !== duration;
+    });
+
+    return hasDifferentDuration ? 'multiple' : duration;
+  },
+  showState() {
```
paulfalgout, line 289: so what if in the app that creates this view it does all of the getting of data and sets the model for this view to include data like:
```js
{
 stateId: null,
 stateMulti: true,
 due_date: null,
 dueMulti: false,
}
```
then in this view we have:
```js
modelEvents: {
  'change:stateMulti': 'showState',
},
getStateComponent() {
  const stateId = this.model.get('stateId');
  if (this.model.get('stateMulti')) {
    return new StateComponent({ 
      stateId,
      viewOptions: {
        className: 'button-secondary w-100',
        template: MultipleStateTemplate,
      },
    });
  }
  return new StateComponent({ stateId });
},
showState() {
  const stateComponent = this.getStateComponent();
  
  this.listenTo(stateComponent, 'change:state', state => {
    this.model.set({ stateId: state, stateMulti: false });
  });

  this.showChildView('state', stateComponent);
}
```

This way we're only overriding `viewOptions` if there's multiple values, and once something is chosen and/or the data is the same, it just rebuilds the regular component and we don't have to duplicate as much of the component functionality.

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
```diff
@@ -0,0 +1,436 @@
...
+      },
+    });
+  },
+});
+
+const BulkEditFlowsHeaderView = View.extend({
+  template: hbs`
+    <div class="flex">
+      <div class="flex-grow">
+        <h3 class="sidebar__heading">{{formatMessage  (intlGet "patients.worklist.bulkEditViews.bulkEditFlowsHeaderView.headingText") itemCount=itemCount}}</h3>
```
paulfalgout: I believe `itemCount` here can be `items.length` and no templateContext is necessary

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
```diff
@@ -0,0 +1,436 @@
...
+        onSelect: _.bind(this.triggerMethod, this, 'confirm:delete'),
+      },
+    ]);
+
+    const optionlist = new Optionlist({
+      ui: this.ui.menu,
+      uiView: this,
+      headingText: i18n.bulkEditActionsHeaderView.menuOptions.headingText,
+      itemTemplate: hbs`<span class="sidebar__delete-icon">{{far "trash-alt"}}</span>{{formatMessage  (intlGet "patients.worklist.bulkEditViews.bulkEditActionsHeaderView.menuOptions.delete") itemCount=itemCount}}`,
+      itemTemplateContext() {
```
paulfalgout: could be `itemTemplateContext: { itemCount },`

paulfalgout (changes requested):

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
LisaManresa: Good call. That would clean things up a lot. And you're right, we only need to know about the multiple attributes when the view first initializes.

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
```diff
@@ -0,0 +1,397 @@
...
+
+    this.listenTo(ownerComponent, 'change:owner', owner => {
+      this.model.set({ owner, ownerMulti: false });
+    });
+
+    this.showChildView('owner', ownerComponent);
+  },
+});
+
+const BulkEditFlowsSuccessView = View.extend({
```
paulfalgout: looks like these aren't needed anymore

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
```diff
@@ -0,0 +1,397 @@
...
+        onSelect: _.bind(this.triggerMethod, this, 'confirm:delete'),
+      },
+    ]);
+
+    const optionlist = new Optionlist({
+      ui: this.ui.menu,
+      uiView: this,
+      headingText: i18n.bulkEditFlowsHeaderView.menuOptions.headingText,
+      itemTemplate: hbs`<span class="sidebar__delete-icon">{{far "trash-alt"}}</span>{{formatMessage  (intlGet "patients.worklist.bulkEditViews.bulkEditFlowsHeaderView.menuOptions.delete") itemCount=itemCount}}`,
+      itemTemplateContext() {
```
paulfalgout: `itemTemplateContext: { itemCount },`

paulfalgout (changes requested):
Looks right.  I dunno if it works, but it's looking solid.  two small things

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -253,6 +279,22 @@ export default App.extend({
 
     return _.union(sortDueOptions, sortUpdateOptions);
   },
+  showDeleteSuccess(itemCount) {
+    if (this.getState().isFlowType()) {
+      Radio.request('alert', 'show:success', renderTemplate(hbs`{{formatMessage  (intlGet "patients.worklist.worklistApp.bulkDeleteFlowsSuccess") itemCount=itemCount}}`, { itemCount }));
```
paulfalgout: I don't like the `renderTemplate(hbs`...intlGet` pattern.. it's so cumbersome, but I can't seem find an alternative.

LisaManresa: Yeah, I wasn't a huge fan of it either. The other option would be to not display the number of items, but that seems like a degradation in the user experience. To clean things up, I could move the templates to the views file and import them here.

paulfalgout: Yeah I think defining the templates over there is better anyhow.. keep as much of intl out of the apps as possible.

PR #321, opened 2020-06-29
Fix check-in data discrepancies & check-in event display bugs

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22743]

- Aligns FE data structure expectations with what is actually returned from the BE. This includes overriding `parseModel` on the check-in entity because BE has to format the response as `data.checkin` instead of `data.attributes`.
- Applies a filter to the patient events request to only retrieve `PatientCheckInCompleted` type.

### DIFF
--- a/src/js/base/moment.js
+++ b/src/js/base/moment.js
@@ -45,7 +45,7 @@ Handlebars.registerHelper({
   formatMoment(date, format, { hash = {} }) {
     if (!date) return new Handlebars.SafeString(hash.defaultHtml || '');
 
-    date = moment(date, hash.inputFormat);
+    date = hash.utc ? moment.utc(date, hash.inputFormat).local() : moment(date, hash.inputFormat);
 
     date = formatDate(date, format);
 
--- a/src/js/entities-service/entities/check-ins.js
+++ b/src/js/entities-service/entities/check-ins.js
@@ -11,6 +11,9 @@ const _Model = BaseModel.extend({
   getContent() {
     return new Backbone.Collection(this.get('content'));
   },
+  parseModel(data) {
+    return data.checkin;
+  },
 });
 
 const Model = Store(_Model, TYPE);
--- a/src/js/views/check-ins/check-in/check-in_views.js
+++ b/src/js/views/check-ins/check-in/check-in_views.js
@@ -12,7 +12,7 @@ const ContextTrailView = View.extend({
   <a class="js-back check-in__context-link">
     {{fas "chevron-left"}}{{ @intl.checkIns.checkInViews.contextTrailView.backBtn }}
   </a>
-  {{fas "chevron-right"}}{{ patient.first_name }} {{ patient.last_name }} - {{formatHTMLMessage (intlGet "checkIns.checkInViews.contextTrailView.checkInCompleted") date=(formatMoment finished_at "LONG" inputFormat="YYYY-MM-DD") }}
+  {{fas "chevron-right"}}{{ patient.first_name }} {{ patient.last_name }} - {{formatHTMLMessage (intlGet "checkIns.checkInViews.contextTrailView.checkInCompleted") date=(formatMoment finishedTs "LONG") utc=true }}
   `,
   templateContext() {
     const patient = this.patient;
--- a/test/integration/check-ins/check-in.js
+++ b/test/integration/check-ins/check-in.js
@@ -4,7 +4,7 @@ import formatDate from 'helpers/format-date';
 
 context('patient check-in', function() {
   specify('display patient check-in', function() {
-    const completedDate = moment().subtract(7, 'days');
+    const completedDate = '2020-06-17 16:05:22';
 
     cy
       .server()
@@ -15,7 +15,7 @@ context('patient check-in', function() {
         return fx;
       })
       .routePatientCheckIn(fx => {
-        fx.data.attributes.finished_at = completedDate.format();
+        fx.data.checkin.finishedTs = completedDate;
 
         return fx;
       })
@@ -27,7 +27,7 @@ context('patient check-in', function() {
       .get('.check-in__frame')
       .find('.check-in__context-trail')
       .should('contain', 'Test Patient')
-      .should('contain', formatDate(completedDate, 'LONG'));
+      .should('contain', formatDate(moment.utc(completedDate).local(), 'LONG'));
 
     cy
       .get('.check-in__frame')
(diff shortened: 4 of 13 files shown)


src/js/entities-service/entities/check-ins.js
```diff
@@ -11,6 +12,13 @@ const _Model = BaseModel.extend({
   getContent() {
     return new Backbone.Collection(this.get('content'));
   },
+  parseModel(data) {
```
paulfalgout, line 14: I think this whole function could be `return data.checkin;`  The rest of this is relatively jsonapi stuff.. the `cached_ts` is not, but these are display only anyhow so won't need that.

test/support/api/check-ins.js
```diff
@@ -8,8 +8,13 @@ Cypress.Commands.add('routePatientCheckIn', (mutator = _.identity) => {
   cy.route({
     url: /api\/patients\/.\/checkins\?filter\[checkin\]/,
     response() {
+      const data = getResource(this.fxCheckInResponse, 'check-ins');
```
paulfalgout: I don't think we need the jsonapi helpers here at all.  `return mutator({ data: this.fxCheckInReponse });`

test/integration/check-ins/check-in.js
```diff
@@ -15,7 +15,7 @@ context('patient check-in', function() {
         return fx;
       })
       .routePatientCheckIn(fx => {
-        fx.data.attributes.finished_at = completedDate.format();
+        fx.data.checkin.finishedTs = completedDate.format();
```
paulfalgout: so this doesn't looks like it is formatted ISO 8601 and is `YY:MM:DD hh:mm:ss` in the data.  I assume these dates are UTC coming from the server and so I _believe_ we'll need to convert them to utc prior to displaying them as well by using `moment.utc(this.model.get('finishedTs'))`

src/js/views/check-ins/check-in/check-in_views.js
```diff
@@ -12,7 +12,7 @@ const ContextTrailView = View.extend({
   <a class="js-back check-in__context-link">
     {{fas "chevron-left"}}{{ @intl.checkIns.checkInViews.contextTrailView.backBtn }}
   </a>
-  {{fas "chevron-right"}}{{ patient.first_name }} {{ patient.last_name }} - {{formatHTMLMessage (intlGet "checkIns.checkInViews.contextTrailView.checkInCompleted") date=(formatMoment finished_at "LONG" inputFormat="YYYY-MM-DD") }}
+  {{fas "chevron-right"}}{{ patient.first_name }} {{ patient.last_name }} - {{formatHTMLMessage (intlGet "checkIns.checkInViews.contextTrailView.checkInCompleted") date=(formatMoment finishedTs "LONG" inputFormat="YYYY-MM-DD") }}
```
paulfalgout: as mentioned below we'll have to convert this date to utc first.  I'm pretty sure since the string doesn't specify it'll read it as the local time.  The inputFormat is wrong as well, but I don't know that we'll need it at all.  I _think_ what we'll want to maybe do is add a `utc` flag to the hash like `inputFormat` like `utc=true` and then if `formatMoment` modify the line like:
```js
date = hash.utc ? moment.utc(date, hash.inputFormat) : moment(date, hash.inputFormat)
```

paulfalgout (changes requested):

src/js/views/check-ins/check-in/check-in_views.js
LisaManresa: This doesn't seem to be clicking for me. If we're always converting the date to utc before rendering, do we still need the `utc` flag?

paulfalgout: this is what would be converting the date string to utc, so this'd be to display an attribute on the model unmodified that's stored in utc but is not designated utc

test/integration/check-ins/check-in.js
```diff
@@ -27,7 +27,7 @@ context('patient check-in', function() {
       .get('.check-in__frame')
       .find('.check-in__context-trail')
       .should('contain', 'Test Patient')
-      .should('contain', formatDate(completedDate, 'LONG'));
+      .should('contain', formatDate(moment.utc(completedDate), 'LONG'));
```
paulfalgout: I'm not sure this is right.. I think it needs to be `moment.utc(completedDate).local()`
```
moment.utc('2020-06-17 16:05:22').format('lll')
"Jun 17, 2020 4:05 PM"
moment.utc('2020-06-17 16:05:22').local().format('lll')
"Jun 18, 2020 1:05 AM"
moment('2020-06-17T16:05:22Z').format('lll')
"Jun 18, 2020 1:05 AM"
```

test/integration/check-ins/check-in.js
```diff
@@ -37,10 +37,10 @@ context('patient check-in', function() {
       .next()
       .find('.check-in__patient-answer');
 
-    cy
-      .get('.check-in__frame')
-      .find('.check-in__context-trail .js-back')
-      .click();
+    // cy
```
paulfalgout: 👉

src/js/views/check-ins/check-in/check-in_views.js
```diff
@@ -12,7 +12,7 @@ const ContextTrailView = View.extend({
   <a class="js-back check-in__context-link">
     {{fas "chevron-left"}}{{ @intl.checkIns.checkInViews.contextTrailView.backBtn }}
   </a>
-  {{fas "chevron-right"}}{{ patient.first_name }} {{ patient.last_name }} - {{formatHTMLMessage (intlGet "checkIns.checkInViews.contextTrailView.checkInCompleted") date=(formatMoment finished_at "LONG" inputFormat="YYYY-MM-DD") }}
+  {{fas "chevron-right"}}{{ patient.first_name }} {{ patient.last_name }} - {{formatHTMLMessage (intlGet "checkIns.checkInViews.contextTrailView.checkInCompleted") date=(formatMoment finishedTs "LONG") }}
```
paulfalgout: this would need a `utc=true` I believe otherwise this will load the date as if it is in the local timezone

src/js/base/moment.js
```diff
@@ -45,7 +45,7 @@ Handlebars.registerHelper({
   formatMoment(date, format, { hash = {} }) {
     if (!date) return new Handlebars.SafeString(hash.defaultHtml || '');
 
-    date = moment(date, hash.inputFormat);
+    date = hash.utc ? moment.utc(date, hash.inputFormat) : moment(date, hash.inputFormat);
```
paulfalgout: I think I was incorrect.. as below I _believe_ this'll need `moment.utc(...).local()` so that the date is read as utc but stored locally as it would be if it included the utc indicator in the string.

paulfalgout (changes requested):

PR #328, opened 2020-07-09
Change bulk edit requests

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch22824]

I'll need to update tests for this PR, but there's a few things in here.
In addition to moving to bulk PATCHes, I changed the checkbox to a button with the icons in the marvel.
Additionally found a few bugs in various data scenarios in the sidebar.

I'm not 100% sure this is it, but it's close.  Probably going to need quite a few more tests for all the edges added here.


src/js/base/collection.js
```diff
@@ -35,16 +35,10 @@ export default Backbone.Collection.extend(_.extend({
...
-      url: _.result(this, 'url'),
-      data: JSON.stringify({
-        data: _.pluck(models, 'id', 'type'),
-      }),
-    }).always(() => {
-      _.each(_.clone(models), function(model) {
-        model.trigger('destroy', model, model.collection, options);
-      });      
-    });
+    const models = _.clone(this.models);
```
LisaManresa, line 38: 👍  Way cleaner.

test/integration/patients/worklist.js
```diff
@@ -1820,48 +1866,78 @@ context('worklist page', function() {
...
-      .parents('.table-list__item')
-      .next()
-      .find('.js-select')
-      .click()
-      .parents('.table-list__item')
-      .next()
-      .find('.js-select')
-      .click();
+      .get('.table-list')
+      .find('.table-list__item')
```
paulfalgout: Does this work?
```js
cy
  .get('.table-list')
  .find('table-list__item .js-select')
  .click({ multiple: true });
```

paulfalgout: @LisaManresa

LisaManresa: I'll give that a try.

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
```diff
@@ -86,12 +86,10 @@ const BulkEditActionsHeaderView = View.extend({
       </div>
       <div>
         <button class="button--icon u-margin--r-8 js-menu">{{far "ellipsis-h"}}</button>
-        <button class="button--icon js-close">{{fas "times"}}</button>
```
paulfalgout, line 95: why was this removed @LisaManresa ?

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
```diff
@@ -270,7 +266,8 @@ const BulkEditActionsBodyView = View.extend({
     }
 
     const time = this.model.get('time');
-    const isDisabled = !this.model.get('date') || (this.model.get('dateMulti') && !time);
+    const isDisabled = (this.model.get('dateMulti') && !time) || !this.model.get('date');
```
paulfalgout, line 269: why flip this around?

LisaManresa: This was mostly for test coverage. But I may be able to rethink the check itself because it may be overkill.

If place after the or, `(this.model.get('dateMulti') && !time)` is never evaluated because `this.model.get('date')` is `null` when `dateMulti` is true. https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/patients/worklist/sidebar/bulk-edit-actions_app.js#L47

I _think_ this check could just be `const isDisabled = !time || !this.model.get('date');`

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
LisaManresa: The sidebar is a modal and already has a close button. This one was covering up that button. Tests were looking for the modal close button and couldn't click it because it was covered.

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
paulfalgout: Time should be disabled when:
- Multiple Dates / Select Time
- Multiple Dates / Multiple Times
- Select Date / SelectTime

Time should be enabled when:
- Jun 30 / Multiple Times
- Jun 30 / 7:00
- Jun 30 / Select Time

paulfalgout: it's like the first check needs to be `(!this.model.get('date') && !this.model.get('dateMulti'))`

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
paulfalgout: so the ellipsis menu fits right up there next to the modal cleanly?  no vertical alignment issues?

paulfalgout (commented):
Worklist bulk editing -- bulk actions editing (failed)
Worklist bulk editing -- bulk flows editing (failed)
Locally I'm getting errors on cypress if I am running docker.  Can we throw some generic PATCH /api/action/*  (or flow) at the top of these tests so they're hitting something in cypress?

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
LisaManresa: This logic is correct for all of those scenarios. This switch just ensures that we hit the `(this.model.get('dateMulti') && !time)` check in the multiple dates scenarios. Otherwise that condition was never getting evaluated (though the result was still correct).

PR #331, opened 2020-07-14
Bulk edit UI bugs

LisaManresa:
- [x] [ch22874]

- Add hover states to checkboxes
- Add sidebar animation to modal sidebar
- Change "Due In" to "Due Date"

### DIFF
--- a/src/js/views/globals/modal/modal_views.js
+++ b/src/js/views/globals/modal/modal_views.js
@@ -6,6 +6,8 @@ import 'sass/modules/modals.scss';
 
 import intl from 'js/i18n';
 
+import { animSidebar } from 'js/anim';
+
 import PreloadRegion from 'js/regions/preload_region';
 
 import ModalTemplate from './modal.hbs';
@@ -73,6 +75,26 @@ const ModalView = View.extend({
   },
 });
 
+const SidebarModalView = ModalView.extend({
+  className: 'modal--sidebar',
+  bodyClass: 'sidebar',
+  onAttach() {
+    animSidebar(this.el);
+  },
+});
+
+const SmallModalView = ModalView.extend({
+  className: 'modal--small',
+  bodyClass: 'modal-content',
+});
+
+const TallModalView = ModalView.extend({
+  className: 'modal--tall',
+  bodyClass: 'modal-content--tall',
+});
 export {
   ModalView,
+  SidebarModalView,
+  SmallModalView,
+  TallModalView,
 };
--- a/src/js/views/patients/worklist/worklist-list.scss
+++ b/src/js/views/patients/worklist/worklist-list.scss
@@ -7,12 +7,13 @@
 }
 
 .worklist-list__item-check {
-  cursor: default;
   padding: 16px 0;
   width: 16px;
+}
 
-  .fa-check-square {
-    color: color(green);
+.work-list__item:hover {
+  .fa-square {
+    color: $black-20;
   }
 }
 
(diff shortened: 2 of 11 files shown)


src/js/views/globals/modal/modal_views.js
```diff
@@ -73,6 +75,13 @@ const ModalView = View.extend({
   },
 });
 
+const SidebarModalView = ModalView.extend({
```
paulfalgout, line 78: since we're making a view here let's move over the 
```js
      className: 'modal--sidebar',
      bodyClass: 'sidebar',	
```
in fact.. we should probably just go ahead a make a tall and a small and move the classes out of that app entirely.

src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -14,6 +14,14 @@
   .fa-check-square {
     color: color(green);
   }
+
+  .fa-square {
+    color: $black-90;
```
paulfalgout: 90 or 80?  90 below is disabled

paulfalgout (changes requested):

src/js/views/patients/worklist/worklist-list.scss
LisaManresa: Ah thanks. Forgot we had two instances. What do you think about making these styles more global. We could add them to buttons.scss and create a `button-checkbox` class.

src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -7,12 +7,13 @@
...
-  cursor: default;
   padding: 16px 0;
   width: 16px;
+}
 
-  .fa-check-square {
-    color: color(green);
+.work-list__item:hover {
+  .fa-square {
+    color: $black-20;
```
paulfalgout, line 16: why is this needed and how is it work-list specific?

src/sass/modules/buttons.scss
```diff
@@ -204,3 +204,26 @@
...
+  .fa-square {
+    color: $black-80;
+
+    &:hover {
+      color: $black-20;
+    }
+  }
+
+  &:disabled {
+    cursor: default;
```
paulfalgout, line 217: is cursor even needed?

src/sass/modules/buttons.scss
```diff
@@ -204,3 +204,26 @@
     color: $black-80;
   }
 }
+
+.button--checkbox {
```
paulfalgout, line 208: would it work to do:
```css
.button--checkbox {
  color: $black-80;
  &:hover...
  &:disabled...
 .fa-check-square...
}
```
without .fa-square?  I can't remember if the icon has to have the color on it entirely.. even if it does maybe `.fa-square` could be color inherit?

paulfalgout (changes requested):

src/js/views/patients/worklist/worklist-list.scss
LisaManresa: Mark asked that the checkbox have the hover state when a user hovers anywhere on the row.

src/sass/modules/buttons.scss
LisaManresa: Yeah, cursor is needed because our reset scss sets the cursor for all buttons to pointer.

src/js/views/patients/worklist/worklist-list.scss
paulfalgout: 🙄

paulfalgout (approved):
for whenever this branch shows the latest update

PR #336, opened 2020-07-17
Use checkin_id to route to patient checkin from data and events

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch22918]

Fixed the patient check-in item in data and events so that it uses the `checkin_id` instead of `id` for routing. Also modified the view to explicitly check for `event_type: PatientCheckInCompleted` since other events won't have a `checkin_id`.


luketlancaster (approved):
👯

paulfalgout (approved):

PR #338, opened 2020-07-24
Support `optout` patient engagement status

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23008]

### DIFF
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -476,9 +476,11 @@ careOptsFrontend:
             status: |
               {status, select,
                 active {Active}
+                disabled {Off}
                 enrolled {Not Scheduled}
                 inactive {Inactive}
                 not_responding {Not Responding}
+                optout {Opted Out}
                 paused {Paused}
                 scheduled {Scheduled}
               }
--- a/src/sass/domain/engagement-status.scss
+++ b/src/sass/domain/engagement-status.scss
@@ -1,4 +1,6 @@
 .engagement-status__icon {
+  color: $black-90;
+
   .icon {
     font-size: 14px;
   }
@@ -7,6 +9,5 @@
   &.enrolled { color: color(blue); }
   &.inactive { color: color(orange); }
   &.not_responding { color: color(yellow); }
-  &.paused { color: $black-90; }
   &.scheduled { color: #8040E6; }
 }


src/js/i18n/en-US.yml
```diff
@@ -481,6 +481,9 @@ careOptsFrontend:
                 not_responding {Not Responding}
                 paused {Paused}
                 scheduled {Scheduled}
+                optout {Opted Out}
+                bad_email {Bad Email}
```
paulfalgout: @luketlancaster `bad_email` `created`?

luketlancaster: the only use I can find for `bad_email` in the old code is for login errors for clinician users...not sure why it was in the i18n for engagement that I sent to @LisaManresa

luketlancaster: also not sure about `created`....old value that's not used anymore maybe? Here's what the old system has: https://github.com/RoundingWell/RWell-rwmicro/blob/master/shared/i18n/en-US.yml#L407-L417

paulfalgout: On legacy frontend support [this](https://github.com/RoundingWell/RWell/blob/b666bbb1030c3324aa931a51a8af5775b5893ef8/mainapp/assets/js/static-options.js#L1):
```js
export const PATIENTSTATUS = {
    ENGAGEMENT_OFF: 'disabled',
    NOT_SCHEDULED: 'enrolled',
    SCHEDULED: 'scheduled',
    ACTIVE: 'active',
    NOT_RESPONDING: 'not_responding',
    INACTIVE: 'inactive',
    PAUSED: 'paused',
    OPTOUT: 'optout'
};
```
so I don't think the others are in use?  Also what about off/disabled?
@luketlancaster

luketlancaster: `'disabled'` is it's own status that comes with an `engagement_off_reason`, which is a string of various values (clinically na, other, no access, etc)

and looks like status is going to be `'disabled'` if engagement is off/on initial load-in of patients to the old system


does that answer any of your questions?

paulfalgout: how does a patient get disabled?... or paused for that matter?  I understand we have the `engagement_off_reason` things but to my knowledge the only way to set that is through the legacy interface.

luketlancaster: so I think the only way for engagement to be disabled, currently, is for either them to stop responding and so eventually the system turns it off, or for them to opt-out in the surveyapp. If they opt out via surveyapp they get a status of `'optout'`.

paulfalgout: I don't think patients go off by themselves.. they go inactive.. _pretty_ sure.. I mean if demo is still working there are tons of inactive patients.

paulfalgout (approved):

PR #341, opened 2020-08-05
Support dynamic patient sidebar widgets

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23034]

### DIFF
--- a/src/js/apps/patients/patient/sidebar/sidebar_app.js
+++ b/src/js/apps/patients/patient/sidebar/sidebar_app.js
@@ -1,8 +1,9 @@
+import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
-import { SidebarView, EngagementStatusPreloadView, EngagementStatusView } from 'js/views/patients/patient/sidebar/sidebar_views';
+import { SidebarView } from 'js/views/patients/patient/sidebar/sidebar_views';
 
 export default App.extend({
   onBeforeStart({ patient }) {
@@ -11,32 +12,34 @@ export default App.extend({
     this.orgEngagementEnabled = currentOrg.getSetting('engagement');
     this.patient = patient;
 
-    this.showView(new SidebarView({ 
+    const widgets = this.getSidebarWidgets();
+
+    this.showView(new SidebarView({
       model: this.patient,
-      orgEngagementEnabled: this.orgEngagementEnabled,
+      collection: widgets,
     }));
-    
-    if (this.orgEngagementEnabled) {
-      this.showChildView('engagement', new EngagementStatusPreloadView());
-    }
   },
   beforeStart({ patient }) {
     return Radio.request('entities', 'fetch:patient:engagementStatus', patient.id);
   },
-  onStart() {
-    this.showEngagementStatus();
-  },
   onFail() {
-    this.showEngagementStatus();
+    this.patient.trigger('status:notAvailable', true);
   },
-  showEngagementStatus() {
-    if (!this.orgEngagementEnabled) return;
-    
-    const engagementStatusView = this.showChildView('engagement', new EngagementStatusView({ model: this.patient }));
+  getSidebarWidgets() {
+    const collection = new Backbone.Collection([
+      { id: '1', widget_type: 'dob', display_name: 'Date of Birth' },
+      { id: '2', widget_type: 'sex', display_name: 'Sex' },
+      { id: '3', widget_type: 'divider' },
+      { id: '4', widget_type: 'engagement', display_name: 'Engagement Status' },
+      { id: '5', widget_type: 'divider' },
+      { id: '6', widget_type: 'groups', display_name: 'Groups' },
+    ]);
 
-    this.listenTo(engagementStatusView, 'click', this.showEngagementSidebar);
-  },
-  showEngagementSidebar() {
-    Radio.request('sidebar', 'start', 'engagement', { patient: this.patient });
+    /* TODO: clean this up once widget config is coming from the BE */
+    if (!this.orgEngagementEnabled) {
+      collection.remove([{ id: '5' }, { id: '6' }]);
+    }
+
+    return collection;
   },
 });
--- a/src/js/entities-service/entities/patients.js
+++ b/src/js/entities-service/entities/patients.js
@@ -12,9 +12,6 @@ const _Model = BaseModel.extend({
   getGroups() {
     return Radio.request('entities', 'groups:collection', this.get('_groups'));
   },
-  getFields() {
-    return Radio.request('entities', 'patientFields:collection', this.get('_patient_fields'));
-  },
 });
 
 const Model = Store(_Model, TYPE);
--- a/src/js/views/patients/patient/sidebar/patient-sidebar.scss
+++ b/src/js/views/patients/patient/sidebar/patient-sidebar.scss
@@ -13,17 +13,16 @@
   max-width: 256px;
   overflow: hidden;
 
-  &:not(:first-of-type):before {
-    border-top: 1px solid $black-90;
-    content: '';
-    display: block;
-    height: 0;
-    margin-bottom: 16px;
-  }
-
   &:last-of-type {
     padding-bottom: 32px;
   }
+}
+
+.patient-sidebar__divider {
+  border-top: 1px solid $black-90;
+  content: '';
+  display: block;
+  height: 0;
 }
 
 .patient-sidebar__heading {
@@ -34,11 +33,6 @@
   &:first-child {
     margin-top: 0;
   }
-}
-
-.patient-sidebar__item:empty:before {
-  // &ndash;
-  content: '–';
 }
 
 .patient-sidebar__item + .patient-sidebar__item {
--- a/src/js/views/patients/patient/sidebar/sidebar_views.js
+++ b/src/js/views/patients/patient/sidebar/sidebar_views.js
@@ -1,142 +1,132 @@
 import _ from 'underscore';
 import anime from 'animejs';
 import moment from 'moment';
+import Radio from 'backbone.radio';
 
 import hbs from 'handlebars-inline-precompile';
-import { View, CollectionView, Region } from 'marionette';
+import { View, CollectionView } from 'marionette';
 
-import intl, { renderTemplate } from 'js/i18n';
 import ModelRender from 'js/behaviors/model-render';
-
-import PatientSidebarTemplate from './patient-sidebar.hbs';
-import PatientSidebarGroupsTemplate from './patient-sidebar-groups.hbs';
 
 import './patient-sidebar.scss';
 import 'sass/domain/engagement-status.scss';
 
-const i18n = intl.patients.patient.sidebar.sidebarViews;
-
-const patientSet = [
-  {
-    heading: i18n.dobLabel,
-    template: hbs`{{formatHTMLMessage (intlGet "patients.patient.sidebar.sidebarViews.dob") dob=(formatMoment dob "LONG" inputFormat="YYYY-MM-DD") age=age}}`,
-    data(patient) {
-      const dob = patient.get('birth_date');
+const sidebarWidgets = {
+  dob: {
+    template: hbs`{{formatHTMLMessage (intlGet "patients.patient.sidebar.sidebarViews.sidebarWidgets.dob") dob=(formatMoment dob "LONG" inputFormat="YYYY-MM-DD") age=age}}`,
+    templateContext() {
+      const dob = this.model.get('birth_date');
 
       const age = moment().diff(moment(dob, 'YYYY-MM-DD'), 'years');
 
       return { dob, age };
     },
   },
-  {
-    heading: i18n.sexLabel,
-    template: hbs`{{formatMessage (intlGet "patients.patient.sidebar.sidebarViews.sex") sex=sex}}`,
+  sex: {
+    template: hbs`{{formatMessage (intlGet "patients.patient.sidebar.sidebarViews.sidebarWidgets.sex") sex=sex}}`,
   },
-];
-
-const InfoView = CollectionView.extend({
-  className: 'patient-sidebar__section',
-  showView(type, text) {
-    /* istanbul ignore next: prevents "null" string */
-    if (!text) text = '';
-
-    const view = new View({
-      className: `patient-sidebar__${ type }`,
-      template: _.constant(_.trim(text)),
-    });
-
-    this.addChildView(view);
+  divider: {
+    template: hbs`<div class="patient-sidebar__divider"></div>`,
   },
-  onRender() {
-    _.each(patientSet, this.showItem, this);
-
-    this.showPatientFields();
-  },
-  showItem({ heading, data = _.constant(this.model.attributes), template }) {
-    this.showView('heading', heading);
-    this.showView('item', renderTemplate(template, data(this.model)));
-  },
-  showPatientFields() {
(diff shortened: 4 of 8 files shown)


src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -11,32 +12,37 @@ export default App.extend({
...
+      status: 'Not Available',
+    });
   },
-  showEngagementStatus() {
-    if (!this.orgEngagementEnabled) return;
-    
-    const engagementStatusView = this.showChildView('engagement', new EngagementStatusView({ model: this.patient }));
+  getSidebarWidgets() {
+    const collection = new Backbone.Collection([
+      { id: '1', widget_type: 'name' },
```
paulfalgout: ah sorry.. I don't think name should be a widget.  It will always be visible and at the top.

src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -11,32 +12,37 @@ export default App.extend({
     this.orgEngagementEnabled = currentOrg.getSetting('engagement');
     this.patient = patient;
 
-    this.showView(new SidebarView({ 
+    const widgets = this.getSidebarWidgets();
+
+    this.showView(new SidebarView({
       model: this.patient,
-      orgEngagementEnabled: this.orgEngagementEnabled,
+      widgets,
```
paulfalgout: I think you could simplify this a little and make it the `collection`

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,151 @@
...
-const SidebarView = View.extend({
+const SidebarView = CollectionView.extend({
   className: 'patient-sidebar flex-region',
-  template: PatientSidebarTemplate,
-  templateContext() {
-    return {
-      engagementEnabled: this.getOption('orgEngagementEnabled'),
-    };
+  template: false,
+  onRender() {
```
paulfalgout: I think if you use `widgets` as the `collection` you can override `childViewOptions` to setup what needs to be passed to the widgetView.. then you can either viewFilter out widgets that don't exist or `widgetView` could handle it.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,151 @@
...
+      templateContext() {
+        return {
+          groups: _.map(this.model.getGroups().models, 'attributes'),
+        };
+      },
+    }),
   },
-});
+  engagement: {
+    view: View.extend({
```
paulfalgout: I think we can probably not nest this under `view`  _and_ in most cases you don't need the `View.extend`  `myRegion.show({ template, templateContext })` will create a new view.. so the only time you need the view.extend is when you're adding things like `getTemplate` or `onRender` functionality.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,151 @@
...
+    const widget = sidebarWidgets[config.get('widget_type')];
+
+    /* istanbul ignore next: prevents widget render error, remove when widgets are no longer hardcoded in app */
+    if (!widget) return;
 
-    this.showChildView('info', new InfoView({ model }));
-    this.showChildView('groups', new GroupsView({ model }));
+    this.addChildView(new widgetView({
+      view: widget.view,
+      config,
```
paulfalgout: I think the model should be config and `patient: this.model` might be better for the `widgetView` itself.. then that view can pass the patient as the model to the widget

test/integration/patients/patient/sidebar.js
```diff
@@ -127,7 +127,7 @@ context('patient sidebar', function() {
       .contains('Engagement settings for this patient could not be found.');
   });
 
-  specify('patient fields', function() {
+  /* specify('patient fields', function() {
```
paulfalgout: this can be removed entirely

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,151 @@
...
+  renderWidget(config) {
+    const widget = sidebarWidgets[config.get('widget_type')];
+
+    /* istanbul ignore next: prevents widget render error, remove when widgets are no longer hardcoded in app */
+    if (!widget) return;
 
-    this.showChildView('info', new InfoView({ model }));
-    this.showChildView('groups', new GroupsView({ model }));
+    this.addChildView(new widgetView({
+      view: widget.view,
```
paulfalgout: maybe call this option `widget` instead of `view`

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,151 @@
...
-      <span class="engagement-status__icon {{ engagement.status }} u-margin--r-4">{{fas "circle"}}</span>{{formatMessage (intlGet "patients.patient.sidebar.sidebarViews.engagementStatusView.status") status=engagement.status}}
-    {{else}}
-      <span class="patient-sidebar__no-engagement">{{ @intl.patients.patient.sidebar.sidebarViews.engagementStatusView.notAvailable }}</span>
-    {{/if}}
-  `,
-  triggers: {
-    'click': 'click',
+  onRender() {
+    const view = this.getOption('view');
+    this.showChildView('content', new view(this.options));
```
paulfalgout: should be able to do `this.showChildView('content', this.getOption('widget'));`

paulfalgout (changes requested):

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,136 @@
...
-      opacity: 0.5,
-      loop: true,
-      easing: 'easeInOutSine',
-      duration: 400,
-      direction: 'alternate',
-    });
+    const region = this.getRegion('content');
+    const widget = this.getOption('widget');
+
+    if (_.isFunction(widget)) {
```
LisaManresa: This is a bit clunky. But since widgets can either be an options object or a view, and also need to have the patient model passed to them, I think we need this check so we know how to append the model.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,136 @@
...
+    templateContext() {
+      const dob = this.model.get('birth_date');
 
       const age = moment().diff(moment(dob, 'YYYY-MM-DD'), 'years');
 
       return { dob, age };
     },
   },
-  {
-    heading: i18n.sexLabel,
```
paulfalgout, line 32: I think the i18n file needs clean up

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,136 @@
...
-  },
-  onRender() {
-    _.each(patientSet, this.showItem, this);
-
-    this.showPatientFields();
-  },
-  showItem({ heading, data = _.constant(this.model.attributes), template }) {
-    this.showView('heading', heading);
-    this.showView('item', renderTemplate(template, data(this.model)));
+  divider: {
```
paulfalgout, line 28: So because there's nothing in the template this _could_ be 
```
divider: '<div class="patient-sidebar__divider"></div>',
```

src/js/views/patients/patient/sidebar/sidebar_views.js
paulfalgout: Yeah awkward...
I think moving the construction to its own function might help
```js
getWidget() {
  const widget = this.getOption('widget');
  const model = this.getOption('patient');
  
  if (_.isFunction(widget)) return new widget({ model });

  return _.extend({ model }, widget);
},
onRender() {
  this.showChildView('content', this.getWidget());
},
```
Also note the extend is reversed so the patient isn't put on the original widget option.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,136 @@
...
+
+    region.show(_.extend(this.getOption('widget'), {
+      model: this.getOption('patient'),
+    }));
   },
 });
 
-const EngagementStatusView = View.extend({
-  className: 'patient-sidebar__engagement-status',
+const SidebarView = CollectionView.extend({
```
paulfalgout, line 108: I think adding
```
viewFilter(model) {
  return model.get('widget_type');
},
```
will ensure only widgets that exist get displayed.  Otherwise you'll get an extra 4px margin _I think_ from the  empty `.patient-sidebar__item`

src/js/views/patients/patient/sidebar/patient-sidebar.scss
```diff
@@ -13,19 +13,18 @@
...
   }
 }
 
+.patient-sidebar__divider {
+  border-top: 1px solid $black-90;
+  content: '';
+  display: block;
+  height: 0;
+}
+
```
paulfalgout, line 27: we also should remove below the `:empty` class.  It isn't going to work with the new widget.  we'll need a new solution with the new widget things.

src/js/entities-service/entities/patients.js
```diff
@@ -13,6 +13,7 @@ const _Model = BaseModel.extend({
     return Radio.request('entities', 'groups:collection', this.get('_groups'));
   },
   getFields() {
+    /* istanbul ignore next: remove ignore when patient fields are supported by widgets in patient/sidebar/sidebar_views.js */
```
paulfalgout: you know what.. let's just remove this whole function.  It's super easy to add, but it's also easy to miss removing the ignore.

src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -11,32 +12,36 @@ export default App.extend({
...
   },
   beforeStart({ patient }) {
     return Radio.request('entities', 'fetch:patient:engagementStatus', patient.id);
   },
-  onStart() {
-    this.showEngagementStatus();
-  },
   onFail() {
-    this.showEngagementStatus();
+    this.patient.set('engagement', {
```
paulfalgout: I'm not sure this is necessary?  It looks to me like engagement is always at least a string, so the not available check in the view could just be checking to see if status is falsy?  If we do need to set a value on the patient for checking elsewhere it should probably be `_statusNotAvailable` or something so that it's a private variable.  But if we can avoid changing the model here it's probably better.

paulfalgout (changes requested):

src/js/apps/patients/patient/sidebar/sidebar_app.js
LisaManresa: At the moment, `engagement` doesn't exist on the patient until the engagement status request succeeds. When the request fails, without this line, it never gets set, so we're unable to differentiate between a loading state and a not available state in the view. And since we're now handling the preload logic in the view rather than in the app, listening for model change events seems to be the cleanest method of handling the preload logic.

I like the `_statusNotAvailable` approach. Are you suggesting that be applied directly to the patient, or nested under `engagement`

paulfalgout: ah so maybe just `this.patient.trigger('status:notavailable')` would be better... then the view can just set a flag on itself.  The patient model will live on in memory so it's awkward to store that state there when it may just be a result of a fluke error.

src/js/views/patients/patient/sidebar/patient-sidebar.scss
LisaManresa: Hmm, yeah. I think the way this is currently architected, that logic would have to live in each view. That's unfortunate.

src/js/views/patients/patient/sidebar/sidebar_views.js
LisaManresa: I think because, even though it doesn't need it, it falls into the `_.extend({ model }, widget)` case of `getWidget` we need to keep the `template` key for this.

paulfalgout: ah true.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,141 @@
...
+          duration: 400,
+          direction: 'alternate',
+        });
+      }
+    },
+    onClick() {
+      Radio.request('sidebar', 'start', 'engagement', { patient: this.model });
+    },
+    onStatusNotAvailable(notAvailable) {
+      this.notAvailable = notAvailable;
```
paulfalgout: can we change this to `this.isNotAvailable`

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,141 @@
...
+          opacity: 0.5,
+          loop: true,
+          easing: 'easeInOutSine',
+          duration: 400,
+          direction: 'alternate',
+        });
+      }
+    },
+    onClick() {
+      Radio.request('sidebar', 'start', 'engagement', { patient: this.model });
```
paulfalgout, line 78: should this `if (!engagement || isNotAvailable) return;`?  Seems so?

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,141 @@
...
+    },
+  },
+  engagement: View.extend({
+    modelEvents: {
+      'change:engagement': 'render',
+      'status:notAvailable': 'onStatusNotAvailable',
+    },
+    getTemplate() {
+      if (!this.model.get('engagement') && !this.notAvailable) {
+        return hbs`<span class="js-loading">{{ @intl.patients.patient.sidebar.sidebarViews.engagementStatusPreloadView.loading }}</span>`;
```
paulfalgout: I think these i18n keys need some updating too

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,141 @@
...
+        groups: _.map(this.model.getGroups().models, 'attributes'),
+      };
+    },
+  },
+  engagement: View.extend({
+    modelEvents: {
+      'change:engagement': 'render',
+      'status:notAvailable': 'onStatusNotAvailable',
+    },
+    getTemplate() {
```
paulfalgout, line 50: I think this could be
```js
if (this.notAvailable) return hbs...
if (!this.model.get('engagment')) return hbs...
return hbs...
```
Also won't need templateContext that way

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -1,142 +1,141 @@
...
+  template: hbs`{{#if display_name}}<div class="patient-sidebar__heading">{{ display_name }}</div>{{/if}}<div class="patient-sidebar__item" data-content-region></div>`,
+  regions: {
+    content: '[data-content-region]',
   },
   onRender() {
-    _.each(patientSet, this.showItem, this);
+    const widget = this.getWidget();
 
-    this.showPatientFields();
+    this.getRegion('content').show(widget);
```
paulfalgout: why this over `this.showChildView('content', widget);`?

paulfalgout (changes requested):

src/js/views/patients/patient/sidebar/sidebar_views.js
LisaManresa: Ooh yes. Thanks for catching that.

PR #342, opened 2020-08-10
Patient sidebar options widget

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23090]

### DIFF
--- a/src/js/apps/patients/patient/sidebar/sidebar_app.js
+++ b/src/js/apps/patients/patient/sidebar/sidebar_app.js
@@ -35,6 +35,60 @@ export default App.extend({
       { id: '6', widget_type: 'groups', display_name: 'Groups' },
     ]);
 
+    /* istanbul ignore if */
+    if (_DEVELOP_) {
+      const display_options = {
+        '1': 'Test Field',
+        'foo': 'Foo',
+        'bar': 'Bar is this one',
+      };
+
+      collection.add([
+        {
+          id: '7',
+          widget_type: 'divider',
+        },
+        {
+          id: '8',
+          widget_type: 'optionsWidget',
+          display_name: 'Populated Option Widget',
+          widget_config: {
+            field_name: 'test-field',
+            display_options,
+          },
+        },
+        {
+          id: '9',
+          widget_type: 'optionsWidget',
+          display_name: 'Empty Option Widget',
+          widget_config: {
+            field_name: 'empty-field',
+            display_options,
+          },
+        },
+        {
+          id: '10',
+          widget_type: 'optionsWidget',
+          display_name: 'Nested Option Widget',
+          widget_config: {
+            field_name: 'nested-field',
+            key: 'foo',
+            display_options,
+          },
+        },
+        {
+          id: '11',
+          widget_type: 'optionsWidget',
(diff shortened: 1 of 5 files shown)


src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -35,6 +35,40 @@ export default App.extend({
...
+          id: '7',
+          widget_type: 'divider',
+        },
+        {
+          id: '8',
+          widget_type: 'optionsWidget',
+          display_name: 'Populated Option Widget',
+          widget_config: {
+            field_name: 'test-field',
+            options,
```
paulfalgout: we should rename this..  I know it's working but it's a little strange with `getOption()` on the view.. because `getOption` is doing `this.options['options'] || this['options'];`

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -101,7 +113,7 @@ const widgetView = View.extend({
 
     if (_.isFunction(widget)) return new widget({ model });
 
-    return _.extend({ model }, widget);
+    return _.extend({ model }, this.model.get('widget_config'), widget);
```
paulfalgout, line 127: we should handle this for when a widget is a view as well.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -82,6 +82,18 @@ const sidebarWidgets = {
       this.render();
     },
   }),
+  optionsWidget: {
+    template: hbs`{{ displayValue }}{{#unless displayValue}}–{{/unless}}`,
```
paulfalgout: I think maybe we should add a class that does the `:empty` thing that we can add here as a `className`. That way in the future if we have a widget with multiple values that could be empty the class could be applied within the template.

paulfalgout (changes requested):
So looking into all of the form patient fields I think we should add support for nested data...  so what do you think about adding an optional `key` to the `widget_config` where it could be `foo.bar` and would handle `bar` if the patient field's value was `{ foo: { bar: 'baz' }}`.   Should be able to use `_.propertyOf(value)(key.split('.'))`

src/js/apps/patients/patient/sidebar/sidebar_app.js
LisaManresa: `display_options`? Kind of sounds like styles, but could work. `value_options`?

src/js/views/patients/patient/sidebar/patient-sidebar.scss
```diff
@@ -35,8 +35,15 @@
   }
 }
 
-.patient-sidebar__item + .patient-sidebar__item {
-  margin-top: 4px;
+.patient-sidebar__item {
```
paulfalgout, line 38: I don't think we should necessarily nest these
`.is-empty` is fine to nest since it is a state modifier, but I don't think  `.patient-sidebar__item + .patient-sidebar__item` should be.

paulfalgout (changes requested):
just one nit-pick

PR #344, opened 2020-08-13
Display 403 failures on action/flow delete

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23105]

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -255,6 +255,7 @@ export default App.extend({
       },
       'delete'() {
         const itemCount = this.selected.length;
+
         this.selected.destroy().then(() => {
           this.showDeleteSuccess(itemCount);
           app.stop();
--- a/test/integration/patients/sidebar/flow-sidebar.js
+++ b/test/integration/patients/sidebar/flow-sidebar.js
@@ -361,6 +361,40 @@ context('flow sidebar', function() {
 
     cy
       .route({
+        status: 403,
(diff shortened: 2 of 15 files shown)


LisaManresa:
This is pending what we decide to do with worklist bulk delete. But we can get 👀  on everything else

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -255,11 +255,18 @@ export default App.extend({
       },
       'delete'() {
         const itemCount = this.selected.length;
-        this.selected.destroy().then(() => {
-          this.showDeleteSuccess(itemCount);
-          app.stop();
-          this.getState().clearSelected();
-        });
+
+        this.selected.destroy({ wait: true })
```
paulfalgout: I think we should revert this, so it's clear _something_ needs to be done.. I don't think this'd what it'd be if we do this.

test/integration/patients/sidebar/flow-sidebar.js
```diff
@@ -373,8 +407,6 @@ context('flow sidebar', function() {
       .find('.js-submit')
       .click();
 
-    cy
```
paulfalgout, line 376: should this get removed?

paulfalgout (changes requested):

test/integration/patients/sidebar/flow-sidebar.js
LisaManresa: No it should not.

paulfalgout (approved):

PR #346, opened 2020-08-17
Add alert queue

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch23136]

### DIFF
--- a/src/js/services/alert.js
+++ b/src/js/services/alert.js
@@ -2,7 +2,7 @@ import _ from 'underscore';
 
 import App from 'js/base/app';
 
-import { AlertView } from 'js/views/globals/alert-box/alert-box_views';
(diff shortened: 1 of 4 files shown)


cypress.json
```diff
@@ -25,5 +25,6 @@
   "projectId": "ep9zr6",
   "viewportWidth": 1280,
   "viewportHeight": 768,
-  "chromeWebSecurity": false
+  "chromeWebSecurity": false,
+  "nodeVersion": "system"
```
LisaManresa: Is this needed for the test changes in the PR. Just wondering if it's a better fit for the upcoming cypress upgrade PR.

LisaManresa (approved):

cypress.json
paulfalgout: For whatever reason I cannot run coverage reports without it.  I suppose we could pull it out, but it's a crippling issue.

PR #347, opened 2020-08-17
Dynamically display Role and Owner in “Owned By” and “Shared By Role" lists

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23118]

Per Mark, this PR:
- Changes "For Role" in app Nav to "Shared by Role"
- Dynamically updates the list titles in the "Owned By" and "Shared by Role" lists with the owner name and role name respectively

### DIFF
--- a/src/js/views/patients/worklist/role_component.js
+++ b/src/js/views/patients/worklist/role_component.js
@@ -1,0 +1,49 @@
+import Radio from 'backbone.radio';
+import hbs from 'handlebars-inline-precompile';
+
+import 'sass/modules/buttons.scss';
+
+import intl from 'js/i18n';
+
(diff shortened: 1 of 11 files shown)


src/js/views/patients/worklist/role_component.js
```diff
@@ -0,0 +1,49 @@
+import Radio from 'backbone.radio';
```
LisaManresa, line 1: This could potentially live in the filter_views file, but I thought for the same of readability, having its own file might make sense.

It's also very similar to the role component in the admin app. If we end up with a third instance of it somewhere it might make sense to create a shared version.

paulfalgout (approved):

PR #349, opened 2020-08-19
Fix parsing on 403 errors for action/flow delete

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23151]

### DIFF
--- a/src/js/apps/admin/sidebar/action-sidebar_app.js
+++ b/src/js/apps/admin/sidebar/action-sidebar_app.js
@@ -41,7 +41,7 @@ export default App.extend({
         this.stop();
       })
       .fail(({ responseJSON }) => {
-        Radio.request('alert', 'show:error', responseJSON.message);
+        Radio.request('alert', 'show:apiError', responseJSON);
       });
   },
   onStop() {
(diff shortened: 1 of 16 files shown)


src/js/apps/admin/sidebar/action-sidebar_app.js
```diff
@@ -41,7 +41,12 @@ export default App.extend({
         this.stop();
       })
       .fail(({ responseJSON }) => {
-        Radio.request('alert', 'show:error', responseJSON.message);
+        const error = responseJSON.errors[0];
```
paulfalgout: Had an idea.  why don't we make a `Radio.request('alert', 'show:apiError', responseJSON)`. so that these are all simpler.  Then in that case I _do_ think we should loop over the errors array.  it wasn't necessary here, but it makes sense for other error types.

paulfalgout (changes requested):

paulfalgout (changes requested):
one more quick fix for this release.  the role filter added to the role worklist needs a down arrow

paulfalgout (approved):

PR #357, opened 2020-09-01
Add test helpers for moment

paulfalgout:
Simplifies date testing by how we use it.  Centralizes “now” to a single location.  Helps enforce utc when necessary.

@LisaManresa too abstract?  I think these would work for 99% of our needs in integration tests and fixtures.  Plus it reduces a lot of `.format()` type things.  If you're :shipit: want to implement some or all of it?

### DIFF
--- a/test/integration/patients/patient/data-and-events.js
+++ b/test/integration/patients/patient/data-and-events.js
@@ -1,6 +1,7 @@
 import _ from 'underscore';
-import moment from 'moment';
+
 import formatDate from 'helpers/format-date';
+import { testTs, testTsSubtract } from 'helpers/test-moment';
 
 context('patient data and events page', function() {
   specify('action, flow and events list', function() {
@@ -33,7 +34,7 @@ context('patient data and events page', function() {
             duration: 0,
             due_date: null,
             due_time: null,
-            updated_at: moment.utc().format(),
+            updated_at: testTs(),
           },
           relationships: {
             patient: { data: { id: '11111' } },
@@ -50,11 +51,11 @@ context('patient data and events page', function() {
 
         fx.data[2].attributes.name = 'Third In List';
         fx.data[2].relationships.state = { data: { id: '55555' } };
-        fx.data[2].attributes.updated_at = moment.utc().subtract(2, 'days').format();
+        fx.data[2].attributes.updated_at = testTsSubtract(2);
 
         fx.data[1].attributes.name = 'Not In List';
         fx.data[1].relationships.state = { data: { id: '33333' } };
-        fx.data[1].attributes.updated_at = moment.utc().subtract(6, 'days').format();
+        fx.data[1].attributes.updated_at = testTsSubtract(6);
 
         return fx;
       }, '1')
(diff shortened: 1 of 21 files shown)


LisaManresa (approved):
I love this. I think you're right that it will cover nearly all of our date testing scenarios. I'm fine with needing to roll custom testing for a few instances here and there.

luketlancaster (approved):
yeah makes sense to me. Also seems way less finicky than doing what's (now) in the helpers all over the place

test/integration/patients/patient/data-and-events.js
```diff
@@ -132,7 +133,7 @@ context('patient data and events page', function() {
       .should('contain', 'Third In List')
       .next()
       .should('contain', 'Check-in completed')
-      .and('contain', formatDate(moment().subtract(4, 'days'), 'TIME_OR_DAY'))
+      .and('contain', formatDate(testDateSubtract(4), 'TIME_OR_DAY'))
```
paulfalgout: so if the date is a timestamp I think this should also be `testDateSubtract(4)`

test/integration/patients/patient/sidebar.js
```diff
@@ -68,7 +70,7 @@ context('patient sidebar', function() {
       .as('patientSidebar')
       .should('contain', 'First Last')
       .should('contain', formatDate(dob, 'LONG'))
-      .should('contain', `Age ${ moment().diff(dob, 'years') }`);
+      .should('contain', `Age ${ moment(testDateSubtract()).diff(dob, 'years') }`);
```
paulfalgout: I think this'd just be `testDate()`

paulfalgout: It's interesting that this works though...  _crazy_ thought.. particularly since we you just did all of this..  _but_ we could support all the same functionality with `testDate()` for now `testDate(1)` for add one day and `testDate(-1)` or subtract one day..  Do you think it's better that we have more functions in the case, or is it faster if we have less api to manage?  Generally I'd say the more explicit API is better, but there's a case that test helpers should be easy to use and malleable. 
But for instance our current API could be written:
```js
function testDate() {
  return testDateAdd();
}

function testDateAdd(time, key = 'days') {
  return moment(date).add(time, key).format('YYYY-MM-DD');
}

function testDateSubtract(time, key = 'days') {
  return testDateAdd(-time, key);
}
```

test/integration/patients/sidebar/flow-sidebar.js
```diff
@@ -41,7 +40,7 @@ context('flow sidebar', function() {
           attributes: {
             first_name: 'First',
             last_name: 'Last',
-            birth_date: moment().subtract(10, 'years').format('YYYY-MM-DD'),
+            birth_date: testTsSubtract(10, 'years'),
```
paulfalgout: This should be date

test/integration/patients/worklist/bulk-edit.js
```diff
@@ -153,6 +154,145 @@ context('Worklist bulk editing', function() {
       .should('contain', 'Tip: To assign a clinician, filter the worklist to a specific group.');
   });
 
+  specify('displaying common groups - actions', function() {
```
paulfalgout, line 157: :+1:

paulfalgout (commented):
@LisaManresa I definitely think it's less important, but we should probably use these helpers in the fixture config as well.

test/integration/patients/patient/sidebar.js
LisaManresa: Hmm. Normally I think I'd go for consolidating the API like that. But I think in this case, being explicit makes the tests more readable and I'd prefer that.

paulfalgout: good enough for me

PR #359, opened 2020-09-15
Add current user to top of owner droplist

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23313]

### DIFF
--- a/src/js/views/patients/shared/components/owner_component.js
+++ b/src/js/views/patients/shared/components/owner_component.js
@@ -1,3 +1,4 @@
+import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 import hbs from 'handlebars-inline-precompile';
 
@@ -72,6 +73,12 @@ export default Droplist.extend({
         collection: getGroupClinicians(group),
         headingText: group.get('name'),
       };
+    });
+
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+
+    this.lists.unshift({
+      collection: new Backbone.Collection([currentUser]),
     });
 
     this.lists.push({
--- a/src/js/views/patients/worklist/filters_views.js
+++ b/src/js/views/patients/worklist/filters_views.js
@@ -1,3 +1,6 @@
+
+import Backbone from 'backbone';
+import Radio from 'backbone.radio';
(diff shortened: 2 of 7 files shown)


src/js/views/patients/shared/components/owner_component.js
```diff
@@ -71,9 +72,17 @@ export default Droplist.extend({
       return {
         collection: getGroupClinicians(group),
         headingText: group.get('name'),
+        className: 'picklist__group owner-component__group',
```
paulfalgout: I'm open to adding the styles to `picklist_group`.   We essentially defaulted to this for every group'd droplist on legacy, and mark likes consistency, so I would assume we'll always to it here too.. we can adjust later if there's some special case.

paulfalgout (changes requested):

src/js/views/patients/shared/components/owner_component.js
```diff
@@ -71,9 +72,17 @@ export default Droplist.extend({
       return {
         collection: getGroupClinicians(group),
         headingText: group.get('name'),
+        className: 'picklist__group',
```
paulfalgout: I'm pretty sure we don't need to add `picklist__group` anywhere now as it's the default?

src/js/views/patients/worklist/filters_views.js
```diff
@@ -52,13 +53,14 @@ const ClinicianDropList = Droplist.extend({
...
-const ClinicianClearButton = View.extend({
-  template: hbs`<button class="button-secondary worklist-list__clear-filter">{{ @intl.patients.worklist.filtersViews.clearClinicianFilter }}</button>`,
-  triggers: {
-    'click': 'click',
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+
+    this.lists.unshift({
+      className: 'picklist__group',
+      collection: new Backbone.Collection([currentUser]),
+      attr: 'name',
```
paulfalgout: Just noticing.. I _think_ `attr` can move up into `picklistOptions` and removed here and in the `groups.map`

paulfalgout (changes requested):

paulfalgout (approved):

PR #361, opened 2020-09-17
Retain intent on add workflow picklist items when they have an icon

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23353]

This also changes the picklist heading, scss file name and class names to account for adding both actions and flows

### DIFF
--- a/src/js/entities-service/entities/clinicians.js
+++ b/src/js/entities-service/entities/clinicians.js
@@ -81,6 +81,7 @@ const Model = Store(_Model, TYPE);
 const Collection = BaseCollection.extend({
   url: '/api/clinicians',
   model: Model,
+  comparator: 'name',
 });
 
 export {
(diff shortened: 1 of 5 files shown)


src/js/views/patients/patient/dashboard/add-workflow.scss
```diff
@@ -0,0 +1,39 @@
...
+.add-workflow__button {
+  padding-right: 36px;
+
+  .add-workflow__arrow {
+    left: auto;
+    right: 16px;
+  }
+}
+
+.add-workflow__picklist-item {
```
paulfalgout: I think we'll want to solve this problem generically for picklists

I think we can do this by adding a `picklist__item--icon` class and rather than wrapping the template in a span just set `itemClassName`

This will also require making sure picklists that current contain `icon`s don't break.

paulfalgout (changes requested):

PR #368, opened 2020-10-01
Moment migration

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch23418]

I set this to a WIP mostly because I didn't want to do a search/replace of the `moment` term to ease reviewing the impact of these changes, but if we go with this, I do think we should update all instances of the "Moment" key.

Up for alternatives, but I looked at luxon, date-fns, and dayjs.. The first two do seem moderately more forward thinking, but only just-so, and dayjs is a much easier transition and is still remarkably small.   It also has a larger active community than the first two, but that may change as people migrate over time from moment.

Thoughts?


LisaManresa:
Wow the size differences between luxon/dayjs and moment are way more substantial than I expected. Dayjs seems like it gives us the best of both worlds in terms of size, performance and ease of transition from moment,

LisaManresa (approved):

PR #370, opened 2020-10-07
Feature/changing render mode on response

wweaver4th:
- [ ] Add QA tasks

### DIFF
--- a/src/sass/formapp.scss
+++ b/src/sass/formapp.scss
@@ -1,11 +1,8 @@
 @import url('https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css');
-@import url('https://bootswatch.com/4/cosmo/bootstrap.min.css');
+
 body {
   font-family: 'ProximaSoft', 'Droid Sans', 'helvetica neue', 'arial', sans-serif !important;
   padding: 24px;
-}
-.btn.disabled, .btn:disabled {
-  display:none;
 }
 .choices.is-disabled .choices__inner, .choices.is-disabled .choices__input {
   background-color: #fff;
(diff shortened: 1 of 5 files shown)


src/sass/formapp.scss
```diff
@@ -1,8 +1,15 @@
 @import url('https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css');
+@import url('https://bootswatch.com/4/cosmo/bootstrap.min.css');
```
paulfalgout: why are we re-importing bootstrap?

src/sass/formapp.scss
```diff
@@ -1,8 +1,15 @@
...
 body {
   font-family: 'ProximaSoft', 'Droid Sans', 'helvetica neue', 'arial', sans-serif !important;
   padding: 24px;
 }
+.btn.disabled, .btn:disabled {
```
paulfalgout: how about `.formio-read-only .btn.disabled, .formio-read-only .btn:disabled`. hiding all disabled buttons everywhere seems like it's going to potentially break all sorts of things.

paulfalgout (changes requested):
Broken test needs to be updated from `not.exist` to `not.be.visible` on line 185 of `test/integration/forms/formapp.js`

src/sass/formapp.scss
wweaver4th: .formio-read-only is not in there.

src/sass/formapp.scss
wweaver4th: Will remove

src/sass/formapp.scss
paulfalgout: well we need to find or do something.  we can't just hide disabled buttons

paulfalgout: ok so that class exists only in form.io 4.10+ and we're on 4.9.  They're up to 4.12.. probably worth updating.

wweaver4th: yep

PR #373, opened 2020-10-08
Switch to coveralls already

paulfalgout:
Depends on #372 
[ch23506]


paulfalgout:
Coveralls is pretty sweet.  https://coveralls.io/builds/34033236. Not though that we get branch misses for every `if (_DEVELOP_)` though :-(. I guess that'll just show us where our flags are.

LisaManresa (approved):

PR #377, opened 2020-10-09
Patient sidebar widget endpoint

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23446]

This change will require that the widget sidebar settings be defined for an organization (see: https://github.com/RoundingWell/care-ops-backend/pull/1000#issuecomment-703777150). Unless we want to define a default widget setting on the FE.

### DIFF
--- a/src/js/apps/patients/patient/sidebar/sidebar_app.js
+++ b/src/js/apps/patients/patient/sidebar/sidebar_app.js
@@ -1,5 +1,5 @@
-import Backbone from 'backbone';
 import Radio from 'backbone.radio';
+import _ from 'underscore';
 
 import App from 'js/base/app';
 
@@ -8,110 +8,12 @@ import { SidebarView } from 'js/views/patients/patient/sidebar/sidebar_views';
 export default App.extend({
   onBeforeStart({ patient }) {
     const currentOrg = Radio.request('bootstrap', 'currentOrg');
-
-    this.orgEngagementEnabled = currentOrg.getSetting('engagement');
+    const widgets = Radio.request('entities', 'widgets:collection', _.collectionOf(currentOrg.getSetting('widget_sidebar'), 'id'));
     this.patient = patient;
 
-    const widgets = this.getSidebarWidgets();
-
     this.showView(new SidebarView({
-      model: this.patient,
+      model: patient,
       collection: widgets,
     }));
-  },
-  beforeStart({ patient }) {
-    return Radio.request('entities', 'fetch:patient:engagementStatus', patient.id);
-  },
-  onFail() {
-    this.patient.trigger('status:notAvailable', true);
-  },
-  getSidebarWidgets() {
-    const collection = new Backbone.Collection([
-      { id: '1', widget_type: 'dob', display_name: 'Date of Birth' },
-      { id: '2', widget_type: 'sex', display_name: 'Sex' },
-      { id: '13', widget_type: 'status', display_name: 'Status' },
-      { id: '3', widget_type: 'divider' },
-      { id: '4', widget_type: 'engagement', display_name: 'Engagement Status' },
-      { id: '5', widget_type: 'divider' },
-      { id: '6', widget_type: 'groups', display_name: 'Groups' },
-    ]);
-
-    /* istanbul ignore next */
-    if (_DEVELOP_) {
-      const display_options = {
-        '1': 'Test Field',
-        'foo': 'Foo',
-        'bar': 'Bar is this one',
-      };
-
-      collection.add([
-        {
-          id: '7',
-          widget_type: 'divider',
-        },
-        {
-          id: '8',
-          widget_type: 'optionsWidget',
-          display_name: 'Populated Option Widget',
-          widget_config: {
-            field_name: 'test-field',
-            display_options,
-          },
-        },
-        {
-          id: '9',
-          widget_type: 'optionsWidget',
-          display_name: 'Empty Option Widget',
-          widget_config: {
-            field_name: 'empty-field',
-            display_options,
-          },
-        },
-        {
-          id: '10',
-          widget_type: 'optionsWidget',
-          display_name: 'Nested Option Widget',
-          widget_config: {
-            field_name: 'nested-field',
-            key: 'foo',
-            display_options,
-          },
-        },
-        {
-          id: '11',
-          widget_type: 'optionsWidget',
-          display_name: 'Empty Nested Option Widget',
-          widget_config: {
-            field_name: 'nested-field',
-            key: 'bar',
-            display_options,
-          },
-        },
-        {
-          id: '12',
-          widget_type: 'templateWidget',
-          display_name: 'Template Widget',
-          widget_config: {
-            template: `
-                       <p>
-                         Test Patient Name: {{ patient.first_name }}
-                       </p>
-                       <p>
-                         Test Field: <span class="widgets-value">{{ fields.test_field }}</span>
-                       </p>
-                       <p>
-                         Nested Field: <span class="widgets-value">{{ fields.nested_field.foo }}</span>
-                       </p>`,
-          },
-        },
-      ]);
-    }
-
-    /* TODO: clean this up once widget config is coming from the BE */
-    if (!this.orgEngagementEnabled) {
-      collection.remove([{ id: '4' }, { id: '5' }]);
-    }
-
-    return collection;
   },
 });
--- a/src/js/views/patients/patient/sidebar/sidebar_views.js
+++ b/src/js/views/patients/patient/sidebar/sidebar_views.js
@@ -66,6 +66,13 @@ const sidebarWidgets = {
     triggers: {
       'click': 'click',
     },
+    initialize() {
+      const patient = this.model;
+      Radio.request('entities', 'fetch:patient:engagementStatus', patient.id)
+        .fail(() => {
+          patient.trigger('status:notAvailable', true);
+        });
+    },
     onRender() {
       if (!this.model.get('engagement')) {
         anime({
@@ -122,9 +129,9 @@ const sidebarWidgets = {
   }),
 };
 
-const widgetView = View.extend({
+const WidgetView = View.extend({
   className: 'patient-sidebar__section',
-  template: hbs`{{#if display_name}}<div class="patient-sidebar__heading">{{ display_name }}</div>{{/if}}<div class="patient-sidebar__item" data-content-region></div>`,
+  template: hbs`{{#if definition.display_name}}<div class="patient-sidebar__heading">{{ definition.display_name }}</div>{{/if}}<div class="patient-sidebar__item" data-content-region></div>`,
   regions: {
     content: '[data-content-region]',
   },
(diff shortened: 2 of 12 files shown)


test/integration/patients/patient/sidebar.js
```diff
@@ -267,15 +366,25 @@ context('patient sidebar', function() {
       .routePrograms()
       .routeAllProgramActions()
       .routeAllProgramFlows()
+      .route({
+        status: 504,
+        method: 'GET',
+        url: 'api/patients/1/engagement-status',
+        response: {},
+        delay: 500,
```
LisaManresa: Not super ideal, but CI executes this too fast for the engagement status loading state to display.

dhrrgn:
> This change will require that the widget sidebar settings be defined for an organization (see: https://github.com/RoundingWell/care-ops-backend/pull/1000#issuecomment-703777150). Unless we want to define a default widget setting on the FE.

If you let me know the base widget definitions I can add them to the BE fixtures, and update the `widget_sidebar` setting fixture to include them all, so dev/preview will have them. I suspect it is just the ones defined in [sidebarWidgets](https://github.com/RoundingWell/care-ops-frontend/blob/80b1ded93a2ab66ba8ce745cfea6a522d6db83cb/src/js/views/patients/patient/sidebar/sidebar_views.js). 

### Side note

We now have field schemas, which you can fetch via `GET /field-schemas`. A field _MAY_ have a schema defined. These definitions are in the [JSONSchema](https://json-schema.org/) format. Any field with a schema defined is guaranteed to have a value that satisfies that schema. 

I mention this because FE could use the `format` annotations in the schema to determine the output formatting ([see here for the spec-defined formats](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3)). The `formatMessage` handlebars helper could be updated (or a new helper `formatField` added) to get the format from the schema and output the field value accordingly. For example, this would eliminate the need to use `formatDateTime` in the `dob` template.

test/integration/patients/patient/sidebar.js
paulfalgout: might be a way to do this with `cy.clock`. Is 500 really needed too? Seems high

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -165,6 +160,32 @@ const SidebarView = CollectionView.extend({
...
+  `,
+  regions: {
+    widgets: {
+      el: '[data-widgets-region]',
+      replaceElement: true,
+    },
+  },
+});
+
+const SidebarWidgetsView = View.extend({
```
paulfalgout: I don't think this view is necessary and this could be done in the SidebarView

test/integration/patients/patient/sidebar.js
```diff
@@ -12,7 +12,111 @@ context('patient sidebar', function() {
...
+            'optionsWidget4',
+            'optionsWidget5',
+            'templateWidget',
+          ],
+        };
+
+        return fx;
+      })
+      .routeWidgets(fx => {
+        const addWidget = _.partial(getResource, _, 'widgets');
```
paulfalgout, line 42: :+1: cool

src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -1,117 +1,30 @@
...
     }));
+
+    this.startChildApp('widgets', {
+      region: this.getRegion('widgets'),
+      patient: this.patient,
+    });
   },
   beforeStart({ patient }) {
-    return Radio.request('entities', 'fetch:patient:engagementStatus', patient.id);
+    return [Radio.request('entities', 'fetch:patient:engagementStatus', patient.id)];
```
paulfalgout: I think we should move this request and the `trigger('status...` into the widget `initialize` for now.  For now this widget is always the gate keeper to this info.  I don't typically like data requests in views, but these widgets are a grey area.  So it'd be `Radio.request(...).fail(() => { trigger`. 

And this app's `beforeStart` could be used for the widget request.. so we don't really need the childapp.  However..  I'm also thinking that until we have a better caching strategy for these long lasting configs like widgets.. I think we should probably just load every widget (no filter on the request) at the bootstrap vs introducing loaders everywhere.  If we ever get out of hand with the number of widgets we can readdress.

paulfalgout (changes requested):

paulfalgout:
@dhrrgn with the exception of the `engagement` widget which should only be added if engagement is on This fixture is probably a good default https://github.com/RoundingWell/care-ops-frontend/pull/377/files#diff-97cf8e606560e4a73d41866963fc19e1R1

I do want to add the JSONSchema stuff to the template widget handling.  It's not a handlebars template though so there's no formatMessage right now.  I haven't quite figured out the best way to approach this with arrays and nested data though.  But using the `format` has been the eventual plan.  Not for dob though because it's not a patient field and also contains the age calculation.

test/fixtures/test/widgets.json
```diff
@@ -0,0 +1,41 @@
...
+    "widget_type": "status",
+    "name": "status",
+    "definition": {
+      "display_name": "Status"
+    }
+  },
+  {
+    "widget_type": "divider",
+    "name": "divider"
+  },
```
dhrrgn, line 27: Not a _huge_ deal since this is a test fixture, but `definition` is defined as NOT NULL. In this case it would be `"definition": {}`

I also specify it is an object in our OpenAPI spec, but this is actually incorrect, it should allow any non-null value. I will fix this

paulfalgout: :+1: should be ok.  Divider won't attempt to use that attribute

src/js/apps/patients/patient/sidebar/sidebar_app.js
LisaManresa: This makes sense.  We'll probably have a few norm-breaking instances like this in these widgets. Added bonus: moving the request to the widget allowed me to remove the delay on the request stub in the test 🎉

src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -1,117 +1,30 @@
...
-    const widgets = this.getSidebarWidgets();
-
     this.showView(new SidebarView({
-      model: this.patient,
-      collection: widgets,
+      model: patient,
     }));
+    this.getRegion('widgets').startPreloader();
   },
   beforeStart({ patient }) {
```
paulfalgout: I don't we should preload or fetch here at all.  We should add the widget fetch to the bootstrap and he we'd just use  `Radio.request('entities', 'widget:collection', _.collectionOf(currentOrg.getSetting('widget_sidebar')));` and then `WidgetCollectionView` could go back to having a template

paulfalgout (changes requested):

src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -8,110 +8,12 @@ import { SidebarView } from 'js/views/patients/patient/sidebar/sidebar_views';
 export default App.extend({
   onBeforeStart({ patient }) {
     const currentOrg = Radio.request('bootstrap', 'currentOrg');
-
-    this.orgEngagementEnabled = currentOrg.getSetting('engagement');
```
dhrrgn, line 12: I may be missing somewhere, but it looks like this setting is no longer checked anywhere? I know you handle errors from `GET /patients/{patientId}/engagement-status` gracefully in the widget, but we should still check if the setting is enabled before even trying to hit that endpoint. 

The reason being is that there is an edge case where we accidentally add the engagement widget to the sidebar, with it disabled. This could, very briefly, show the engagement widget then disappear. This would not only look odd, but Will has said indicated in the past that we don't want customers who don't have it to even know it's a thing unless we tell them about it. 

If I'm missing it somewhere, sorry for the long comment and carry on.

dhrrgn (dismissed):

test/integration/patients/patient/sidebar.js
```diff
@@ -287,16 +404,23 @@ context('patient sidebar', function() {
   specify('organization engagement disabled', function() {
```
paulfalgout: this test can be removed entirely.

test/integration/patients/patient/sidebar.js
```diff
@@ -267,15 +375,24 @@ context('patient sidebar', function() {
...
+      .wait('@routePatient')
+      .wait('@routeWidgets');
 
     cy
       .get('.patient__sidebar')
       .find('.patient-sidebar__heading')
       .contains('Engagement Status')
       .next()
-      .should('contain', 'Loading...');
+      .should('contain', 'Loading...')
```
paulfalgout, line 394: Oddly, relooking at this, this test doesn't seem correct to me.. shouldn't it say "Not Available" or something like that?

src/js/apps/patients/patient/sidebar/sidebar_app.js
paulfalgout: We used that flag to filter out the engagement widget from the static set, not to prevent rendering an existing widget.  That'd be a bit unnecessarily complicated for an edge case I think.  If we're worried about the edge case of adding it accidentally, maybe the endpoint , as the source of truth, should filter it out before sending it for display.

src/js/views/patients/patient/sidebar/patient-sidebar.scss
```diff
@@ -1,5 +1,10 @@
 .patient-sidebar {
   padding: 40px 24px 0;
+  width: 100%;
+}
+
+.patient-sidebar__sections .spinner {
```
paulfalgout: this isn't being used.

src/js/views/patients/patient/sidebar/patient-sidebar.scss
```diff
@@ -1,5 +1,10 @@
 .patient-sidebar {
   padding: 40px 24px 0;
+  width: 100%;
```
paulfalgout: is this needed? isn't `.patient-sidebar` a `div`?

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -137,20 +144,20 @@ const widgetView = View.extend({
...
+    if (_.isFunction(widget)) return new widget(_.extend({ model }, this.model.get('definition')));
 
-    return _.extend({ model }, this.model.get('widget_config'), widget);
+    return _.extend({ model }, this.model.get('definition'), widget);
   },
 });
 
 const SidebarView = CollectionView.extend({
-  className: 'patient-sidebar flex-region',
+  className: 'patient-sidebar',
```
paulfalgout: Not sure why `flex-region` was on here..  racking my brain for why it was.. but I guess it's ok to remove it.

paulfalgout (changes requested):

test/integration/patients/patient/sidebar.js
LisaManresa: It shows "Loading..." while the request is being made, then displays "Not Available" if it fails. That check is in the block below this on line 400.

paulfalgout: ah gotcha.  yep.. makes sense.

src/js/views/patients/patient/sidebar/patient-sidebar.scss
LisaManresa: It is a div, but it's inside a flex div which caused it to shrink to the width of the text inside it. Visually, it's not noticeable. But once the spinner was added it was off center. I'll remove this now that we don't need the spinner.

src/js/views/patients/patient/sidebar/sidebar_views.js
LisaManresa: Oh... per my comment above, that's why flex-region was on there. So I broke the sidebar display by removing this because I thought it was superfluous, then wonkily fixed it that `width: 100%`. I'll add this back on there to ensure that this element fills the width of the container.

PR #379, opened 2020-10-14
Add nested widgets within templateWidget

paulfalgout:
Use `{{ widget.widgetNameId }}` where `widgetNameId` is the id string of the widget coming from the widgets endpoint.

- [x] Add the Clubhouse Story ID: [ch23563]


LisaManresa (approved):
second_widget

paulfalgout:
image

PR #381, opened 2020-10-15
Account for multiple divider models in the sidebar widget collection

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23586]

### DIFF
--- a/src/js/entities-service/entities/widgets.js
+++ b/src/js/entities-service/entities/widgets.js
@@ -1,3 +1,4 @@
+import _ from 'underscore';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
@@ -12,6 +13,9 @@ const Model = Store(_Model, TYPE);
 const Collection = BaseCollection.extend({
   url: '/api/widgets',
   model: Model,
+  modelId(attr) {
+    return _.uniqueId(`${ attr.id }-`);
+  },
 });
 
 export {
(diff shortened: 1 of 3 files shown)


src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -8,7 +8,14 @@ import { SidebarView } from 'js/views/patients/patient/sidebar/sidebar_views';
 export default App.extend({
   onBeforeStart({ patient }) {
     const currentOrg = Radio.request('bootstrap', 'currentOrg');
-    const widgets = Radio.request('entities', 'widgets:collection', _.collectionOf(currentOrg.getSetting('widget_sidebar'), 'id'));
+    const widgetSettings = _.map(currentOrg.getSetting('widget_sidebar'), setting => {
```
paulfalgout: dan's suggestion is pretty interesting and it seems like it will work in this case:
If we add this to the widget collection in the entities.
```js
modelId(attr) {
    return _.uniqueId(`${ attr.id }-`);
},
```
This will allow us to add the same ids and have identical model copies..  it's the simplest solution for now.. if we ever have to `.get` or manipulate the sidebar live, we'll have to change it, but it seems like an improvement.. thoughts?

paulfalgout (changes requested):

src/js/apps/patients/patient/sidebar/sidebar_app.js
LisaManresa: Yeah just saw that suggestion. Agreed, it does sound much simpler.

PR #382, opened 2020-10-19
Update to webpack v5

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch23555]
For reviewing this one @LisaManresa if you could pull it down and run all of the various npm scripts through to make sure this doesn't break any of your day-to-day.

There's a few small known issues addressed here, but I'm comfortable moving forward.
`webpack-dev-server`'s hot reload is currently failing when the target is set to browserlist which is automatically true for our config.  But we can set it to `web` for local development re: https://github.com/webpack/webpack-dev-server/issues/2758#issuecomment-710086019. I don't _love_ this as it means we're not compiling the exact same settings for dev as we are for prod _however_ they're currently working on a fix and we probably don't really have a situation in which this would break something...

`webpack-dev-server`'s `--open` cli flag is broken, so that configuration needed to move to `webpack.config.js` and called with `webpack serve` https://github.com/webpack/webpack-dev-server/issues/2765#issuecomment-708347211

`html-webpack-plugin` wasn't using the correct hooks for webpack v5 and is currently working on fixing that in their alpha release.  The changes are small and the alpha addresses the issue for our use case, so I'm pretty confident in making that shift as otherwise it seems to work fine.  https://github.com/jantimon/html-webpack-plugin/issues/1527#issuecomment-710825524


BenCoffeed (approved):
🎉

LisaManresa (approved):
I don't see any red flags after running everything locally.

PR #383, opened 2020-10-22
Clean up node_modules sourcemaps warnings

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23612]

This addition gets our dev sourcemaps warning down to one:
```
DevTools failed to load SourceMap: Could not load content for webpack://care-ops-frontend/node_modules/backbone.store/dist/backbone.store.mjs.map: HTTP error: status code 404, net::ERR_UNKNOWN_URL_SCHEME
```

I'm still poking around to see why this sourcemap doesn't get picked up by the filter. It looks like this file never passes through the loader.

### DIFF
--- a/config/webpack.rules.js
+++ b/config/webpack.rules.js
@@ -46,7 +46,7 @@ const postcssLoader = {
         ['autoprefixer', { cascade: false }],
         ['cssnano', { preset: 'default' }],
       ],
-    }
+    },
   },
 };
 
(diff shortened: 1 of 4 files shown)


paulfalgout:
It's because for whatever reason backbone.store uses `.mjs` as it's extension..  some sort of legacy es6 naming that never stuck.  Apparently it works, but it's skipping your filter.  Honestly I think this is good and we should fix that over on https://github.com/RoundingWellOS/backbone.store.

PR #384, opened 2020-10-23
Add help link to nav menu

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23636]


src/js/apps/globals/nav_app.js
```diff
@@ -21,6 +21,15 @@ const appNavMenu = new Backbone.Collection([
     icon: 'window',
     text: i18n.app.patients,
   },
+  {
+    onSelect() {
+      window.open('https://roundingwell.freshdesk.com/');
```
paulfalgout: This should be `help.roundingwell.com`

src/js/apps/globals/nav_app.js
```diff
@@ -132,8 +141,8 @@ export default App.extend({
     const currentUser = Radio.request('bootstrap', 'currentUser');
     if (!currentUser.can('admin')) appNavMenu.remove('AdminApp');
 
-    // If there's only 2 options, only sign out
-    if (appNavMenu.length === 2) appNavMenu.shift();
+    // If there's only 3 options, only sign out
+    if (appNavMenu.length === 3) appNavMenu.shift();
```
paulfalgout: what is this change?

paulfalgout (changes requested):

src/js/apps/globals/nav_app.js
LisaManresa: Yeah, I thought this seemed a little fragile. This is removing the "Workspace" menu item if the user isn't an admin. Although, it could be cleaned up by adding `appNavMenu.remove('PatientsApp')` in the `currentUser.can('admin')` check above.

src/js/apps/globals/nav_app.js
```diff
@@ -130,10 +139,11 @@ export default App.extend({
   },
   onStart() {
     const currentUser = Radio.request('bootstrap', 'currentUser');
-    if (!currentUser.can('admin')) appNavMenu.remove('AdminApp');
 
-    // If there's only 2 options, only sign out
-    if (appNavMenu.length === 2) appNavMenu.shift();
+    if (!currentUser.can('admin')) {
+      appNavMenu.remove('AdminApp');
+      appNavMenu.remove('PatientsApp');
```
paulfalgout, line 145: I still don't understand this change.  Why are we removing the patient section for non-admins?  That doesn't seem correct at all.  It's also the default route, so this only breaks the menu.  

And the `shift` as far as I am aware is there so that if the user can only go to one section the menu isn't the one section they can go to and sign out.. it's only sign out, since the other option wouldn't functionally do anything.

paulfalgout: Really the only thing that needs to occur is the `shift` needs to be removed because that scenario is impossible.  the only section a non-admin can't get to is admin

paulfalgout (changes requested):

src/js/apps/globals/nav_app.js
LisaManresa: I do think retaining this logic makes sense. If AdminApp option is removed and there are no other spaces within the app to navigate to, why show the PatientsApp option in the menu if the user can't act on it?

The only downside is that whenever we add another space, we have to keep this in mind, but that's not something that will happen frequently.

paulfalgout: ah yeah true because the help link doesn't route anywhere.  ok.  works for me.

paulfalgout (approved):

PR #387, opened 2020-10-28
Support for form response history

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23624]

### DIFF
--- a/src/js/apps/forms/form/form-history_app.js
+++ b/src/js/apps/forms/form/form-history_app.js
@@ -1,0 +1,56 @@
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+
+import { LayoutView, HistoryBarView, FormView } from 'js/views/forms/form/form-history_views';
+
+export default App.extend({
+  viewEvents: {
+    'select:response': 'onSelectResponse',
+  },
+  stateEvents: {
+    'change:response': 'showIframeView',
+  },
+  onBeforeStart() {
+    this.showView(new LayoutView());
+    this.getRegion('iframe').startPreloader();
+  },
+  beforeStart({ action }) {
+    return action.fetch();
+  },
+  onStart({ form }, action) {
+    this.form = form;
+    const formResponses = action.getFormResponses();
+
+    const historyBarView = this.showChildView('history', new HistoryBarView({
+      formResponses,
+      state: {
+        selected: formResponses.first(),
+      },
+    }));
+
+    this.listenTo(historyBarView, 'click:current', this.onClickCurrent);
+
+    this.setState('response', formResponses.first());
+
+    const iframeView = this.getChildView('iframe');
+    Radio.reply(`form${ this.form.id }`, 'print:form', () => {
+      iframeView.print();
+    });
+  },
+  onStop() {
+    Radio.stopReplying(`form${ this.form.id }`);
+  },
+  onClickCurrent() {
+    this.stop();
+  },
+  onSelectResponse(response) {
+    this.setState('response', response);
+  },
+  showIframeView() {
+    this.showChildView('iframe', new FormView({
+      form: this.form,
+      model: this.getState('response'),
+    }));
+  },
+});
--- a/src/js/apps/forms/form/form-update_app.js
+++ b/src/js/apps/forms/form/form-update_app.js
@@ -1,0 +1,43 @@
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+
+import { Layout } from 'js/views/forms/form/form-update_views';
+
+export default App.extend({
+  viewEvents: {
+    'click:update': 'onClickUpdate',
+  },
+  stateEvents: {
+    'change:shouldUpdate': 'showLayout',
+  },
+  onStart({ form, response, action }) {
+    this.form = form;
+    this.response = response;
+    this.action = action;
+
+    this.setState('shouldUpdate', false);
+    this.showLayout();
+
+    const layoutView = this.getView();
+    Radio.reply(`form${ this.form.id }`, 'print:form', () => {
+      layoutView.print();
+    });
+  },
+  onStop() {
+    Radio.stopReplying(`form${ this.form.id }`);
+  },
+  onClickUpdate() {
+    this.setState({
+      shouldUpdate: true,
+    });
+  },
+  showLayout() {
+    this.showView(new Layout({
+      model: this.form,
+      response: this.response,
+      state: this.getState(),
+      action: this.action,
+    }));
+  },
+});
--- a/src/js/apps/forms/form/form_app.js
+++ b/src/js/apps/forms/form/form_app.js
@@ -5,12 +5,16 @@ import App from 'js/base/app';
 import intl from 'js/i18n';
 
 import PatientSidebarApp from 'js/apps/patients/patient/sidebar/sidebar_app';
+import FormHistoryApp from 'js/apps/forms/form/form-history_app';
+import FormUpdateApp from 'js/apps/forms/form/form-update_app';
 
 import { LayoutView } from 'js/views/forms/form/form_views';
 
 export default App.extend({
   childApps: {
     patient: PatientSidebarApp,
+    history: FormHistoryApp,
+    update: FormUpdateApp,
   },
   onBeforeStart() {
     this.getRegion().startPreloader();
@@ -28,29 +32,40 @@ export default App.extend({
   onStart(options, [action], [patient]) {
     this.patient = patient;
     this.action = action;
+    this.formResponses = action.getFormResponses();
+    this.form = this.action.getForm();
 
-    this.listenTo(action, 'destroy', () => {
-      Radio.request('alert', 'show:success', intl.forms.form.formApp.deleteSuccess);
-      Radio.trigger('event-router', 'default');
+    this.listenTo(action, {
+      'destroy'() {
+        Radio.request('alert', 'show:success', intl.forms.form.formApp.deleteSuccess);
+        Radio.trigger('event-router', 'default');
+      },
+      'change:_form_responses'() {
+        this.formResponses = action.getFormResponses();
+      },
     });
 
-    const form = this.action.getForm();
-
     this.showView(new LayoutView({
-      model: form,
+      model: this.getState(),
       action,
       patient,
-      state: this.getState(),
+      form: this.form,
     }));
 
-    this.setState('sidebar', 'action');
+    this.setState({
+      sidebar: 'action',
+      historyResponseId: null,
+    });
   },
   stateEvents: {
     'change:sidebar': 'onChangeSidebar',
+    'change:historyResponseId': 'onChangeHistoryResponseId',
   },
   viewEvents: {
     'click:sidebarButton': 'onClickSidebarButton',
     'click:expandButton': 'onClickExpandButton',
+    'click:historyButton': 'onClickHistoryButton',
+    'click:printButton': 'onClickPrintButton',
   },
   onClickSidebarButton() {
     const sidebar = this.getState('sidebar');
@@ -72,6 +87,17 @@ export default App.extend({
 
     this.setState('sidebar', this.getState().previous('sidebar'));
   },
+  onClickHistoryButton() {
+    if (this.getChildApp('history').isRunning()) {
+      this.setState('historyResponseId', null);
+      return;
+    }
+
+    this.setState('historyResponseId', this.formResponses.first().id);
+  },
+  onClickPrintButton() {
+    Radio.request(`form${ this.form.id }`, 'print:form');
+  },
   onChangeSidebar(state, sidebar) {
     this.stopChildApp('patient');
     Radio.request('sidebar', 'close');
@@ -90,6 +116,16 @@ export default App.extend({
       this.showActionSidebar();
     }
   },
+  onChangeHistoryResponseId() {
+    const historyResponseId = this.getState('historyResponseId');
+
+    if (!historyResponseId) {
+      this.startUpdateApp();
+      return;
+    }
+
+    this.startHistoryApp();
+  },
   showActionSidebar() {
     const sidebarApp = Radio.request('sidebar', 'start', 'action', { action: this.action, isShowingForm: true });
 
@@ -99,6 +135,32 @@ export default App.extend({
       if (sidebar === 'action') {
         this.setState('sidebar', 'patient');
       }
+    });
+  },
+  startUpdateApp() {
+    this.stopChildApp('history');
+
+    this.startChildApp('update', {
+      region: this.getRegion('form'),
+      form: this.action.getForm(),
+      response: this.formResponses.first(),
+      action: this.action,
+      state: this.getState(),
+    });
+  },
+  startHistoryApp() {
+    this.stopChildApp('update');
+
+    const historyApp = this.startChildApp('history', {
+      region: this.getRegion('form'),
+      action: this.action,
+      form: this.form,
+    });
+
+    this.listenTo(historyApp, {
+      stop() {
+        this.setState('historyResponseId', null);
+      },
     });
   },
 });
--- a/src/js/entities-service/entities/actions.js
+++ b/src/js/entities-service/entities/actions.js
@@ -31,13 +31,12 @@ const _Model = BaseModel.extend({
     if (!formId) return;
     return Radio.request('entities', 'forms:model', formId);
   },
-  getRecentResponse() {
-    const formResponses = Radio.request('entities', 'formResponses:collection', this.get('_form_responses'), {
+  getFormResponses() {
+    return Radio.request('entities', 'formResponses:collection', this.get('_form_responses'), {
       comparator(responseA, responseB) {
         return alphaSort('desc', responseA.get('_created_at'), responseB.get('_created_at'));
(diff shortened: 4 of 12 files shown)


src/js/apps/forms/form/form_app.js
```diff
@@ -36,6 +36,12 @@ export default App.extend({
 
     const form = this.action.getForm();
 
+    this.setState({
+      'response': action.getRecentResponse(),
```
paulfalgout: don't need quotes for `setState` keys

src/js/views/forms/form/form.scss
```diff
@@ -88,25 +88,55 @@
...
+
+    .icon {
+      color: $black-60;
+    }
+  }
+
 }
 
-.form__update-bar {
+.form__update-bar,
```
paulfalgout: maybe both of these classes can be something like `form__action-bar` or something

src/js/views/forms/form/form.scss
```diff
@@ -88,25 +88,55 @@
...
 
-  .icon {
-    margin-right: 8px;
-  }
+.form__update-bar .icon {
+  margin-right: 8px;
 }
 
-.form__update-button {
+.form__update-button,
```
paulfalgout: similarly to above.. maybe `.form__action-button`

src/js/views/forms/form/form.scss
```diff
@@ -88,25 +88,55 @@
...
+.form__current-version-button {
   margin-left: auto;
 }
 
+.form__button-response {
```
paulfalgout: I think we can safely add the default background color to `.button-filter`

The rest of the style updates I'm not so sure about.  It seems like perhaps only the width is needed?  The icon in the marvel is 16px as it is in button-filter.  I don't want to introduce a variant to where to arrow is placed on a filter... if we're changing it, we're changing it everywhere.

src/js/views/forms/form/form.scss
```diff
@@ -88,25 +88,55 @@
...
   font-weight: 600;
   line-height: 1;
   margin-bottom: 8px;
   padding: 8px 16px;
+}
 
-  .icon {
-    margin-right: 8px;
-  }
+.form__update-bar .icon {
```
paulfalgout: we should probably add a class to that particular icon so that we don't need the separate parent classes.

src/js/views/forms/form/form_views.js
```diff
@@ -30,36 +32,68 @@ const ContextTrailView = View.extend({
...
+      disableHistory,
     };
   },
   onRender() {
     this.renderSidebarTooltip();
     this.renderPrintTooltip();
     this.renderExpandTooltip();
+    this.renderHistoryTooltip();
+  },
+  renderHistoryTooltip() {
```
paulfalgout: This tooltip is new to me, but I'm not sure it's a good idea.  The form iframe is a black box.  We don't know when a form has successfully created a response, so at no point can we understand when to enable history.  For instance currently when you submit a form, you will not see the update bar until you leave the form and come back.

src/js/views/forms/form/form_views.js
```diff
@@ -88,6 +122,72 @@ const ContextTrailView = View.extend({
...
+  triggers: {
+    'click @ui.update': 'click:update',
+  },
+});
+
+const HistoryDroplist = Droplist.extend({
+  popWidth: 224,
+  viewOptions: {
+    className: 'button-filter form__button-response',
+    template: hbs`<span class="u-margin--r-4">{{far "history"}}</span> {{formatDateTime _created_at "AT_TIME"}}{{far "angle-down"}}`,
```
paulfalgout: Ideally we don't have to wrap the icon in a utility class.

src/js/views/forms/form/form_views.js
```diff
@@ -88,6 +122,72 @@ const ContextTrailView = View.extend({
...
+  },
+  picklistOptions: {
+    itemTemplate: hbs`{{formatDateTime _created_at "AT_TIME"}}`,
+  },
+  onChangeSelected(selected) {
```
paulfalgout: We can probably skip this listener and the HistoryView can just listen to `change:selected`

src/js/views/forms/form/form_views.js
```diff
@@ -156,9 +266,32 @@ const LayoutView = View.extend({
   onClickUpdate() {
     this.shouldUpdate = true;
     const sidebar = this.detachChildView('sidebar');
+    // If iframe is expanded, there is no sidebar
+    if (!sidebar) return;
```
paulfalgout: I think this probably would need to be:
```js
if (sidebar) this.showChildView('sidebar', sidebar);
```
preventing the render is breaking the update.

src/js/views/forms/form/form_views.js
```diff
@@ -156,9 +266,32 @@ const LayoutView = View.extend({
...
     this.render();
     this.showChildView('sidebar', sidebar);
   },
+  toggleHistory() {
```
paulfalgout: I think this view might be doing too much work and we should be putting this back in the app.

src/js/entities-service/entities/actions.js
```diff
@@ -31,12 +31,15 @@ const _Model = BaseModel.extend({
     if (!formId) return;
     return Radio.request('entities', 'forms:model', formId);
   },
-  getRecentResponse() {
-    const formResponses = Radio.request('entities', 'formResponses:collection', this.get('_form_responses'), {
+  getFormResponses() {
```
paulfalgout, line 34: I think this is the correct update and we can remove `getRecentResponse` entirely and just use `.first` when needed

paulfalgout (changes requested):
I think we need to pull the iframe out of the layout and re-render it (or use multiple small views) along with adding the appropriate action bar from the app instead of inside the layout view.  
The layoutview can then be passed the app's state as it's model, specifically for handling the top right icon states.

Then based on those states in the app, the app can decide what should be displayed.
The app should get the collection of responses and can store the id of the currently selected history one.  My suggestion would be to remove `showHistory` and just have a `historyResponseId` that can be null.  The app can pass the responses collection to the layout to that the history button can disable based on length or toggle to the first id instead of needing a boolean.

I'll ping mark, but I'm not sure as mentioned in line, that we should be utilizing that tooltip.  At some point maybe we could be listening to change events on the responses collection, but as it stands, we're completely unaware of what happens in the iframe, and so once the form app has loaded, the only state that can change is the top right icons the user can click on.  We might want to allow for the history view regardless and just have the one option as being the least confusing option.

paulfalgout:
OK updated states:

When there is no response, hide the history icon.
Otherwise no need to disable, we'll allow for the user to go to the "history mode" even if there's a single response.

Then I think we should split up the current form display/response/update stuff from the history stuff into two child apps.  The current form stuff will be fairly simple, but for history let's go ahead and have it refetch the action so that whenever the user goes into history, we have the latest list of form responses to display.  And then the icons/form-app can be using the state to decide which form childApp to run.

src/js/views/forms/form/form.scss
LisaManresa: Yeah changing the icon height was hacky. The issue is that the button increases in height by 2px when it contains an icon. Which made the history bar 2 pixels larger than the update bar view and there was a visible jump in button/text placement. But I should be able to make adjustments to the bar to accommodate that.

src/js/apps/forms/form/form_app.js
```diff
@@ -72,6 +82,19 @@ export default App.extend({
...
+      this.setState('historyResponseId', null);
+      return;
+    }
+
+    this.setState('historyResponseId', this.formResponses.first().id);
+  },
+  onClickPrintButton() {
+    const formChannel = Radio.channel(`form${ this.form.id }`);
+
+    formChannel.trigger('print:form');
```
paulfalgout: This should be `request`.  Can do `Radio.request(`form${ this.form.id }`, 'print:form');`

src/js/views/forms/form/form-history_views.js
```diff
@@ -0,0 +1,94 @@
...
+  },
+});
+
+const LayoutView = View.extend({
+  className: 'form__history-region flex-region',
+  template: hbs`
+    <div data-history-region></div>
+    <div class="form__iframe--has-bar" data-iframe-region></div>
+  `,
+  templateContext() {
```
paulfalgout: doesn't seem to be in use

src/js/views/forms/form/form.scss
paulfalgout: still this I think

src/js/apps/forms/form/form-history_app.js
```diff
@@ -0,0 +1,56 @@
...
+      },
+    }));
+
+    this.listenTo(historyBarView, 'click:current', this.onClickCurrent);
+
+    this.showIframeView();
+
+    const formChannel = Radio.channel(`form${ this.form.id }`);
+    const iframeView = this.getChildView('iframe');
+    this.listenTo(formChannel, 'print:form', () => {
```
paulfalgout: Should change to `reply` and will need a `stopReplying` in the `onStop`

src/js/apps/forms/form/form-history_app.js
```diff
@@ -0,0 +1,56 @@
...
+  onBeforeStart({ response }) {
+    this.showView(new LayoutView());
+    this.getRegion('iframe').startPreloader();
+
+    this.setState('response', response);
+  },
+  beforeStart({ action }) {
+    return action.fetch();
+  },
+  onStart({ patient, form }, action) {
```
paulfalgout: can remove patient

src/js/apps/forms/form/form_app.js
```diff
@@ -101,4 +134,36 @@ export default App.extend({
...
+    const historyApp = this.startChildApp('history', {
+      region: this.getRegion('form'),
+      action: this.action,
+      response: this.formResponses.first(),
+      form: this.form,
+    });
+
+    this.listenTo(historyApp, {
+      stop() {
+        this.formResponses = this.action.getFormResponses();
```
paulfalgout: I understand why this is happening, but.. I dunno.. don't love it.    I think either we should just `getFormResponses()` everywhere instead of storing `formResponses` on `this` _or_ we should listen to the action's `change:_form_response` and `this.formResponses = this.action.getFormResponses();`

Thoughts?

src/js/views/forms/form/form-history_views.js
```diff
@@ -0,0 +1,94 @@
...
+
+import Droplist from 'js/components/droplist';
+
+import PreloadRegion from 'js/regions/preload_region';
+
+const HistoryDroplist = Droplist.extend({
+  popWidth: 224,
+  viewOptions: {
+    className: 'button-filter form__button-response',
+    template: hbs`<span class="u-margin--r-4">{{far "history"}}</span> {{formatDateTime _created_at "AT_TIME"}}{{far "angle-down"}}`,
```
paulfalgout: I'm hoping this icon doesn't need the u-margin class

src/js/apps/forms/form/form-update_app.js
```diff
@@ -0,0 +1,42 @@
...
+    const formChannel = Radio.channel(`form${ this.form.id }`);
+    const layoutView = this.getView();
+    this.listenTo(formChannel, 'print:form', () => {
+      layoutView.print();
+    });
+  },
+  onClickUpdate() {
+    this.setState({
+      'shouldUpdate': true,
+      'historyResponseId': false,
```
paulfalgout: no need for historyResponseId?  Also no need for quotes on the keys

src/js/views/forms/form/form_views.js
```diff
@@ -16,53 +16,60 @@ const ContextTrailView = View.extend({
...
+    'click @ui.historyButton': 'click:historyButton',
   },
   onClickBack() {
     Radio.request('history', 'go:back');
   },
   templateContext() {
     const patient = this.getOption('patient');
-    const isSidebarVisible = !!this.state.get('sidebar');
+    const isSidebarVisible = !!this.model.get('sidebar');
+    const showHistory = this.model.get('historyResponseId') !== false;
```
paulfalgout: I think rather that having null and false, this could always be either null or the id.  Then `showHistory` would be based off of `!!action.getFormResponses().length`

src/js/apps/forms/form/form-history_app.js
```diff
@@ -0,0 +1,56 @@
...
+    'change:response': 'onChangeResponse',
+  },
+  stateEvents: {
+    'change:response': 'showIframeView',
+  },
+  onBeforeStart({ response }) {
+    this.showView(new LayoutView());
+    this.getRegion('iframe').startPreloader();
+
+    this.setState('response', response);
```
paulfalgout: I think we should probably not worry about passing the response in here.  Just set it based on `action.getFromResponses().first()` in the `onStart` as that will ensure the latest response it what is shown.

paulfalgout (changes requested):

src/js/apps/forms/form/form_app.js
LisaManresa: Ah yes, I meant to write a note about this. I just thought I'd throw it in there to see if we wanted it. Listening to the change event seems to be the cleanest solution to me.

src/js/views/forms/form/form_views.js
LisaManresa: In this case just relying on `!!action.getFormResponses().length` isn't enough. The history button is also hidden when the form is being updated. In that case we need some indication that we're in that state, hence the `false` value.

src/js/views/forms/form/form-history_views.js
LisaManresa: How should we handle the case where there are two icons inside a filter button? In the `button-filter` case it makes sense to target the `angle-down` icon specifically because that will be a consistent icon on filter dropdown menus. But the icon the left could be anything (in this case the history icon). So do we target those individually as we need them since we can't just target `.icon`? That doesn't seem very scalable.

src/js/apps/forms/form/form-history_app.js
LisaManresa: Do you think we should be passing the selected response to the `showIframeView` method then? The reason `response` is set on the state here is because we pass `this.getState('response')`. I can move it to onStart, though.

src/js/views/forms/form/form.scss
LisaManresa: So it turns out the issue has nothing to do with button borders (all buttons already have a border), it comes down to font size/line height. In the update bar, the button has a line height of 1 and font size of 14px. The `button-filter` would too, except that the icon is 16px. So I think we either need to give the bar a set height of 50px, or give all the buttons in it a line-height of 16px. Thoughts?

test/integration/forms/form.js
```diff
@@ -129,6 +134,147 @@ context('Patient Form', function() {
...
+      .get('@historyBtn')
+      .click();
+
+    cy
+      .get('@historyBtn')
+      .click();
+
+    // Cypress is unable to stub the contentWindow API in Firefox
+    // https://github.com/cypress-io/cypress/issues/136#issuecomment-658574403
+    if (!Cypress.isBrowser('firefox')) {
```
paulfalgout, line 160: can probably combine these two `if`s into the same `if`

src/js/views/forms/form/form_views.js
paulfalgout: Why are we hiding it during an update?  It doesn't seem like it would need to be.

src/js/apps/forms/form/form-history_app.js
```diff
@@ -0,0 +1,56 @@
...
+    return action.fetch();
+  },
+  onStart({ form }, action) {
+    this.form = form;
+    const formResponses = action.getFormResponses();
+
+    const historyBarView = this.showChildView('history', new HistoryBarView({
+      formResponses,
+      state: {
+        selected: action.getFormResponses().first(),
```
paulfalgout: I think this and below should just be `formReponses.first()` since we got it a couple lines up.

But yeah I think this setting of the state is probably better rather than passing it since what `first` is could have changed from the fetch

src/js/views/forms/form/form_views.js
paulfalgout: really though this never actually needs to be the id as it'll only ever be `first()` and then we're not really using it anyways..  it's either 2 states.. show history section or not..  or it's three.. those 2 plus a hidden state..  but I'm kinda hoping hidden is only based on if there's 0 history to show.

paulfalgout (changes requested):

test/integration/forms/form.js
LisaManresa: I think they need to be separate as the `.form__update-region` and `.form__history-region` elements are only present when their corresponding state is enabled. And we need two separate stubs to test the print for each.

paulfalgout: I'm not sure I'm following.  But I'm suggesting removing the } on line 167 and the `if` on line 170?

paulfalgout: ah I see the update-region above.. really the test and the stub could be combined there as well?

LisaManresa: Oh I see what you're saying 👍

paulfalgout (approved):

PR #392, opened 2020-11-04
Created column on worklists

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23664]

### DIFF
--- a/src/js/views/patients/worklist/worklist-list.scss
+++ b/src/js/views/patients/worklist/worklist-list.scss
@@ -52,6 +52,10 @@
   white-space: normal;
 }
 
+.worklist-list__meta {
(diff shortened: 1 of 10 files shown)


src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -52,6 +52,10 @@
   white-space: normal;
 }
 
+.worklist-list__meta {
```
LisaManresa, line 55: This fixes a wonky ellipsis that displays in the meta cell when the browser width is narrow.

Screen Shot 2020-11-04 at 10 45 27 AM

paulfalgout (approved):

PR #393, opened 2020-11-06
Design tweaks

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23663]


paulfalgout (changes requested):
Are `program-flow__action-attachment` and `flow__action-attachment` still in use?  If not they should be removed.

LisaManresa:
Nope, I missed pulling out the `program-flow__action-attachment` scss.

paulfalgout (approved):

PR #395, opened 2020-11-10
Change “Attachment” references to “Form”

LisaManresa:
[x] [ch23757]
This does not go as far as to change classnames, component names and `data-attachment-region` names. It's mainly i18n changes. 

@paulfalgout, do you think we should also change region names to `data-form-region` in this PR?

Sofia also found an existing button with the attachment component disabled state. That's been fixed in this PR as well.

### DIFF
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -96,11 +96,11 @@ careOptsFrontend:
       components:
         accessComponent:
           headingText: Update Access Type
-        attachmentComponent:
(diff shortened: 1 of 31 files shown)


paulfalgout (changes requested):
Yeah I had assumed we had attachment references in the data, but it appears we don't.  Why don't we remove attachment altogether.. region names, any classes, and the component can all be form.

paulfalgout (approved):

PR #399, opened 2020-11-16
Update dependency including form.io

paulfalgout:
Form.io starting with version 12 needs an arbitrary wait time so that cypress does not start testing before the form is ready. 😿

- [x] Add the Clubhouse Story ID: [ch23767]


paulfalgout:
related https://github.com/formio/formio.js/issues/3489

LisaManresa (approved):

PR #400, opened 2020-11-17
Chore/named imports

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23766]

I also have a branch migrating all underscore references to named imports. But this PR is big enough without those, so we'll start with this.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -1,14 +1,15 @@
 import $ from 'jquery';
 import _ from 'underscore';
 import Radio from 'backbone.radio';
-import 'js/utils/formatting';
+
+import startsWith from 'js/utils/formatting/starts-with';
 
 import 'sass/app-root.scss';
 
 const configVersion = '3';
 
 function start(opts) {
-  const isForm = _.startsWith(location.pathname, '/formapp/');
+  const isForm = startsWith(location.pathname, '/formapp/');
 
   if (isForm) {
     startForm();
--- a/src/js/utils/formatting/dasherize.js
+++ b/src/js/utils/formatting/dasherize.js
@@ -1,0 +1,5 @@
+import trim from 'js/utils/formatting/trim';
+
+export default str => {
+  return trim(str).replace(/([A-Z])/g, '-$1').replace(/[-_\s]+/g, '-').toLowerCase();
+};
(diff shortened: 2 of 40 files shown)


src/js/index.js
```diff
@@ -87,3 +88,15 @@ document.addEventListener('DOMContentLoaded', () => {
 
   getConfig();
 });
+
+// For use of escaping a string for within a regex
```
LisaManresa: Might not be the right place for these. Any other suggestions @paulfalgout?

src/js/index.js
```diff
@@ -87,3 +88,15 @@ document.addEventListener('DOMContentLoaded', () => {
...
+// http://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex#6969486
+RegExp.escape = function(str) {
+  return str.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
+};
+
+// Underscore uses {{ }} and {{{ }}}
+_.templateSettings = {
```
paulfalgout: I think maybe we can remove this entirely?  Are we really using underscore templates somewhere?

src/js/index.js
paulfalgout: Yeah I think `base/setup` would be better

src/js/utils/formatting/dasherize.js
```diff
@@ -0,0 +1,5 @@
+import trim from 'js/utils/formatting/trim';
```
paulfalgout: we don't actually seem to be using this.. Should we keep it around?  Maybe

paulfalgout (changes requested):

src/js/index.js
LisaManresa: I thought I remembered talking about using underscore templates for the patient fields sidebar work. But looking at `patient-template.js` that doesn't seem to be the case and there are no other instances of _.template. So I'll remove this.

src/js/utils/formatting/dasherize.js
LisaManresa: Ah right, it's used by `slugify()`, but we removed that as well. I can remove this

PR #401, opened 2020-11-19
Migrate to named underscore imports

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23792]

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -1,5 +1,5 @@
 import $ from 'jquery';
-import _ from 'underscore';
+import { extend } from 'underscore';
 import Radio from 'backbone.radio';
 
 import startsWith from 'js/utils/formatting/starts-with';
@@ -74,7 +74,7 @@ document.addEventListener('DOMContentLoaded', () => {
   };
 
   if ((_DEVELOP_ || _E2E_) && sessionStorage.getItem('cypress')) {
-    $.ajaxSetup(_.extend(ajaxSetup, {
+    $.ajaxSetup(extend(ajaxSetup, {
       beforeSend(xhr) {
         xhr.setRequestHeader('Authorization', `Bearer ${ sessionStorage.getItem('cypress') }`);
       },
--- a/src/js/utils/patient-template.js
+++ b/src/js/utils/patient-template.js
@@ -1,4 +1,7 @@
-import _ from 'underscore';
+
+import { each, first, propertyOf, reduce } from 'underscore';
+
+export const _ = { propertyOf, escape };
 
 // {{ fields.field_name.deep_nest }}
 const fieldRegEx = /{{\s*fields.([\w-.]+?)\s*}}/g;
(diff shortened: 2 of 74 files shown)


src/js/index.js
```diff
@@ -88,3 +88,15 @@ document.addEventListener('DOMContentLoaded', () => {
 
   getConfig();
 });
+
```
paulfalgout: This came back

src/js/utils/patient-template.js
```diff
@@ -27,11 +27,11 @@ function escapeChar(match) {
 function deepGetTemplate(dataKey, nestedKeys) {
   let keys = '';
 
-  _.each(nestedKeys.split('.'), key => {
+  each(nestedKeys.split('.'), key => {
     keys += `'${ key }',`;
   });
 
-  return `'+\n((__t=(_.propertyOf(data.${ dataKey })([${ keys }])))==null?'':_.escape(__t))+\n'`;
+  return `'+\n((__t=(_.propertyOf(data.${ dataKey })([${ keys }])))==null?'':escape(__t))+\n'`;
```
paulfalgout: this looks problematic to me.. either `propertyOf` will work without underscore just importing it in or escape won't work.  If this is a linter issue we can add a comment to disable that import error.  I kinda think `_` isn't necessary here, but what we could do if this break on compile is something like:
```
import { propertyOf, escape, each, first, reduce } from underscore;
export const _ = { propertyOf, escape };
Then that `_` should be available to the template string.
```

src/js/views/admin/shared/components/published_component.js
```diff
@@ -36,7 +36,7 @@ export default Droplist.extend({
     this.setState({ selected: this.collection.get(status) });
   },
   // Overridden for flow component
-  isPublishDisabled: _.constant(false),
+  isPublishDisabled: constant(false),
```
paulfalgout: does our linter complain about `isPublishedDisabled: () => false,`.   Might be better in this case.

paulfalgout (changes requested):

src/js/index.js
LisaManresa: Thanks, thought I caught all the bad merges

src/js/utils/patient-template.js
LisaManresa: The only issue was that our linter complains that `propertyOf` is defined but never used because it's wrapped in the string. Thoughts on disabling eslint for the import line?

src/js/utils/patient-template.js
```diff
@@ -73,12 +76,12 @@ export default function patientTemplate(text) {
 
   source = `var __t;\nreturn '${ source }';\n`;
 
-  const render = new Function('data', '_', source);
+  const render = new Function('data', '{ propertyOf, escape }', source);
```
LisaManresa, line 79: Ok, this was the key to getting `propertyOf` working.

PR #403, opened 2020-12-02
Add patient modal

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23812]

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -8,6 +8,7 @@ import intl from 'js/i18n';
 
 import { AppNavView, AppNavCollectionView, MainNavDroplist, PatientsAppNav, AdminAppNav } from 'js/views/globals/app-nav/app-nav_views';
 import { PatientSearchModal } from 'js/views/globals/search/patient-search_views';
+import { AddPatientModal } from 'js/views/globals/add-patient/add-patient_views';
 
 const i18n = intl.globals.nav;
 
@@ -108,6 +109,9 @@ export default App.extend({
   },
   stateEvents: {
     'change:currentApp': 'onChangeCurrentApp',
+  },
+  viewEvents: {
+    'click:addPatient': 'showAddPatientModal',
   },
   selectNav(appName, event, eventArgs) {
     this.setState('currentApp', appName);
@@ -221,5 +225,30 @@ export default App.extend({
     Radio.request('modal', 'show:custom', patientSearchModal);
 
     navView.triggerMethod('search:active', true);
+  },
+  showAddPatientModal() {
+    const addPatientView = new AddPatientModal({
+      model: Radio.request('entities', 'patients:model', {}),
+    });
+
+    const addPatientModal = Radio.request('modal', 'show:custom', addPatientView);
+
+    this.listenTo(addPatientModal, {
+      'save'({ model }) {
+        model.saveAll(model.attributes)
+          .then(({ data }) => {
+            Radio.trigger('event-router', 'patient:dashboard', data.id);
+            addPatientModal.destroy();
+          })
+          .fail(({ responseJSON }) => {
+            // This assumes that only the similar patient error is handled on the server
+            addPatientView.setState({
+              backend_errors: {
+                name: responseJSON.errors[0].detail,
+              },
+            });
+          });
+      },
+    });
   },
 });
--- a/src/js/base/jsonapi-mixin.js
+++ b/src/js/base/jsonapi-mixin.js
@@ -72,7 +72,8 @@ export default {
 
     if (isArray(entity)) {
       return {
-        data: map(entity, ({ id }) => {
+        data: map(entity, item => {
+          const id = item.id ? item.id : item;
           return { id, type: entityType };
         }),
       };
--- a/src/js/entities-service/entities/patients.js
+++ b/src/js/entities-service/entities/patients.js
@@ -1,19 +1,50 @@
-import { intersection, pluck } from 'underscore';
+import { extend, intersection, pluck } from 'underscore';
 import Radio from 'backbone.radio';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
+
+import trim from 'js/utils/formatting/trim';
 
 const TYPE = 'patients';
 
 const _Model = BaseModel.extend({
   type: TYPE,
   urlRoot: '/api/patients',
+  defaults: {
+    _groups: [],
+  },
+  validate(attrs) {
+    if (!trim(attrs.first_name)) {
+      return 'A patient first name is required';
+    }
+
+    if (!trim(attrs.last_name)) {
+      return 'A patient last name is required';
+    }
+
+    if (!trim(attrs.birth_date)) {
+      return 'A patient date of birth is required';
+    }
+
+    if (!attrs.sex) {
+      return 'A patient sex is required';
+    }
+  },
   getGroups() {
     return Radio.request('entities', 'groups:collection', this.get('_groups'));
   },
   getFields() {
     return Radio.request('entities', 'patientFields:collection', this.get('_patient_fields'));
+  },
+  saveAll(attrs) {
+    if (this.isNew()) attrs = extend({}, this.attributes, attrs);
+
+    const relationships = {
+      'groups': this.toRelation(attrs._groups, 'groups'),
+    };
+
+    return this.save(attrs, { relationships }, { wait: true });
   },
 });
 
--- a/src/js/entities-service/entities/patients.js
+++ b/src/js/entities-service/entities/patients.js
@@ -1,41 +1,45 @@
-import { extend, intersection, pluck } from 'underscore';
+import { extend, intersection, pluck, reject } from 'underscore';
 import Radio from 'backbone.radio';
 import Store from 'backbone.store';
+import dayjs from 'dayjs';
+
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
-
-import trim from 'js/utils/formatting/trim';
 
 const TYPE = 'patients';
 
 const _Model = BaseModel.extend({
   type: TYPE,
   urlRoot: '/api/patients',
-  defaults: {
-    _groups: [],
+  validate(attrs, { preSave } = {}) {
+    if (preSave) {
+      return this.preSaveValidate(attrs);
+    }
+
+    if (!attrs.first_name || !attrs.last_name) return { name: 'required' };
+    if (!attrs.birth_date) return { birth_date: 'required' };
+    if (!attrs.sex) return { sex: 'required' };
   },
-  validate(attrs) {
-    if (!trim(attrs.first_name)) {
-      return 'A patient first name is required';
-    }
+  preSaveValidate(attrs) {
+    const birthDate = dayjs(attrs.birth_date);
 
-    if (!trim(attrs.last_name)) {
-      return 'A patient last name is required';
-    }
-
-    if (!trim(attrs.birth_date)) {
-      return 'A patient date of birth is required';
-    }
-
-    if (!attrs.sex) {
-      return 'A patient sex is required';
-    }
+    if (dayjs().isBefore(birthDate)) return { birth_date: 'invalidDate' };
   },
   getGroups() {
     return Radio.request('entities', 'groups:collection', this.get('_groups'));
   },
   getFields() {
     return Radio.request('entities', 'patientFields:collection', this.get('_patient_fields'));
+  },
+  addGroup(group) {
+    const groups = this.get('_groups') || [];
+
+    this.set('_groups', groups.concat(group.id));
+  },
+  removeGroup(group) {
+    const groups = this.get('_groups');
+
+    this.set('_groups', reject(groups, groupId => groupId === group.id));
   },
   saveAll(attrs) {
     if (this.isNew()) attrs = extend({}, this.attributes, attrs);
--- a/src/js/views/globals/add-patient/add-patient_views.js
+++ b/src/js/views/globals/add-patient/add-patient_views.js
@@ -1,0 +1,351 @@
+import { isEmpty, keys } from 'underscore';
+import Backbone from 'backbone';
+import hbs from 'handlebars-inline-precompile';
+import { View } from 'marionette';
+import { mixinState } from 'marionette.toolkit';
+
+import Datepicker from 'js/components/datepicker';
+import Droplist from 'js/components/droplist';
+import GroupsManagerComponent from 'js/components/groups-manager';
+
+import InputWatcherBehavior from 'js/behaviors/input-watcher';
+
+import intl from 'js/i18n';
+import trim from 'js/utils/formatting/trim';
+
+import AddPatientTemplate from './add-patient.hbs';
+
+import './add-patient.scss';
+
+const i18n = intl.globals.addPatient.addPatientViews;
+
+const InputView = View.extend({
+  template: hbs`<input class="input-primary w-100 js-input {{#if hasError}}has-error{{/if}}" value="{{ value }}" />`,
+  templateContext() {
+    const errors = this.getOption('state').get('backend_errors');
+
+    return {
+      hasError: errors && errors[this.getOption('errorField')],
+      value: this.model.get(this.getOption('attr')),
+    };
+  },
+  behaviors: [InputWatcherBehavior],
+  ui: {
+    input: '.js-input',
+  },
+  initialize({ state, errorField }) {
+    this.listenTo(state, {
+      'change:backend_errors'(currentState, errors) {
+        if (errors && errors[errorField]) {
+          this.render();
+        } else {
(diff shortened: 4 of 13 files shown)


src/js/views/globals/app-nav/app-nav.scss
```diff
@@ -72,3 +74,13 @@ $menu-hover-color: #375467;
     font-weight: 700;
   }
 }
+
+.app-nav__add-patient {
```
paulfalgout: maybe this could be `app-nav__bottom-button`

src/js/entities-service/entities/patients.js
```diff
@@ -1,20 +1,51 @@
...
+  validate(attrs) {
+    if (!trim(attrs.first_name)) {
+      return 'A patient first name is required';
+    }
+
+    if (!trim(attrs.last_name)) {
+      return 'A patient last name is required';
+    }
+
+    if (!trim(attrs.birth_date)) {
```
paulfalgout: I don't think a trim should be needed here

src/js/entities-service/entities/patients.js
```diff
@@ -1,20 +1,51 @@
...
+    }
+  },
   getGroups() {
     return Radio.request('entities', 'groups:collection', this.get('_groups'));
   },
   getFields() {
     return Radio.request('entities', 'patientFields:collection', this.get('_patient_fields'));
   },
+  saveAll(attrs) {
+    if (this.isNew()) attrs = extend({}, this.attributes, attrs);
```
paulfalgout, line 45: I think this could be `extend({ _groups:  [] }, ` instead of adding the default array.   Because that array is created without a closure I think that default creates an odd scoping issue where
```
const a = new _Model();
const b = new _Model();
a.attributes._groups.push('hello');
console.log(b.attributes._groups[0]); // hello
```

src/js/base/jsonapi-mixin.js
```diff
@@ -72,7 +72,8 @@ export default {
 
     if (isArray(entity)) {
       return {
-        data: map(entity, ({ id }) => {
+        data: map(entity, item => {
+          const id = item.id ? item.id : item;
```
paulfalgout, line 76: these lower level changes make me nervous, but I can't think of a good reason not to support this.  Previously we would've done `toRelation(collectionOf(attr._groups)` but the "just use the id" was so useful on model creation.. this is probably a really good change..   Anything that wouldn't work here, just plain wouldn't work as a relationship anyhow as far as I can tell.

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,298 @@
...
+        const patientGroups = this.model.get('_groups');
+
+        this.model.set('_groups', reject(patientGroups, groupId => groupId === group.id));
+      },
+    });
+  },
+  showError() {
+    const error = this.state.get('error');
+
+    if (error === 'Similar patient exists') {
```
paulfalgout: we should not hardcode error strings.  The way to determine what field a json api applies to is the pointer.

However in this case because will wants to add the link to search for a patient, we should prevent the birthdate error from occurring by not allowing a submission with a future birthdate.  We can do this by either checking it on the validate and if the birthdate is in the future render our own local error, or we can modify the birthdate picker to prevent dates in the future.

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,298 @@
...
+  },
+  childViewTriggers: {
+    'save': 'save',
+    'cancel': 'close',
+  },
+  modelEvents: {
+    'change': 'showSave',
+  },
+  template: AddPatientTemplate,
+  initialize({ state }) {
```
paulfalgout: rather than passing in a state model can we do 
```
import { mixinState } from 'marionette.toolkit';


...

AddPatientModel({
 initialize(options) {
    this.initState(options);
 }
})

mixinState(AddPatientModal);
```

Then up in the app do
`addPatientModal.setState(`

That'll also provide `this.getState()` and `stateEvents`

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,298 @@
...
+    }));
+  },
+  showGroupsComponent() {
+    const groupsManager = this.showChildView('groups', new GroupsManagerComponent({
+      member: this.model,
+    }));
+
+    this.listenTo(groupsManager, {
+      'add:member'(patient, group) {
+        const patientGroups = this.model.get('_groups');
```
paulfalgout: can we change this to `this.model.addGroup(group)` and `this.model.removeGroup`?

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,298 @@
...
+        text: 'Male',
+        value: 'm',
+      },
+      {
+        text: 'Female',
+        value: 'f',
+      },
+      {
+        text: 'Other',
+        value: 'other',
```
paulfalgout: I don't believe "other" is the actual value.  I think the value is null. In ICU message `select` `other {}` is any unlisted value.  We may need to double check with BE as to what value they'll support.  This'd also need to be reflected in the validate on the model

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,298 @@
...
+        this.model.set('birth_date', date.format('YYYY-MM-DD'));
+        datepicker.destroy();
+      },
+    });
+
+    datepicker.show();
+  },
+});
+
+const FirstNameView = View.extend({
```
paulfalgout: I think there's an opportunity here to make an input view that has like `attr: 'first_name', shouldFocus: true, errorField, 'name'` or something.

src/js/views/globals/add-patient/add-patient_views.js
paulfalgout: actually because we're so married to handling only one particular error for the moment, we could ignore the "errorField" altogether and do something like `this. showFirstNameView({ error: true })` and avoid needing to pass state models

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,298 @@
...
+    this.ui.input.focus();
+  },
+  onWatchKeydown(evt) {
+    if (evt.which === ENTER_KEY) {
+      evt.preventDefault();
+      return;
+    }
+  },
+  onWatchChange(text) {
+    const newText = removeNewline(text);
```
paulfalgout: is newline an issue here?  I kinda thing an `input` won't matter and that this is needed for textareas or contenteditables

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,298 @@
...
+  initialize() {
+    this.listenTo(this.getOption('state'), {
+      'change:errorField': this.render,
+    });
+  },
+  onDomRefresh() {
+    this.ui.input.focus();
+  },
+  onWatchKeydown(evt) {
+    if (evt.which === ENTER_KEY) {
```
paulfalgout: I dunno if we need to care about the enter key?

paulfalgout (changes requested):

src/js/base/jsonapi-mixin.js
LisaManresa: FWIW, I did a search for `toRelation` in the whole app and this is the first instance of it being used with an array as far as I can tell.

src/js/views/globals/add-patient/add-patient_views.js
LisaManresa: OK. I had a feeling this would be the case. I think checking the date on validate would be a good idea. In case we end up with a date field that allows users to type out the date in addition to using a picker.

src/js/apps/globals/nav_app.js
```diff
@@ -222,4 +226,30 @@ export default App.extend({
 
     navView.triggerMethod('search:active', true);
   },
+  showAddPatientModal() {
+    const addPatientView = new AddPatientModal({
+      model: Radio.request('entities', 'patients:model', {}),
+    });
+
+    // addPatientView.setState(addPatientState);
```
paulfalgout: can remove

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,285 @@
...
+    return {
+      hasError: errors && errors[this.getOption('errorField')],
+      value: this.model.get(this.getOption('attr')),
+    };
+  },
+  behaviors: [InputWatcherBehavior],
+  ui: {
+    input: '.js-input',
+  },
+  initialize() {
```
paulfalgout: could probably `initialize({ state, errorField })` here rather than getOption in the `initialize`

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,285 @@
...
+    this.listenTo(this.getOption('state'), {
+      'change:errors'(state, errors) {
+        if (errors.name) {
```
paulfalgout: this should be using `errors[errorField]` I think

src/js/entities-service/entities/patients.js
```diff
@@ -9,12 +11,64 @@ const TYPE = 'patients';
 const _Model = BaseModel.extend({
   type: TYPE,
   urlRoot: '/api/patients',
+  validate(attrs) {
```
paulfalgout: we probably want to restructure this to be like:
```
const errors = {};

if (!attrs.first_name) errors.first_name = ...

return isEmpty(errors) ? null : errors;
```

Mostly because you can create a birth_date error that's legitimate, but not see it because something above it failed.

Also really user-facing errors here need to be i18n'd which could be done in a couple of ways.  These could continue to be some form of dev-strings that we i18n when handled in the view, or we can i18n them here.  

But for the most part these are functionally dev-related anyhow, so I'm ok with communicating these issues really however we want... for instance it could be

```
if(!every(['first_name', 'last_name', birth_date', 'sex'], attr => has(attrs, attr))) errors.cannotSave = true;
```
And in this case we never really need to check `errors.cannotSave` directly.. it's just ensuring that errors isn't empty.

Then the `birth_date` error could be specific i18n'd text or similarly just a flag indicating how the view should handle it.

FWIW we could also split the validation by passing options..  
```
validate(attrs, { preSave } = {}) {
  if (preSave) {
    // check birth_date error when the user hits the save button
   return
  }
  // check for can save errors to enable/disable save button
}
```
And so in the code some places we'd use `isValid()` some we'd use `isValid({ preSave: true })`

For the birthdate it's probably ok to have an onChange check, so we don't necessarily need to split it, but this'd be important for things where we want to allow the save button the be enabled but don't want to show an error like "input must be 3 characters long" before the user could possibly type 3 characters.

src/js/views/globals/add-patient/add-patient_views.js
paulfalgout: It's possible we don't need to check the particular error on these listens at all?  For instance.. let's say you get the birthday error.. fix it and submit, but the server returns the similar patient error.. does the birthday error ever clear itself?  maybe this check should just be done in the view above and re-render the whole form as the simplest way to address previous errors?

src/js/apps/globals/nav_app.js
```diff
@@ -222,4 +226,30 @@ export default App.extend({
...
+      'save'({ model }) {
+        model.saveAll(model.attributes)
+          .then(({ data }) => {
+            Radio.trigger('event-router', 'patient:dashboard', data.id);
+            addPatientModal.destroy();
+          })
+          .fail(({ responseJSON }) => {
+            addPatientView.setState({
+              errors: {
+                name: responseJSON.errors[0].detail,
```
paulfalgout, line 247: Lets add a comment here that says that our mvp implementation assumes handling only the similar-patient server error.

paulfalgout (changes requested):

src/js/apps/globals/nav_app.js
```diff
@@ -216,10 +221,54 @@ export default App.extend({
...
+        model.saveAll(model.attributes)
+          .then(({ data }) => {
+            Radio.trigger('event-router', 'patient:dashboard', data.id);
+            addPatientModal.destroy();
+          })
+          .fail(({ responseJSON }) => {
+            // This assumes that only the similar patient error is handled on the server
+            addPatientView.setState({
+              errors: {
+                name: 'duplicate',
```
LisaManresa: This may not be ideal, I'd assume we want to surface BE errors as they are. But to do that here, we'll need a BE error state and a FE error state so that we can differentiate the errors message string from the BE and the error code string from the FE.

src/js/apps/globals/nav_app.js
```diff
@@ -206,6 +210,7 @@ export default App.extend({
     const patientSearchModal = new PatientSearchModal({
       collection: Radio.request('entities', 'searchPatients:collection'),
       prefillText,
+      isFromAddPatient: !!this.getState('addPatientModal'),
```
LisaManresa: In talking with Mark, a nice-to-have is the ability to go back to the add patient modal from patient search modal.

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,325 @@
...
+      hasError: errors && errors[this.getOption('errorField')],
+      value: this.model.get(this.getOption('attr')),
+    };
+  },
+  behaviors: [InputWatcherBehavior],
+  ui: {
+    input: '.js-input',
+  },
+  initialize({ state, errorField }) {
+    this.listenTo(state, 'change:errors', () => {
```
LisaManresa: Moving this functionality to the parent view causes focus to be hijacked when a user is correcting an error. For example: They submit and get the `similar patient exists` error, then they try to modify the last name. When the error is cleared after the first letter they type, the whole form rerenders and changes the focus to the beginning of the first name input.

src/js/apps/globals/nav_app.js
paulfalgout: We need to be displaying the string passed up by the backend.  So let's differentiate by either setting two different states, or perhaps by setting the i18n string in the state rather than the code, or maybe just `i18n.errorView[field][errorCode] || errorCode`

src/js/apps/globals/nav_app.js
```diff
@@ -216,10 +221,54 @@ export default App.extend({
...
       },
+      'show:addPatient': this.showAddPatientModal,
     });
 
     Radio.request('modal', 'show:custom', patientSearchModal);
 
     navView.triggerMethod('search:active', true);
   },
+  showAddPatientModal() {
+    const storedAddPatientModal = this.getState('addPatientModal');
```
paulfalgout: the modal is destroyed when the search modal is shown.  We should not store it and reuse it.

I think it's ideal if we don't introduce this right now, and if mark wants it down the line we can make a new feature

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,325 @@
...
+  `,
+  triggers: {
+    'click .js-cancel': 'cancel',
+    'click .js-save': 'click:save',
+  },
+  templateContext() {
+    const state = this.getOption('state');
+
+    return {
+      isDisabled: !this.model.isValid() || !isEmpty(state.get('errors')) || this.isDisabled,
```
paulfalgout: I think the middle check is redundant.. and I don't think the `isEmpty` is necessary.  errors should never be `{}`

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,325 @@
...
+      },
+    });
+
+    this.listenTo(datepicker, {
+      'change:selectedDate'(date) {
+        const state = this.getOption('state');
+        const errors = state.get('errors');
+
+        state.set('errors', omit(errors, 'birth_date'));
+        this.model.set('birth_date', date.format('YYYY-MM-DD'));
```
paulfalgout, line 155: I think it would be better to do `this.model.set('birth_date', date.format('YYYY-MM-DD'), { validate: true })` Then I suppose `state.set('errors', this.model.validationErrors);` however what might be better in general is in the patientmodal doing a `modelEvents` for the event `invalid` and setting the state anytime that event occurs.

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -0,0 +1,325 @@
...
+});
+
+const ErrorView = View.extend({
+  className: 'add-patient__error',
+  template: hbs`{{ error }}{{#if showSearch}}<span class="add-patient__search js-search">{{@intl.globals.addPatient.addPatientViews.errorView.searchBtn}}</span>{{/if}}`,
+  templateContext() {
+    const state = this.getOption('state');
+    const errors = state.get('errors');
+
+    if (isEmpty(errors)) return;
```
paulfalgout: I don't think `errors` will ever be an empty object

paulfalgout (changes requested):

paulfalgout (approved):

PR #407, opened 2020-12-12
Birth date select component

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23867]

There are some outstanding UI concerns that I'll work with Mark on. Some behavior may change. But there's still a good amount that can be reviewed and shouldn't change.


src/js/components/dateselect/dateselect_views.js
```diff
@@ -0,0 +1,42 @@
...
+
+import DateSelectTemplate from './date-select.hbs';
+
+const LayoutView = View.extend({
+  template: DateSelectTemplate,
+  regions: {
+    selectRegion: '[data-select-region]',
+  },
+  ui: {
+    'cancel': '.js-cancel',
```
paulfalgout: left side doesn't need quotes

src/js/components/dateselect/index.js
```diff
@@ -0,0 +1,165 @@
...
+  const month = dayjs().month(idx);
+
+  return {
+    value: idx,
+    text: `${ month.format('MMMM') } (${ (idx + 1) })`,
+  };
+}));
+
+const selectListOptions = ['buttonText', 'buttonClassName'];
+const SelectList = Droplist.extend({
```
paulfalgout, line 36: I think we want the `isSelectlist: true` so you can type out the date

src/js/components/dateselect/index.js
```diff
@@ -0,0 +1,165 @@
...
+import Component from 'js/base/component';
+import Droplist from 'js/components/droplist';
+
+import { LayoutView } from './dateselect_views';
+
+import './date-select.scss';
+
+const i18n = intl.components.dateSelect;
+
+const monthsCollection = new Backbone.Collection(times(12, idx => {
```
paulfalgout, line 17: Maybe the year opts should move up here too since it'll be the same every time.

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -4,7 +4,8 @@ import hbs from 'handlebars-inline-precompile';
 import { View } from 'marionette';
 import { mixinState } from 'marionette.toolkit';
 
-import Datepicker from 'js/components/datepicker';
+import DateSelectComponent from 'js/components/dateselect';
+// import Datepicker from 'js/components/datepicker';
```
paulfalgout: can remove

test/integration/globals/app-nav.js
```diff
@@ -334,6 +367,11 @@ context('App Nav', function() {
   });
 
   specify('add patient failure', function() {
+    // Flake Warning:
```
paulfalgout:
```
const testDate = new Date(2019, 1, 1).getTime(); // Jan 1, 2019 timestamp

cy.clock(testDate);
// test date relative to testDate
cy.clock().invoke('restore');
```

src/js/views/globals/add-patient/add-patient_views.js
```diff
@@ -227,10 +201,6 @@ const AddPatientModal = View.extend({
   onClose() {
     this.destroy();
   },
-  onModelChange() {
```
paulfalgout, line 230: not needed?

paulfalgout (changes requested):

src/js/components/dateselect/index.js
paulfalgout: I think this is holding up a release, but otherwise the other changes could be made in a chore.

test/integration/globals/app-nav.js
LisaManresa: I ended up needed to go with `cy.clock(testDate, ['Date'])` otherwise the app wouldn't start up in the test.

paulfalgout (approved):

src/js/views/globals/add-patient/add-patient_views.js
LisaManresa: It was never actually being triggered. I accidentally left it in in the first round of dev.

PR #411, opened 2020-12-16
Shared Component tests

LisaManresa:
- [ ] Add the Clubhouse Story ID: [ch23817] [ch23887]


paulfalgout (approved):

paulfalgout:
@LisaManresa looks good but there's a conflict

paulfalgout (changes requested):
After looking into another campaign related to components I think we should move this to `js/views/shared/components/`  similar to the `patient/shared/components` and keep everything in `js/components` completely non-domain.

PR #414, opened 2020-12-21
Add owner filter to all time worklists

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch23947]
- [ ] Add QA tasks

### DIFF
--- a/src/js/apps/patients/worklist/filters_app.js
+++ b/src/js/apps/patients/worklist/filters_app.js
@@ -97,6 +97,8 @@ export default App.extend({
     const ownerFilter = new OwnerDroplist({
       owner,
       groups: this.groups,
+      isFilter: true,
+      headingText: intl.patients.worklist.filtersApp.ownerFilterHeadingText,
     });
 
     this.listenTo(ownerFilter, 'change:owner', ({ id, type }) => {
--- a/src/js/apps/patients/worklist/filters_app.js
+++ b/src/js/apps/patients/worklist/filters_app.js
@@ -4,6 +4,7 @@ import intl from 'js/i18n';
 
 import App from 'js/base/app';
 
+import OwnerDroplist from 'js/views/patients/shared/components/owner_component';
 import { FiltersView, GroupsDropList, ClinicianDropList, TypeToggleView, RoleComponent } from 'js/views/patients/worklist/filters_views';
 
 export default App.extend({
@@ -29,6 +30,10 @@ export default App.extend({
 
     if (this.shouldShowRole) {
       this.showRolesFilterView();
+    }
+
+    if (!this.shouldShowClinician && !this.shouldShowRole) {
+      this.showOwnerFilterView();
     }
   },
   showGroupsFilterView() {
@@ -83,6 +88,26 @@ export default App.extend({
     });
 
     this.showChildView('role', roleFilter);
+  },
+  showOwnerFilterView() {
+    const owner = this.getState('clinicianId') ?
+      Radio.request('entities', 'clinicians:model', this.getState('clinicianId')) :
+      Radio.request('entities', 'roles:model', this.getState('roleId'));
+
+    const ownerFilter = new OwnerDroplist({
+      owner,
+      groups: this.groups,
+    });
+
+    this.listenTo(ownerFilter, 'change:owner', ({ id, type }) => {
+      if (type === 'roles') {
+        this.setState({ roleId: id, clinicianId: null });
+      } else {
+        this.setState({ clinicianId: id, roleId: null });
+      }
+    });
+
+    this.showChildView('owner', ownerFilter);
   },
   _getGroups() {
     const groups = this.groups.clone();
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -85,6 +85,12 @@ const StateModel = Backbone.Model.extend({
       },
     };
 
+    if (!clinicianId) {
+      filters[this.id].role = roleId;
+    } else {
+      filters[this.id].clinician = clinicianId;
+    }
+
     return filters[this.id];
   },
   getSelectedList() {
--- a/src/js/views/patients/shared/components/owner_component.js
+++ b/src/js/views/patients/shared/components/owner_component.js
@@ -15,6 +15,7 @@ const i18n = intl.patients.shared.components.ownerComponent;
 const OwnerItemTemplate = hbs`{{matchText name query}} <span class="owner-component__role">{{matchText short query}}</span>`;
 const OwnerButtonTemplate = hbs`{{far "user-circle"}}{{ name }}`;
 const OwnerShortButtonTemplate = hbs`{{far "user-circle"}}{{ short }}`;
+const FilterButtonTemplate = hbs`{{far "user-circle"}}{{ name }}{{far "angle-down"}}`;
 
 let rolesCollection;
 
@@ -36,10 +37,15 @@ function getGroupClinicians(group) {
 
 export default Droplist.extend({
   isCompact: false,
+  isFilter: false,
+  headingText: i18n.headingText,
+  placeholderText: i18n.placeholderText,
+  hasRoles: true,
   popWidth() {
+    const isFilter = this.getOption('isFilter');
     const isCompact = this.getOption('isCompact');
 
-    return isCompact ? null : this.getView().$el.outerWidth();
+    return (isFilter || isCompact) ? null : this.getView().$el.outerWidth();
   },
   picklistOptions() {
     return {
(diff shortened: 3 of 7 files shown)


luketlancaster:
it's been a minute, so rip it up. I think my test setup is a bit overkill, and not sure about my solution for actually adding the filters 🤷‍♂️

paulfalgout (changes requested):
Ok so I'm realizing we need the owner filter to use the clinicianId / roleId on the state.  We should not store the owner model on the state model as we need to be able to dump the data to the cookie.  

And just to get ahead of the game let's not introduce `shouldShowOwner` and rather use if both shouldClinician / shouldRole are true for now.   After this we can I think assume that at least one is true.

luketlancaster:
ok...made some changes that aren't pretty, but I think they work. My computer keeps crapping out when I try to run the tests so mostly pushing it up to see if they pass. Lemme know what needs changing this time around @paulfalgout

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -4,13 +4,14 @@ import intl from 'js/i18n';
...
 
+import OwnerDroplist from 'js/views/patients/shared/components/owner_component';
 import { FiltersView, GroupsDropList, ClinicianDropList, TypeToggleView, RoleComponent } from 'js/views/patients/worklist/filters_views';
 
 export default App.extend({
   stateEvents: {
     'change:clinicianId': 'showFilters',
   },
-  onStart({ shouldShowClinician, shouldShowRole }) {
+  onStart({ shouldShowClinician, shouldShowRole, shouldShowOwner }) {
```
paulfalgout: `- shouldShowOwner`

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -84,6 +89,27 @@ export default App.extend({
 
     this.showChildView('role', roleFilter);
   },
+  showOwnerFilterView() {
+    const clinician = Radio.request('entities', 'clinicians:model', this.getState('clinicianId'));
+    const role = Radio.request('entities', 'roles:model', this.getState('roleId'));
+
+    const ownerFilter = new OwnerDroplist({
+      owner: clinician.isValid() ? clinician : role,
```
paulfalgout: maybe just `clinicianId ?`. `isValid` is just checking to see if the model can be saved to the server.  Really you could make a `_getOwner` perhaps that `if (clinicianId) return Radio.request.. ; return Radio.request(... roles`

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -85,6 +85,12 @@ const StateModel = Backbone.Model.extend({
       },
     };
 
+    if (!clinicianId) {
```
paulfalgout, line 88: do you even need the if/else here?  can probably just set them both either way?

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -84,6 +89,27 @@ export default App.extend({
...
+      groups: this.groups,
+    });
+
+    this.listenTo(ownerFilter, 'change:owner', ({ id, type }) => {
+      if (type === 'roles') {
+        this.setState('roleId', id);
```
paulfalgout: let's do these in a single `setState({})`

paulfalgout (changes requested):

src/js/apps/patients/worklist/filters_app.js
luketlancaster: oh yeah...I did this before I thought about unsetting the clinicianId/roleId on state. That makes sense

src/js/apps/patients/worklist/worklist_app.js
luketlancaster: If one of them is `undefined` than the filter that gets sent to the API is empty and it sure wasn't happy about that

wweaver4th:
@luketlancaster this dropdown menu looks a bit different than the others. Font size, menu width, lack of an arrow...
Screen Shot 2021-01-02 at 8 23 41 AM
Screen Shot 2021-01-02 at 8 23 52 AM

wweaver4th:
Screen Shot 2021-01-02 at 8 26 58 AM
Screen Shot 2021-01-02 at 8 27 06 AM

luketlancaster:
@wweaver4th yeah, the [card](https://app.clubhouse.io/roundingwell/story/23947/add-owner-filter-to-time-worklists) said ' For this card let's not worry about styling it as a button-filter quite yet.' so I didn't focus on styling, just the functionality

paulfalgout:
@luketlancaster  we could add more cards, but since they'll all be blocked by the next, I'm fine with just iterating here as well.

Either way, I think these todos open up the next bits:
- [ ] `picklistOptions` I _think_ should always have `className` `u-text--nowrap`. (I think this'll prevent the role from wrapping after the name and the picklist should widen)
- [x] Move `headingText` and `placeHolderText` to class like `infoText` and store the current i18n as defaults on the class itself like `isCompact`
- [x] Add `isFilter: false` to the class for template / class selection (don't necessarily need to use it yet)
- [x] Default the groups in initialize options to `[]`
- [x] The currentUser should only be added `if (this.lists.length)`
- [x] Add a `hasRoles: true` options to the class and wrap adding the roles in `if (this.getOption('hasRoles')`)
  - [x] Add `rolesHeadingText` only if `this.lists.length`

These changes should allow for the context styling and for swapping the current user/role filters on the other pages so that we just have one configurable owner filter.  which we can do in additional commits or in additional cards.

luketlancaster:
@paulfalgout sounds good to me, happy to keep working right here as well 👍

paulfalgout:
Ok to finish this off for a release:

- [ ] We need to implement `isFilter`:
  - [x] `popWidth` should be null for `isFilter === true`
  - [x] `viewOptions` should be broken up into `if (isFilter) return {... if (isCompact) return {... return {`  - this will include a new template and using the button-filter class
  - [x] `headingText` needs to be passed to the component by the view.  It should probably say, "Filter by Owner" to match the role.

I think everything else can be done in the refactor work.

luketlancaster (commented):
apparently this one comment was pending for like two weeks 🤦

luketlancaster:
@paulfalgout lemme know what you think there...I bumbled through it best I could, and it at least looks fine when manually testing. Pushed up to get tests run/coverage done too because I wanted to speed things along if I could

src/js/views/patients/shared/components/owner-component.scss
```diff
@@ -1,3 +1,12 @@
+.owner-component {
```
paulfalgout: let's remove this class for the moment.. looking into alternatives

src/js/views/patients/shared/components/owner_component.js
```diff
@@ -51,23 +57,38 @@ export default Droplist.extend({
...
   },
   viewOptions() {
     const isCompact = this.getOption('isCompact');
+    const isFilter = this.getOption('isFilter');
     const selected = this.getState('selected');
     const isRole = selected.type === 'roles';
 
+    if (isFilter) {
+      return {
+        className: 'owner-component button-filter',
```
paulfalgout: I have an idea where this can probably just be button-filter.  more soon

src/js/views/patients/shared/components/owner_component.js
```diff
@@ -51,23 +57,38 @@ export default Droplist.extend({
...
+        template: FilterButtonTemplate,
+      };
+    }
+
+    if (isCompact) {
+      return {
+        className: 'owner-component--list-item button-secondary--compact w-100',
```
paulfalgout: for the moment let's remove `owner-component--list-item` here

src/js/views/patients/shared/components/owner_component.js
```diff
@@ -51,23 +57,38 @@ export default Droplist.extend({
...
     const isRole = selected.type === 'roles';
 
+    if (isFilter) {
+      return {
+        className: 'owner-component button-filter',
+        template: FilterButtonTemplate,
+      };
+    }
+
+    if (isCompact) {
```
paulfalgout, line 74: might as well move the `const selected ... const isRole` into this if

src/js/views/patients/shared/components/owner_component.js
```diff
@@ -77,14 +98,18 @@ export default Droplist.extend({
...
 
-    this.lists.unshift({
-      collection: new Backbone.Collection([currentUser]),
-    });
-
-    this.lists.push({
-      collection: getRoles(),
-      headingText: i18n.rolesHeadingText,
-    });
+    if (this.lists.length) {
```
paulfalgout, line 98: the currentUser const can move in here

paulfalgout (changes requested):

luketlancaster:
ok @paulfalgout your changes are in, and things are looking good. I think the failures over on bulk edit do need to be fixed, but running them out here real quick to make sure

paulfalgout (approved):

src/js/views/patients/shared/components/owner_component.js
```diff
@@ -51,40 +57,57 @@ export default Droplist.extend({
         };
       },
       isSelectlist: true,
-      headingText: i18n.headingText,
```
paulfalgout, line 54: this issue is that these need to be `headingText: this.getOption('headingText'),`. @luketlancaster

PR #415, opened 2020-12-22
Style overdue time

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23869]

### DIFF
--- a/src/js/views/patients/patient/dashboard/dashboard_views.js
+++ b/src/js/views/patients/patient/dashboard/dashboard_views.js
@@ -125,7 +125,11 @@ const ActionItemView = View.extend({
   },
   showDueDay() {
     const isDisabled = this.model.isNew();
-    const dueDayComponent = new DueComponent({ date: this.model.get('due_date'), isCompact: true, state: { isDisabled } });
+    const dueDayComponent = new DueComponent({
+      date: this.model.get('due_date'),
+      isCompact: true, state: { isDisabled },
+      isOverdue: this.model.isOverdue(),
+    });
 
     this.listenTo(dueDayComponent, 'change:due', date => {
       this.model.saveDueDate(date);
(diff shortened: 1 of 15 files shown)


src/js/views/patients/patient/dashboard/dashboard_views.js
```diff
@@ -135,7 +135,11 @@ const ActionItemView = View.extend({
   },
   showDueTime() {
     const isDisabled = this.model.isNew() || !this.model.get('due_date');
-    const dueTimeComponent = new TimeComponent({ time: this.model.get('due_time'), isCompact: true, state: { isDisabled } });
+    const dueTimeComponent = new TimeComponent({
+      time: this.model.get('due_time'),
+      isCompact: true, state: { isDisabled },
+      date: this.model.get('due_date'),
```
paulfalgout: so.. I think we're going to have to do something like `isOverdue: this.model.isOverdue()` and we'll have to pass that to both the due and time components and then re-render both depending when due_time or due_date change.  Because now the time component is dependent on the whole timestamp, but the date isn't.

paulfalgout (changes requested):

src/js/views/patients/shared/components/due_component.js
```diff
@@ -68,7 +69,7 @@ export default Component.extend({
         defaultText: isCompact ? '' : i18n.defaultText,
         dateFormat: isCompact ? 'SHORT' : 'LONG',
         date: selected,
-        isOverdue: selected ? dayjs(selected).isBefore(dayjs(), 'day') : false,
+        isOverdue: (!isDisabled && selected) ? dayjs(selected).isBefore(dayjs(), 'day') : false,
```
paulfalgout: This component will have to use the same overdue logic now as the time one

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
```diff
@@ -277,8 +286,13 @@ const BulkEditActionsBodyView = View.extend({
 
     const time = this.model.get('time');
     const isDisabled = (this.model.get('dateMulti') && !time) || !this.model.get('date');
+    const isOverdue = getIsOverdue(time, this.model.get('date'));
```
paulfalgout, line 294: `this.model.isOverdue()` won't work here?

paulfalgout (changes requested):

src/js/views/patients/worklist/bulk-edit/bulk-edit_views.js
LisaManresa: In this case, `this.model` is the bulk edit app state so we don't have access to the method.

src/js/views/patients/shared/components/due_component.js
LisaManresa: Jumped the gun pushing those changes up. Just about wrapped up the due component change.

paulfalgout (approved):

PR #417, opened 2020-12-30
Datepicker month select

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23965]

### DIFF
--- a/src/js/components/datepicker/datepicker_state.js
+++ b/src/js/components/datepicker/datepicker_state.js
@@ -3,32 +3,48 @@ import Backbone from 'backbone';
 import dayjs from 'dayjs';
 
 export default Backbone.Model.extend({
-  initialize() {
-    this.set('selectedDate', this.dayjs('selectedDate'));
-    this.setCurrentMonth(this.dayjs('currentMonth'));
-
-    this.on({
-      'change:selectedDate': this.onChangeSelectedDate,
-      'change:currentMonth': this.onChangeCurrentMonth,
-    });
-  },
-  onChangeCurrentMonth(model, currentMonth) {
-    if (!(currentMonth instanceof dayjs)) {
-      this.set('currentMonth', dayjs(currentMonth));
-    }
-  },
-  onChangeSelectedDate(model, selectedDate) {
-    if (selectedDate && !(selectedDate instanceof dayjs)) {
-      this.set('selectedDate', dayjs(selectedDate));
-      return;
-    }
-    this.setCurrentMonth(selectedDate);
+  initialize({ currentMonth, selectedDate, selectedMonth }) {
+    this.setCurrentMonth(currentMonth);
+    this.setSelectedDate(selectedDate);
+    this.setSelectedMonth(selectedMonth);
   },
   setCurrentMonth(date) {
     // If date === null dayjs(date) will fail
     const currentMonth = date ? dayjs(date) : dayjs();
 
     this.set('currentMonth', currentMonth.startOf('month'));
+  },
+  setSelectedDate(date) {
+    if (!date) {
+      this.set({
+        selectedDate: null,
+      });
+      return;
+    }
+
+    date = dayjs(date);
+
+    this.set({
+      selectedDate: date.startOf('day'),
+      currentMonth: date.startOf('month'),
+      selectedMonth: null,
+    });
+  },
+  setSelectedMonth(date) {
+    if (!date) {
+      this.set({
+        selectedMonth: null,
+      });
+      return;
+    }
+
+    const selectedMonth = dayjs(date).startOf('month');
+
+    this.set({
+      selectedMonth,
+      currentMonth: selectedMonth,
+      selectedDate: null,
+    });
   },
   getCurrentMonth() {
     return this.dayjs('currentMonth');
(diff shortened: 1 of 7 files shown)


src/js/components/datepicker/index.js
```diff
@@ -94,6 +102,11 @@ export default Component.extend({
     const state = this.getState();
     state.setCurrentMonth(state.getPrevMonth());
   },
+  onSelectMonth() {
+    const state = this.getState();
+    state.setSelectedMonth(state.getCurrentMonth());
+    this.selectDate(null);
```
paulfalgout: this line doesn't seem necessary.

However `this.selectDate` should now call a function on the state model that both sets the date and clears the month.

test/integration/components/datepicker.js
```diff
@@ -34,15 +38,29 @@ context('Datepicker', function() {
...
+        'change:selectedDate': date => {
+          state.selectedDate = date;
+          this.model.set({
+            date,
+            selectedMonth: null,
+          });
+
+          datepicker.destroy();
+        },
+        'change:selectedMonth': selectedMonth => {
```
paulfalgout, line 63: I think `datepicker.destroy()` is probably ideal here too like above

src/js/components/datepicker/index.js
```diff
@@ -66,11 +70,15 @@ export default Component.extend({
   getMonthPickerView() {
     const model = this.getState();
 
-    const monthPickerView = new MonthPickerView({ model });
+    const monthPickerView = new MonthPickerView({
+      model,
+      selectMonth: this.getOption('selectMonth'),
```
paulfalgout: can we change this option name to `canSelectMonth`?

test/integration/components/datepicker.js
```diff
@@ -14,7 +14,11 @@ context('Datepicker', function() {
     },
     template: hbs`
       <button class="button--blue u-margin--t-16 u-margin--l-16">
-        {{formatDateTime date "LONG" defaultHtml="Select Date"}}
+        {{#unless selectedMonth}}
```
paulfalgout: for this test let's do `{{#if selectedMonth}}` and `{{#if date}}`  only one should be non-null at a time, so let's test that.

paulfalgout (changes requested):
I suspect we need some additional logic to handle the currentMonth based on the selectedMonth as well.  If you change the datepicker to a previous month and reopen it, it should open back up to the selected month.

src/js/components/datepicker/datepicker_state.js
```diff
@@ -30,6 +30,26 @@ export default Backbone.Model.extend({
...
+      date = dayjs(date).startOf('day');
+    }
+
+    this.set({
+      'selectedDate': date,
+      'selectedMonth': null,
+    });
+  },
+  setSelectedMonth(date) {
+    // If date === null dayjs(date) will fail
```
paulfalgout: I think this should be 
```js
if (!date) {
  this.set('selectedMonth', null);
  return;
}
```

src/js/components/datepicker/datepicker_state.js
```diff
@@ -30,6 +30,26 @@ export default Backbone.Model.extend({
...
+    const selectedMonth = date ? dayjs(date) : dayjs();
```
paulfalgout: Then this can be `const selectedMonth = dayjs(date).startOf('month');`

src/js/components/datepicker/datepicker_state.js
```diff
@@ -30,6 +30,26 @@ export default Backbone.Model.extend({
 
     this.set('currentMonth', currentMonth.startOf('month'));
   },
+  setSelectedDate(date) {
+    if (date) {
+      date = dayjs(date).startOf('day');
+    }
+
+    this.set({
```
paulfalgout, line 27: So let's make this more like the month one and update both..
I think it'll need to be

```js
if (!date) {
  this.set({
    selectedDate: null,
    currentMonth: dayjs().startOf('month'),
  });
  return;
}

date = dayjs(date);

this.set({
  selectedDate: date.startOf('day'),
  currentMonth: date.startOf('month'),
  selectedMonth: null,
});
```

Then I think we can remove the two change events above and rely on this new api.
but it means we should add both
`setSelectedMonth` and `setSelectedDate` as the first things in the initialize and remove the current `set` so that any strings are converted to dayjs.  And that should remove the need to call `setCurrentMonth`

paulfalgout (changes requested):

src/js/components/datepicker/datepicker_state.js
```diff
@@ -4,31 +4,47 @@ import dayjs from 'dayjs';
 
 export default Backbone.Model.extend({
   initialize() {
-    this.set('selectedDate', this.dayjs('selectedDate'));
-    this.setCurrentMonth(this.dayjs('currentMonth'));
+    this.setCurrentMonth(this.get('currentMonth'));
```
LisaManresa: May not be the correct approach, but we do have a scenario where a `currentMonth` can be passed in without a `selectedDate` or `selectedMonth`. This would account for that, but be overridden if either selected option is passed.

paulfalgout: That works.  FWIW.. I believe the first attribute in initialize is the data so rather than the `get`s it could be `initialize({ currentMonth, selectedDate, selectedMonth })`

paulfalgout (changes requested):
one tiny thing

PR #420, opened 2021-01-05
Fix for various owner component bugs

luketlancaster:


paulfalgout (approved):

luketlancaster:
I think the bulk edit test failures are still here...looking into those

paulfalgout (changes requested):
let's change this to hasClinicians

luketlancaster:
@paulfalgout something like that?

paulfalgout (approved):

PR #421, opened 2021-01-06
Worklist date filter

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23973]


paulfalgout (changes requested):
we should also be handling `change:selectedMonth` and turn on the ability to select a month.

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -5,15 +5,17 @@ import intl from 'js/i18n';
 import App from 'js/base/app';
 
 import OwnerDroplist from 'js/views/patients/shared/components/owner_component';
-import { FiltersView, GroupsDropList, ClinicianDropList, TypeToggleView, RoleComponent } from 'js/views/patients/worklist/filters_views';
+import { FiltersView, GroupsDropList, ClinicianDropList, TypeToggleView, RoleComponent, DateFilterView } from 'js/views/patients/worklist/filters_views';
 
 export default App.extend({
   stateEvents: {
     'change:clinicianId': 'showFilters',
+    'change:date change:relativeDate change:month': 'showDateFilterView',
```
paulfalgout: I think this entire stateEvents should probably be removed.  I don't think the `change:clinicianId` is doing anything.

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -111,6 +117,37 @@ export default App.extend({
 
     this.showChildView('owner', ownerFilter);
   },
+  showDateFilterView() {
+    const dateFilter = this.showChildView('date', new DateFilterView({
```
paulfalgout: I'm now leaning toward this being a component that has it's own state where this app is listening to state events so that it can make a new filtered request vs re-rendering and passing options.

src/js/views/patients/worklist/filters_views.js
```diff
@@ -64,6 +69,83 @@ const ClinicianDropList = Droplist.extend({
   },
 });
 
+const DateFilterView = View.extend({
+  template: hbs`
+    <button class="button-filter js-date">
```
paulfalgout: I'm a little on the fence about this.. typically when there's a single child tag this'd be better described as `tagName: 'button', className: 'button-filter', triggers: { 'click': 'click' }`

In a component `ViewClass` can be a function that uses the state to determine which view to pass.. _so_ we could remove the need for template logic and templateContext as well... there'd be a little bit of duplication.. particularly when the "Added:" label changes..  so I'm not 100% sold on that.

paulfalgout (changes requested):
I think this is pretty much there.. but I think we should go ahead and make this a component with it's own state, and which updates based on state changes, and listen to the state changes.

src/js/views/patients/worklist/filters_views.js
LisaManresa: I like the ViewClass approach. But once the value of the date type can be change, we may still need to rely on the template context. I think that could still be a fairly clean solution, though.

LisaManresa:
I believe I tackled most concerns in migrating the date filter to a Component.

I was also thinking that the worklist directory was getting cluttered and there were enough filters-specific files to warrant their own directory. Happy to undo that change if that's not the type of organization we want.

src/js/views/patients/worklist/filters/date-filter_component.js
```diff
@@ -0,0 +1,92 @@
...
+
+    if (state.get('selectedMonth')) return MonthView;
+
+    if (state.get('relativeDate')) return RelativeDateView;
+
+    return DefaultDateView;
+  },
+  viewOptions() {
+    return {
+      model: this.getState(),
```
paulfalgout, line 51: ah this is nice

paulfalgout (approved):

PR #422, opened 2021-01-06
Use OwnerComponent/droplist instead of role/clinician one one worklists

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch23983]
- [ ] Add QA tasks

so I'm still gonna need to update the tests here, but wanted some quick feedback while those were running. Should the `shouldShowClinician`/`shouldShowRole` logic get worked out a bit? Didn't know how many threads you were thinking I should pull at w/this one.

### DIFF
--- a/src/js/apps/patients/worklist/filters_app.js
+++ b/src/js/apps/patients/worklist/filters_app.js
@@ -5,7 +5,7 @@ import intl from 'js/i18n';
 import App from 'js/base/app';
 
 import OwnerDroplist from 'js/views/patients/shared/components/owner_component';
-import { FiltersView, GroupsDropList, ClinicianDropList, TypeToggleView, RoleComponent } from 'js/views/patients/worklist/filters_views';
+import { FiltersView, GroupsDropList, TypeToggleView } from 'js/views/patients/worklist/filters_views';
 
 export default App.extend({
   stateEvents: {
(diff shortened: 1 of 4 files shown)


src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -62,9 +62,8 @@ const StateModel = Backbone.Model.extend({
     const status = 'queued,started';
 
     const filters = {
-      'owned-by': { clinician: clinicianId, status, group },
+      'owned-by': { status, group },
```
paulfalgout, line 65: let's format `owned-by` and `shared-by` the same.  one line seems ok.

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -24,16 +24,10 @@ export default App.extend({
...
-      this.showCliniciansFilterView();
-    }
-
-    if (this.shouldShowRole) {
-      this.showRolesFilterView();
-    }
-
     if (!this.shouldShowClinician && !this.shouldShowRole) {
       this.showOwnerFilterView();
+    } else {
```
paulfalgout: This section should just be `this.showOwnerFilterView(...)`. no if/else.  If the `shouldShowClinician` flag isn't working let's change those.  Let's make them `shouldHideClinicians` and then we can always pass them and always rely on them being set accurately for the list.

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -64,41 +58,26 @@ export default App.extend({
...
-  showOwnerFilterView() {
-    const owner = this.getState('clinicianId') ?
-      Radio.request('entities', 'clinicians:model', this.getState('clinicianId')) :
-      Radio.request('entities', 'roles:model', this.getState('roleId'));
+  showOwnerFilterView(hasRoles = true, hasClinicians = true) {
+    // first render: has both role id and clinician id
+    // is role if both hasRoles and hasClinicians is true
+    let owner;
+
+    if (this.getState('clinicianId') && this.getState('roleId') && hasRoles && !hasClinicians) {
```
paulfalgout: Would this not work? `const owner = (!hasClinicians || !this.getState('clinicianId'))? <request role> : <request clinician>;`

paulfalgout (changes requested):

paulfalgout (approved):

PR #424, opened 2021-01-07
Update created/due labels

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch23968]
- [ ] Add QA tasks

### DIFF
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -30,7 +30,7 @@ careOptsFrontend:
           title: Programs
           programHeader: Program
           stateHeader: Program State
-          updatedHeader: Last Updated
+          updatedHeader: Updated
       reportsAllViews:
         layoutView:
           nameHeader: Name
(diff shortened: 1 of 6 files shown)


luketlancaster:
using the extra 🐴 power of the ☁️ to run the tests

src/js/i18n/en-US.yml
```diff
@@ -216,7 +216,7 @@ careOptsFrontend:
             cancelBtn: Cancel
             saveBtn: Save
           timestampsView:
-            createdAt: Created
+            createdAt: Added
```
paulfalgout, line 219: One thing that this surfaced...  We should change all of the "Last Updated"s to just "Updated". In this case if there are any keys that use "last" we should remove that.

paulfalgout (changes requested):

paulfalgout (approved):

paulfalgout (approved):

PR #426, opened 2021-01-08
Filter list by date

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch23981]
- [ ] Update filter tests

Wanted to get this in before the weekend for 👀 , but I think this is all working it just needs tests.

### DIFF
--- a/src/js/apps/patients/worklist/filters_app.js
+++ b/src/js/apps/patients/worklist/filters_app.js
@@ -1,3 +1,5 @@
+import { clone } from 'underscore';
+
 import Radio from 'backbone.radio';
 
 import intl from 'js/i18n';
@@ -86,8 +88,8 @@ export default App.extend({
     });
 
     this.listenTo(dateFilterComponent.getState(), {
-      'change'(state) {
-        this.setState(state);
+      'change'({ attributes }) {
+        this.setState(clone(attributes));
       },
     });
 
--- a/src/js/apps/patients/worklist/worklist_state.js
+++ b/src/js/apps/patients/worklist/worklist_state.js
@@ -1,0 +1,132 @@
+import { clone, extend, reduce, keys } from 'underscore';
+import dayjs from 'dayjs';
+import store from 'store';
+
+import Backbone from 'backbone';
+import Radio from 'backbone.radio';
+
+export default Backbone.Model.extend({
+  defaults() {
+    return {
+      actionsSortId: 'sortCreatedDesc',
+      flowsSortId: 'sortCreatedDesc',
+      filters: {
+        type: 'flows',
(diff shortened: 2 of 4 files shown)


src/js/apps/patients/worklist/filters_app.js
```diff
@@ -86,8 +88,8 @@ export default App.extend({
     });
 
     this.listenTo(dateFilterComponent.getState(), {
-      'change'(state) {
-        this.setState(state);
+      'change'({ attributes }) {
```
paulfalgout, line 91: I think this was technically broken, but either way needed the clone to trigger a change event after the 1st change

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -0,0 +1,169 @@
...
+    return this.getType() === 'flows';
+  },
+  getSort() {
+    return this.get(`${ this.getType() }SortId`);
+  },
+  setSort(sortId) {
+    this.set(`${ this.getType() }SortId`, sortId);
+  },
+  getEntityDateFilter() {
+    // NOTE: Default dateType handles missing filter in current localStorage
```
paulfalgout: Added a note here, but because filters is nested on the state, it gets cleared by anything in the localstorage.  So far this isn't a problem for the other state since it's all null, but `dateType` being null here was a problem.  @LisaManresa you may need to do the same.  I'm not sure though.. maybe we should be extending the localstorage filter with the defaults so this kind of thing isn't necessary?

src/js/apps/patients/worklist/filters_app.js
LisaManresa: Oooh, you're referring to the filters app state change event right? So the change event listener in worklist_app wasn't getting fired?

src/js/apps/patients/worklist/worklist_state.js
LisaManresa: I think I would lean toward extending localStorage with the defaults. Especially if we're going to have to make this accommodation in more than one place.

src/js/apps/patients/worklist/filters_app.js
paulfalgout: well the original line was setting the state to the state model and not it's attributes, I changed it to attributes but it would only get the first change event fired because every subsequent update was just updating the attributes object with the same object

paulfalgout:
@LisaManresa @luketlancaster 👀

luketlancaster (approved):
don't see anything that seems wrong...but, you know, it's me so 🤷‍♂️

LisaManresa (approved):
Seems good to go, p

PR #427, opened 2021-01-08
Add previous & next buttons to worklist date filter

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23975]
- [ ] Add QA tasks

This was a doozy to test, but I think I hit all the use cases. Just one note about test coverage below...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@auth0/auth0-spa-js": {
-      "version": "1.13.5",
-      "resolved": "https://registry.npmjs.org/@auth0/auth0-spa-js/-/auth0-spa-js-1.13.5.tgz",
-      "integrity": "sha512-wHdQf5NU1u/ZL6+vK8zrT3PV/7J7SFVQVHeQ8yrJJyaVsqMEHtr/lD9PLnhZTxPjRxo5wb3d+CO3jb2GTVNRKQ==",
+      "version": "1.13.6",
+      "resolved": "https://registry.npmjs.org/@auth0/auth0-spa-js/-/auth0-spa-js-1.13.6.tgz",
+      "integrity": "sha512-o+Mb0YXrasP364j+WB5hoALoL9sn8SZC0rDFLcyks6HUnQH/vzupdoWWFAIl1+EocAFB3CGMgSc425wpHNB4pw==",
       "requires": {
         "abortcontroller-polyfill": "^1.5.0",
         "browser-tabs-lock": "1.2.9",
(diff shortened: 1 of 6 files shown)


src/js/views/patients/worklist/filters/date-filter_state.js
```diff
@@ -22,4 +23,63 @@ export default Backbone.Model.extend({
...
+      this.setDate(dayjs().subtract(1, 'day'));
+      return;
+    }
+
+    if (relativeDate === 'yesterday') {
+      this.setDate(dayjs().subtract(2, 'days'));
+      return;
+    }
+
+    if (this._isCurrentMonth()) {
```
LisaManresa: This and the same conditional below leave branches untested: the `!this._isCurrentMonth` case.

The solution is to just make the `setMonth` call the default fallback, but I was hesitant to do that in case there was some case I was missing. I've looked this over a dozen times and haven't come up with a scenario where `this._isCurrentMonth` wouldn't be true at this point in the method. So it may be safe to remove the check.

But I'll take a look on Monday with fresh eyes and verify later. Or maybe there's a better way to structure this so we don't have an untested branch.

paulfalgout: yep I think it can be removed

paulfalgout (changes requested):
The buttons should also include a tooltip displaying the date (or month year) for if clicked.

package-lock.json
```diff
@@ -5371,7 +5371,7 @@
     "getos": {
       "version": "3.2.1",
       "resolved": "https://registry.npmjs.org/getos/-/getos-3.2.1.tgz",
-      "integrity": "sha512-U56CfOK17OKgTVqozZjUKNdkfEv6jk5WISBJ8SHoagjE6L69zOwl3Z+O8myjY9MEW3i2HPWQBt/LTbCgcC973Q==",
+      "integrity": "sha1-ATTR9OAOtGFExanArE3Ah8uyfcU=",
```
paulfalgout:
```
rm -rf node_modules package-lock.json
npm cache clear -f
```

Fixed npm switching to sha1 fo me.

src/js/views/patients/worklist/filters/date-filter_state.js
```diff
@@ -22,4 +23,54 @@ export default Backbone.Model.extend({
       selectedMonth: null,
     });
   },
+  incrementBackward() {
+    if (this.get('selectedDate')) {
+      this.setDate(dayjs(this.get('selectedDate')).subtract(1, 'day'));
```
paulfalgout: I just realized.. we do have that model helper so these could be `this.dayjs('selectedDate').`.. etc.  I dunno if it's worth it.  I almost think we should remove it.. not _that_ useful.  Thoughts?

src/js/views/patients/worklist/filters/date-filter_views.js
```diff
@@ -56,6 +59,80 @@ const DefaultDateView = View.extend({
...
+      ui: this.ui.prev,
+    });
+
+    new Tooltip({
+      message: tooltipMessages.nextMessage,
+      uiView: this,
+      ui: this.ui.next,
+    });
+  },
+  getTooltipMessages() {
```
paulfalgout, line 105: how about 2 functions here where this one is `if(this.model.get('selectedDate') return this._getTooltipMessage(this.model.dayjs('selectedDate'), 'day' 'MM/DD/YYYY');`

And the other is something like: 
```
return {
  prevMessage: dayjs(ts).subtract(1, unit).format(format),
  nextMessage: dayjs(ts).add(1, unit).format(format),
};
```

paulfalgout (changes requested):

src/js/views/patients/worklist/filters/date-filter_state.js
LisaManresa: I think this will be useful enough to keep, mainly for the null/undefined check. I think there are at least a few instances where we can simplify some logic with that.

paulfalgout: ah.. yeah good point.  well might as well use it then?

src/js/views/patients/worklist/filters/date-filter.scss
```diff
@@ -0,0 +1,27 @@
+.datepicker__type-filter {
```
paulfalgout: I don't think this css should be in this PR.  But I do think there needs to be some spacing for the left/right arrows.

paulfalgout (changes requested):
Couple more things.  I think there's a little bit of styling.  The filter has a right-aligned down arrow.  And the spacing is different with the left/right arrows.

src/js/views/patients/worklist/filters/date-filter.scss
LisaManresa: Thanks for catching this. I stashed some changes and this lingered.

paulfalgout (approved):

PR #430, opened 2021-01-12
Bust iframe if login needed

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch24073]

We don't test the auth file for various reasons, but if someone wants to double check me and pull this down and login/logout various ways that'd be swell.


luketlancaster (approved):
pulled down and did various...things. Logged out and refreshed another tab, cleared site data, etc. Looks good to me

PR #432, opened 2021-01-13
Worklist date type filter

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23979]

### DIFF
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -622,7 +622,14 @@ careOptsFrontend:
             month: This Month
             today: Today
             yesterday: Yesterday
-          added: 'Added:'
+          datepickerLayoutView:
+            added: Added
+            due: Due
+            updated: Updated
+          dateTypes:
+            created_at: 'Added:'
+            due_date: 'Due:'
+            updated_at: 'Updated:'
           defaultDateView:
             thisMonth: This Month
           relativeDateView:
--- a/src/js/views/patients/worklist/filters/date-filter_component.js
+++ b/src/js/views/patients/worklist/filters/date-filter_component.js
@@ -3,10 +3,35 @@ import { noop } from 'underscore';
 import Component from 'js/base/component';
 import Datepicker from 'js/components/datepicker';
 
-import { DateView, MonthView, RelativeDateView, DefaultDateView, ControllerView, ActionsView } from './date-filter_views';
+import { DateView, MonthView, RelativeDateView, DefaultDateView, ControllerView, ActionsView, FilterTypeView } from './date-filter_views';
 import StateModel from './date-filter_state';
 
 const DateFilterPickerComponent = Datepicker.extend({
+  stateEvents: {
+    'change': 'show',
+    'change:selectedDate': 'onChangeStateSelectedDate',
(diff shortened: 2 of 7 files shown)


src/js/i18n/en-US.yml
```diff
@@ -622,12 +622,18 @@ careOptsFrontend:
             month: This Month
             today: Today
             yesterday: Yesterday
-          added: 'Added:'
+          created_at: 'Added:'
```
paulfalgout: Is this key in use?

src/js/views/patients/worklist/filters/date-filter_views.js
```diff
@@ -23,8 +54,13 @@ const MonthView = View.extend({
   className: 'button-filter js-date',
   template: hbs`
     {{far "calendar-alt"}}{{~ remove_whitespace ~}}
-    {{ @intl.patients.worklist.filters.dateFilterViews.added }} {{formatDateTime selectedMonth "MMM YYYY"}}{{far "angle-down"}}
+    {{label}} {{formatDateTime selectedMonth "MMM YYYY"}}{{far "angle-down"}}
   `,
+  templateContext() {
+    return {
+      label: intl.patients.worklist.filters.dateFilterViews[this.model.get('dateType')],
```
paulfalgout: There's a few of these now can we make an `const i18n = intl.patients.worklist.filters;` up top?

src/js/views/patients/worklist/filters/date-filter_component.js
```diff
@@ -68,7 +99,7 @@ export default Component.extend({
     const datePicker = new DateFilterPickerComponent({
       ui: this.getView().$el,
       uiView: this.getView(),
-      state: state.pick('selectedDate', 'selectedMonth'),
+      state: state.pick('selectedDate', 'selectedMonth', 'dateType'),
```
paulfalgout: So I think `onClickDate()` will need to change to
`this.showDatePicker(this.getState().pick('selectedDate', 'selectedMonth', 'dateType'));`

and inside showDatePicker if the `dateType` changes we'll need to `this.showDatePicker({ dateType });` with the new dateType but without the current state _or_ if we want to be really fancy we'd only use the current state if the current `dateType` was re-chosen.  This is necessary so that you can change the type and then choose the same currently selected date.
Then all of the `setRelatveDate`, `setMonth`, `setDate`s need to take a second argument which is the `dateType` so that the `dateType` updates on the component state at the same time.  And we should test that `data-date-filter-region` can change to another labeling via the component.

paulfalgout (changes requested):

src/js/i18n/en-US.yml
LisaManresa: `added` no longer should be. I found one extraneous instance. Fixing that.

paulfalgout: Ahh I see why I was missing it.. could the types be nested one more time under like `dataTypes:` or something?  I didn't realize it was `dateFilterViews[]` when searching.

paulfalgout (changes requested):

paulfalgout (approved):

PR #435, opened 2021-01-17
Add cleared flag filter

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch24091]


paulfalgout:
So this is already in the 20211117 hotfix release.  It's probably an easy review.

luketlancaster (approved):

PR #437, opened 2021-01-21
Initial Schedule app/route

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24084]

Because I am the queen of scope creep, I also have the initial collection view locally. But broke it out to make this easier to digest and align with the clubhouse card.

### DIFF
--- a/src/js/apps/patients/schedule/schedule_app.js
+++ b/src/js/apps/patients/schedule/schedule_app.js
@@ -1,0 +1,83 @@
+import { clone, extend } from 'underscore';
+import Radio from 'backbone.radio';
+import store from 'store';
+
+import App from 'js/base/app';
+
+import StateModel from './schedule_state';
+
+import FiltersApp from './filters_app';
+
+import { LayoutView, ScheduleTitleView, TableHeaderView } from 'js/views/patients/schedule/schedule_views';
+
+export default App.extend({
+  StateModel,
+  childApps: {
+    filters: {
+      AppClass: FiltersApp,
+      regionName: 'filters',
+      restartWithParent: false,
+      getOptions: ['currentClinician'],
+    },
+  },
+  stateEvents: {
+    'change:filters': 'restart',
+  },
+  initListState() {
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    const storedState = store.get(`schedule_${ currentUser.id }`);
+    const filters = this.getState('filters');
+
+    // NOTE: Allows for new defaults to get added to stored filters
+    if (storedState) storedState.filters = extend({}, filters, storedState.filters);
+
+    this.setState(extend({ id: `schedule_${ currentUser.id }` }, storedState));
+  },
+  onBeforeStart() {
+    if (this.isRestarting()) {
+      this.showScheduleTitle();
+      this.getRegion('list').startPreloader();
+      return;
+    }
+
+    this.initListState();
+    this.showView(new LayoutView({
+      state: this.getState(),
+    }));
+
+    this.getRegion('list').startPreloader();
+
+    this.showScheduleTitle();
+    this.startFiltersApp();
+    this.showTableHeaders();
+  },
+  beforeStart() {
+    const filter = this.getState().getEntityFilter();
+    return Radio.request('entities', 'fetch:actions:collection:groupByDate', { filter });
+  },
+  onStart(options, collection) {
+    this.collection = collection;
+  },
+  showScheduleTitle() {
+    const filters = this.getState().get('filters');
+    const owner = Radio.request('entities', 'clinicians:model', filters.clinicianId);
+
+    this.showChildView('title', new ScheduleTitleView({
+      model: owner,
+    }));
+  },
+  startFiltersApp() {
+    const filtersApp = this.startChildApp('filters', {
+      state: this.getState().getFilters(),
+    });
+
+    this.listenTo(filtersApp.getState(), 'change', ({ attributes }) => {
+      this.setState({ filters: clone(attributes) });
+    });
+  },
+  showTableHeaders() {
+    const tableHeadersView = new TableHeaderView();
+
+    this.showChildView('table', tableHeadersView);
+  },
+});
--- a/src/js/apps/patients/schedule/schedule_state.js
+++ b/src/js/apps/patients/schedule/schedule_state.js
@@ -1,0 +1,86 @@
+import { clone, extend } from 'underscore';
+import store from 'store';
+import Backbone from 'backbone';
+import Radio from 'backbone.radio';
+import dayjs from 'dayjs';
+
+export default Backbone.Model.extend({
+  defaults() {
+    return {
+      filters: {
+        groupId: null,
+        clinicianId: this.currentClinician.id,
+        selectedDate: null,
+        selectedMonth: null,
+        relativeDate: null,
+      },
+      selectedActions: {},
+    };
+  },
+  preinitialize() {
+    this.currentClinician = Radio.request('bootstrap', 'currentUser');
+    this.groups = this.currentClinician.getGroups();
+  },
+  initialize() {
+    this.on('change', this.onChange);
+  },
+  onChange() {
+    store.set(`schedule_${ this.currentClinician.id }`, this.attributes);
+  },
+  getFilters() {
+    return clone(this.get('filters'));
+  },
+  getEntityDateFilter() {
+    const { selectedDate, selectedMonth, relativeDate } = this.getFilters();
+
+    if (selectedDate) {
+      return {
+        due_date: this.formatDateRange(selectedDate, 'day'),
+      };
+    }
+
+    if (selectedMonth) {
+      return {
+        due_date: this.formatDateRange(selectedMonth, 'month'),
+      };
+    }
+
+    if (relativeDate === 'today') {
+      return {
+        due_date: this.formatDateRange(dayjs(), 'day'),
+      };
+    }
+
+    if (relativeDate === 'yesterday') {
+      const yesterday = dayjs().subtract(1, 'days');
+      return {
+        due_date: this.formatDateRange(yesterday, 'day'),
+      };
+    }
+
+    // This month
+    return {
+      due_date: this.formatDateRange(dayjs(), 'month'),
+    };
+  },
+  getEntityFilter() {
+    const { groupId, clinicianId } = this.getFilters();
+    const group = groupId || this.groups.pluck('id').join(',');
+    const status = 'queued,started';
+
+    const dateFilter = this.getEntityDateFilter();
+
+    const filters = extend({
+      clinician: clinicianId,
+      status,
+      group,
+    }, dateFilter);
+
+    return filters;
+  },
+  formatDateRange(date, rangeType) {
+    const dateFormat = 'YYYY-MM-DD';
+
+    return `${ dayjs(date).startOf(rangeType).format(dateFormat) },${ dayjs(date).endOf(rangeType).format(dateFormat) }`;
+  },
+});
--- a/src/js/entities-service/actions.js
+++ b/src/js/entities-service/actions.js
@@ -1,3 +1,6 @@
+import Backbone from 'backbone';
+import { keys, reduce } from 'underscore';
+
 import BaseEntity from 'js/base/entity-service';
 import { _Model, Model, Collection } from './entities/actions';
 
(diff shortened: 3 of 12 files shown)


src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -0,0 +1,78 @@
...
+      this.setState({ filters: clone(attributes) });
+    });
+  },
+  showTableHeaders() {
+    const tableHeadersView = new TableHeaderView();
+
+    this.showChildView('table', tableHeadersView);
+  },
+  _getScheduledByDate() {
+    const groupedCollection = this.collection.groupBy('due_date');
```
LisaManresa: This is what I'm thinking for formatting the original action collection into one grouped by date for the collectionView to display. We can use buildChildView to send actions to a child collection view as a collection to display actions by date.

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -0,0 +1,88 @@
...
+      getOptions: ['currentClinician'],
+    },
+  },
+  stateEvents: {
+    'change:filters': 'restart',
+  },
+  onBeforeStart() {
+    if (this.isRestarting()) {
+      this.showScheduleTitle();
+      this.showTableHeaders();
```
paulfalgout: I don't think the table headers should be in the restart

paulfalgout: The title restarting makes sense I suppose because it changes based on the owner state.. _but_ that view itself has code to change based on the state, but we don't pass the state to that view...

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -0,0 +1,75 @@
...
+  },
+  templateContext() {
+    return {
+      owner: this.getOption('owner').get('name') || this.getOption('role').get('name'),
+    };
+  },
+  initialize({ state }) {
+    this.state = state;
+
+    this.listenTo(state, {
```
paulfalgout: as mentioned above.. we should either be doing this and passing the state down, or we should be re-showing it in the `restart`. I think this was copied from the worklist and should be cleaned up there as well

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -0,0 +1,75 @@
...
+  template: hbs`
+    {{formatMessage (intlGet "patients.schedule.scheduleViews.scheduleTitleView.title") role=role owner=owner}}{{~ remove_whitespace ~}}
+    <span class="list-page__header-icon js-title-info">{{fas "info-circle"}}</span>
+  `,
+  ui: {
+    tooltip: '.js-title-info',
+  },
+  templateContext() {
+    return {
+      owner: this.getOption('owner').get('name') || this.getOption('role').get('name'),
```
paulfalgout: It wasn't clear in the marvel campaign because of a last minute change, but we're only going to do clinicians and not roles for the schedule to match the "owned by" worklist.

src/js/apps/patients/schedule/schedule_state.js
```diff
@@ -0,0 +1,89 @@
...
+  initialize() {
+    this.on('change', this.onChange);
+  },
+  onChange() {
+    store.set(`schedule_${ this.currentClinician.id }`, this.attributes);
+  },
+  getFilters() {
+    return clone(this.get('filters'));
+  },
+  getEntityDateFilter() {
```
paulfalgout, line 33: this can probably be simplified since it's only `due_date` on the schedule

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -0,0 +1,88 @@
...
+  },
+  onStart(options, collection) {
+    this.collection = collection;
+  },
+  showScheduleTitle() {
+    const filters = this.getState().getFilters();
+    const owner = Radio.request('entities', 'clinicians:model', filters.clinicianId);
+    const role = Radio.request('entities', 'roles:model', filters.roleId);
+
+    this.showChildView('title', new ScheduleTitleView({
```
paulfalgout, line 65: we can probably pass the clinician as the model for this view, if we're not passing the state

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -0,0 +1,88 @@
...
+      owner,
+      role,
+    }));
+  },
+  startFiltersApp() {
+    const filtersApp = this.startChildApp('filters', {
+      state: this.getState().getFilters(),
+      shouldShowClinician: this.getState().id !== 'shared-by',
```
paulfalgout: all of these conditionals are based off of worklist state ids

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -0,0 +1,88 @@
...
+  },
+  _getScheduledByDate() {
+    const groupedCollection = this.collection.groupBy('due_date');
+    const collectionByDate = reduce(keys(groupedCollection), (collection, key) => {
+      return collection.concat({
+        date: key,
+        actions: groupedCollection[key],
+      });
+    }, []);
+    return new Backbone.Collection(collectionByDate);
```
paulfalgout: should probably sort this by date.

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -0,0 +1,88 @@
...
+    const tableHeadersView = new TableHeaderView();
+
+    this.showChildView('table', tableHeadersView);
+  },
+  _getScheduledByDate() {
+    const groupedCollection = this.collection.groupBy('due_date');
+    const collectionByDate = reduce(keys(groupedCollection), (collection, key) => {
+      return collection.concat({
+        date: key,
+        actions: groupedCollection[key],
```
paulfalgout: `groupedCollection[key]` is an array correct? not a collection?  We might want to move this function into the actions entity service an do a `fetch:actions:groupByDueDate` or something and have those actions be a collection so we can pass them right along as you suggested.

src/js/views/patients/schedule/layout.hbs
```diff
@@ -0,0 +1,9 @@
+<div class="list-page__header">
```
paulfalgout, line 1: There's a lot of classes in here that aren't used for anything.  Can we clean that up?  Also if it can be avoided I'd like to not use a class for just cypress.  and prefer`.js-` or region selectors over style ones

src/js/views/patients/schedule/filters_views.js
```diff
@@ -0,0 +1,37 @@
+import hbs from 'handlebars-inline-precompile';
+import { View } from 'marionette';
+
+import 'sass/modules/buttons.scss';
+
+import Droplist from 'js/components/droplist';
+import OwnerDroplist from 'js/views/patients/shared/components/owner_component';
+
+const FiltersView = View.extend({
+  className: 'worklist-list__filters',
```
paulfalgout: shouldn't reuse the workist-list classes here.  Either we need to module these or duplicate them.

paulfalgout (changes requested):
We can simplify to owner being only a clinician and date being only due across this PR.

src/js/apps/patients/schedule/schedule_app.js
paulfalgout: can we pass `model: owner` instead of needing `templateContext` on this view?

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -0,0 +1,73 @@
...
+    const owner = Radio.request('entities', 'clinicians:model', filters.clinicianId);
+
+    this.showChildView('title', new ScheduleTitleView({
+      owner,
+    }));
+  },
+  startFiltersApp() {
+    const filtersApp = this.startChildApp('filters', {
+      state: this.getState().getFilters(),
+      shouldShowClinician: true,
```
paulfalgout: honestly we probably don't need these should flags at all for now.  It's definitely not currently using them and maybe never will.

src/js/apps/patients/schedule/schedule_state.js
```diff
@@ -0,0 +1,87 @@
...
+import dayjs from 'dayjs';
+
+export default Backbone.Model.extend({
+  defaults() {
+    return {
+      filters: {
+        groupId: null,
+        clinicianId: this.currentClinician.id,
+        roleId: this.currentClinician.getRole().id,
+        dateType: 'due_date',
```
paulfalgout: can we remove `dateType`?

src/js/apps/patients/schedule/schedule_state.js
```diff
@@ -0,0 +1,87 @@
...
+import Radio from 'backbone.radio';
+import dayjs from 'dayjs';
+
+export default Backbone.Model.extend({
+  defaults() {
+    return {
+      filters: {
+        groupId: null,
+        clinicianId: this.currentClinician.id,
+        roleId: this.currentClinician.getRole().id,
```
paulfalgout: `roleId` doesn't seem necessary

src/js/apps/patients/schedule/schedule_state.js
```diff
@@ -0,0 +1,87 @@
...
+  },
+  onChange() {
+    store.set(`schedule_${ this.currentClinician.id }`, this.attributes);
+  },
+  getFilters() {
+    return clone(this.get('filters'));
+  },
+  getEntityDateFilter() {
+    const { selectedDate, selectedMonth, relativeDate } = this.getFilters();
+    const dateFormat = 'YYYY-MM-DD';
```
paulfalgout: we can probably just move this variable into the function

src/js/apps/patients/schedule/schedule_state.js
```diff
@@ -0,0 +1,87 @@
...
+  },
+  getEntityFilter() {
+    const { groupId, clinicianId } = this.getFilters();
+    const group = groupId || this.groups.pluck('id').join(',');
+    const status = 'queued,started';
+
+    const dateFilter = this.getEntityDateFilter();
+
+    const filters = extend({
+      clinician: clinicianId || this.currentClinician.id,
```
paulfalgout: I'm not sure I see the need for the `||` There's defaults on the model and it should never be nullable.

src/js/apps/patients/schedule/schedule_state.js
```diff
@@ -0,0 +1,87 @@
...
+  },
+  preinitialize() {
+    this.currentClinician = Radio.request('bootstrap', 'currentUser');
+    this.groups = this.currentClinician.getGroups();
+  },
+  initialize() {
+    this.on('change', this.onChange);
+  },
+  onChange() {
+    store.set(`schedule_${ this.currentClinician.id }`, this.attributes);
```
paulfalgout, line 28: I see that we're setting the store, but where are we getting it?

src/js/entities-service/actions.js
```diff
@@ -27,6 +33,30 @@ const Entity = BaseEntity.extend({
...
+  fetchActionsByDate({ filter }) {
+    const data = { filter };
+
+    return this.fetchCollection({ data })
+      .then(response => {
+        const groupedCollection = response.groupBy('due_date');
+        const arrayByDate = reduce(keys(groupedCollection), (collection, key) => {
+          return collection.concat({
+            date: key,
+            actions: new Backbone.Collection(groupedCollection[key], {
```
paulfalgout: I don't think we want this to be a generic backbone collection.  Seems like it should be `Collection` in this context would be an actions collection.

Also it might be worth passing the `new Backbone.Collection` as the memo instead of `[]` to the reduce such that instead of array.concat it'd be `collection.add` then `return collection`

Also, rethinking this, let's sort both on the collectionviews and not in the data.

paulfalgout (changes requested):

src/js/entities-service/actions.js
```diff
@@ -27,6 +32,24 @@ const Entity = BaseEntity.extend({
...
+  fetchActionsByDate({ filter }) {
+    const data = { filter };
+
+    return this.fetchCollection({ data })
+      .then(response => {
+        const groupedCollection = response.groupBy('due_date');
+        const arrayByDate = reduce(keys(groupedCollection), (collection, key) => {
+          collection.add({
+            date: key,
+            actions: Radio.request('entities', 'actions:collection', groupedCollection[key]),
```
paulfalgout: I think this can just be `this.getCollection(groupdCollection[key])`

src/js/entities-service/actions.js
```diff
@@ -27,6 +32,24 @@ const Entity = BaseEntity.extend({
 
     return this.fetchCollection({ url });
   },
+  fetchActionsByDate({ filter }) {
+    const data = { filter };
+
+    return this.fetchCollection({ data })
+      .then(response => {
+        const groupedCollection = response.groupBy('due_date');
+        const arrayByDate = reduce(keys(groupedCollection), (collection, key) => {
```
paulfalgout: I think this could be `return reduce...`

paulfalgout (changes requested):
small thing then GTG

PR #438, opened 2021-01-22
Impliment strange formio workaround

paulfalgout:
Hopefully solved by:

https://github.com/formio/formio.js/issues/3684

- [x] Add the Clubhouse Story ID: [ch24208]


paulfalgout:
@wweaver4th tested this on a sandbox so we're going to push this 🐮 👕 through.

PR #442, opened 2021-01-26
Fix expand view bug in forms

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch24081]
- [ ] Add QA tasks

### DIFF
--- a/src/js/apps/forms/form/form_app.js
+++ b/src/js/apps/forms/form/form_app.js
@@ -53,12 +53,13 @@ export default App.extend({
     }));
 
     this.setState({
-      sidebar: 'action',
+      isActionSidebar: true,
+      isExpanded: false,
       historyResponseId: null,
     });
   },
   stateEvents: {
-    'change:sidebar': 'onChangeSidebar',
+    'change': 'onChangeState',
     'change:historyResponseId': 'onChangeHistoryResponseId',
   },
   viewEvents: {
@@ -68,24 +69,18 @@ export default App.extend({
     'click:printButton': 'onClickPrintButton',
   },
   onClickSidebarButton() {
-    const sidebar = this.getState('sidebar');
+    const isActionSidebar = this.getState('isActionSidebar');
+    const isExpanded = this.getState('isExpanded');
 
-    if (sidebar === 'action') {
-      this.setState('sidebar', 'patient');
+    if (isExpanded) {
+      this.setState({ isActionSidebar: true, isExpanded: false });
       return;
     }
 
-    this.setState('sidebar', 'action');
+    this.setState({ isActionSidebar: !isActionSidebar, isExpanded: false });
   },
   onClickExpandButton() {
-    const sidebar = this.getState('sidebar');
-
-    if (sidebar) {
-      this.setState('sidebar', null);
-      return;
-    }
-
-    this.setState('sidebar', this.getState().previous('sidebar'));
+    this.toggleState('isExpanded');
   },
   onClickHistoryButton() {
     if (this.getChildApp('history').isRunning()) {
@@ -98,23 +93,30 @@ export default App.extend({
   onClickPrintButton() {
     Radio.request(`form${ this.form.id }`, 'print:form');
   },
-  onChangeSidebar(state, sidebar) {
-    this.stopChildApp('patient');
-    Radio.request('sidebar', 'close');
+  onChangeState(model) {
+    if (!model.hasChanged('isExpanded') && !model.hasChanged('isActionSidebar')) return;
 
-    if (!sidebar) {
+    const isActionSidebar = model.get('isActionSidebar');
+    const isExpanded = model.get('isExpanded');
+
+    if (!isActionSidebar || isExpanded) {
+      Radio.request('sidebar', 'close');
+    }
+
+    if (isExpanded) {
+      this.stopChildApp('patient');
       this.getRegion('sidebar').empty();
       return;
+    }
+
+    if (isActionSidebar) {
+      this.showActionSidebar();
     }
 
     this.startChildApp('patient', {
       region: this.getRegion('sidebar'),
       patient: this.patient,
     });
-
-    if (sidebar === 'action') {
-      this.showActionSidebar();
-    }
   },
   onChangeHistoryResponseId() {
     const historyResponseId = this.getState('historyResponseId');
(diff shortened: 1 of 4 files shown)


src/js/apps/forms/form/form_app.js
```diff
@@ -54,11 +54,13 @@ export default App.extend({
 
     this.setState({
       sidebar: 'action',
+      expanded: false,
```
paulfalgout: can we change this to `isExpanded`

src/js/apps/forms/form/form_app.js
```diff
@@ -131,8 +128,9 @@ export default App.extend({
 
     this.listenTo(sidebarApp, 'stop', () => {
```
paulfalgout, line 134: I think this handler can be simplified.  it's always going to be patient now, so we only care about expanded.  So this can be 
```
if expanded return;
this.setState...patient
```

src/js/views/forms/form/form_views.js
```diff
@@ -68,8 +69,18 @@ const ContextTrailView = View.extend({
     this.renderHistoryTooltip();
   },
   renderSidebarTooltip() {
-    const shouldShowActionSidebar = this.model.get('sidebar') !== 'action';
-    const message = shouldShowActionSidebar ? intl.forms.form.formViews.showActionSidebar : intl.forms.form.formViews.hideActionSidebar;
+    // todo: fixme
+    let message;
```
paulfalgout: I notice a fixme, but I generally consider `let` and `else` code smells.  not always, but generally.  some sort of `getMessage` is probably better here.

src/js/apps/forms/form/form_app.js
```diff
@@ -54,11 +54,13 @@ export default App.extend({
 
     this.setState({
       sidebar: 'action',
```
paulfalgout: Since for the moment this is functioning as a boolean, let's make it so and do `isActionSidebar`

src/js/apps/forms/form/form_app.js
```diff
@@ -69,6 +71,7 @@ export default App.extend({
   },
   onClickSidebarButton() {
     const sidebar = this.getState('sidebar');
+    this.setState('expanded', false);
```
paulfalgout: I think we should set expanded and sidebar in a single change here and not separately from sidebar

src/js/apps/forms/form/form_app.js
```diff
@@ -54,11 +54,13 @@ export default App.extend({
 
     this.setState({
       sidebar: 'action',
+      expanded: false,
       historyResponseId: null,
     });
   },
   stateEvents: {
     'change:sidebar': 'onChangeSidebar',
```
paulfalgout: so this'll probably be easier if we make a `change` handler and check if the event is `change:sidebar` or `change:expanded`   Otherwise we run into the handler getting called twice.. which is doable, but probably in this case more complicated

src/js/apps/forms/form/form_app.js
```diff
@@ -98,11 +94,10 @@ export default App.extend({
   onClickPrintButton() {
     Radio.request(`form${ this.form.id }`, 'print:form');
   },
-  onChangeSidebar(state, sidebar) {
-    this.stopChildApp('patient');
-    Radio.request('sidebar', 'close');
-
-    if (!sidebar) {
+  onChangeSidebar(state) {
+    if (state.get('expanded')) {
```
paulfalgout: I think you could first check if this `!isActionSidebar || isExpanded` and request sidebar close
Then if `isExpanded` stop the patient/empty return
Then if `isActionSidebar` show
start patient sidebar

basically to put when the sidebar to close in one bit of logic.

I don't see any situation where calling these things when not needed is an issue, so it should all be ok.

paulfalgout (changes requested):

luketlancaster:
@paulfalgout ok this _mostly_ works.

Only issue right now: it doesn't re-expand with the action sidebar open, it always opens back to the patient sidebar. Is that the behavior we want? Or the other way around? Or try to "remember" the last state?

**edit

turns out it does work if you open and close the action sidebar, then expand/shrink the view

**another edit

wait I guess it works as expected now? Not sure why I was seeing the issue I was seeing but I'm not anymore so... 🤘 ?

luketlancaster:
(also i still have tests to write for this, but wanted them to run out on the ☁️ first)

src/js/apps/forms/form/form_app.js
```diff
@@ -98,22 +93,30 @@ export default App.extend({
   onClickPrintButton() {
     Radio.request(`form${ this.form.id }`, 'print:form');
   },
-  onChangeSidebar(state, sidebar) {
-    this.stopChildApp('patient');
-    Radio.request('sidebar', 'close');
+  onChangeState(model) {
+    if (model.hasChanged('isExpanded') || model.hasChanged('isActionSidebar')) {
```
paulfalgout: rather than wrap the whole thing, let's make this a guard:
`if (!hasChanged('isExpanded.. &&!hasChange('isAction..) return;`

src/js/apps/forms/form/form_app.js
```diff
@@ -98,22 +93,30 @@ export default App.extend({
...
+      const isActionSidebar = model.get('isActionSidebar');
+      const isExpanded = model.get('isExpanded');
 
-    if (!sidebar) {
-      this.getRegion('sidebar').empty();
-      return;
-    }
+      this.startChildApp('patient', {
```
paulfalgout: I think this `startChildApp` should go after the `isExpanded`.  Really it can probably be the very last line, but the `if (isActionSidebar` wouldn't `return;`

src/js/views/forms/form/form_views.js
```diff
@@ -50,10 +51,10 @@ const ContextTrailView = View.extend({
     Radio.request('history', 'go:back');
   },
   templateContext() {
-    const isSidebarVisible = !!this.model.get('sidebar');
+    const isSidebarVisible = !this.model.get('isExpanded');
```
paulfalgout: let's replace `isSidebarVisible` with `isExpanded` and switch the logic in the template.  And replace the `expanded` below with `isExpanded` as well.

paulfalgout (changes requested):
Looks about done.

paulfalgout (approved):

PR #446, opened 2021-01-27
Change only the role of the assignee on flow assignment

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch23892]

Can I get 👀 on this PR and this https://app.clubhouse.io/roundingwell/story/23892/flow-owner-assignment-should-assign-flow-actions-too to make sure there's not a 4th version of this somewhat small change.

### DIFF
--- a/src/js/apps/patients/patient/flow/flow_app.js
+++ b/src/js/apps/patients/patient/flow/flow_app.js
@@ -32,7 +32,6 @@ export default SubRouterApp.extend({
   onStart({ currentRoute }, [flow], [actions]) {
     this.flow = flow;
     this.actions = actions;
-    this.currentClinician = Radio.request('bootstrap', 'currentUser');
 
     this.showChildView('contextTrail', new ContextTrailView({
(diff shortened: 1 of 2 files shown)


luketlancaster (approved):
so the rules (based on my read of this code):

1. Don't change the flow actions owner if you're setting the flow owner to a role
2. Don't change "done" flow action owners ever
3. Only change the flow action owner if the current flow action owner is a role and is the same role as the new flow clinician owner

is that what we're going for?

paulfalgout:
To specify #3 the action owner must be a role to change it.

But honestly I dunno.  I think that's correct, but then I thought the last 2 sets of rules were correct as well.. so what do I know?

luketlancaster:
yeah I mostly just wanted to write all that out so it was written all out and someone who knew the rules could eventually look at it

LisaManresa (approved):
If we're understanding the requirements correctly, this logic looks fine... I think 😬

PR #447, opened 2021-01-28
Schedule List View

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24238]

### DIFF
--- a/src/js/apps/patients/schedule/schedule_app.js
+++ b/src/js/apps/patients/schedule/schedule_app.js
@@ -9,7 +9,7 @@ import StateModel from './schedule_state';
 import FiltersApp from './filters_app';
 
 import DateFilterComponent from 'js/views/patients/shared/components/date-filter';
-import { LayoutView, ScheduleTitleView, TableHeaderView } from 'js/views/patients/schedule/schedule_views';
+import { LayoutView, ScheduleTitleView, TableHeaderView, ScheduleListView } from 'js/views/patients/schedule/schedule_views';
 
 export default App.extend({
   StateModel,
@@ -58,7 +58,12 @@ export default App.extend({
     return Radio.request('entities', 'fetch:actions:collection:groupByDate', { filter });
   },
   onStart(options, collection) {
-    this.collection = collection;
+    const collectionView = new ScheduleListView({
+      collection,
+      state: this.getState(),
+    });
+
+    this.showChildView('list', collectionView);
   },
   showScheduleTitle() {
     const filters = this.getState().get('filters');
--- a/src/js/views/patients/schedule/schedule.scss
+++ b/src/js/views/patients/schedule/schedule.scss
@@ -6,7 +6,3 @@
 .schedule__filter:not(:empty) {
   margin-right: 16px;
 }
-
-.schedule-list__header {
-  padding-left: 0;
-}
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -1,7 +1,10 @@
-import { View } from 'marionette';
+import Radio from 'backbone.radio';
+import { View, CollectionView } from 'marionette';
+import dayjs from 'dayjs';
 import hbs from 'handlebars-inline-precompile';
 
-import intl from 'js/i18n';
+import { alphaSort } from 'js/utils/sorting';
+import intl, { renderTemplate } from 'js/i18n';
 
 import 'sass/modules/list-pages.scss';
 import 'sass/modules/table-list.scss';
@@ -11,6 +14,8 @@ import PreloadRegion from 'js/regions/preload_region';
 import Tooltip from 'js/components/tooltip';
 
 import LayoutTemplate from './layout.hbs';
+
+import './schedule-list.scss';
 
 const LayoutView = View.extend({
   className: 'flex-region',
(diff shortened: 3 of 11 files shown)


src/js/views/patients/schedule/schedule_views.js
```diff
@@ -48,16 +51,150 @@ const ScheduleTitleView = View.extend({
...
+    </td>
+  `,
+});
+
+
+const ScheduleListView = CollectionView.extend({
+  tagName: 'table',
+  className: 'schedule-list__table',
+  childView: DayListView,
+  buildChildView(child, ChildViewClass) {
```
paulfalgout: viewOptions is probably better here

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -48,16 +51,150 @@ const ScheduleTitleView = View.extend({
...
+  },
+  regions: {
+    actions: {
+      el: '[data-actions-region]',
+      replaceElement: true,
+    },
+  },
+  childViewContainer: '[data-actions-region]',
+  viewComparator(viewA, viewB) {
+    // nullVal of 24 to ensure null due_time is last in list and due_time never exceeds 23:59:59
```
paulfalgout, line 160: interesting.  works for me

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -48,16 +51,150 @@ const ScheduleTitleView = View.extend({
...
+  templateContext() {
+    const date = dayjs(this.model.get('date'));
+    const today = dayjs();
+
+    return {
+      rowspan: this.collection.length,
+      isToday: date.isSame(today, 'year') && date.isSame(today, 'month') && today.date() === date.date(),
+    };
+  },
+  regions: {
```
paulfalgout: is this in use?

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -48,16 +51,150 @@ const ScheduleTitleView = View.extend({
...
+        <table class="w-100" data-actions-region></table>
+      </div>
+    </td>
+  `,
+  templateContext() {
+    const date = dayjs(this.model.get('date'));
+    const today = dayjs();
+
+    return {
+      rowspan: this.collection.length,
```
paulfalgout: not in use?

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -48,16 +51,150 @@ const ScheduleTitleView = View.extend({
...
+      isToday: date.isSame(today, 'year') && date.isSame(today, 'month') && today.date() === date.date(),
```
paulfalgout: `date.isSame(today, 'day')` should work here.  `isSame(..'month'` also checks the year

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -59,6 +59,13 @@ export default App.extend({
   },
   onStart(options, collection) {
     this.collection = collection;
+
+    const collectionView = new ScheduleListView({
+      collection: this.collection,
```
paulfalgout: is this the only use of `this.collection`?  Might not need to attach it to the app

src/js/views/patients/schedule/schedule.scss
```diff
@@ -7,6 +7,112 @@
   margin-right: 16px;
 }
 
+.schedule__filter-date {
```
paulfalgout: is this in use?

paulfalgout: There's also a case to be made that this could just be `..flex--end-row`

src/js/views/patients/schedule/schedule.scss
```diff
@@ -7,6 +7,112 @@
...
+  margin-left: auto;
+}
+
 .schedule-list__header {
```
paulfalgout: I think we should go ahead and make here down in a `schedule-list.scss` file.  They're highly related, but I think the block name being the filename is proving useful.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -48,16 +51,150 @@ const ScheduleTitleView = View.extend({
...
+    };
+  },
+  regions: {
+    actions: {
+      el: '[data-actions-region]',
+      replaceElement: true,
+    },
+  },
+  childViewContainer: '[data-actions-region]',
```
paulfalgout, line 158: historically we've used `.js-` selector for these, but they're kinda region-y..  I kinda like it.. more obvious in the templates. hmm

paulfalgout (changes requested):
looks pretty good.

src/js/views/patients/schedule/schedule_views.js
LisaManresa: It works, but the empty ones are sorted to the top of the list if we use ascending sort. This was a quick way to sort them to the bottom instead.

src/js/apps/patients/schedule/schedule_app.js
LisaManresa: It is now. I'll remove. We'll be using `this.collection` all over the place with bulk edit, though.

src/js/views/patients/schedule/schedule.scss
LisaManresa: Yeah this isn't needed at all now. Will remove

src/js/views/patients/schedule/schedule_views.js
LisaManresa: I kind of prefer the region reference as well. But happy to change it to fit the pattern if you prefer.

paulfalgout: let's switch to the region selector idea and refactor older ones if we see them

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -48,16 +53,144 @@ const ScheduleTitleView = View.extend({
...
+        <table class="w-100" data-actions-region></table>
+      </div>
+    </td>
   `,
+  templateContext() {
+    const date = dayjs(this.model.get('date'));
+    const today = dayjs();
+
+    return {
+      isToday: date.isSame(today, 'month') && date.isSame(today, 'day'),
```
paulfalgout: This is still redundant.  `dayjs('2021-02-01').isSame(dayjs('2021-01-01'), 'day') === false`.  `isSame(today, 'day')` checks year, month, and day.

paulfalgout (changes requested):
Looks good to me. Just one small thing and :shipit:

LisaManresa:
Tests were flakey because of the dynamic dates, so I made some modifications to make them more stable.

LisaManresa:
Actually, let's hold off on merging this for now. I'm going over a few design items with Mark and would like to add them to this PR

LisaManresa:
Ok, style changes applied.

paulfalgout (approved):

PR #450, opened 2021-01-29
Add Schedule and spacers to patient workflow left nav

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch24240]
- [ ] Add QA tasks

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -79,9 +79,16 @@ const patientsAppWorkflowsNav = new Backbone.Collection([
     eventArgs: ['owned-by'],
   },
   {
+    text: i18n.patientsApp.worklists.schedule,
+    event: 'worklist',
+    eventArgs: ['schedule'],
+    className: 'app-nav__spacer',
+  },
+  {
     text: i18n.patientsApp.worklists.sharedByRole,
     event: 'worklist',
     eventArgs: ['shared-by'],
+    className: 'app-nav__spacer',
   },
   {
     text: i18n.patientsApp.worklists.newPastDay,
--- a/src/js/views/globals/app-nav/app-nav.scss
+++ b/src/js/views/globals/app-nav/app-nav.scss
@@ -75,6 +75,10 @@ $menu-hover-color: #375467;
   }
 }
 
+.app-nav__spacer {
+  margin-bottom: 16px;
+}
+
 .app-nav__bottom-button {
   bottom: 0;
   padding: 16px;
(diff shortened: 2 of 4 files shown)


src/js/apps/globals/nav_app.js
```diff
@@ -78,11 +78,13 @@ const patientsAppWorkflowsNav = new Backbone.Collection([
     event: 'worklist',
     eventArgs: ['owned-by'],
   },
+  {},
```
luketlancaster: feels dirty but works

src/js/views/globals/app-nav/app-nav.scss
```diff
@@ -75,6 +75,10 @@ $menu-hover-color: #375467;
   }
 }
 
+.app-nav__spacer {
+  height: 8px;
```
luketlancaster: couldn't figure out the correct spacing from the marvel, but this looked right

src/js/apps/globals/nav_app.js
LisaManresa: Rather than add a new model just for spacers, we can add support for a `className` on the existing models.

`NavItemView` would need something like this:
```
  className() {
    const className = 'app-nav__link';

    if (this.model.get('className')) {
      return `${ className } ${ this.model.get('className') }`;
    }

    return className;
  },
```
That will allow us to use the standard spacing classnames.

paulfalgout: Yeah let's add something like `type: 'spacer'` and check for type where you're checking for text.  And then we the default can be the current links.. don't have to type the rest.

src/js/apps/globals/nav_app.js
```diff
@@ -100,6 +102,15 @@ const patientsAppWorkflowsNav = new Backbone.Collection([
   },
 ]);
 
+// /* istanbul ignore next */
+if (_DEVELOP_) {
```
paulfalgout: I think we can go ahead and remove develop.. looks like this'll be done soon

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -97,16 +97,20 @@ const NavItemView = View.extend({
...
-    role() {
-      const clinician = Radio.request('bootstrap', 'currentUser');
-      return clinician.getRole().get('name');
-    },
-  },
+});
+
+const SpacerNavItemView = View.extend({
+  className: 'app-nav__spacer',
+  tagName: 'div',
```
paulfalgout: div is the default value

paulfalgout (changes requested):

src/js/apps/globals/nav_app.js
paulfalgout: Ah yeah.  what @LisaManresa said.  Sounds good.

luketlancaster: yeah what @LisaManresa said it way better than what I did, I'll go with that

LisaManresa (approved):

PR #451, opened 2021-02-01
Fix nav event for schedule item

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch24413]
- [ ] Add QA tasks

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -80,8 +80,8 @@ const patientsAppWorkflowsNav = new Backbone.Collection([
   },
   {
     text: i18n.patientsApp.worklists.schedule,
-    event: 'worklist',
-    eventArgs: ['schedule'],
+    event: 'schedule',
+    eventArgs: [],
     className: 'app-nav__spacer',
   },
   {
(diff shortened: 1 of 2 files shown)


src/js/apps/globals/nav_app.js
```diff
@@ -80,8 +80,7 @@ const patientsAppWorkflowsNav = new Backbone.Collection([
   },
   {
     text: i18n.patientsApp.worklists.schedule,
-    event: 'worklist',
-    eventArgs: ['schedule'],
+    event: 'schedule',
```
paulfalgout, line 83: does this work?

paulfalgout (changes requested):

src/js/apps/globals/nav_app.js
luketlancaster: nope sure didn't! Pushed up a simple fix. Lemme know if you want me to make it so the items don't require `eventArgs` at all

paulfalgout: I also think that `this.startRoute('schedule');` is causing an error when clicking on the schedule nav when on the page.  This is because the schedule app isn't a subrouter so it should be `this.startCurrent('schedule')`

An empty array is ok with me for now

paulfalgout (approved):

PR #452, opened 2021-02-01
Feature/schedule list bulk edit

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24414] [ch24415]

### DIFF
--- a/src/js/apps/patients/schedule/schedule_app.js
+++ b/src/js/apps/patients/schedule/schedule_app.js
@@ -113,6 +113,9 @@ export default App.extend({
       },
     });
   },
+  onClickBulkCancel() {
+    this.getState().clearSelected();
+  },
   showScheduleList() {
     const collectionView = new ScheduleListView({
       collection: this.collection.groupByDate(),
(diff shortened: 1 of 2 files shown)


src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -94,6 +94,17 @@ export default App.extend({
           .done(() => {
             Radio.request('alert', 'show:success', renderTemplate(BulkEditActionsSuccessTemplate, { itemCount: this.selected.length }));
             app.stop();
+
+            if (saveData.due_date) {
+              const isDueDateChanged = this.selected.some(action => action.changed.due_date);
+
+              if (isDueDateChanged) {
+                this.getState().clearSelected();
```
paulfalgout: Does this line need to happen after `this.selected.some`?  I suspect not, in which case, this and L108 could just move up to L97

paulfalgout (changes requested):
mostly a question/nitpick

src/js/apps/patients/schedule/schedule_app.js
LisaManresa: Yeah, that's true. I thought the `some` was going to be multi-lined, so I separated them out. I'll consolidate.

paulfalgout: oh yeah the `if`s could be combined, but I was looking at `this.getState().clearSelected();` called here and then below.  it can probably happen before the `if`

paulfalgout (approved):

PR #453, opened 2021-02-01
Schedule List bug fixes and design tweaks

LisaManresa:
- [ ] Add the Clubhouse Story ID: [ch24436] [ch24399] [ch24416]

We may want to delay merging this until after I've had a chance to review the schedule page with Mark.


src/js/views/patients/schedule/schedule-list.scss
```diff
@@ -52,7 +52,7 @@
 }
 
 .schedule-list__day-list-row {
-  cursor: default;
+  cursor: pointer;
 
   &:hover {
     background: $black-90;
```
paulfalgout: I would assume the background on hover is blue and this grey when mult-selected

test/integration/patients/worklist/bulk-edit.js
```diff
@@ -1071,9 +1071,13 @@ context('Worklist bulk editing', function() {
 
     cy
       .get('.datepicker')
-      .find('.is-today')
-      .parent()
-      .prev()
+      .find('.datepicker__header .js-prev')
```
paulfalgout, line 1074: this makes sense to me.  Always a date in the past

paulfalgout (commented):
This is probably ok.  It won't mess with the disparity in browser timezone... but it's still a bit awkward.  The time is stored as a string and is explicitly relative to a timezone, so while this works because `cy.clock` is being set, it's a little bit round about.  I think if we're dealing with dates and not timestamp we should probably do as much as possible to keep the date stored as a string and only use `dayjs` as a last resort for some form of calculation, likely within a `formatDate` helper.  But I want to fix the broken tests and get schedule going, so we don't need to take this any further beyond final styling.

src/js/views/patients/schedule/schedule-list.scss
LisaManresa: Ooh, good point. I forgot about that feature of bulk edit. I'll confirm with Mark.

LisaManresa:
Ok, went over design tweaks with Mark. This is good to go. But I'll follow up on the test date issues.

PR #454, opened 2021-02-03
Fix date testing in worklists and schedule

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24480]

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -27,5 +27,8 @@
   "viewportHeight": 768,
   "chromeWebSecurity": false,
   "nodeVersion": "system",
-  "retries": 1
+  "retries": {
+    "runMode": 1,
+    "openMode": 0
+  }
 }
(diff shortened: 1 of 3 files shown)


LisaManresa:
Also added new `retries` settings. This may not quite be the route we want to go. Another option could be to use the `isCI` environment variable and conditionally set `Cypress.env('RETRIES')` in `defaults.js`. I have not tried that yet.

LisaManresa:
[ch24456]

cypress.json
```diff
@@ -27,5 +27,8 @@
   "viewportHeight": 768,
   "chromeWebSecurity": false,
   "nodeVersion": "system",
-  "retries": 1
+  "retries": {
+    "runMode": 3,
```
paulfalgout: yeah I don't want to increase the regular runMode above 1 because we're most likely just increasing the wait time of failing tests vs preventing flake.

paulfalgout (changes requested):

cypress.json
LisaManresa: Right... we discussed that. Sorry. Will keep investigating.

paulfalgout: This can be merged if this is `runMode: 1` and we could keep the other card open

paulfalgout (approved):

PR #455, opened 2021-02-04
Enable default_html option for option and phone widgets

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch24466]
- [ ] Add QA tasks

### DIFF
--- a/src/js/views/patients/patient/sidebar/sidebar_views.js
+++ b/src/js/views/patients/patient/sidebar/sidebar_views.js
@@ -105,12 +105,13 @@ const sidebarWidgets = {
   }),
   optionsWidget: {
     className: 'widgets-value',
-    template: hbs`{{ displayValue }}`,
+    template: hbs`{{ displayValue }}{{#unless displayValue}}{{{ defaultHtml }}}{{/unless}}`,
     templateContext() {
       const fields = this.model.getFields();
       const currentField = fields.find({ name: this.getOption('field_name') });
+      const defaultHtml = this.getOption('default_html');
 
-      if (!currentField) return;
+      if (!currentField) return { defaultHtml };
 
       const fieldValue = currentField.get('value');
       const key = this.getOption('key');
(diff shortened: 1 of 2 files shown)


LisaManresa (approved):

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -105,38 +105,44 @@ const sidebarWidgets = {
...
     className: 'widgets-value',
-    template: hbs`{{ displayValue }}`,
+    template: hbs`{{ displayValue }}{{#unless displayValue}}{{{ defaultHtml }}}{{/unless}}`,
     templateContext() {
       const fields = this.model.getFields();
       const currentField = fields.find({ name: this.getOption('field_name') });
+      const defaultHtml = this.getOption('default_html');
 
-      if (!currentField) return;
+      if (!currentField && !defaultHtml) return;
```
paulfalgout: I think this could be:
```js
if (!currentField) return { defaultHtml };
```
which would remove the need for the `? ... : null` below

paulfalgout (changes requested):

src/js/views/patients/patient/sidebar/sidebar_views.js
luketlancaster: I guess I need clarity on the rules/expectations here. If there's `default_html`, is the expectation that there won't be a field to use? Because in my tests the fields exist, but have empty values, so we display some html instead of the dashes (does that make sense?)

paulfalgout: yes that's the main goal.  That we'd allow for default html when the value of a field is empty.  But I think a null value or no current field are functionally the same for what we're doing.

luketlancaster: so here are my rules:

1. If there's a `currentField` that has a value, display it
2. If there's a `currentField` w/o a value and there is `defaultHtml`, display the `defaultHtml`
3. If there's a `currentField` w/o a value and **no** `defaultHtml`, return and display the `–`
4. If there's no `currentField`, return and display the `–`

given those, it seems like the `if () return` is going to get pretty complex right? Essentially moving a lot of the logic above it, just so I can have two returns?

paulfalgout: Which of 1-4 does this not account for?
```js
templateContext() {
      const fields = this.model.getFields();
      const currentField = fields.find({ name: this.getOption('field_name') });
      const defaultHtml = this.getOption('default_html');

      if (!currentField) return { defaultHtml };

      const fieldValue = currentField.get('value');
      const key = this.getOption('key');
      const displayOptions = this.getOption('display_options');

      const value = key ? propertyOf(fieldValue)(key.split('.')) : fieldValue;

      return {
        displayValue: displayOptions[value] || value,
        defaultHtml,
      };
    },
```

luketlancaster: oh hm. I guess for some reason I blanked on still needing `defaultHtml` in the second return 👍

PR #457, opened 2021-02-04
Add field widget

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch24467]
- [ ] Add QA tasks

### DIFF
--- a/src/js/views/patients/patient/sidebar/sidebar_views.js
+++ b/src/js/views/patients/patient/sidebar/sidebar_views.js
@@ -22,6 +22,20 @@ function buildWidget(widget, patient, widgetModel, options) {
   return extend({ model: patient }, options, widgetModel.get('definition'), widget);
 }
 
+function getFieldValue(fields, name, key) {
+  const currentField = fields.find({ name });
+
+  if (!currentField) return;
+
+  const fieldValue = currentField.get('value');
+
+  if (key) {
+    return propertyOf(fieldValue)(key.split('.'));
+  }
+
+  return fieldValue;
+}
+
 const sidebarWidgets = {
   dob: {
     template: hbs`{{formatHTMLMessage (intlGet "patients.patient.sidebar.sidebarViews.sidebarWidgets.dob") dob=(formatDateTime dob "LONG" inputFormat="YYYY-MM-DD") age=age}}`,
@@ -107,17 +121,9 @@ const sidebarWidgets = {
     className: 'widgets-value',
     template: hbs`{{ displayValue }}{{#unless displayValue}}{{{ defaultHtml }}}{{/unless}}`,
     templateContext() {
-      const fields = this.model.getFields();
-      const currentField = fields.find({ name: this.getOption('field_name') });
       const defaultHtml = this.getOption('default_html');
-
-      if (!currentField) return { defaultHtml };
-
-      const fieldValue = currentField.get('value');
-      const key = this.getOption('key');
+      const value = getFieldValue(this.model.getFields(), this.getOption('field_name'), this.getOption('key'));
       const displayOptions = this.getOption('display_options');
-
-      const value = key ? propertyOf(fieldValue)(key.split('.')) : fieldValue;
 
       return {
         displayValue: displayOptions[value] || value,
@@ -129,16 +135,11 @@ const sidebarWidgets = {
     className: 'widgets-value',
     template: hbs`{{ displayValue }}{{#unless displayValue}}{{{ defaultHtml }}}{{/unless}}`,
     templateContext() {
-      const fields = this.model.getFields();
-      const currentField = fields.find({ name: this.getOption('field_name') });
       const defaultHtml = this.getOption('default_html');
+      const value = getFieldValue(this.model.getFields(), this.getOption('field_name'), this.getOption('key'));
 
-      if (!currentField) return { defaultHtml };
+      if (!value) return { defaultHtml };
 
-      const fieldValue = currentField.get('value');
-      const key = this.getOption('key');
-
-      const value = key ? propertyOf(fieldValue)(key) : fieldValue;
       const phone = parsePhoneNumber(value, 'US');
 
       return {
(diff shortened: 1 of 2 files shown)


luketlancaster:
like this? Think the tests need to be a bit more...robust? Or this kinda what the use-case is gonna look like?

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -146,6 +146,27 @@ const sidebarWidgets = {
...
+      const fields = this.model.getFields();
+      const currentField = fields.find({ name: this.getOption('field_name') });
+      const defaultHtml = this.getOption('default_html');
+
+      if (!currentField) return { defaultHtml };
+
+      const fieldValue = currentField.get('value');
+      const key = this.getOption('key');
+
+      const displayValue = key ? propertyOf(fieldValue)(key) : fieldValue;
```
paulfalgout: Oh hmm I just noticed this should be `key.split('.')` and so should phone number.  It should support `some.deeply.nested.key` but without the split it's just a single level down.

I think it's time to DRY this a bit.

I think we should make a standalone function and call it like 
`const value = getFieldValue(this.model.getFields(), this.getOption('field_name'));`

Then we can do with `value` what we need.

paulfalgout (changes requested):
Yeah let's tested a nested key.  Also right now the nested tests are one level.  Can we update all nested key tests to be `foo.bar`

luketlancaster:
so this is gonna fail. But I wasn't sure what the intention for some of these test fields was. i.e. :
```
          addWidget({
            id: 'optionsWidget1',
            widget_type: 'optionsWidget',
            definition: {
              display_name: 'Populated Option Widget',
              field_name: 'test-field',
              display_options,
            },
          }),
```
the test says that it's expecting to display `Test Field`, but the value of `test-field` is `1`...so is the goal to take that value, and check the `display_options` for that key?

Also looked like `optionsWidget` had some more exceptions, so I pulled back from using `getFieldValue` for the time being

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -118,9 +129,10 @@ const sidebarWidgets = {
       const displayOptions = this.getOption('display_options');
 
       const value = key ? propertyOf(fieldValue)(key.split('.')) : fieldValue;
+      const opts = key ? propertyOf(displayOptions)(key.split('.')) : null;
```
paulfalgout: I think this widget is:
```js
templateContext() {
  const defaultHtml = this.getOption('default_html');
  const value = getFieldValue(this.model.getFields(), this.getOption('field_name'), this.getOption('key'));

  if (!value) return { defaultHtml };
  const displayOptions = this.getOption('display_options');

  return {
    displayValue: displayOptions[value] || value,
    defaultHtml,
  };
}
```

paulfalgout (changes requested):

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -22,6 +22,17 @@ function buildWidget(widget, patient, widgetModel, options) {
   return extend({ model: patient }, options, widgetModel.get('definition'), widget);
 }
 
+function getFieldValue(fields, name, key) {
+  const currentField = fields.find({ name });
+
+  if (!currentField) return;
+
+  const fieldValue = currentField.get('value');
+  const value = key ? propertyOf(fieldValue)(key.split('.')) : fieldValue;
```
paulfalgout: FWIW I think this can also be:
```js
if (key) {
  return propertyOf(fieldValue)(key.split('.'));
}
return fieldValue;
```

paulfalgout (changes requested):

src/js/views/patients/patient/sidebar/sidebar_views.js
luketlancaster: yeah I tried that and it didn't work with nested `displayOptions` keys IIRC. I'll give it another shot though

paulfalgout: Oh display options themselves shouldn't be nested. only the patient field values will be nested

paulfalgout: The problem with nesting and the optionsWidget was this: `const value = key ? propertyOf(fieldValue)(key) : fieldValue;`. where `key` could not be `deeply.nested.key` for the `fieldValue` because it was missing the `.split`

luketlancaster:
so the tests pass...but might be worth giving the cases another 👀 to make sure that I understood the correct use-cases. I.e can `display_options` be nested? Is the nesting that I did correct, etc

paulfalgout:
To be more clear display_options should always be:
```
{
  'simple_value': 'Simple Display',
  1: 'One',
}
```
But the `key` of a widget is used if the value of the patient field we're looking at is nested.  `key = 'foo'` means the value of the patient field we're using is nested one level deep ie: `{ foo: 1 }` and `foo.bar` means the value of is `{ foo: { bar: 1 }}` on the patient field..  so once we get that value, that value will either match a display option and in this case show "One" for either example above, or if the value was `NOTAKEYONDISPLAYOPTIONS` it would just show that value.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -147,6 +150,21 @@ const sidebarWidgets = {
...
     },
   },
+  fieldWidget: {
+    className: 'widgets-value',
+    template: hbs`{{ displayValue }}{{#unless displayValue}}{{{ defaultHtml }}}{{/unless}}`,
+    templateContext() {
+      const defaultHtml = this.getOption('default_html');
+      const displayValue = getFieldValue(this.model.getFields(), this.getOption('field_name'), this.getOption('key'));
+
+      if (!displayValue) return { defaultHtml };
```
paulfalgout: So we can remove this check.  It isn't doing anything in this widget.  We can probably go ahead and remove it everywhere and it'd be just fine.  Can you test that?

paulfalgout (changes requested):

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -129,24 +135,29 @@ const sidebarWidgets = {
...
-
-      if (!currentField) return { defaultHtml };
-
-      const fieldValue = currentField.get('value');
-      const key = this.getOption('key');
-
-      const value = key ? propertyOf(fieldValue)(key) : fieldValue;
-      const phone = parsePhoneNumber(value, 'US');
+      const value = getFieldValue(this.model.getFields(), this.getOption('field_name'), this.getOption('key'));
+      const phone = parsePhoneNumber(value || '', 'US');
```
luketlancaster: so `parsePhoneNumber` needs a string or else it gets real sad

paulfalgout: ok well let's keep the `if (!value) return..` for instances like this one

paulfalgout (approved):

PR #460, opened 2021-02-08
Worklist Find in List

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24468]

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -16,7 +16,8 @@ import BulkEditActionsApp from 'js/apps/patients/sidebar/bulk-edit-actions_app';
 import BulkEditFlowsApp from 'js/apps/patients/sidebar/bulk-edit-flows_app';
 
 import DateFilterComponent from 'js/views/patients/shared/components/date-filter';
-import { ListView, SelectAllView, LayoutView, ListTitleView, TableHeaderView, SortDroplist, sortCreatedOptions, sortDueOptions, sortUpdateOptions } from 'js/views/patients/worklist/worklist_views';
+
+import { ListView, SelectAllView, LayoutView, ListTitleView, TableHeaderView, SortDroplist, TypeToggleView, sortCreatedOptions, sortDueOptions, sortUpdateOptions } from 'js/views/patients/worklist/worklist_views';
 import { BulkEditButtonView, BulkEditFlowsSuccessTemplate, BulkEditActionsSuccessTemplate, BulkDeleteFlowsSuccessTemplate, BulkDeleteActionsSuccessTemplate } from 'js/views/patients/shared/bulk-edit/bulk-edit_views';
 
 export default App.extend({
@@ -32,7 +33,7 @@ export default App.extend({
     bulkEditFlows: BulkEditFlowsApp,
   },
   stateEvents: {
-    'change:filters change:actionsDateFilters change:flowsDateFilters': 'restart',
+    'change:type change:filters change:actionsDateFilters change:flowsDateFilters': 'restart',
     'change:actionsSortId': 'onChangeStateSort',
     'change:flowsSortId': 'onChangeStateSort',
     'change:selectedFlows': 'onChangeSelected',
@@ -57,6 +58,7 @@ export default App.extend({
   onBeforeStart({ worklistId }) {
     if (this.isRestarting()) {
       this.showListTitle();
+      this.showTypeToggleView();
       this.showSortDroplist();
       this.showTableHeaders();
       this.showDateFilter();
@@ -77,6 +79,7 @@ export default App.extend({
     this.showSortDroplist();
     this.showTableHeaders();
     this.showListTitle();
+    this.showTypeToggleView();
     this.showDateFilter();
     this.startFiltersApp();
   },
@@ -258,6 +261,17 @@ export default App.extend({
       worklistId: this.worklistId,
       isFlowList: this.getState().isFlowType(),
     }));
+  },
+  showTypeToggleView() {
+    const typeToggleView = new TypeToggleView({
+      isFlowList: this.getState('type') === 'flows',
+    });
+
+    this.listenTo(typeToggleView, 'toggle:listType', type => {
+      this.setState('type', type);
+    });
+
+    this.showChildView('toggle', typeToggleView);
   },
   showBulkEditButtonView() {
     const bulkEditButtonView = this.showChildView('filters', new BulkEditButtonView({
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -16,6 +16,7 @@ import BulkEditActionsApp from 'js/apps/patients/sidebar/bulk-edit-actions_app';
 import BulkEditFlowsApp from 'js/apps/patients/sidebar/bulk-edit-flows_app';
 
 import DateFilterComponent from 'js/views/patients/shared/components/date-filter';
+import SearchComponent from 'js/views/patients/shared/components/list-search';
 
 import { ListView, SelectAllView, LayoutView, ListTitleView, TableHeaderView, SortDroplist, TypeToggleView, sortCreatedOptions, sortDueOptions, sortUpdateOptions } from 'js/views/patients/worklist/worklist_views';
 import { BulkEditButtonView, BulkEditFlowsSuccessTemplate, BulkEditActionsSuccessTemplate, BulkDeleteFlowsSuccessTemplate, BulkDeleteActionsSuccessTemplate } from 'js/views/patients/shared/bulk-edit/bulk-edit_views';
@@ -80,6 +81,7 @@ export default App.extend({
     this.showTableHeaders();
     this.showListTitle();
     this.showTypeToggleView();
+    this.showSearchView();
     this.showDateFilter();
     this.startFiltersApp();
   },
@@ -98,6 +100,7 @@ export default App.extend({
 
     this.showChildView('list', collectionView);
 
+    this.showSearchView();
     this.toggleBulkSelect();
   },
   startFiltersApp() {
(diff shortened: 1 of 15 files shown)


src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -270,4 +298,11 @@ export default App.extend({
       'click:edit': this.onClickBulkEdit,
     });
   },
+  setSearchState(state, searchQuery) {
+    if (searchQuery.length === 1) return;
```
paulfalgout: I'm not sure I understand this line

src/js/views/patients/shared/components/list-search/list-search_views.js
```diff
@@ -0,0 +1,53 @@
...
+    const className = 'list-search__container';
+
+    if (this.model.get('isDisabled')) {
+      return `${ className } is-disabled`;
+    }
+
+    return className;
+  },
+  template: InputTemplate,
+  templateContext() {
```
paulfalgout, line 22: model attributes are already available to the template, this context isn't needed

src/js/views/patients/shared/components/list-search/index.js
```diff
@@ -0,0 +1,38 @@
...
+  viewOptions() {
+    return {
+      model: this.getState(),
+    };
+  },
+  viewEvents: {
+    'watch:change': 'onViewWatchChange',
+    'clear': 'onViewClear',
+  },
+  stateEvents: {
```
paulfalgout: I don't think this is doing anything

src/js/views/patients/shared/components/list-search/index.js
```diff
@@ -0,0 +1,38 @@
...
+    'change:query': 'onChangeQuery',
+  },
+  onViewClear() {
+    this.setState('query', '');
+  },
+  onViewWatchChange(query) {
+    this.setState('query', query);
+  },
+  initialize({ state }) {
```
paulfalgout: no need to add this, it is the default behavior

src/js/views/patients/shared/components/list-search/list-search_views.js
```diff
@@ -0,0 +1,53 @@
...
+  <span class="list-search__search-icon">{{far "search"}}</span>
+  <input class="list-search__input input-primary--small js-input w-100" type="text" {{#if isDisabled}}disabled{{/if}} placeholder="{{@intl.patients.shared.components.listSearch.listSearchViews.placeholder}}"/>
+  <span class="list-search__clear-icon js-clear is-hidden">{{fas "times-circle"}}</span>
+`;
+
+const SearchView = View.extend({
+  behaviors: {
+    InputWatcherBehavior,
+  },
+  className() {
```
paulfalgout: className is not updated on render, so we might want to use `onRender` with `toggleClass`

src/js/views/patients/shared/components/list-search/index.js
```diff
@@ -0,0 +1,38 @@
...
+    'watch:change': 'onViewWatchChange',
+    'clear': 'onViewClear',
+  },
+  stateEvents: {
+    'change:query': 'onChangeQuery',
+  },
+  onViewClear() {
+    this.setState('query', '');
+  },
+  onViewWatchChange(query) {
```
paulfalgout, line 24: does `onClear`'s `this.ui.input.val('');` trigger this? if so the `onViewClear` is redundant

src/js/views/patients/shared/components/list-search/list-search-component.scss
```diff
@@ -0,0 +1,41 @@
...
+  &.is-disabled {
+    .icon {
+      opacity: 0.5;
+    }
+  }
+}
+
+.list-search__search-icon {
+  left: 16px;
+  line-height: 25px;
```
paulfalgout: 25?

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -83,6 +86,33 @@ const SelectAllView = View.extend({
...
+  triggers: {
+    'click .js-toggle-actions': 'click:toggleActions',
+    'click .js-toggle-flows': 'click:toggleFlows',
+  },
+  ui: {
+    buttons: 'button',
+  },
+  onClickToggleActions() {
+    this.triggerMethod('toggle:listType', 'actions');
+    // this.ui.buttons.toggleClass('button--blue');
```
paulfalgout: can remove

test/integration/patients/worklist/worklist.js
```diff
@@ -1996,6 +1999,151 @@ context('worklist page', function() {
...
+      .wait('@routeFlows');
+
+    cy
+      .get('.list-page__header')
+      .find('[data-search-region] .js-input')
+      .as('listSearch')
+      .should('be.disabled')
+      .should('have.attr', 'placeholder', 'Find in List...');
+
+    // Cypress moves just a little too fast for the transition to enabled search
```
paulfalgout: hmm this is funny.  I'm not sure why the above is disabled at all as it occurs after the wait for the endpoint.  I don't think we should need this wait and it likely indicates an issue elsewhere

src/js/views/patients/shared/components/list-search/index.js
```diff
@@ -0,0 +1,38 @@
...
+    query: '',
+  },
+});
+
+export default Component.extend({
+  StateModel,
+  ViewClass: SearchView,
+  viewOptions() {
+    return {
+      model: this.getState(),
```
paulfalgout: So the state attributes are added to the view already, so if we don't expect the view to react to state changes then the view should probably be use those https://github.com/RoundingWellOS/marionette.toolkit/blob/master/src/component.js#L214

src/js/views/patients/shared/components/list-search/list-search_views.js
paulfalgout: But as said in the component, maybe we shouldn't be using the model here anyhow

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -95,6 +101,7 @@ export default App.extend({
 
     this.showChildView('list', collectionView);
 
+    this.setState('isSearchDisabled', !collection || !collection.length);
```
paulfalgout: I'm not sure this is a state attribute that is necessary.  I think `showSearchView` can be called here and it can just use `isDisabled = !this.collection || !this.collection.length`

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: oh is this to prevent filtering unless there's a number of characters?  I think it's 3 or more would filter.  Also I think it needs to be `setState({ searchQuery ? searchQuery.length > 2 : searchQuery : '' }`. because it needs to un-search when backspacing.

paulfalgout (changes requested):

src/js/apps/patients/worklist/worklist_app.js
LisaManresa: Ah, I thought it was 2 or more characters, hence the `=== 1` to handle the setting of the state when the query is empty. I'll adjust this.

test/integration/patients/worklist/worklist.js
LisaManresa: I was wondering about this. It does seem to smell a bit.

src/js/views/patients/shared/components/list-search/list-search-component.scss
LisaManresa: Yeah... This is probably fine with 24px, but the circle of the magnifying glass was so nicely centered with the circle of the border at 25px. I'm pretty certain 24px will have the whole unit vertically centered though, so I'll change it.

src/js/views/patients/shared/components/list-search/index.js
LisaManresa: Manually changing the value doesn't trigger the `watch:change` event.

test/integration/patients/worklist/worklist.js
LisaManresa: The flakiness of this test was really odd. But I replaced the wait with an explicit check for `.js-input:not([disabled])` so that it would wait for that element to exist.

src/js/views/patients/shared/components/list-search/list-search-component.scss
paulfalgout: hmm yeah I wonder if it turns out mark wants that adjustment or something if using the absolute positioning top would solve it with a less flag raising number, probably not though.  if we did end up doing something non-4px we should add a comment as to why, but hopefully a 4px works for mark.  we can see if it comes up in review.

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -40,7 +41,7 @@ export default Backbone.Model.extend({
     this.on('change', this.onChange);
   },
   onChange() {
-    store.set(`${ this.id }_${ this.currentClinician.id }`, this.attributes);
+    store.set(`${ this.id }_${ this.currentClinician.id }`, omit(this.attributes, 'searchQuery', 'isSearchDisabled'));
```
paulfalgout: can remove isSearchDisabled

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -32,11 +34,12 @@ export default App.extend({
     bulkEditFlows: BulkEditFlowsApp,
   },
   stateEvents: {
-    'change:filters change:actionsDateFilters change:flowsDateFilters': 'restart',
+    'change:type change:filters change:actionsDateFilters change:flowsDateFilters': 'restart',
     'change:actionsSortId': 'onChangeStateSort',
     'change:flowsSortId': 'onChangeStateSort',
     'change:selectedFlows': 'onChangeSelected',
     'change:selectedActions': 'onChangeSelected',
+    'change:isSearchDisabled': 'showSearchView',
```
paulfalgout: can remove

paulfalgout (changes requested):
one bit of clean up.  Then maybe worth walking Mark through this.

PR #461, opened 2021-02-08
Form update bar on submission and form communication refactor

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch24498]

So there are certainly things about the WIP commit that will change from here.. so all of that with a grain of salt.

But particularly
https://github.com/RoundingWell/care-ops-frontend/commit/db918b1fd158a06d6d29785830c904d73de6d8c8
https://github.com/RoundingWell/care-ops-frontend/commit/96766dc94b668535254356bd615e4b4958df3f68
frame the intention of the change.

The previous form architecture relied on storing the state of the form in the iframe's url, but this shifts it to a state service.  The current form child app structure isn't setup to update the service, so this would currently require some significant changes, particularly to the history app... so there's possibly the need to either update that which I'm investigating or to pull back some of the changes.


paulfalgout:
@luketlancaster @LisaManresa lots of 👀 here when possible.  iframe testing now has to be done within the iframe, which works.. even on firefox, though no captures are made, so you only see the iframe while the tests are running, but no snapshots afterwards.

LisaManresa (approved):
Pulled it down and looked through it locally. Looking really good.

PR #462, opened 2021-02-09
Worklist find in list improvements

LisaManresa:
- [ ] Add the Clubhouse Story ID: [ch24509]

Adds the find in list-specific empty view.

I also added a proposal for improving searching of owner roles and states. Since those are abbreviated, searching for something like "Nurse" returned no results.

### DIFF
--- a/src/js/views/patients/worklist/worklist-list.scss
+++ b/src/js/views/patients/worklist/worklist-list.scss
@@ -116,3 +116,6 @@
   vertical-align: middle;
 }
 
+.worklist-list__search-helper {
+  display: none;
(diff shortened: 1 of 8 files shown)


src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -116,3 +116,6 @@
   vertical-align: middle;
 }
 
+.worklist-list__search-helper {
```
paulfalgout, line 119: :+1:

paulfalgout (approved):

PR #465, opened 2021-02-11
Support select all with find in list 

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch24542]
- [x] Needs tests
Replaces https://github.com/RoundingWell/care-ops-frontend/pull/464

I couldn't quite put my finger on revision requests.  The end result is relatively the same, with the exception of modifying selected to be built from the visible rows rather than comparing the visible to the selected later.  If you want to take this over for tests and anything I missed @LisaManresa ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
       "faInfoCircle",
       "faKeyboard",
       "faLifeRing",
+      "faMinusSquare",
       "faSignOutAlt",
       "faTimes",
       "faTimesCircle",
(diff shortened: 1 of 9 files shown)


LisaManresa:
I'll pull this and double check tests.

LisaManresa:
Mark requested that we have a partially selected state for the select all checkbox when some items are selected. Added that in one commit and added updated tests.

paulfalgout:
I need to talk with Mark about that, I don't quite see the point

LisaManresa (approved):

PR #466, opened 2021-02-11
Feature/schedule find in list

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24515]

### DIFF
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -92,31 +92,33 @@ const DayItemView = View.extend({
     <td class="schedule-list__action-list-cell schedule-list__due-time {{#if isOverdue}}is-overdue{{/if}}">
       <button class="button--checkbox u-margin--r-8 js-select">{{#if isSelected}}{{fas "check-square"}}{{else}}{{fal "square"}}{{/if}}</button>
       {{#if due_time}}
-        {{formatDateTime due_time "TIME" inputFormat="HH:mm:ss"}}
+        {{formatDateTime due_time "TIME" inputFormat="HH:mm:ss"}}&#8203;
       {{else}}
-        <span class="schedule-list__no-time">{{ @intl.patients.schedule.scheduleViews.dayItemView.noTime }}</span>
+        <span class="schedule-list__no-time">{{ @intl.patients.schedule.scheduleViews.dayItemView.noTime }}</span>&#8203;
       {{/if}}
     </td>
     <td class="schedule-list__action-list-cell schedule-list__patient">
       <div class="schedule-list__state-patient">
-        <span class="schedule-list__action-state action--{{ stateOptions.color }}">{{fa stateOptions.iconType stateOptions.icon}}</span>{{~ remove_whitespace ~}}
-        <span class="schedule-list__patient-name js-patient">{{ patient.first_name }} {{ patient.last_name }}</span>
+        <span class="schedule-list__action-state action--{{ stateOptions.color }}">{{fa stateOptions.iconType stateOptions.icon}}</span><span class="schedule-list__search-helper">{{ state }}</span>&#8203;{{~ remove_whitespace ~}}
+        <span class="schedule-list__patient-name js-patient">{{ patient.first_name }} {{ patient.last_name }}</span>&#8203;
       </div>
     </td>
     <td class="schedule-list__action-list-cell schedule-list__action">
-      <span class="schedule-list__action-name js-action">{{ name }}</span>
+      <span class="schedule-list__action-name js-action">{{ name }}</span>&#8203;
     </td>
     <td class="schedule-list__action-list-cell schedule-list__action-form">
       {{#if form}}<span class="js-form schedule-list__action-form-icon">{{far "poll-h"}}</span>{{/if}}
     </td>
   `,
   templateContext() {
+    const state = this.model.getState();
+
     return {
       isOverdue: this.model.isOverdue(),
-      stateOptions: this.model.getState().get('options'),
+      state: state.get('name'),
+      stateOptions: state.get('options'),
       patient: this.model.getPatient().attributes,
       form: this.model.getForm(),
-      owner: this.model.getOwner().attributes,
       isSelected: this.state.isSelected(this.model),
     };
   },
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -1,9 +1,11 @@
+import { debounce, every, map } from 'underscore';
 import Radio from 'backbone.radio';
 import { View, CollectionView } from 'marionette';
 import dayjs from 'dayjs';
 import hbs from 'handlebars-inline-precompile';
 
 import { alphaSort } from 'js/utils/sorting';
+import words from 'js/utils/formatting/words';
 import intl, { renderTemplate } from 'js/i18n';
 
 import 'sass/modules/list-pages.scss';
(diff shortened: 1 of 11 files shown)


src/js/views/patients/schedule/schedule_views.js
```diff
@@ -230,16 +251,62 @@ const ScheduleListView = CollectionView.extend({
...
+      this.children.each(cv => {
+        cv.removeFilter();
+      });
+      this.setVisibleChildren();
+      return;
+    }
+
+    const matchers = this._buildMatchers(searchQuery);
+
+    this.children.each(cv => {
```
LisaManresa: Just a quick note about this: Putting this logic on the `DayListView` causes this filter to execute after the `ScheduleListView` filter. We need to ensure this filter happens first so the `ScheduleListView` filter can determine whether or not to hide the day row.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -207,6 +212,13 @@ const DayListView = CollectionView.extend({
   initialize({ state }) {
     this.state = state;
   },
+  childViewTriggers: {
+    'render': 'listItem:render',
+  },
+  onListItemRender(view) {
+    const date = dayjs(this.model.get('date'));
+    view.searchString = `${ date.format('D') }&#8203;${ date.format('MMM, ddd') }&#8203;${ view.$el.text() }`;
```
paulfalgout: can just use regular spaces here.  just need the `&#8203;` so that `.text()` doesn't squish two text areas together with no breaking character.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -230,16 +251,62 @@ const ScheduleListView = CollectionView.extend({
...
   initialize({ state }) {
     this.state = state;
 
-    this.listenTo(state, {
-      'select:all': this.render,
-      'select:none': this.render,
+    this.listenTo(state, 'change:searchQuery', this.searchList);
+  },
+  searchList(state, searchQuery) {
+    // Must remove the parent view's filter so we have access to previously filtered children
```
paulfalgout: I think we should pass the state down to the children so they can watch for the searchQuery change so each day works just like the regular worklist / matcher / setFilter
In this collectionview I think we should always use
```js
viewFilter(view) {
  return !view.isEmpty();
}
```
Then we can do `childViewEvents` in this view  with `'filter': `onChildFilter` where `onChildFilter` calls `this.filter()`.  Then in the `initialize` we should debounce the `onChildFilter` to something like 60.
Then what's in `setVisibleChildren` should be in an event like `onRenderChildren`

This way if things are added/removed from these collections for other reasons they'll follow the same logic.

paulfalgout (changes requested):

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -206,6 +211,39 @@ const DayListView = CollectionView.extend({
...
+    'render': 'listItem:render',
+  },
+  onListItemRender(view) {
+    const date = dayjs(this.model.get('date'));
+    view.searchString = `${ date.format('D') } ${ date.format('MMM, ddd') } ${ view.$el.text() }`;
+  },
+  searchList(state, searchQuery) {
+    if (!searchQuery) {
+      this.removeFilter();
+      this.triggerMethod('filter');
```
LisaManresa: As far as I could tell, `removeFilter` does not emit any events, but we need to emit one in order for the parent collection view to know that the searchQuery has been cleared. There might be a better way to do this though.

paulfalgout: Ah yeah.  Use `render:children` instead of `filter`. The `filter` event only occurs if views are filtered, but if there is no `filter` it doesn't attempt to filter.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -230,17 +277,44 @@ const ScheduleListView = CollectionView.extend({
       state: this.state,
     };
   },
-  emptyView: EmptyView,
+  childViewEvents: {
+    'filter'() {
```
paulfalgout: this can just be `'filter': 'debounceChildFilter'`.. err I guess a `render:children` version

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -230,17 +277,44 @@ const ScheduleListView = CollectionView.extend({
...
+    }
+
+    return EmptyView;
+  },
   viewComparator(viewA, viewB) {
     return alphaSort('asc', viewA.model.get('date'), viewB.model.get('date'));
   },
+  viewFilter(view) {
+    if (this.state.get('searchQuery')) {
+      return !view.children.isEmpty();
```
paulfalgout: I think just `view.isEmpty()` is ideal.  This allows the collectionview to decide when it is empty, if for some reason we inject some other view in that collectionview.

paulfalgout (changes requested):

src/js/views/patients/schedule/schedule_views.js
LisaManresa: We can't call `view.isEmpty()` because that view is never empty since the action item children are rendered into a container within that view. So we need to check the children.

However, there seems to be an order of operations issue here. When the list first renders `viewFilter` is called before any of the children are rendered. So `view.children.isEmpty()` is always true and the list renders empty. Which is the reason for the `searchQuery` check.

LisaManresa:
Ok, I think I have this sorted. I did add a minor modification to routeFlows and routeActions to pass an optional delay to them. I was seeing lots of test flake because those responses were coming back too quickly to test that the search was disabled while the list was loading. But, fwiw, it never seemed to flake in CI.

test/integration/patients/worklist/schedule.js
```diff
@@ -481,7 +481,6 @@ context('schedule page', function() {
 
     cy
       .get('[data-select-all-region]')
-      .find('button')
```
LisaManresa, line 484: The select all icon gets re-rendered frequently now and it was causing a lot of flake in these tests because the element was removed from the dom. Hence this update to just click on the parent.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -206,6 +211,39 @@ const DayListView = CollectionView.extend({
...
+    'render': 'listItem:render',
+  },
+  onListItemRender(view) {
+    const date = dayjs(this.model.get('date'));
+    view.searchString = `${ date.format('D') } ${ date.format('MMM, ddd') } ${ view.$el.text() }`;
+  },
+  searchList(state, searchQuery) {
+    if (!searchQuery) {
+      this.removeFilter();
+      this.triggerMethod('render:children');
```
paulfalgout: I cannot see a need for this trigger.  `removeFilter` does `setFilter(null)`  and as long as the `viewFilter` isn't already `null` then it will call `filter` which will always trigger `render:children`

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -230,17 +277,41 @@ const ScheduleListView = CollectionView.extend({
       state: this.state,
     };
   },
-  emptyView: EmptyView,
+  childViewEvents: {
+    'filter': 'onChildFilter',
+    'render:children': 'onChildFilter',
```
paulfalgout, line 280: When a collection view is `.filter()`d either manually or by a change of its filter it _may_ filter but it will always `render:children`.  This `filter` event should never be necessary.

test/integration/patients/worklist/schedule.js
paulfalgout: yeah I bet this is our 10ms debounce in action.

paulfalgout (changes requested):
Two small changes. I pulled this down and it tests pass without the two lines suggested.

src/js/views/patients/schedule/schedule_views.js
LisaManresa: Let me look into this. Something is up when we clear the search query. And the filter never fires. Not sure how tests were passing for you. But I think I can find a fix for this.

PR #471, opened 2021-02-16
Shared By Worklist no owner toggle

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24502]

### DIFF
--- a/src/js/apps/patients/worklist/filters_app.js
+++ b/src/js/apps/patients/worklist/filters_app.js
@@ -5,12 +5,13 @@ import intl from 'js/i18n';
 import App from 'js/base/app';
 
 import OwnerDroplist from 'js/views/patients/shared/components/owner_component';
-import { FiltersView, GroupsDropList } from 'js/views/patients/worklist/filters_views';
+import { FiltersView, GroupsDropList, NoOwnerToggleView } from 'js/views/patients/worklist/filters_views';
 
 export default App.extend({
-  onStart({ shouldShowClinician, shouldShowRole }) {
+  onStart({ shouldShowClinician, shouldShowRole, shouldShowOwnerToggle }) {
     this.shouldShowClinician = shouldShowClinician;
     this.shouldShowRole = shouldShowRole;
+    this.shouldShowOwnerToggle = shouldShowOwnerToggle;
     this.currentClinician = Radio.request('bootstrap', 'currentUser');
     this.groups = this.currentClinician.getGroups();
 
@@ -20,6 +21,7 @@ export default App.extend({
   showFilters() {
     this.showGroupsFilterView();
     this.showOwnerFilterView();
+    this.showOwnerToggleView();
   },
   showGroupsFilterView() {
     if (this.groups.length < 2) return;
@@ -61,6 +63,17 @@ export default App.extend({
     });
 
     this.showChildView('owner', ownerFilter);
+  },
+  showOwnerToggleView() {
+    if (!this.shouldShowOwnerToggle) return;
+
+    const ownerView = this.showChildView('ownerToggle', new NoOwnerToggleView({
+      model: this.getState(),
+    }));
+
+    this.listenTo(ownerView, 'click', () => {
+      this.toggleState('noOwner');
+    });
   },
   _getGroups() {
     const groups = this.groups.clone();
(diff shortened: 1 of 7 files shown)


src/js/apps/patients/worklist/filters_app.js
```diff
@@ -62,6 +64,22 @@ export default App.extend({
...
+
+    if (this.getState('noOwner')) {
+      ownerView = this.showChildView('ownerToggle', new NoOwnerActiveView());
+    } else {
+      ownerView = this.showChildView('ownerToggle', new NoOwnerInactiveView());
+    }
+
+    this.listenTo(ownerView, 'click', () => {
+      this.setState('noOwner', !this.getState('noOwner'));
+      this.showOwnerToggleView();
```
paulfalgout: I think this could be one view that gets passed `model: this.getState()` and maybe just to simplify things puts the button / classes in the template.  You could use `getTemplate` if it's cleaner than template logic and the view can render for `modelEvents` `change:noOwner`

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -128,10 +129,14 @@ export default Backbone.Model.extend({
       },
     };
 
-    if (this.id === 'shared-by' || !clinicianId) {
-      filters[this.id].role = roleId;
+    if (noOwner) {
```
paulfalgout: I don't think this is correct.  We still want to filter by the role, we just want to add the clinician filter.  so this `if (noOwner)` should be inside the `if(this.id === 'shared-by' ||...` such that role and clinician are both set.

paulfalgout (changes requested):

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -62,6 +64,17 @@ export default App.extend({
...
   },
+  showOwnerToggleView() {
+    if (!this.shouldShowOwnerToggle) return;
+
+    const ownerView = this.showChildView('ownerToggle', new NoOwnerToggleView({
+      model: this.getState(),
+    }));
+
+    this.listenTo(ownerView, 'click', () => {
+      this.setState('noOwner', !this.getState('noOwner'));
```
paulfalgout: this.toggleState('noOwner')

test/integration/patients/worklist/worklist.js
```diff
@@ -1114,6 +1114,41 @@ context('worklist page', function() {
...
+    cy
+      .get('.app-frame__nav')
+      .find('.app-nav__link')
+      .contains('Shared By Role')
+      .click()
+      .wait('@routeFlows');
+
+    cy
+      .get('[data-owner-toggle-region]')
+      .find('.worklist-filter__no-owner')
```
paulfalgout: can we use `contains('No Owner')` instead of a style class here and below?

src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -119,3 +119,14 @@
 .worklist-list__search-helper {
   display: none;
 }
+
+.worklist-filter__no-owner {
```
paulfalgout: I'm having second thoughts about this class.  I don't want this button to break in the future if `button-primary` changes.  Let's make a new `button-filter-toggle` class that only extends `.button` in the buttons module and we can copy any other needed styles.  We can also go ahead and style the icon there like we do in button filter with the `.icon.fa-` so it's specific.  Then we're just toggling the `button-blue` class

paulfalgout (changes requested):

src/sass/modules/buttons.scss
```diff
@@ -126,11 +153,13 @@
 
   &:hover:enabled {
     background-color: color(blue, light);
+    color: $white;
```
LisaManresa, line 156: A little redundant, but needed in order to override the `.button-filter-toggle` and this seemed like a cleaner option than nesting `&.button-blue` inside `.button-filter-toggle`

paulfalgout (approved):
This looks good, but I'm gonna sit on the merge until it's releasable with BE unless we see a possible merge conflict.

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -130,6 +131,10 @@ export default Backbone.Model.extend({
 
     if (this.id === 'shared-by' || !clinicianId) {
       filters[this.id].role = roleId;
+
+      if (noOwner) {
+        filters[this.id].clinician = '00000000-0000-0000-0000-000000000000';
```
paulfalgout: I just ran across this https://www.npmjs.com/package/uuid#uuidnil. We should use it

paulfalgout (changes requested):

paulfalgout (approved):

PR #472, opened 2021-02-16
Show details icon and tooltip on schedule and worklists

luketlancaster:
- [x] Add the Clubhouse Story ID: [ch24576]

### DIFF
--- a/src/js/views/patients/schedule/schedule-list.scss
+++ b/src/js/views/patients/schedule/schedule-list.scss
@@ -14,11 +14,12 @@
 }
 
 .schedule-list__header-action {
-  width: 55%;
+  width: calc(68% - 84px);
 }
 
 .schedule-list__header-form {
-  width: 5%;
+  padding-left: 32px;
+  width: 84px;
 }
 
 .schedule-list__table {
@@ -99,11 +100,19 @@
 }
 
 .schedule-list__action {
-  width: 63%;
+  width: calc(68% - 84px);
+}
+
+.schedule-list__action-details {
+  color: #999;
+  padding: 0 16px 0 0;
+  text-align: center;
+  width: 32px;
 }
 
 .schedule-list__action-form {
-  width: 5%;
+  padding: 0;
+  width: 48px;
 }
 
 .schedule-list__state-patient,
@@ -150,6 +159,11 @@
 
 .action-tooltip__action-name {
   display: table-cell;
+}
+
+.action-tooltip__action-details {
+  font-size: 12px;
+  margin: 8px 0 0 20px;
 }
 
 .schedule-list__patient-name,
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -6,7 +6,7 @@ import hbs from 'handlebars-inline-precompile';
 
 import { alphaSort } from 'js/utils/sorting';
 import words from 'js/utils/formatting/words';
-import intl, { renderTemplate } from 'js/i18n';
+import intl from 'js/i18n';
 
 import 'sass/modules/list-pages.scss';
 import 'sass/modules/table-list.scss';
@@ -14,6 +14,8 @@ import 'sass/modules/table-list.scss';
 import PreloadRegion from 'js/regions/preload_region';
 
 import Tooltip from 'js/components/tooltip';
+
+import { DetailsTooltip } from 'js/views/patients/shared/actions_views';
 
 import LayoutTemplate from './layout.hbs';
 
@@ -103,10 +105,14 @@ const DayItemView = View.extend({
       <span class="schedule-list__action-name js-action">{{ name }}</span>&#8203;{{~ remove_whitespace ~}}
       <span class="schedule-list__search-helper">{{ flow }}</span>&#8203;{{~ remove_whitespace ~}}
     </td>
+    <td class="schedule-list__action-list-cell schedule-list__action-details" data-details-region></td>
     <td class="schedule-list__action-list-cell schedule-list__action-form">
       {{#if form}}<span class="js-form schedule-list__action-form-icon">{{far "poll-h"}}</span>{{/if}}
     </td>
   `,
+  regions: {
+    details: '[data-details-region]',
+  },
   templateContext() {
     const state = this.model.getState();
 
@@ -139,22 +145,7 @@ const DayItemView = View.extend({
     });
   },
   onRender() {
-    const template = hbs`
-      {{#if flowName}}<p class="action-tooltip__flow"><span class="action-tooltip__flow-icon">{{fas "folder"}}</span>{{ flowName }}</p>{{/if}}
-      <p><span class="action-tooltip__action-icon">{{far "file-alt"}}</span><span class="action-tooltip__action-name">{{ name }}</span></p>
-    `;
-
-    this.$el.toggleClass('is-selected', this.state.isSelected(this.model));
-
-    new Tooltip({
-      className: 'tooltip tooltip--wide',
-      messageHtml: renderTemplate(template, {
-        name: this.model.get('name'),
-        flowName: this.flow ? this.flow.get('name') : null,
-      }),
-      uiView: this,
-      ui: this.ui.actionName,
-    });
+    this.showDetailsTooltip();
   },
   onClickSelect() {
     this.state.toggleSelected(this.model, !this.state.isSelected(this.model));
@@ -173,6 +164,11 @@ const DayItemView = View.extend({
     }
 
     Radio.trigger('event-router', 'patient:action', this.model.get('_patient'), this.model.id);
+  },
+  showDetailsTooltip() {
+    if (!this.model.get('details') && !this.model.getFlow()) return;
+
+    this.showChildView('details', new DetailsTooltip({ model: this.model }));
   },
 });
 
--- a/src/js/views/patients/shared/actions_views.js
+++ b/src/js/views/patients/shared/actions_views.js
@@ -2,7 +2,13 @@ import Radio from 'backbone.radio';
 import hbs from 'handlebars-inline-precompile';
 import { View } from 'marionette';
 
+import { renderTemplate } from 'js/i18n';
+
 import 'sass/modules/buttons.scss';
+
+import Tooltip from 'js/components/tooltip';
+
+import trim from 'js/utils/formatting/trim';
 
 import StateComponent from './components/state_component';
 import OwnerComponent from './components/owner_component';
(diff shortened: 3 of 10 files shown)


src/js/views/patients/schedule/schedule-list.scss
```diff
@@ -102,6 +102,12 @@
   width: 63%;
 }
 
+.schedule-list__action-details {
+  color: $black-60;
+  padding: 0;
+  width: 16px;
```
paulfalgout: So _probably_ the form column should also be a px setting and the action column should be `calc(68% -32px)`.  And the headings would also need to accommodate for that.  I'm not sure what adding 16px is doing without any percentage adjustments, but something somewhere ain't quite right.

src/js/views/patients/shared/actions_views.js
```diff
@@ -22,11 +24,26 @@ const FormButton = View.extend({
...
+const DetailsTooltip = View.extend({
+  template: hbs`<span class="js-details">{{far "info-circle"}}</span>`,
+  ui: {
+    tooltip: '.js-details',
+  },
+  onRender() {
+    new Tooltip({
+      message: this.model.get('details'),
+      uiView: this,
+      ui: this.ui.tooltip,
```
paulfalgout: So I think this could probably just be `this.$el` and then no wrapping `<span>` / `ui` is necessary

src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -24,6 +24,12 @@
   width: 16px;
 }
 
+.worklist-list__details {
+  color: $black-60;
+  padding: 16px 0;
+  width: 16px;
```
paulfalgout, line 32: Same here with injecting a fixed width inside the percentage columns.    It's possible that this should really go inside the meta since these spacing wise are really grouped together.  Makes the headings a little more interesting, but probably just a `padding-left: 24px` or something like that

paulfalgout (changes requested):

src/js/views/patients/worklist/worklist-list.scss
LisaManresa: I'm thinking it should either live in the beginning of the meta cell or absolutely positioned on the right side of the action name cell. It would avoid wonkiness with pixels/percents in the table.

paulfalgout: I can't remember what breaks the white-space ellipsis stuff on the action. Would an absolute positioned element mess with that?

luketlancaster:
alrighty so watcha' think about that @paulfalgout?

src/js/views/patients/worklist/worklist-list.scss
LisaManresa: As long as the container's padding accounts for the absolutely positioned element, putting it in the padded area, the ellipsis should work as expected.

src/js/views/patients/schedule/schedule-list.scss
```diff
@@ -99,11 +99,19 @@
 }
 
 .schedule-list__action {
-  width: 63%;
+  width: 58%;
```
paulfalgout: why would this not be `calc(68% - 32px)`?  It seems like if 16px is less than 5% then the action column width will not fill the space.

src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -24,6 +24,13 @@
   width: 16px;
 }
 
+.worklist-list__details {
```
paulfalgout, line 27: Let's try:
```css
    color: #999;
    display: inline-block;
    margin-right: 16px;
    vertical-align: middle;
    width: 16px;
```

src/js/views/patients/worklist/action-item.hbs
```diff
@@ -8,6 +8,7 @@
   </div>
 </td>
 <td class="table-list__cell worklist-list__meta w-40">
+  <span class="table-list__meta worklist-list__details" data-details-region></span>&#8203;{{~ remove_whitespace ~}}
```
paulfalgout: can remove `table-list__meta`

src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -69,6 +76,10 @@
   text-overflow: inherit;
 }
 
+.worklist-list__meta-header {
+  padding-left: 36px;
```
paulfalgout: with the above requests this'd be 48px

src/js/views/patients/schedule/schedule-list.scss
```diff
@@ -99,11 +99,19 @@
...
+}
+
```
paulfalgout, line 105: So I think this is it:
```css
.schedule-list__header-action {
  width: calc(68% - 84px);
}
.schedule-list__header-form {
  padding-left: 32px;
  width: 84px;
}
.schedule-list__action {
  width: calc(68% - 84px);
}
.schedule-list__action-details {
  color: #999;
  padding: 0 16px 0 0;
  text-align: center;
  width: 32px;
}
.schedule-list__action-form {
  padding: 0;
  width: 48px;
}
```

paulfalgout (changes requested):
Not related, but can you remove the `u-margin--r-16` from the data-date-filter-region on the schedule layout.hbs?

src/js/views/patients/schedule/schedule-list.scss
luketlancaster: yeah I tried that, but couldn't get any `calc` to work at all. Just ended up filling too much space and pushing the items to the right way off the table

luketlancaster:
ok lookin' pretty good. Only thing I saw is this: 
Screen Shot 2021-02-17 at 1 09 43 PM

the form button gets cut off when the window gets small. I think `w-40` isn't `w` enough since we keep throwing more icons into that cell

paulfalgout:
Yeah that's a problem.  Particularly. Sept 23 at 12:00PM I believe is worst case width.. might be time for 2 custom column widths.. this one and for action name instead of the `w-`

src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -73,8 +73,13 @@
   white-space: normal;
 }
 
+.worklist-list__item-name {
+  width: 27%;
+}
+
 .worklist-list__meta {
   text-overflow: inherit;
+  width: 43%;
```
luketlancaster, line 83: just did these two just to get them up. Sizing all adds up to 100% now so...yay?`

luketlancaster:
alrighty this is up to date with current requests. Have at it

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -174,6 +165,15 @@ const DayItemView = View.extend({
 
     Radio.trigger('event-router', 'patient:action', this.model.get('_patient'), this.model.id);
   },
+  showDetailsTooltip() {
+    const flow = this.model.getFlow();
```
paulfalgout: I don't think this `const flow` is necessary.  The tooltip view isn't using it

src/js/views/patients/worklist/table-header.hbs
```diff
@@ -7,7 +7,7 @@
     {{ @intl.patients.worklist.tableHeaderTemplate.actionHeader }}
   {{/if}}
 </td>
-<td class="table-list__header w-40">
+<td class="table-list__header w-40 worklist-list__meta-header">
```
paulfalgout: it doesn't seem like this can remain `w-40` if the column isn't.  The header should match the widths and then be adjusted

src/js/views/patients/shared/actions_views.js
```diff
@@ -22,11 +26,42 @@ const FormButton = View.extend({
...
+      uiView: this,
+      ui: this.$el,
+    });
+  },
+  _formatDetails(details) {
+    if (!details) return;
+
+    const sliced = details.slice(0, 140);
+
+    return `${ sliced }${ sliced.length > 139 ? '...' : '' }`;
```
paulfalgout: I think this should check `details.length > 140`?

paulfalgout (changes requested):
This is going to conflict with the test cleanup PR, but we should probably get that in first.

src/js/views/patients/schedule/schedule_views.js
luketlancaster: oh yeah whoops forgot to take it out here

luketlancaster:
sounds good. I'll wait on that PR getting merged in before I push any more changes up, but I'm ready to merge those in locally whenever

src/js/views/patients/shared/actions_views.js
```diff
@@ -22,11 +26,42 @@ const FormButton = View.extend({
...
+      uiView: this,
+      ui: this.$el,
+    });
+  },
+  _formatDetails(details) {
+    if (!details) return;
+
+    const sliced = details.slice(0, 140);
+
+    return `${ sliced }${ sliced.length > 140 ? '...' : '' }`;
```
paulfalgout: I still don't understand `sliced.length` here.  `fooString.slice(0,140).length` can _never_ be greater than 140.

Let's just simplify this function and make it
```js
if (!details || details.length <= 140) return details;

return `${ details.slice(0, 140) }...`;
```

paulfalgout (changes requested):
Other PR was merged. I think once we get this merge fixed with one small change, this can go in and we can run a release.

src/js/views/patients/shared/actions_views.js
paulfalgout: Also there needs to be a test for greater than 140

test/integration/patients/worklist/schedule.js
```diff
@@ -242,7 +245,7 @@ context('schedule page', function() {
 
     cy
       .url()
-      .should('contain', 'patient/1/action/1')
+      .should('contain', 'flow/1/action/1')
```
paulfalgout: It looks like this change reduced coverage by no longer testing if a patient action links to the correct place.

paulfalgout (changes requested):
coverage needs to be 100%

src/js/views/patients/shared/actions_views.js
paulfalgout: @luketlancaster

luketlancaster: ah crap missed it sorry

src/js/views/patients/shared/actions_views.js
```diff
@@ -22,11 +26,40 @@ const FormButton = View.extend({
...
+        details: this._formatDetails(this.model.get('details')),
+      }),
+      uiView: this,
+      ui: this.$el,
+    });
+  },
+  _formatDetails(details) {
+    if (!details || details.length <= 140) return details;
+
+    return `${ details.slice(0, 140).trimEnd() }...`;
```
paulfalgout: Let's use `import trim from 'js/utils/formatting/trim';` as `trimEnd` has an alias in Edge and I don't want to deal with it or check if babel auto-fixes it etc.

paulfalgout (changes requested):

src/js/views/patients/shared/actions_views.js
luketlancaster: 👍 just pushed up so circle ci can run the tests because I don't think mine can handle a video call and cypress at the same time 🤞

paulfalgout (approved):

PR #478, opened 2021-02-22
Feature/dashboards

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch23448]
- [ ] Add QA tasks

It looks like the dashboards list page has already been set up. This completes the display of the iframe on the dashboard page and migrates all instances of "report" to "dashboard".


paulfalgout (commented):
This seems alright to me.  Though I think it might be good on this PR to test it with live data before merging.  If it works, I think we can remove the _DEVELOP_ flag on the menu item.

LisaManresa:
Thanks to @BenCoffeed & @shadowhand, this is working on QA now. A small style tweak to center the spinner is added here, but not reflected on QA.

test/integration/admin/dashboard/dashboard.js
```diff
@@ -0,0 +1,56 @@
...
+      .find('.dashboard__context-trail .js-back')
+      .click();
+
+    cy
+      .url()
+      .should('not.contain', 'dashboards/1')
+      .should('contain', 'dashboards');
+  });
+
+  specify('dashboard does not exist', function() {
```
paulfalgout, line 37: I don't understand why this test works. how does returning null fail?  I think this should be testing for if the route returns a 404 or something.. but I suspect this is "working" because of a javascript error of the data returning something in an unexpected format which should really never happen.

paulfalgout (changes requested):

test/integration/admin/dashboard/dashboard.js
LisaManresa: Yeah... not sure what I was thinking there.

PR #480, opened 2021-03-02
Test reliability

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24659] 

We can't test the length of the search result list when searching by date since that length will vary, but we can test that expected results are in the list.

### DIFF
--- a/test/integration/patients/worklist/worklist.js
+++ b/test/integration/patients/worklist/worklist.js
@@ -2030,13 +2030,36 @@ context('worklist page', function() {
   });
 
   specify('find in list', function() {
+    const currentYear = dayjs().year();
+
+    localStorage.setItem('owned-by_11111', JSON.stringify({
+      actionsSortId: 'sortUpdateDesc',
+      flowsSortId: 'sortUpdateDesc',
+      filters: {
+        type: 'flows',
+        groupId: null,
+        clinicianId: '11111',
+      },
+      flowsDateFilters: {
+        selectedMonth: `${ currentYear }-01-01`,
+        dateType: 'created_at',
+      },
+      selectedActions: {},
+      selectedFlows: {},
+    }));
+
     cy
       .server()
       .routeGroupsBootstrap(_.identity, testGroups)
       .routeFlows(fx => {
+        _.each(fx.data, function(flow) {
+          flow.attributes.created_at = `${ currentYear }-01-30`;
+          flow.attributes.updated_at = `${ currentYear }-01-31`;
+        });
+
         fx.data[0].attributes.name = 'Test Flow';
-        fx.data[0].attributes.created_at = testTsSubtract(1, 'month');
-        fx.data[0].attributes.updated_at = testTsSubtract(20);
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -142,43 +142,3 @@ workflows:
       - successPreviewStatus:
           requires:
             - addPreviewSubdomainAuth0
-
-  test_cross_browser:
```
paulfalgout: I don't want to remove these

test/integration/patients/worklist/worklist.js
```diff
@@ -2207,13 +2193,19 @@ context('worklist page', function() {
...
-      .get('[data-select-all-region]')
-      .find('.fa-check-square');
+      .get('@flowList')
+      .find('.work-list__item')
+      .contains('Test Flow');
 
     cy
-      .get('[data-filters-region]')
-      .find('.js-bulk-edit')
-      .should('contain', 'Edit 3 Flows');
```
paulfalgout: I do think we _could_ test that this has the correct number by not introducing fixture data that could randomly break particular tests.

paulfalgout (changes requested):

.circleci/config.yml
```diff
@@ -166,9 +167,9 @@ workflows:
           parallel: true
           parallelism: 2
           group: 'care ops chrome'
-          start: 'npm start'
```
LisaManresa: This was a guess. But I figured since the custom command is calling `npm start`, we didn't need this. Or we could leave this and remove the the `npm start` from the command

test/integration/patients/worklist/worklist.js
LisaManresa: Yeah, there was no need to dynamic date in this test. Made them static.

.circleci/config.yml
```diff
@@ -156,6 +156,7 @@ workflows:
           pre-steps:
             - font_awesome
       - cypress/run:
+          command: npm start & cypress run --config
```
paulfalgout: I'm not convinced this is going to work.  I don't really know why we're seeing coverage errors on these tests, but here's the error specifically: https://dashboard.cypress.io/projects/ep9zr6/runs/2234/test-results/5c7a4b7b-c872-4496-a645-94d24c06ef96?statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D

I think we need to prove the fix works before we change these by probably temporarily modifying the main action or adding a temporary new one.  Maybe remove this change from this PR and a separate PR can come out where we can make sure that the coverage reports don't run on CI?  It will probably take forcing an error to prove that no after each is called when tests are run on circleci.

paulfalgout (changes requested):

.circleci/config.yml
LisaManresa: That's fair. I thought it would be a much more minor change. I'll put up a separate PR

paulfalgout (approved):

PR #481, opened 2021-03-02
Adding pre-wrap for disabled textareas

wweaver4th:
- [ ] Add the Clubhouse Story ID: [ch24674]
- [ ] Add QA tasks

### DIFF
--- a/src/sass/formapp.scss
+++ b/src/sass/formapp.scss
@@ -57,9 +57,10 @@ body {
 .comment__message {
   white-space: pre-wrap;
 }
-.form-control:disabled, .form-control[readonly] {
+.form-control:disabled, .form-control[readonly], .formio-editor-read-only-content {
   background-color: inherit;
   border: none !important;
+  white-space: pre-wrap;
 }
 .formio-read-only .btn.disabled, .formio-read-only .btn:disabled {
   display:none;
--- a/src/sass/formapp.scss
+++ b/src/sass/formapp.scss
@@ -63,7 +63,7 @@ body {
 }
 
 // NOTE: Fixed in upcoming release https://github.com/formio/formio.js/issues/2471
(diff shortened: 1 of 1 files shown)


src/sass/formapp.scss
```diff
@@ -57,9 +57,10 @@ body {
 .comment__message {
   white-space: pre-wrap;
 }
-.form-control:disabled, .form-control[readonly] {
+.form-control:disabled, .form-control[readonly], .formio-editor-read-only-content {
```
paulfalgout: @wweaver4th What is causing this to be needed?  It looks like this style already exists. https://github.com/formio/formio.js/blob/1b58fdc181d3f717de129fc3c35dbbefd77bb945/dist/formio.form.css#L1031

paulfalgout: OK this is fixed in theory in 4.13.x which is still a release candidate.

https://github.com/formio/formio.js/issues/2471

We should probably make this a style on its own line, not with these other 2 styles with a `// NOTE:` comment mentioning this comment so that it can be removed when the form.io fix is released.

So I think this change should probably be:
```css
// NOTE: Fixed in upcoming release https://github.com/formio/formio.js/issues/2471
.formio-component-textarea div.formio-editor-read-only-content[ref="input"] {
  white-space: pre-wrap; 
}
```

Then if this selector doesn't solve the problem we should update the open issue.

wweaver4th: The textareas in the response view of the forms do not respect line breaks added by the user.

wweaver4th: Word. I'll change a push up an update.

wweaver4th: Bah, hold up.

paulfalgout (approved):

PR #482, opened 2021-03-02
Edit patient modal

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24642]


src/js/entities-service/entities/patients.js
```diff
@@ -35,22 +35,27 @@ const _Model = BaseModel.extend({
...
+    this.set('_groups', groups.concat({
+      id: group.id,
+    }));
   },
-  removeGroup(group) {
+  removeGroup(removedGroup) {
     const groups = this.get('_groups');
 
-    this.set('_groups', reject(groups, groupId => groupId === group.id));
+    this.set('_groups', reject(groups, group => group.id === removedGroup.id));
```
LisaManresa, line 45: This was broken, the format of `removedGroup` was incorrect, so nothing would ever get rejected.

src/js/entities-service/entities/patients.js
```diff
@@ -35,22 +35,27 @@ const _Model = BaseModel.extend({
...
   },
-  saveAll() {
-    const attrs = extend({}, this.attributes);
+  saveAll(attrs) {
+    attrs = omit(extend({}, this.attributes, attrs), 'engagement', 'source');
```
LisaManresa: This was an assumption that we would not want to include `engagement` and `source`. I could see leaving `source`, but engagement comes from a separate endpoint and is not treated as a true relationship.

test/integration/patients/patient/patient.js
```diff
@@ -11,10 +11,20 @@ context('patient page', function() {
 
         return fx;
       })
-      .routePatientActions(_.identity, '1')
-      .routePatientFlows()
+      .routePatientActions(fx => {
+        fx.included = [];
```
LisaManresa: Since the context trail now re-renders more frequently, we needed to wait for these requests to complete before looking for the context trail element.

However, the data from these requests was overriding the test patient data. Since we aren't specifically testing anything related to actions/flows, emptying out the included data fixes that. But we are using it in a number of places, so perhaps a more universal fix is needed.

src/js/entities-service/entities/patients.js
paulfalgout: There should not be a problem returning `source` on the model.  `engagement` could be an issue.. I'm not sure.  if it is, then we should be storing this on the patient as `_engagement` rather than omitting it here.

src/js/views/globals/patient-modal/patient-modal_views.js
```diff
@@ -58,26 +63,45 @@ const InputView = View.extend({
...
+    this.isDisabled = !this.model.isNew() && !this.model.hasChanged();
+  },
   onClickSave() {
     this.isDisabled = true;
     this.render();
   },
+  onModelUpdate() {
+    if (this.model.isNew()) return;
+
+    this.isDisabled = !this.model.hasChanged() && !this.model.isValid({ preSave: true });
```
paulfalgout: I'm not entirely sure what changing this value is doing without rerendering.  _but_ I think storing this value was a cheap way to handle the submit button, but probably isn't ideal for more complex state.  perhaps we should change the `isDisabled` to `isSubmitting` and the check on if the view should be disabled should be within `templateContext` and the `modelEvents` can be `'change': 'render'`

src/js/views/globals/patient-modal/patient-modal_views.js
```diff
@@ -169,15 +194,22 @@ const AddPatientModal = View.extend({
     'click:search': 'click:search',
   },
   modelEvents: {
-    'change': 'onChange',
+    'change change:_groups': 'onChange',
```
paulfalgout: when `_groups` changes this event will get called twice.  The `change` event will always trigger if `change:_groups` is triggered.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -236,10 +241,40 @@ const SidebarView = CollectionView.extend({
       patient: this.model,
     };
   },
+  modelEvents: {
+    'change:first_name change:last_name change:birth_date change:sex change:_groups': 'render',
```
paulfalgout: I think we can safely put `'change': 'render'` here

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -236,10 +241,40 @@ const SidebarView = CollectionView.extend({
...
   },
   viewComparator: false,
+  onClickMenu() {
+    const menuOptions = new Backbone.Collection([
+      {
+        onSelect: bind(this.triggerMethod, this, 'click:accountDetails'),
+      },
+    ]);
+
+    const itemTemplate = this.model.get('source') === 'manual' ? hbs`{{ @intl.patients.patient.sidebar.sidebarViews.menuOptions.edit }}` : hbs`{{ @intl.patients.patient.sidebar.sidebarViews.menuOptions.view }}`;
```
paulfalgout: I think this could be `this.model.canEdit()`

test/integration/patients/patient/patient.js
paulfalgout: This has been the source of some flake for sure.  It's hard though because removing it universally can also break things for non-included models.  Is it because `/test/` fixtures are in the includes? Or are we modifying regular fixture data that gets sometimes overridden by what is actually in the fixtures?

src/js/entities-service/entities/patients.js
paulfalgout: Hmm yeah I can see how this would work for a new patient, but break for a patient that already has groups

src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -3,17 +3,50 @@ import Radio from 'backbone.radio';
...
     const widgets = Radio.request('entities', 'widgets:collection', collectionOf(currentOrg.getSetting('widget_sidebar'), 'id'));
-    this.patient = patient;
 
-    this.showView(new SidebarView({
+    if (!this.isRestarting()) {
+      this.patient = patient;
+    }
+
+    const sidebarView = this.showView(new SidebarView({
       model: patient,
```
paulfalgout: If the above change is correct, should this be `this.patient`?

src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -3,17 +3,50 @@ import Radio from 'backbone.radio';
...
 import { SidebarView } from 'js/views/patients/patient/sidebar/sidebar_views';
 
 export default App.extend({
   onBeforeStart({ patient }) {
     const currentOrg = Radio.request('bootstrap', 'currentOrg');
     const widgets = Radio.request('entities', 'widgets:collection', collectionOf(currentOrg.getSetting('widget_sidebar'), 'id'));
-    this.patient = patient;
 
-    this.showView(new SidebarView({
+    if (!this.isRestarting()) {
```
paulfalgout: when does this app restart?

paulfalgout (changes requested):

src/js/apps/patients/patient/sidebar/sidebar_app.js
LisaManresa: Ah, it did, but then I refactored to use model change events in views. This can be removed.

test/integration/patients/patient/patient.js
LisaManresa: Ok, dug into this again and realized this change is unnecessary. Just passing any id other than '1' as the second argument fixes the problem (I thought I had tried that earlier, but must have missed it). 

So the root of the issue is that `routePatientActions` and `routePatientFlows` default to assigning an id of '1' to all included `patients`. Might not be such a big issue since it's easy enough to pass a different id or use a different id in the test.

paulfalgout (changes requested):
The cypress tests look like they're failing on a missing `wait('@route...` that is rerendering the sidebar mid-test or something like that.

src/js/views/globals/patient-modal/patient-modal_views.js
```diff
@@ -280,11 +320,6 @@ const AddPatientModal = View.extend({
   onClickSave() {
     this.model.isValid({ preSave: true });
 
-    if (!isEmpty(this.model.validationError)) {
```
LisaManresa, line 283: The modal SaveView rerenders on model change and updates the disabled attribute accordingly now, so no need to check for errors on save

src/js/views/globals/patient-modal/patient-modal_views.js
```diff
@@ -58,28 +63,43 @@ const InputView = View.extend({
...
     const state = this.getOption('state');
+    const isDisabled = this.isSubmitting || !this.model.hasChanged() || this.model.hasChanged() && !this.model.isValid({ preSave: true });
 
     return {
-      isDisabled: state.get('errors') || this.isDisabled,
+      isDisabled: state.get('errors') || isDisabled,
     };
   },
+  initialize() {
+    this.isDisabled = !this.model.isNew() && !this.model.hasChanged();
```
paulfalgout: can remove

src/js/views/globals/patient-modal/patient-modal_views.js
```diff
@@ -58,28 +63,43 @@ const InputView = View.extend({
...
   triggers: {
     'click .js-cancel': 'cancel',
     'click .js-save': 'click:save',
   },
+  modelEvents: {
+    'change': 'render',
+  },
   templateContext() {
     const state = this.getOption('state');
+    const isDisabled = this.isSubmitting || !this.model.hasChanged() || this.model.hasChanged() && !this.model.isValid({ preSave: true });
```
paulfalgout: I think the `this.model.hasChanged() &&` might not be needed?  you're only ever getting to that or if it has changed because of the previous or

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -242,6 +249,56 @@ const SidebarView = CollectionView.extend({
...
+    widgets: '[data-widgets-region]',
+  },
+  onRender() {
+    this.showChildView('name', new NameView({
+      model: this.model,
+    }));
+
+    this.showChildView('widgets', new WidgetCollectionView({
+      model: this.model,
+      collection: this.getOption('collection'),
```
paulfalgout: `this.collection` should be fine here.

paulfalgout (changes requested):
This is looking pretty solid

PR #483, opened 2021-03-08
Skip coverage on cross-browser tests

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24646]

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,6 +158,7 @@ workflows:
       - cypress/run:
           pre-steps:
             - browser-tools/install-chrome
+            - run: echo 'export CYPRESS_COVERAGE=false' >> $BASH_ENV
           name: Chrome
           browser: chrome
           requires:
@@ -172,6 +173,7 @@ workflows:
           pre-steps:
             - jq/install
             - browser-tools/install-firefox
+            - run: echo 'export CYPRESS_COVERAGE=false' >> $BASH_ENV
           name: Firefox
           browser: firefox
           requires:


package.json
```diff
@@ -15,6 +15,7 @@
     "coverage:report:ci": "nyc report --reporter=lcovonly || true",
     "dev": "webpack serve",
     "dev:coverage": "npm start & cypress open --env COVERAGE=open",
+    "cypress:nocoverage": "npm start & cypress run --env COVERAGE=false --config video=false",
```
paulfalgout: we do want video

paulfalgout (changes requested):

LisaManresa:
Ok, I think this may be the solution. Was able to verify the command worked in the test deploy run.

package.json
```diff
@@ -15,6 +15,7 @@
     "coverage:report:ci": "nyc report --reporter=lcovonly || true",
     "dev": "webpack serve",
     "dev:coverage": "npm start & cypress open --env COVERAGE=open",
+    "cypress:nocoverage": "npm start & cypress run --env COVERAGE=false",
```
paulfalgout: `npm start` isn't needed here.  It's called by the orb.

paulfalgout (changes requested):
which run?

LisaManresa:
[This is the run](https://dashboard.cypress.io/projects/ep9zr6/runs/2263/test-results/c0141922-949f-4a78-9b66-575eb3c2bc43) that had the change. It failed because I didn't remove the coveralls post-step and there was nothing to upload. Watching the videos you can catch some that show no coverage called at the end of each test. But I can run another one with an explicit failure if that'll make it more clear.

package.json
LisaManresa: Hmm, would it be better to leave the npm start here and remove the start command from the orb? In case this is used elsewhere?

paulfalgout:
I think having the start separate is slightly better.  Having start in a separate command doesn't supress the logs of the web server.
Also I don't think it is working yet.  I wonder if `false` is "false".  Maybe we need to try `=0`.  Not sure how to make it falsy.  

Screen Shot 2021-03-09 at 4 31 49 AM

package.json
```diff
@@ -15,6 +15,7 @@
     "coverage:report:ci": "nyc report --reporter=lcovonly || true",
     "dev": "webpack serve",
     "dev:coverage": "npm start & cypress open --env COVERAGE=open",
+    "cypress:nocoverage": "cypress run --env COVERAGE=false",
```
paulfalgout: So now cypress isn't recording the run.. After looking at the orb I think we should probably rely on how it works and use the config rather than overriding the command: https://github.com/cypress-io/circleci-orb/blob/master/src/orb.yml#L1356

paulfalgout (changes requested):

LisaManresa:
In testing with the `test-deploy` run, `COVERAGE` always equals `1` regardless of what we pass (or don't pass, it's still 1 even when the field is blank) in the config field. 

The latest iteration is an attempt to use an `executor` to establish the environment variables based on [this example](https://circleci.com/developer/orbs/orb/cypress-io/cypress#usage-env-vars). No luck so far. But I'll keep trying.

paulfalgout (changes requested):
Ok I think this can be done by adding this to the `pre-steps:`

`- run: echo 'export CYPRESS_COVERAGE=false' >> $BASH_ENV`

For whatever reason using the config to set env vars breaks the record.. orb doesn't seem 100%.

Anyhow.  I did just realize in the cypress dashboard sidebar, where I was playing the video you can expand the "Test Definition" bar and it will show if there is an `afterEach` or not, so a bug in the test isn't needed to slow down the video.

LisaManresa:
Ok, using `pre-steps` worked. Thanks @paulfalgout. [Here is the run](https://dashboard.cypress.io/projects/ep9zr6/runs/2281/test-results/d70f779f-4ccf-409d-a6dd-f8a2f02929ec?modificationDateRange=%7B%22startDate%22%3A%222021-03-09%22%2C%22endDate%22%3A%222021-03-09%22%7D&modificationDateRangeId=run&modificationScope=RUN&modificationTypes=%5B%22MODIFIED_BODY%22%2C%22MODIFIED_BODY_AND_HOOKS%22%2C%22MODIFIED_HOOKS%22%5D) that verified no `afterEach` was called with `pre-steps` configured.

PR #484, opened 2021-03-08
Handle formio errors

paulfalgout:
form.io only expects an array of string and doesn’t tie errors to specific fields.

So this in theory actually handles any error, if we ever throw errors for the form itself.

- [x] Add the Clubhouse Story ID: [ch24696]

### DIFF
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -1,5 +1,5 @@
 import $ from 'jquery';
-import { extend } from 'underscore';
+import { extend, pluck } from 'underscore';
 
 import Radio from 'backbone.radio';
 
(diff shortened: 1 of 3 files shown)


src/js/services/forms.js
```diff
@@ -72,8 +72,8 @@ export default App.extend({
     formResponse.saveAll()
       .then(() => {
         this.trigger('success', formResponse);
-      }).fail(errors => {
-        /* istanbul ignore next: Don't need to test error handler */
```
paulfalgout, line 76: oops

paulfalgout:
@LisaManresa 👀

src/js/services/forms.js
LisaManresa: 😬

LisaManresa (approved):

PR #488, opened 2021-03-15
Support key in patient sidebar dateTimeWidget

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24734]


paulfalgout (changes requested):
Can you add it to the documentation as well?

paulfalgout (approved):

PR #492, opened 2021-03-17
Cypress upgrade

LisaManresa:
- [x] Add the Clubhouse Story ID: [ch24726]


LisaManresa:
This potentially fixes the Firefox issue in [ch24775]. But we'll need to wait for the cross-browser test run so see.

paulfalgout (approved):

PR #494, opened 2021-03-19
Ship sourcemaps for uploading

BenCoffeed:
- [x] Add the Clubhouse Story ID: [ch24739]
- [ ] Add QA tasks


paulfalgout (changes requested):
I think we need to add the line:
`devtool: isProduction ? 'nosources-source-map' : 'eval',` and not change the `mode`.

paulfalgout:
Let's assume this works 😎

PR #502, opened 2021-03-29
Add traffic cone icon and grey color for new state option

wweaver4th:
- [ ] Add the Clubhouse Story ID: [ch24933]
- [ ] Add QA tasks


wweaver4th:
added wrong CH [ch24934]

LisaManresa (approved):

PR #512, opened 2021-04-12
Form save outside of form.io

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch24769] [ch24760]

Also includes removing the response redraw which required `refreshOnChange: true` for html and content components


paulfalgout:
@wweaver4th attempting to go ahead and update form.io here.  4.13.0 allows for us to remove some warnings including one new one.  https://github.com/RoundingWell/care-ops-frontend/pull/513

PR #516, opened 2021-04-22
Make font awesome a dependency

paulfalgout:
And update form io styles organization

- [x] Add the Clubhouse Story ID: [ch25091]

- Updates font-awesome to a dependency
- Splits up the form styles
- Utilizes some base styles


paulfalgout:
This failure is due to coveralls, not failing tests.

paulfalgout:
Also this includes the font-awesome v4 shim which I believe should confidently solve icon issues seen while testing for even icons we aren't currently using.

PR #520, opened 2021-05-03
Close modal on route change

paulfalgout:
If the route change the modal cancels.

- [x] Add the Clubhouse Story ID: [ch25095]

### DIFF
--- a/src/js/apps/patients/patients-main_app.js
+++ b/src/js/apps/patients/patients-main_app.js
@@ -34,7 +34,7 @@ export default RouterApp.extend({
 
     this.routeAction(defaultRoute, () => {
       defer(()=> {
-        this.replaceRoute(defaultRoute, defaultWorklist);
+        this.navigateRoute(defaultRoute, defaultWorklist);
         Radio.request('nav', 'select', this.routerAppName, defaultRoute, [defaultWorklist]);
         this.setLatestList(defaultRoute, [defaultWorklist]);
         this.showPatientsWorklist(defaultWorklist);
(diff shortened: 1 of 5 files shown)


src/js/apps/patients/patients-main_app.js
```diff
@@ -34,7 +34,7 @@ export default RouterApp.extend({
 
     this.routeAction(defaultRoute, () => {
       defer(()=> {
-        this.replaceRoute(defaultRoute, defaultWorklist);
+        this.navigateRoute(defaultRoute, defaultWorklist);
```
paulfalgout, line 37: so we were actually swallowing this route in history.  The "default" even wasn't causing an issue with double-history

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -127,6 +127,7 @@ const PatientsAppNav = View.extend({
     search: '.js-search',
   },
   onSearchActive(isActive) {
+    if (this.isDestroyed()) return;
```
paulfalgout, line 130: In this case the nav was switching away from the patients nav, and then the search modal was destroying which triggered this event causing an error.

src/js/views/globals/root_views.js
```diff
@@ -81,6 +81,7 @@ const ModalRegionView = TopRegionView.extend({
     this.listenTo(hotkeyCh, 'close', this.empty);
   },
   onRegionShow() {
+    this.listenTo(historyCh, 'change:route', this.empty);
```
paulfalgout, line 84: If there's a browser back/forward, the modal should cancel

PR #549, opened 2021-07-29
Add snippet to form

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch25791]

You'll need this to test snippets on your sandbox.
They'll kinda work without this I think but they'll all have the label "Snippet"


paulfalgout:
This isn't going to break anything, so I'm just going to merge it.

PR #554, opened 2021-08-05
tweak padding for data-column style

wweaver4th:
- [ ] Add QA tasks


paulfalgout (changes requested):
This is not valid css. The only length that does not need a unit is `0` Everything else needs `px`, `%`, `em` etc

PR #558, opened 2021-08-14
Add patient, action, program-action ids to form submission

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch25928]
Nested like so:
```
{
  patient: {
    first_name: 'foo',
    fields: {...}
  },
  clinician: { ... },
  patient_id: <uuid>,
  patient_action_id: <uuid>,
  program_action_id: <uuid>
}
```

### DIFF
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -1,5 +1,5 @@
 import $ from 'jquery';
-import { extend, pluck } from 'underscore';
+import { extend, pluck, get } from 'underscore';
 
 import Radio from 'backbone.radio';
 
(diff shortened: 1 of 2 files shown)


wweaver4th:
@paulfalgout is this ready for PR? I'm not seeing these IDs in response payload but maybe I didn't set it up right. I added them as `prefilled_fields` and tried referencing them in a field.

paulfalgout:
@wweaver4th the ids are always attached if they exist no matter what prefilled you setup.  And they are setup directly on the `submission.data` object.
```
{
  patient: {
    first_name: 'foo',
    fields: {...}
  },
  clinician: { ... },
  patient_id: <uuid>,
  patient_action_id: <uuid>,
  program_action_id: <uuid>
}
```

paulfalgout:
This works even if we don't use it now, it might as well be on the submission object.

PR #580, opened 2021-10-21
Bugfix add icons and filter disabled clinicians from active

paulfalgout:
- [x] Add the Clubhouse Story ID: [ch26654]


paulfalgout:
[ch26655]

PR #582, opened 2021-10-26
Extend the fields if patient exists

paulfalgout:
Fields may not exist in the prefill, but patient names may.

- [x] Add the Clubhouse Story ID: [ch26692]


shadowhand (commented):
Nice find. 👍

PR #592, opened 2021-12-06
Fixes flow meta changing when both done states

paulfalgout:
Previous check assumed a single done state.  Should have been this from the start.

- [x] Add the Clubhouse Story ID: [sc-27025]


paulfalgout:
[sc-27025]

PR #600, opened 2021-12-16
Allow customizing FA options

shadowhand:
Shortcut Story ID: [sc-27157]

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -14,8 +14,11 @@ const appNavMenu = new Backbone.Collection([
       Radio.trigger('event-router', 'default');
     },
     id: 'PatientsApp',
-    iconType: 'far',
-    icon: 'window',
+    icon: {
+      type: 'far',
+      name: 'window',
+      classes: ['fa-fw'],
+    },
     text: i18n.mainNav.patients,
   },
   {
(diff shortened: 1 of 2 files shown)


src/js/base/fontawesome.js
```diff
@@ -6,9 +6,9 @@ import '@fortawesome/fontawesome-svg-core/styles.css';
 config.autoAddCss = false;
 config.replacementClass = 'svg-inline--fa icon';
 
-function faHelper(prefix, iconName, { hash = {} }) {
+function faHelper(prefix, iconName, params = {}) {
```
shadowhand: I'm not sure why this was written as `{ hash = {} }` before... seems like a typo?

src/js/apps/globals/nav_app.js
```diff
@@ -17,6 +17,9 @@ const appNavMenu = new Backbone.Collection([
     iconType: 'far',
     icon: 'window',
     text: i18n.mainNav.patients,
+    params: {
+      classes: ['fa-fw'],
```
shadowhand, line 20: Parameter definition found [here][1].

[1]: https://github.com/FortAwesome/Font-Awesome/blob/7d3d774145ac38663f6d1effc6def0334b68ab7e/js-packages/%40fortawesome/fontawesome-svg-core/index.d.ts#L91

src/js/base/fontawesome.js
paulfalgout: This is a handlebars thing that is probably worth keeping. https://handlebarsjs.com/guide/block-helpers.html#hash-arguments
Also this change is definitely breaking a `fa*` template helper somewhere, but it isn't going to be caught by tests as it's a style change.

paulfalgout (changes requested):

src/js/base/fontawesome.js
shadowhand: I don't follow. I'm pretty sure that the handlebar docs are saying that the final argument for a helper should be a hash (aka plain object) with a default of `{}`, like I am showing here.

The `{ hash = {} }` makes a required object with an optional `hash` property... right?

paulfalgout: For instance for the handlebars block `if` `{{#if foo includeZero=true}}`  is checked on this line: https://github.com/handlebars-lang/handlebars.js/blob/6790c080c641ef2b44e663800e1794fae180977a/lib/handlebars/helpers/if.js#L16

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -39,7 +39,7 @@ const MainNavDroplist = Droplist.extend({
       lists: [{
         collection: this.collection,
         itemTemplate: hbs`
-          <span class="app-nav__picklist-icon">{{fa iconType icon}}</span>
+          <span class="app-nav__picklist-icon">{{fa iconType icon params}}</span>
```
paulfalgout: I think `{{fa iconType icon classes="fa-fw"}}` or `{{fa iconType icon classes=params.classes}}` would do this without changing the helper.

src/js/apps/globals/nav_app.js
```diff
@@ -14,53 +14,71 @@ const appNavMenu = new Backbone.Collection([
       Radio.trigger('event-router', 'default');
     },
     id: 'PatientsApp',
-    iconType: 'far',
-    icon: 'window',
+    icon: {
+      type: 'far',
+      name: 'window',
+      classes: ['fa-fw'],
```
shadowhand, line 20: @paulfalgout how about this?

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -39,7 +39,7 @@ const MainNavDroplist = Droplist.extend({
       lists: [{
         collection: this.collection,
         itemTemplate: hbs`
-          <span class="app-nav__picklist-icon">{{fa iconType icon}}</span>
+          <span class="app-nav__picklist-icon">{{fa icon.type icon.name classes=icon.classes}}</span>
```
shadowhand, line 42: It would be really nice if we could do something like `{{fa ...icon}}` here, to automatically fill the prefix/name. I suppose the args would have to be `prefix, iconName` instead of `type, name`.

paulfalgout (approved):

PR #602, opened 2021-12-21
Change dataset to directory

paulfalgout:
And update to API results schema

Shortcut Story ID: [sc-27219]

@wweaver4th this works with the sandbox fixture
```json
{
  "components": [
    {
      "label": "Select",
      "widget": "choicesjs",
      "tableView": true,
      "dataSrc": "custom",
      "data": {
        "custom": "values = getData('facilities');"
      },
      "template": "<span>{{ item.name }} - {{ item.location }}</span>",
      "key": "select",
      "type": "select",
      "input": true
    }
  ]
}
```


paulfalgout:
[ch27186]

PR #614, opened 2022-01-19
Remove deprecations

paulfalgout:
Shortcut Story ID: [sc-27474] [sc-27475]

### DIFF
--- a/src/js/entities-service/actions.js
+++ b/src/js/entities-service/actions.js
@@ -6,10 +6,17 @@ const Entity = BaseEntity.extend({
   radioRequests: {
     'actions:model': 'getModel',
     'actions:collection': 'getCollection',
-    'fetch:actions:model': 'fetchModel',
+    'fetch:actions:model': 'fetchAction',
     'fetch:actions:collection': 'fetchActions',
     'fetch:actions:collection:byPatient': 'fetchActionsByPatient',
     'fetch:actions:collection:byFlow': 'fetchActionsByFlow',
+  },
+  fetchAction(id) {
+    const include = [
+      'program-action.program',
+      'flow.program-flow.program',
+    ].join();
+    return this.fetchModel(id, { data: { include } });
   },
   fetchActions({ filter }) {
     const data = { filter };
--- a/src/js/entities-service/actions.js
+++ b/src/js/entities-service/actions.js
@@ -6,7 +6,7 @@ const Entity = BaseEntity.extend({
   radioRequests: {
     'actions:model': 'getModel',
     'actions:collection': 'getCollection',
-    'fetch:actions:model': 'fetchCachedModel',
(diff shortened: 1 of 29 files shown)


src/js/entities-service/actions.js
```diff
@@ -6,11 +6,15 @@ const Entity = BaseEntity.extend({
...
     'actions:collection': 'getCollection',
-    'fetch:actions:model': 'fetchCachedModel',
+    'fetch:actions:model': 'fetchAction',
     'fetch:actions:collection': 'fetchActions',
     'fetch:actions:collection:byPatient': 'fetchActionsByPatient',
     'fetch:actions:collection:byFlow': 'fetchActionsByFlow',
   },
+  fetchAction(id) {
+    const data = { include: 'program-action.program,flow.program-flow.program' };
+    return this.fetchModel(id, { data });
```
shadowhand: Would it make sense for this to be ... ?

```suggestion
    const include = [
      'program-action.program',
      'flow.program-flow.program',
    ];
    return this.fetchModel(id, { include: include.join(',') });
```

That way it is easier to add additional includes later?

src/js/entities-service/flows.js
```diff
@@ -6,10 +6,14 @@ const Entity = BaseEntity.extend({
...
     'flows:model': 'getModel',
     'flows:collection': 'getCollection',
-    'fetch:flows:model': 'fetchCachedModel',
+    'fetch:flows:model': 'fetchFlow',
     'fetch:flows:collection': 'fetchFlows',
     'fetch:flows:collection:byPatient': 'fetchFlowsByPatient',
   },
+  fetchFlow(id) {
+    const data = { include: 'program-flow,program-flow.program,program-flow.program-actions' };
+    return this.fetchModel(id, { data });
```
shadowhand: Same question here.

src/js/outreach/index.js
```diff
@@ -169,9 +169,12 @@ const FormApp = App.extend({
       .fail(({ responseJSON }) => {
         this.showFormSave();
         /* istanbul ignore next: Don't handle non-API errors */
-        if (!responseJSON) return;
+        if (!responseJSON) {
+          this.channel.request('send', 'form:errors', ['Unknown Error']);
```
shadowhand: Doesn't this need a `return` after to prevent the next `responseJson.errors` access from showing a warning?

src/js/entities-service/actions.js
paulfalgout: yeah that's reasonable.  I think we used to do that too prior to pulling all of the includes

src/js/outreach/index.js
paulfalgout: I'm mostly just messing with a test that will not fail locally for any reason, but is really flaky on CI

PR #635, opened 2022-02-23
Expose formio evaluator errors as error

paulfalgout:
[sc-27773]

Currently errors that occur in our custom javascript added to forms either in the definition or context scripts/reducers are caught and delivered as warnings instead of errors by form.io

This prevents us from seeing errors on production and makes it such that real errors are not obvious.

I’ve asked form.io to formally allow for caught errors to throw as errors here:
https://github.com/formio/formio.js/issues/4613

Shortcut Story ID: [sc-###]

### DIFF
--- a/test/fixtures/test/forms.json
+++ b/test/fixtures/test/forms.json
@@ -9,5 +9,26 @@
     "options": {
       "read_only": true
     }
+  },
+  {
+    "id": "33333",
+    "name": "Script and Reducers",
+    "options": {
+      "context": [
+        "return { foo() { return 'foo'; } }"
+      ],
+      "reducers": [
+        "formSubmission.storyTime = foo()\nreturn formSubmission"
+      ]
+    }
+  },
+  {
+    "id": "44444",
+    "name": "Reducer Error",
+    "options": {
+      "reducers": [
+        "syntaxError('foo;"
+      ]
+    }
   }
 ]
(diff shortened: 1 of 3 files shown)


test/integration/forms/form.js
```diff
@@ -951,6 +951,52 @@ context('Patient Form', function() {
       .find('textarea[name="data[storyTime]"]');
   });
 
+  specify('form scripts and reducers', function() {
```
paulfalgout, line 954: We weren't actually previously testing scripts or reducers in this app.  So this simple test checks that the reducers are working and have access to the context scripts.

paulfalgout:
I think the coveralls report missed something.  No changes were made around the issue it's addressing.

test/fixtures/test/forms.json
```diff
@@ -9,5 +9,26 @@
...
+        "formSubmission.storyTime = foo()\nreturn formSubmission"
+      ]
+    }
+  },
+  {
+    "id": "44444",
+    "name": "Reducer Error",
+    "options": {
+      "reducers": [
+        "syntaxError('foo;"
```
paulfalgout, line 30: @zfixler @nmajor25 regarding our recent conversation.  The point of this PR is that the code in this reducer currently would just throw a console warning instead of an error.  Not a terrible problem for very simple logic as we'll catch stuff, but for anything mildly complicated we could be getting errors in production that we neglected to test but no one currently is aware.

nmajor25 (approved):

PR #639, opened 2022-03-02
Expand forms by default and store user preference in localStorage

nmajor25:
Shortcut Story ID: [sc-27808]

When a form is viewed, change the default to be expanded width mode. When a user selects an option to have the form expanded or not, remember that preference across pages and app reloads.

The user's preference is persisted by storing an `isExpanded: boolean` key-value in localStorage.

This feature caused some tests to break, so those were updated to pass. And a few new tests were added to ensure the localStorage feature works properly.

### DIFF
--- a/src/js/apps/forms/form/form-patient_app.js
+++ b/src/js/apps/forms/form/form-patient_app.js
@@ -1,5 +1,7 @@
 import Radio from 'backbone.radio';
 import dayjs from 'dayjs';
+import store from 'store';
+import { extend } from 'underscore';
 
 import App from 'js/base/app';
 
@@ -17,11 +19,16 @@ export default App.extend({
       getOptions: ['patient'],
     },
   },
+  initFormState() {
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    const storedState = store.get(`form-state_${ currentUser.id }`);
+
+    this.setState(extend({ isExpanded: true }, storedState));
+  },
   onBeforeStart() {
     this.getRegion().startPreloader();
-    this.setState({
-      isExpanded: false,
-    });
+
+    this.initFormState();
   },
   beforeStart({ formId, patientId }) {
     return [
@@ -36,6 +43,7 @@ export default App.extend({
     this.patient = patient;
     this.form = form;
     this.isReadOnly = this.form.isReadOnly();
+    this.currentUser = Radio.request('bootstrap', 'currentUser');
 
     this.startFormService();
 
@@ -71,7 +79,13 @@ export default App.extend({
     });
   },
   stateEvents: {
+    'change': 'onChangeState',
     'change:isExpanded': 'showSidebar',
+  },
+  onChangeState(state) {
+    if (!state.hasChanged('isExpanded')) return;
+
+    store.set(`form-state_${ this.currentUser.id }`, { isExpanded: this.getState('isExpanded') });
   },
   showFormStatus(response) {
     if (this.isReadOnly) return;
--- a/src/js/apps/forms/form/form-patient_app.js
+++ b/src/js/apps/forms/form/form-patient_app.js
@@ -79,11 +79,14 @@ export default App.extend({
     });
   },
   stateEvents: {
-    'change': 'onChangeState',
-    'change:isExpanded': 'showSidebar',
+    'change:isExpanded': 'onChangeIsExpanded',
   },
-  onChangeState(state) {
-    store.set(`form-state_${ this.currentUser.id }`, { isExpanded: this.getState('isExpanded') });
+  onChangeIsExpanded() {
+    const isExpanded = this.getState('isExpanded');
+
+    store.set(`form-state_${ this.currentUser.id }`, { isExpanded: isExpanded });
+
+    this.showSidebar();
   },
   showFormStatus(response) {
     if (this.isReadOnly) return;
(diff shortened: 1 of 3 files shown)


test/integration/forms/form.js
```diff
@@ -778,6 +772,72 @@ context('Patient Action Form', function() {
       .get('iframe')
       .should('have.attr', 'src', '/formapp/');
   });
+
+  specify('store expanded state in localStorage', function() {
```
nmajor25, line 786: A quick explanation of what's going on in this test:

1. Set localStorage to have `isExpanded: false` value.

2. Load a page with a read only form (`/patient-action/1/form/22222`).

3. Verify that the correct state is shown on the initial page load (expanded vs. not expanded).

    - Check initial page state to verify that the form is NOT expanded

    - Verify that the sidebar is in the correct state:
      - Sidebar should be shown
      - Sidebar toggle button should be in the active state (has class `"is-selected"`)

    - Verify that the expand button is in the correct state:
      - Tooltip text should be `"Increase Width"`
      - Favicon image being used should be `"fa-expand-alt"`

4. Check that localStorage is updated correctly when UI state is changed (expanded <=> not expanded).

    - Switch from not expanded => expanded:
      - Change state by clicking on the `"Increase Width"` button (form should now be fully expanded)
      - Verify localStorage isExpanded value was switched to `true`

    - Switch from expanded => not expanded
      - Change state by clicking on the `"Decrease Width"` button (form should now be NOT expanded)
      - Verify localStorage isExpanded value was switched to `false`

src/js/apps/forms/form/form-patient_app.js
```diff
@@ -36,6 +43,7 @@ export default App.extend({
     this.patient = patient;
     this.form = form;
     this.isReadOnly = this.form.isReadOnly();
+    this.currentUser = Radio.request('bootstrap', 'currentUser');
```
paulfalgout: might as well do this on line 30 in the `onBeforeStart` and use it in `initFormState`

test/integration/forms/form.js
```diff
@@ -1058,6 +1117,61 @@ context('Patient Form', function() {
       .find('.alert')
       .contains('Insufficient permissions');
   });
+
+  specify('store expanded state in localStorage', function() {
```
nmajor25, line 1131: Same as test on line [ 776](https://github.com/RoundingWell/care-ops-frontend/pull/639/commits/4dfc2b57084f5f47afff0382b23fc2096c18e338#diff-f35721f6910e84ec70dafbba4d97ec92f15f390497573168f0d4ad9624035377R776) but for the `/patient/1/form/22222` route.

test/integration/forms/form.js
```diff
@@ -1162,4 +1276,4 @@ context('Preview Form', function() {
     cy
       .go('back');
   });
-});
+});
```
paulfalgout: missing ending newline.  I think http://EditorConfig.org would prevent this

src/js/apps/forms/form/form-patient_app.js
```diff
@@ -1,5 +1,7 @@
 import Radio from 'backbone.radio';
 import dayjs from 'dayjs';
+import store from 'store';
+import { extend } from 'underscore';
```
paulfalgout: super nitpick, but I think the underscore import should be first above Radio

paulfalgout (changes requested):
Looks great.  only some small stuff

src/js/apps/forms/form/form-patient_app.js
nmajor25: Ahhh, that's a good way to do it. I hemmed and hawed a lot on where to put that.

test/integration/forms/form.js
nmajor25: Okay, I'll get EditorConfig configured.

src/js/apps/forms/form/form-patient_app.js
nmajor25: 10-4, I'll get that updated.

paulfalgout (changes requested):
I'm not sure what happened to coveralls the 1st time around, but the report is legit:

https://coveralls.io/builds/47020913/source?filename=src%2Fjs%2Fapps%2Fforms%2Fform%2Fform-patient_app.js#L86
and
https://coveralls.io/builds/47020913/source?filename=src%2Fjs%2Fapps%2Fforms%2Fform%2Fform_app.js#L149

src/js/apps/forms/form/form-patient_app.js
```diff
@@ -79,11 +79,14 @@ export default App.extend({
...
-    'change': 'onChangeState',
-    'change:isExpanded': 'showSidebar',
+    'change:isExpanded': 'onChangeIsExpanded',
   },
-  onChangeState(state) {
-    store.set(`form-state_${ this.currentUser.id }`, { isExpanded: this.getState('isExpanded') });
+  onChangeIsExpanded() {
+    const isExpanded = this.getState('isExpanded');
+
+    store.set(`form-state_${ this.currentUser.id }`, { isExpanded: isExpanded });
```
paulfalgout: super nit-pick.  This can be `{ isExpanded }`

paulfalgout (changes requested):

src/js/apps/forms/form/form-patient_app.js
nmajor25: Good call, that's cleaner.

paulfalgout (approved):

PR #640, opened 2022-03-04
Move sidebar widgets to generic patient widgets

paulfalgout:
Shortcut Story ID: [sc-27967]

### DIFF
--- a/src/js/views/patients/widgets/widgets.js
+++ b/src/js/views/patients/widgets/widgets.js
@@ -1,0 +1,239 @@
+import { each, map, propertyOf, reduce, extend, isFunction } from 'underscore';
+import Radio from 'backbone.radio';
+import { View, CollectionView } from 'marionette';
+import dayjs from 'dayjs';
+import parsePhoneNumber from 'libphonenumber-js/min';
+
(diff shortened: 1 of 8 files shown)


src/js/views/patients/widgets/widgets.js
```diff
@@ -0,0 +1,239 @@
+import { each, map, propertyOf, reduce, extend, isFunction } from 'underscore';
+import { View, CollectionView } from 'marionette';
+import dayjs from 'dayjs';
+import Radio from 'backbone.radio';
```
nmajor25: A nitpicky question... Is the import order correct here?

Should `Radio` be above `Marionette`?

paulfalgout: oh yeah probably so

nmajor25 (commented):
Everything looks good to me!

As I get more acquainted with the codebase, I'll be able to provide more detailed/valuable reviews. But, everything you're doing in this makes sense and the code in each file looks good.

And the tests run and pass on my machine locally, and the functionality in the sidebar works after you split the widget into it's own file.

PR #648, opened 2022-03-14
Add Ipad touch support for tooltips

nmajor25:
Shortcut Story ID: [sc-27936]

Adds better support for tooltips on touch devices. The change was specifically made to accommodate Ipad devices.

Currently, the `mouseenter` event is being used to open tooltips. Which works smoothly on non-touch devices, but not as well on touch screens. This pull request uses `pointerenter` to open tooltips on conventional desktop devices. And also adds a `pointerdown` event listener to open tooltips on touch screens. These changes add touch support without breaking anything on the non-touch device experience.

These code changes caused many tests to break where tooltips were opened. To fix that, instances of `trigger('mouseover')` were replaced with `trigger('pointerover)` in the tests.

Also, some additional testing was added to the tooltip component to ensure the tooltips open on touch screens using the `pointerdown` event trigger.

### DIFF
--- a/test/integration/components/tooltip.js
+++ b/test/integration/components/tooltip.js
@@ -61,7 +61,21 @@ context('Tooltip', function() {
         .get('@hook')
         .contains(model.id)
         .as('button')
-        .trigger('mouseover');
+        .trigger('pointerover');
+
+      cy
+        .get('.tooltip')
+        .contains(model.id);
+
+      cy
+        .get('@button')
+        .trigger('mouseout');
+
+      cy
+        .get('@hook')
+        .contains(model.id)
+        .as('button')
+        .trigger('pointerdown');
 
       cy
         .get('.tooltip')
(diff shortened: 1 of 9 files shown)


test/integration/components/tooltip.js
```diff
@@ -61,7 +61,21 @@ context('Tooltip', function() {
         .get('@hook')
         .contains(model.id)
         .as('button')
-        .trigger('mouseover');
+        .trigger('pointerover');
+
+      cy
```
nmajor25, line 66: Add additional test to make sure the tooltips open when the `pointerdown` event is fired on the element.

test/integration/components/tooltip.js
```diff
@@ -89,7 +103,21 @@ context('Tooltip', function() {
...
```
nmajor25, line 108: Add additional test to make sure the tooltips open when the `pointerdown` event is fired on the element.

Same as what's done for the `'Displaying vertical positioning'` section above.

paulfalgout (approved):

PR #649, opened 2022-03-14
WIP form change reducers and before submit

paulfalgout:
Shortcut Story ID: [sc-28065]

### DIFF
--- a/src/js/formapp.js
+++ b/src/js/formapp.js
@@ -8,7 +8,7 @@ import 'sass/formapp/bootstrap.min.css';
 
 import 'sass/formapp-core.scss';
 
-import { extend, map, reduce } from 'underscore';
+import { extend, map, reduce, debounce } from 'underscore';
 import Backbone from 'backbone';
 
 import { versions } from './config';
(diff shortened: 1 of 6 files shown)


src/js/formapp.js
```diff
@@ -106,15 +105,53 @@ function getSubmission(formData, formSubmission, reducers, evalContext) {
...
+    context.hasChanged = FormioUtils.evaluate(hasChangedFunction, context);
+    return FormioUtils.evaluate(reducer, context) || memo;
+  }, structuredClone(form.submission.data));
+
+  form.setSubmission({ data });
+
+  prevSubmission = structuredClone(form.submission.data);
+
+  isChanging = false;
+}, 300);
```
nmajor25, line 127: What's the purpose of the `300` milliseconds wait before calling the `onChange()` function?

Not questioning the choice, just curious what it does.

nmajor25 (approved):
Looks good to me!

src/js/formapp.js
paulfalgout: I think I got the number from the debounce that form.io is doing on their change event, but essentially what this does is allow for a bunch of changes while only calling onChange once every 300ms.. So for instance if we removed this we'd run all of the onChanges on every keystroke inside of an input on a form.  It might be better to be a bit quicker actually, but for now it should mirror current response times on data changes.  As opposed to throttle, it's triggered at the end of the time so that you always get the latest change when the event is triggered.

nmajor25: Interesting... okay, thanks!

PR #654, opened 2022-03-17
Update Font Awesome to version 6.1.0

nmajor25:
Shortcut Story ID: [sc-28100]

This pull request updates the Font Awesome dependency version from v5 to v6. Version 6 should be backwards compatible. So styling in the app should not be effected.

Since some icons were renamed in Version 6, some CSS selectors need to be updated in the app to match those changes.

This also causes a number of Cypress tests to break. Therefore, this pull request includes fixes for those tests.

[Full list of changed icon names](https://fontawesome.com/docs/web/setup/upgrade/whats-changed#icons-renamed-in-version-6).


paulfalgout (changes requested):
Ah ok so some of the class names applied on the svgs have changed.  This is actually problematic.  The tests needed to be updated, but we also need to look through all of our styles for any user of these classes to make sure they're updated as well.  For instance search for `.fa-expand-alt` and you should find that one of the widget styles refers to this class.

nmajor25:
Missed those, okay I'll go through and get those classes updated.

src/js/views/globals/prelogin/prelogin.css
```diff
@@ -102,7 +102,7 @@
 }
 
 
-.prelogin__continue-button .fa-sign-in-alt {
+.prelogin__continue-button .fa-right-to-bracket {
```
nmajor25, line 105: We could keep this the same and it wouldn't modify any of the styling. But I figured we'd update it to match the new name Font Awesome gave the icon.

It coincides with the Handlebars template in the `src/js/views/globals/prelogin/login-prompt.hbs` file.

src/js/views/globals/prelogin/login-prompt.hbs
```diff
@@ -4,7 +4,7 @@
 
 <button class="prelogin__continue-button button--green js-login">
   Continue to sign in
-  <svg aria-hidden="true" focusable="false" data-prefix="fas" data-icon="sign-in-alt" role="img" xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 512 512" class="svg-inline--fa fa-sign-in-alt">
+  <svg aria-hidden="true" focusable="false" data-prefix="fas" data-icon="right-to-bracket" role="img" xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 512 512" class="svg-inline--fa fa-right-to-bracket">
```
nmajor25, line 7: Coincides with the change to the CSS selector in the `src/js/views/globals/prelogin/prelogin.css` file.

paulfalgout (approved):

PR #656, opened 2022-03-22
Only fetch patient fields assigned with widgets for the sidebar

paulfalgout:
Shortcut Story ID: [sc-28116]


nmajor25 (approved):
Everything looks good to me! 

The sidebar widgets still work as expected and the tests run successfully. And each file change you made looks good as well.

PR #657, opened 2022-03-22
Add org setting for reduced schedule

nmajor25:
Shortcut Story ID: [sc-28114]

Add a `reduced_patient_schedule` organization setting to a clinician. When a clinician has a `reduced_patient_schedule` setting to `true`, only the `Schedule` element will be shown in the  `nav_app` sidebar.

This pull request also includes a Cypress test that verifies that the `nav_app` sidebar appears correctly for a clinician with this setting.

### DIFF
--- a/test/integration/patients/worklist/schedule.js
+++ b/test/integration/patients/worklist/schedule.js
@@ -579,6 +579,43 @@ context('schedule page', function() {
       .should('be.empty');
   });
 
+  specify('reduced patient schedule employee', function() {
+    cy
+      .server()
+      .routeCurrentClinician(fx => {
+        fx.data.id = '123456';
+        fx.data.attributes.access = 'employee';
+        fx.data.attributes.enabled = true;
+        return fx;
+      })
+      .routeSettings(fx => {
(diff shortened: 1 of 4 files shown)


test/integration/patients/worklist/schedule.js
```diff
@@ -579,6 +579,43 @@ context('schedule page', function() {
...
+        return fx;
+      })
+      .routeSettings(fx => {
+        const reducedPatientSchedule = _.find(fx.data, setting => setting.id === 'reduced_patient_schedule');
+        reducedPatientSchedule.attributes.value = true;
+
+        return fx;
+      })
+      .routeActions()
+      .visit('/schedule')
```
paulfalgout, line 598: This is correct, but just a note that we should be able to change this to `visit()` when the epic is complete.

paulfalgout (approved):

test/integration/patients/worklist/schedule.js
nmajor25: Okay, good call. Would have probably missed that. Thanks!

PR #659, opened 2022-03-25
Add reduced patients schedule app

nmajor25:
Shortcut Story ID: [sc-28115]

If a clinician has the `reduced_patient_schedule = true` setting, we'll show them a reduced version of the Schedule app. In this scenario, the nav sidebar will only contain the `Schedule` element and the bulk edit and filter options will be disabled and not shown to the user.

Also, the `"/schedule"` URL route will be used by default for these clinicians. Instead of the conventional `"/worklist/owned-by"`.

**Note**: The `reduced_patient_schedule` setting will eventually be replaced by RBAC.

### DIFF
--- a/src/js/apps/patients/reduced-patients-main_app.js
+++ b/src/js/apps/patients/reduced-patients-main_app.js
@@ -6,50 +6,38 @@ import RouterApp from 'js/base/routerapp';
 
 import FlowApp from 'js/apps/patients/patient/flow/flow_app';
 import PatientApp from 'js/apps/patients/patient/patient_app';
-import WorklistApp from 'js/apps/patients/worklist/worklist_app';
-import ScheduleApp from 'js/apps/patients/schedule/schedule_app';
+import ReducedScheduleApp from 'js/apps/patients/schedule/reduced_schedule_app';
 
 export default RouterApp.extend({
   routerAppName: 'PatientsApp',
-
   childApps() {
-    const worklistApp = WorklistApp;
-
     return {
       flow: FlowApp,
       patient: PatientApp,
-      ownedBy: worklistApp,
-      forRole: worklistApp,
-      newPastDay: worklistApp,
-      pastThree: worklistApp,
-      lastThirty: worklistApp,
-      schedule: ScheduleApp,
+      schedule: ReducedScheduleApp,
     };
   },
-
   defaultRoute() {
-    const defaultRoute = 'worklist';
-    const defaultWorklist = 'owned-by';
+    const defaultRoute = 'schedule';
 
     this.routeAction(defaultRoute, () => {
       defer(()=> {
-        this.navigateRoute(defaultRoute, defaultWorklist);
-        Radio.request('nav', 'select', this.routerAppName, defaultRoute, [defaultWorklist]);
-        this.setLatestList(defaultRoute, [defaultWorklist]);
-        this.showPatientsWorklist(defaultWorklist);
+        this.navigateRoute(defaultRoute);
+        Radio.request('nav', 'select', this.routerAppName, defaultRoute);
+        this.setLatestList(defaultRoute);
+        this.showSchedule();
       });
     });
   },
-
   eventRoutes() {
     return {
       'default': {
         action: 'defaultRoute',
         route: '',
       },
-      'worklist': {
-        action: 'showPatientsWorklist',
-        route: 'worklist/:id',
+      'schedule': {
+        action: 'showSchedule',
+        route: 'schedule',
         isList: true,
       },
       'patient:dashboard': {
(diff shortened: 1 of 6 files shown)


src/js/views/patients/schedule/schedule_views.js
```diff
@@ -116,15 +116,18 @@ const DayItemView = View.extend({
...
+    const isReduced = this.state.get('isReduced');
+
     return {
       isOverdue: this.model.isOverdue(),
       state: state.get('name'),
       stateOptions: state.get('options'),
       patient: this.model.getPatient().attributes,
       form: this.model.getForm(),
-      isSelected: this.state.isSelected(this.model),
+      isSelected: !isReduced && this.state.isSelected(this.model),
```
nmajor25, line 127: When the `StateModel` is removed from the `reduced_schedule_app.js` file, this line of code causes an error when the current clinician is a reduced schedule user. It looks for the `this.state.isSelected()` function but can’t find it since it usually comes from the `StateModel` in the `schedule_state.js` file.

Adding `!isReduced` stops the `this.state.isSelected()` function from being called when the view is in the `isReduced` state (when the `StateModel` isn’t present). And keeps it functioning the same in the normal, non-reduced state.

test/integration/globals/default-route.js
```diff
@@ -28,6 +28,26 @@ context('patient page', function() {
       .should('contain', 'worklist/owned-by');
   });
 
+  specify('current clinician has reduced patient schedule access', function() {
```
nmajor25, line 31: Verifies that the page routes to `”/schedule”` by default if the clinician has `”employee”` access and has the `reduced_patient_schedule = true` setting.

test/integration/globals/default-route.js
```diff
@@ -28,6 +28,26 @@ context('patient page', function() {
...
+
+        return fx;
+      })
+      .visit('/');
+
+    cy
+      .url()
+      .should('contain', 'schedule');
+  });
+
```
nmajor25, line 50: Verifies that the page routes to `”/schedule”` by default if the clinician has `”employee”` access and has the `reduced_patient_schedule = true` setting.

test/integration/patients/worklist/schedule.js
```diff
@@ -614,6 +618,34 @@ context('schedule page', function() {
...
+      .its('length')
+      .should('be.gt', 0);
+
+    cy
+      .get('.schedule-list__table')
+      .find('.schedule-list__list-row .schedule-list__day-list')
+      .first()
+      .find('.schedule-list__day-list-row .js-select')
+      .should('not.exist');
   });
```
nmajor25, line 649: Overview of the tests:

- Verify the page is routed to “/schedule” by default.
- Verify sidebar only contains the “Schedule” element and that is has the `is-selected` class name.
- Verify the filtering and bulk edit elements are not shown.
  - Bulk edit button, owner filter, and date filter should all not exist.
- Verify the list of actions are loaded into the page.
  - Not checking anything specific here. Just that the list contains more than `0` actions.
- Verify that individual list items do not contain the checkbox element.

test/integration/patients/worklist/schedule.js
```diff
@@ -844,7 +876,6 @@ context('schedule page', function() {
       })
       .as('patchActionFail');
 
-
```
nmajor25, line 847: Just removed an extra space that I noticed elsewhere in the file.

src/js/apps/patients/patients-main_app.js
```diff
@@ -13,7 +13,6 @@ export default RouterApp.extend({
   routerAppName: 'PatientsApp',
 
   childApps() {
-    /* istanbul ignore next */
```
paulfalgout, line 16: :+1:  I bet this was added in a previous feature where there was logic here

src/js/apps/patients/reduced-patients-main_app.js
```diff
@@ -0,0 +1,78 @@
...
+
+import RouterApp from 'js/base/routerapp';
+
+import FlowApp from 'js/apps/patients/patient/flow/flow_app';
+import PatientApp from 'js/apps/patients/patient/patient_app';
+import ReducedScheduleApp from 'js/apps/patients/schedule/reduced_schedule_app';
+
+export default RouterApp.extend({
+  routerAppName: 'PatientsApp',
+  childApps() {
```
paulfalgout, line 13: you can actually define this and `eventRoutes` as POJOs since there's no extra javascript going on.  Although sometimes they make sense to keep as functions if we might be adding logic in the near future or in this case because it's consistent.. might as well.  Just mentioning to point out the simplest form that could be in this case, but no changes needed for this PR

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -116,15 +116,18 @@ const DayItemView = View.extend({
...
       flow: this.model.getFlow() && this.model.getFlow().get('name'),
       hasOutreach: this.model.hasOutreach(),
+      isReduced: isReduced,
```
paulfalgout, line 130: could be just `isReduced,`

src/js/views/patients/schedule/schedule_views.js
paulfalgout: Seems like a reasonable solution for this mini-fork.  I hope we don't regret the simplification later, but works great for now.

paulfalgout (approved):
Solid work.  I'm questioning my thoughts around making a `reduced-patients-main_app`. Looking at this it may have been better to just override the default and the schedule route in` patients-main_app`. Though I suppose the complexity there is also removing the routes not supported.. I dunno.  Good for now.  We can refactor later if a better idea arises or we run into issues.

src/js/views/patients/schedule/schedule_views.js
nmajor25: Ah yes, that would be cleaner...

src/js/apps/patients/reduced-patients-main_app.js
nmajor25: Okay makes sense. I'd be happy to come back to this and change them into a plain object format. Simpler the better.

src/js/views/patients/schedule/schedule_views.js
nmajor25: Why would we regret it in the future? Out of curiosity.

PR #666, opened 2022-04-01
Display form widgets container

nmajor25:
Shortcut Story ID: [sc-28205]

Display a widget container if there are widgets defined on a form. If there are no widgets defined for a form, the section should be hidden.

The feature should look like the image below:

Screen Shot 2022-03-24 at 4 43 30 PM

Cypress tests were also updated to ensure a basic version of the widgets container displays properly on the page. As this feature is built out more in the future, more detailed tests will be added for the functionality of the individual widgets inside the container.

### DIFF
--- a/src/js/apps/forms/form/form-patient_app.js
+++ b/src/js/apps/forms/form/form-patient_app.js
@@ -45,8 +45,7 @@ export default App.extend({
     this.form = form;
     this.isReadOnly = this.form.isReadOnly();
 
-    const formWidgets = this.form.getWidgets();
-    const widgets = Radio.request('entities', 'widgets:collection', formWidgets);
+    const widgets = this.form.getWidgets();
 
     this.startFormService();
 
--- a/src/js/apps/forms/form/form-patient_app.js
+++ b/src/js/apps/forms/form/form-patient_app.js
@@ -3,8 +3,6 @@ import Radio from 'backbone.radio';
 import dayjs from 'dayjs';
 import store from 'store';
 
-import collectionOf from 'js/utils/formatting/collection-of';
-
 import App from 'js/base/app';
 
 import FormsService from 'js/services/forms';
@@ -48,7 +46,7 @@ export default App.extend({
     this.isReadOnly = this.form.isReadOnly();
 
     const formWidgets = this.form.getWidgets();
-    const widgets = Radio.request('entities', 'widgets:collection', collectionOf(formWidgets, 'id'));
+    const widgets = Radio.request('entities', 'widgets:collection', formWidgets);
 
     this.startFormService();
 
--- a/src/js/apps/forms/form/form-patient_app.js
+++ b/src/js/apps/forms/form/form-patient_app.js
@@ -3,6 +3,8 @@ import Radio from 'backbone.radio';
 import dayjs from 'dayjs';
 import store from 'store';
 
+import collectionOf from 'js/utils/formatting/collection-of';
+
 import App from 'js/base/app';
 
 import FormsService from 'js/services/forms';
@@ -45,9 +47,12 @@ export default App.extend({
     this.form = form;
     this.isReadOnly = this.form.isReadOnly();
 
+    const formWidgets = this.form.getWidgets();
+    const widgets = Radio.request('entities', 'widgets:collection', collectionOf(formWidgets, 'id'));
+
     this.startFormService();
 
-    this.showView(new LayoutView({ model: this.form, patient }));
+    this.showView(new LayoutView({ model: this.form, patient, widgets }));
 
     this.showForm();
 
--- a/src/js/views/forms/form/form.scss
+++ b/src/js/views/forms/form/form.scss
@@ -37,11 +37,6 @@
   margin-bottom: 16px;
   padding: 8px 8px 0 8px;
 
-  > div:first-of-type {
-    display: flex;
-    flex-wrap: wrap;
-  }
-
   .form__widgets-section {
     margin-bottom: 8px;
     padding: 0 8px;
--- a/src/js/views/forms/form/form.scss
+++ b/src/js/views/forms/form/form.scss
@@ -30,6 +30,24 @@
   }
 }
 
+.form__widgets {
+  background-color: $white;
+  border: 1px solid $black-90;
+  border-radius: 4px;
+  margin-bottom: 16px;
+  padding: 8px 8px 0 8px;
+
+  > div:first-of-type {
+    display: flex;
+    flex-wrap: wrap;
+  }
+
+  .form__widgets-section {
+    margin-bottom: 8px;
+    padding: 0 8px;
+  }
+}
+
 .form__sidebar {
   display: flex;
   flex: initial;
@@ -47,7 +65,7 @@
   border: 1px solid $black-90;
   border-bottom: 0;
   border-radius: 4px 4px 0 0;
-  height: calc(100% - 80px);
+  height: 100%;
   padding: 8px;
 
   iframe {
--- a/src/js/views/forms/form/form_views.js
+++ b/src/js/views/forms/form/form_views.js
@@ -179,10 +179,8 @@ const LayoutView = View.extend({
     widgets: '[data-widgets-region]',
   },
   templateContext() {
-    this.hasFormWidgets = this.getOption('widgets').length;
-
     return {
-      hasFormWidgets: this.hasFormWidgets,
+      hasFormWidgets: this.hasFormWidgets(),
     };
   },
   onRender() {
@@ -191,7 +189,7 @@ const LayoutView = View.extend({
       action: this.getOption('action'),
     }));
 
-    if (this.hasFormWidgets) {
+    if (this.hasFormWidgets()) {
       this.showChildView('widgets', new WidgetCollectionView({
         model: this.getOption('patient'),
         collection: this.getOption('widgets'),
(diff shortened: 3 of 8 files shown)


src/js/views/forms/form/form.scss
```diff
@@ -47,7 +65,7 @@
   border: 1px solid $black-90;
   border-bottom: 0;
   border-radius: 4px 4px 0 0;
-  height: calc(100% - 80px);
+  height: 100%;
```
nmajor25, line 63: From my manual testing in my local dev environment and what I could see in Cypress, the height of the iframe element maintains the correct height using simply `height: 100%`. 

There may be other side effects of this I'm not aware of in the UI...

src/js/views/forms/form/form.scss
```diff
@@ -30,6 +30,24 @@
...
 }
 
+.form__widgets {
+  background-color: $white;
+  border: 1px solid $black-90;
+  border-radius: 4px;
+  margin-bottom: 16px;
+  padding: 8px 8px 0 8px;
+
+  > div:first-of-type {
```
nmajor25: We discussed possibly replacing this with a class name, but I couldn't figure out how to do that. This works, but doesn't feel like the cleanest solution.

 Maybe you could find a better solution instead of what I did.

This is the current HTML structure of the form widgets container:

```html
<div class="form__widgets flex">
  <div> <! –– automatically added somewhere by the widgets view? ––>
    <div class="form__widgets-section">
      <! –– widget heading and item ––>
    </div>
    <div class="form__widgets-section">
      <! –– widget heading and item ––>
    </div>
  </div>
</div>
```

That empty `<div>` is added somewhere. By the widgets view? I'm not sure. But I couldn't find a place to attach a class name to it.

test/integration/forms/form.js
```diff
@@ -598,6 +598,12 @@ context('Patient Action Form', function() {
       .contains('Read Only')
       .should('be.disabled');
 
+    cy
```
nmajor25, line 601: A simple test that verifies the form widgets container is shown and that it contains a few widgets.

Same test was also added for the patient form as well.

src/js/views/forms/form/form.scss
paulfalgout: This is interesting.  The calc I know is a hold-over style from the previous app which supported old-IE.  So it's possibly related.  We should make sure forms are tested on iPad for this release

src/js/apps/forms/form/form-patient_app.js
```diff
@@ -45,9 +47,12 @@ export default App.extend({
     this.form = form;
     this.isReadOnly = this.form.isReadOnly();
 
+    const formWidgets = this.form.getWidgets();
+    const widgets = Radio.request('entities', 'widgets:collection', collectionOf(formWidgets, 'id'));
```
paulfalgout: I know the `collectionOf` stuff was probably in the patient sidebar app, but I think we should move this line into the `getWidgets` such that the function returns the instantiated widgets collection.  The stuff on the form options and org settings is a little un-standard so there's not clear patterns but I think it makes sense to work with the widgets as a collection at the earliest possible opportunity.

src/js/views/forms/form/form_views.js
```diff
@@ -154,6 +156,11 @@ const LayoutView = View.extend({
           </div>
         </div>
       </div>
+      {{#if formHasWidgets}}
```
paulfalgout: let's make this `hasFormWidgets` where `has` `is` etc are the prefix for the check

src/js/views/forms/form/form_views.js
```diff
@@ -169,12 +176,31 @@ const LayoutView = View.extend({
       replaceElement: false,
     },
     status: '[data-status-region]',
+    widgets: '[data-widgets-region]',
+  },
+  templateContext() {
+    return {
+      formHasWidgets: this.formHasWidgets,
```
paulfalgout: I think we should move the logic from initialize to here for the widgets even though it doesn't change, I think it's good practice.  Also I believe at this point widgets is always defined and will have a length.  So I suspect that the `widget &&` is a missing else branch in the coverage tests.

src/js/views/forms/form/form_views.js
```diff
@@ -169,12 +176,31 @@ const LayoutView = View.extend({
...
+    };
+  },
+  initialize({ patient, widgets }) {
```
paulfalgout: Nothing is using the attached patient and only the `templateContext` will use `widgets` so it could probably just be a `this.getOption` in `templateContext` or we should convert the other `getOption`s. There's no hard and fast rule as to when to use `getOption` over attaching at `initialize`. They should always produce the same results.. So it's sort of a judgement call to whether a few `getOption`s are cleaner than an initialize.  I typically use initialize if more than a couple methods will be using the option.  But generally if we attach something we shouldn't use `getOption` as well.

src/js/views/forms/form/form_views.js
```diff
@@ -169,12 +176,31 @@ const LayoutView = View.extend({
...
+    this.formHasWidgets = widgets && widgets.length;
   },
   onRender() {
     this.showChildView('contextTrail', new ContextTrailView({
       patient: this.getOption('patient'),
       action: this.getOption('action'),
     }));
+
+    if (this.formHasWidgets) {
+      this.showChildView('widgets', new WidgetCollectionView({
```
paulfalgout, line 193: I think you could add a `className: 'flex-wrap',` here to cover the style issue.  Or if every instance of `WidgetCollectionView` would benefit from the class that same line can be added to it's definition in the view file.

test/integration/forms/form.js
```diff
@@ -598,6 +598,12 @@ context('Patient Action Form', function() {
...
+      .get('.form__widgets')
+      .find('.form__widgets-section')
+      .its('length')
+      .should('be.gte', 0);
```
paulfalgout: I'm not sure I'm following the `gte` of 0.  Seems like if the `find` works this'd always be true?    I think `should('exist')` might cover this rather than needing the length.  (Although technically should exist is also redundant it's a little clearer that it wasn't just an unfinished line)

paulfalgout (changes requested):
Cool. Pretty clean.

src/js/views/forms/form/form.scss
nmajor25: I'll make a note of this in the Shortcut story for Sofia.

src/js/apps/forms/form/form-patient_app.js
nmajor25: Makes sense to me. Get the widgets data in the correct format as soon as possible. Will update that for both `form_app.js` and `form-patient_app.js`.

src/js/views/forms/form/form_views.js
nmajor25: Will get this updated.

src/js/views/forms/form/form_views.js
nmajor25: I'll use the `getOption()` method to keep consistent with what's used elsewhere in the file. Don't have a preference to either one. Would rather be consistent.

test/integration/forms/form.js
nmajor25: You're right, definitely redundant... I'll replace it with the `should('exist')`.

src/js/apps/forms/form/form-patient_app.js
```diff
@@ -45,9 +45,12 @@ export default App.extend({
     this.form = form;
     this.isReadOnly = this.form.isReadOnly();
 
+    const formWidgets = this.form.getWidgets();
```
paulfalgout: Actually I think ideally this is just `const widget = this.form.getWidget();` where the entity request is in the form model

src/js/views/forms/form/form_views.js
```diff
@@ -169,12 +176,29 @@ const LayoutView = View.extend({
       replaceElement: false,
     },
     status: '[data-status-region]',
+    widgets: '[data-widgets-region]',
+  },
+  templateContext() {
+    this.hasFormWidgets = this.getOption('widgets').length;
```
paulfalgout: this can probably be `const hasFormWidgets =` then the return can be ` { hasFormWidgets }` no need for the `:`

paulfalgout (changes requested):
couple of nit-picks

src/js/apps/forms/form/form-patient_app.js
nmajor25: Yes it definitely can. Would be cleaner that way.

src/js/views/forms/form/form_views.js
nmajor25: `this.hasFormWidgets` is used in the `onRender()` method:

```js
 if (this.hasFormWidgets) {
  this.showChildView('widgets', new WidgetCollectionView({
    model: this.getOption('patient'),
    collection: this.getOption('widgets'),
    className: 'flex flex-wrap',
    itemClassName: 'form__widgets-section',
  }));
}
```

If I switch it to `const hasFormWidgets`, `this.hasFormWidgets` won't be available there. And therefore will always be `false` and not display the widget data.

paulfalgout: ah I see.  yeah so I guess there's nothing particularly wrong with this.. but it's a little unorthodox to attach something in the `templateContext` for use elsewhere.  I think either we make a `hasFormWidgets` method on the view and use it twice or it'd also be ok I think to change the `if` in the `onRender` to just be the same check of `this.getOption('widgets').length`

nmajor25: Okay, makes sense. I think adding the `hasFormWidgets` method is a good idea.

paulfalgout (approved):

paulfalgout:
The failing test is a flake.  we're good here

PR #669, opened 2022-04-06
Release fixes

paulfalgout:
Shortcut Story ID: [sc-28459]

Two small but unrelated commits here.

The first is a caching issue we're spitballing solutions for with @BenCoffeed that I'm sneaking in

Second is a guess that seems to work, but form.io is such spaghetti with bad documentation that I'm not sure exactly what this flag is doing or if this is correct...  _but_ it seems like it will solve the current `setTimeout` issue without breaking the reason why this `if` was there.  But I can't explain why or if this covers every use case.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,7 +12,8 @@
   "globals": {
     "_PRODUCTION_": true,
     "_DEVELOP_": true,
-    "_E2E_": true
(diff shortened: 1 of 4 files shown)


nmajor25 (approved):
Looks good to me. Hopefully that fixes the form bug!

PR #674, opened 2022-04-06
Remove actions with no due date from the schedule worklist

nmajor25:
Shortcut Story ID: [sc-28337]

This pull request excludes unscheduled actions from being included in the schedule worklist. It does so by filtering out any actions that have a due date value equal to `"null"`.

This is a bug that resulting from this pull request: https://github.com/RoundingWell/care-ops-frontend/pull/659.

### DIFF
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -294,6 +294,7 @@ const ScheduleListView = CollectionView.extend({
       return !view.isEmpty();
     }
 
+    // 'null' string is a key from groupBy
     if (view.model.get('date') === 'null') {
       return false;
     }
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -294,6 +294,10 @@ const ScheduleListView = CollectionView.extend({
       return !view.isEmpty();
     }
 
+    if (view.model.get('date') === 'null') {
+      return false;
+    }
+
     return true;
   },
   initialize({ state }) {


src/js/views/patients/schedule/schedule_views.js
```diff
@@ -294,6 +294,10 @@ const ScheduleListView = CollectionView.extend({
       return !view.isEmpty();
     }
 
+    if (view.model.get('date') === 'null') {
+      return false;
+    }
+
```
nmajor25, line 301: When an action has no due date, its `date` value is equal to the `"null"` string. This code just adds a check for that and tells the `ScheduleListView` to not include that action in the worklist if that's the case.

This change will apply to both the normal and reduced versions of the schedule.

shadowhand: A ... string `'null'` ... ? That doesn't seem right.

paulfalgout: Yeah it's odd, but it's a result of the `groupBy` for a case that's not supposed to happen.  I think we need a comment here as otherwise I'll be tempted to remove the string in the future.  This is a key from the `groupBy`

paulfalgout (changes requested):
This needs a test.  Might as well add https://app.shortcut.com/roundingwell/story/28429/add-tests-for-test-coverage to this work as well since we're in the tests for this app.

src/js/views/patients/schedule/schedule_views.js
paulfalgout: Also @shadowhand this was a quick way to solve an edge case, but attempting to leave the `max` null on the endpoint didn't seem to work for not getting actions with `due_dates`.  Since this situation is apparently not a use case, I wasn't too worried about a small FE filter solving the issue vs looking into the endpoint further, but probably figured it might be worth mentioning.

nmajor25:
Added that comment. And I made a note in the https://app.shortcut.com/roundingwell/story/28429/add-tests-for-test-coverage story about including this in the test coverage updates.

PR #679, opened 2022-04-13
Load form widgets with field requests

nmajor25:
Shortcut Story ID: [sc-28372]

Update the form header widgets to include field requests.

The feature should look like the image below:

161309535-146cbb74-d72b-4a51-8de4-02f6710f7eac

### DIFF
--- a/src/js/apps/forms/widgets/widgets_header_app.js
+++ b/src/js/apps/forms/widgets/widgets_header_app.js
@@ -1,0 +1,25 @@
+import { get, map } from 'underscore';
+import Radio from 'backbone.radio';
+
+import collectionOf from 'js/utils/formatting/collection-of';
+import App from 'js/base/app';
+
+import { FormWidgetsHeaderView } from 'js/views/forms/form/widgets/widget_header_view';
+
+export default App.extend({
+  beforeStart({ patient, form }) {
+    const formWidgets = form.getWidgets();
+    const widgets = Radio.request('entities', 'widgets:collection', collectionOf(get(formWidgets, 'widgets'), 'id'));
+    const requests = map(get(formWidgets, 'fields'), fieldName => patient.fetchField(fieldName));
+    requests.unshift(widgets);
+    return requests;
+  },
+  onStart({ patient }, widgets) {
+    if (!widgets.length) return;
+
+    this.showView(new FormWidgetsHeaderView({
+      model: patient,
+      collection: widgets,
+    }));
+  },
+});
--- a/src/js/apps/forms/widgets/widgets_header_app.js
+++ b/src/js/apps/forms/widgets/widgets_header_app.js
@@ -1,20 +1,19 @@
-import { get, map } from 'underscore';
+import { map } from 'underscore';
 import Radio from 'backbone.radio';
 
-import collectionOf from 'js/utils/formatting/collection-of';
 import App from 'js/base/app';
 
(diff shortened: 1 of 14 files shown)


src/js/apps/forms/widgets/widgets_header_app.js
```diff
@@ -0,0 +1,25 @@
...
+import collectionOf from 'js/utils/formatting/collection-of';
+import App from 'js/base/app';
+
+import { FormWidgetsHeaderView } from 'js/views/forms/form/widgets/widget_header_view';
+
+export default App.extend({
+  beforeStart({ patient, form }) {
+    const formWidgets = form.getWidgets();
+    const widgets = Radio.request('entities', 'widgets:collection', collectionOf(get(formWidgets, 'widgets'), 'id'));
+    const requests = map(get(formWidgets, 'fields'), fieldName => patient.fetchField(fieldName));
```
paulfalgout: Check the shortcut card.  I'd like to refactor and remove `fetchField` here and elsewhere

paulfalgout (changes requested):

src/js/entities-service/entities/forms.js
```diff
@@ -36,9 +36,14 @@ const _Model = BaseModel.extend({
     return get(this.get('options'), 'beforeSubmit', defaultBeforeSubmit);
   },
   getWidgets() {
-    const widgets = get(this.get('options'), 'widgets', []);
+    const formWidgets = get(this.get('options'), 'widgets', { fields: [], widgets: [] });
```
paulfalgout: You don't really need the defaults on this `get`

src/js/views/forms/form/widgets/widget_header.scss
```diff
@@ -0,0 +1,12 @@
+.form__widgets {
```
paulfalgout: I think we should probably change this file name to `form-widgets.scss` and this class to `form-widgets` and `form-widgets__section`. That way when you see the class name you can find the file by the block

test/integration/forms/form.js
```diff
@@ -853,6 +855,87 @@ context('Patient Action Form', function() {
...
+        return fx;
+      }, 'test-field');
+
+    cy
+      .visit('/patient-action/1/form/55555')
+      .wait('@routeFormDefinition')
+      .wait('@routePatientByAction')
+      .wait('@routeAction')
+      .wait('@routeWidgets')
+      .wait('@routePatientFieldtest-field');
```
paulfalgout: Since we're introducing `test-field` for this test can we rename it so it's camelCase if only for this `wait`.  I don't actually think there are any hyphenated field names other than the ones in our tests, so I think it's ok to change it here.  the `-field` here is just a little awkward looking

paulfalgout (changes requested):
Looks good.. just a couple of nit-picks.

test/integration/forms/form.js
nmajor25: Makes sense, I'll change that to camelCase.

src/js/views/forms/form/widgets/widget_header.scss
nmajor25: Will get that updated!

src/js/entities-service/entities/forms.js
nmajor25: True, just tested it and everything works without the defaults there.

paulfalgout (approved):

PR #683, opened 2022-04-19
Move patient name from breadcrumbs to form title

nmajor25:
Shortcut Story ID: [sc-28568]

Move the patient name from the breadcrumbs section to the form title.

It should look like the image below:

Screen Shot 2022-04-19 at 10 27 35 AM

The form title should use an ellipsis text overflow instead of wrapping to a new line.

The ellipsis text overflow should look like this:

Screen Shot 2022-04-19 at 10 26 47 AM

### DIFF
--- a/src/js/views/forms/form/form.scss
+++ b/src/js/views/forms/form/form.scss
@@ -15,17 +15,10 @@
   width: 100%;
 }
 
-.form__crumbs-title-section {
-  overflow: hidden;
-}
-
 .form__title {
   font-size: 24px;
   line-height: 1.2;
-  overflow: hidden;
   padding: 16px;
-  text-overflow: ellipsis;
-  white-space: nowrap;
 }
 
 .form__title-icon {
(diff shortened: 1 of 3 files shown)


src/js/views/forms/form/form.scss
```diff
@@ -15,10 +15,17 @@
   width: 100%;
 }
 
+.form__crumbs-title-section {
```
paulfalgout: there's an `overflow--hidden` class that'd probably work here.   One of a few that should probably be prefixed with `u-`  but none-the-less probably a good use case here.

However why does this div need a hidden overflow?

src/js/views/forms/form/form.scss
```diff
@@ -15,10 +15,17 @@
...
+  overflow: hidden;
+}
+
 .form__title {
```
paulfalgout: similarly you can probably throw `.u-text--overflow` on the form title element.

paulfalgout (changes requested):
~~I'm assuming a super long field title does the `...`s successfully?~~

Nevermind just saw the screenshots

paulfalgout (approved):

PR #685, opened 2022-04-22
Add tests for test coverage

nmajor25:
Shortcut Story ID: [sc-28429]

Overall test coverage dropped as a result of code changes made for the reduced schedule app epic (https://github.com/RoundingWell/care-ops-frontend/pull/659 & https://github.com/RoundingWell/care-ops-frontend/pull/674).

This pull request updates the Cypress tests for the reduced patient schedule app to fix the coverage drop. And some other dead code is also removed.

### DIFF
--- a/src/js/apps/patients/schedule/reduced_schedule_app.js
+++ b/src/js/apps/patients/schedule/reduced_schedule_app.js
@@ -10,12 +10,6 @@ import { LayoutView, ScheduleTitleView, TableHeaderView, ScheduleListView } from
 
 export default App.extend({
   onBeforeStart() {
-    if (this.isRestarting()) {
-      this.showScheduleTitle();
-      this.getRegion('list').startPreloader();
-      return;
-    }
-
     this.setState({ isReduced: true });
 
     this.showView(new LayoutView({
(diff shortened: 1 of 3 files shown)


src/js/apps/patients/schedule/reduced_schedule_app.js
```diff
@@ -10,12 +10,6 @@ import { LayoutView, ScheduleTitleView, TableHeaderView, ScheduleListView } from
 
 export default App.extend({
   onBeforeStart() {
-    if (this.isRestarting()) {
```
nmajor25, line 13: We're not restarting the app, so this was dead code.

src/js/entities-service/entities/patients.js
```diff
@@ -30,9 +30,6 @@ const _Model = BaseModel.extend({
   getFields() {
     return Radio.request('entities', 'patientFields:collection', this.get('_patient_fields'));
   },
-  fetchField(name) {
```
nmajor25, line 33: Dead code, so this was also removed.

test/integration/patients/worklist/schedule.js
```diff
@@ -582,10 +582,12 @@ context('schedule page', function() {
   specify('reduced patient schedule employee', function() {
```
nmajor25, line 582: I updated this test to load three actions into the reduced schedule list:

1. One for linking to a patient action.
2. One for linking to a flow action.
3. One that does not have a due date.

Using those three actions, I added the following to the test:

* Verify the search filter works.
* Verify that the list only includes two items. The patient action and flow action should show. But the action with no due date should not be included in the list.
* Verify the links for the patient and flow actions work as expected and take the user to the correct pages.

PR #687, opened 2022-04-26
Remove linking from the reduced schedule

nmajor25:
Shortcut Story ID: [sc-28553]

Remove row linking and patient name linking from the reduced version of the schedule. Link/hover CSS styling should also be removed.

Only the form link should be functioning for each day item in the schedule list.

Although links to the patient/flow are removed from the schedule list, clinicians should still be able to access those pages via direct link, search, or otherwise.

### DIFF
--- a/src/js/views/patients/schedule/schedule-list.scss
+++ b/src/js/views/patients/schedule/schedule-list.scss
@@ -79,7 +79,11 @@
     background: $greyed;
   }
 
-  &:hover:not(.is-selected) {
+  &.is-reduced {
+    cursor: default;
+  }
+
+  &:hover:not(.is-selected, .is-reduced) {
     background-color: color(light_blue, light);
   }
 }
@@ -166,7 +170,11 @@
 .schedule-list__action-form-icon {
   cursor: pointer;
 
-  &:hover {
+  &.is-reduced {
+    cursor: default;
+  }
+
+  &:hover:not(.is-reduced) {
     color: color(blue);
   }
 }
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -98,11 +98,11 @@ const DayItemView = View.extend({
     <td class="schedule-list__action-list-cell schedule-list__patient">
       <div class="schedule-list__state-patient">
         <span class="schedule-list__action-state action--{{ stateOptions.color }}">{{fa stateOptions.iconType stateOptions.icon}}</span><span class="schedule-list__search-helper">{{ state }}</span>&#8203;{{~ remove_whitespace ~}}
-        <span class="schedule-list__patient-name js-patient">{{ patient.first_name }} {{ patient.last_name }}</span>&#8203;
+        <span class="schedule-list__patient-name {{#unless isReduced}}js-patient{{/unless}} {{#if isReduced}}is-reduced{{/if}}">{{ patient.first_name }} {{ patient.last_name }}</span>&#8203;
       </div>
     </td>
     <td class="schedule-list__action-list-cell">
-      <span class="schedule-list__action-name js-action">{{ name }}</span>&#8203;{{~ remove_whitespace ~}}
+      <span class="schedule-list__action-name {{#unless isReduced}}js-action{{/unless}}">{{ name }}</span>&#8203;{{~ remove_whitespace ~}}
       <span class="schedule-list__search-helper">{{ flow }}</span>&#8203;{{~ remove_whitespace ~}}
     </td>
     <td class="schedule-list__action-list-cell schedule-list__action-details" data-details-region></td>
@@ -116,18 +116,16 @@ const DayItemView = View.extend({
   templateContext() {
     const state = this.model.getState();
 
-    const isReduced = this.state.get('isReduced');
-
     return {
       isOverdue: this.model.isOverdue(),
       state: state.get('name'),
       stateOptions: state.get('options'),
       patient: this.model.getPatient().attributes,
       form: this.model.getForm(),
-      isSelected: !isReduced && this.state.isSelected(this.model),
+      isSelected: !this.isReduced && this.state.isSelected(this.model),
       flow: this.model.getFlow() && this.model.getFlow().get('name'),
       hasOutreach: this.model.hasOutreach(),
-      isReduced: isReduced,
+      isReduced: this.isReduced,
     };
   },
   ui: {
@@ -142,6 +140,7 @@ const DayItemView = View.extend({
   initialize({ state }) {
     this.state = state;
     this.flow = this.model.getFlow();
+    this.isReduced = state.get('isReduced');
 
     this.listenTo(state, {
       'select:all': this.render,
@@ -149,6 +148,8 @@ const DayItemView = View.extend({
     });
   },
   onRender() {
+    if (this.isReduced) this.$el.addClass('is-reduced');
+
     this.showDetailsTooltip();
   },
   onClickSelect() {
@@ -162,6 +163,8 @@ const DayItemView = View.extend({
     Radio.trigger('event-router', 'form:patientAction', this.model.id, this.model.getForm().id);
   },
   onClick() {
+    if (this.isReduced) return;
+
     if (this.flow) {
       Radio.trigger('event-router', 'flow:action', this.flow.id, this.model.id);
       return;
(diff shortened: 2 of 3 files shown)


src/js/views/patients/schedule/schedule_views.js
```diff
@@ -142,13 +140,16 @@ const DayItemView = View.extend({
...
     this.flow = this.model.getFlow();
+    this.isReduced = state.get('isReduced');
 
     this.listenTo(state, {
       'select:all': this.render,
       'select:none': this.render,
     });
   },
   onRender() {
+    if (this.isReduced) this.$el.addClass('is-reduced');
```
nmajor25: If `isReduced === true`, this adds a `.is-reduced` classname to the `DayItemView` view. So, in that case the entire class for it would be `"schedule-list__day-list-row is-reduced"`.

That classname is used in the `src/js/views/patients/schedule/schedule-list.scss` file to remove the light blue background hover effect on the row. But the hover effect is only removed if the schedule is reduced.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -98,11 +98,11 @@ const DayItemView = View.extend({
     <td class="schedule-list__action-list-cell schedule-list__patient">
       <div class="schedule-list__state-patient">
         <span class="schedule-list__action-state action--{{ stateOptions.color }}">{{fa stateOptions.iconType stateOptions.icon}}</span><span class="schedule-list__search-helper">{{ state }}</span>&#8203;{{~ remove_whitespace ~}}
-        <span class="schedule-list__patient-name js-patient">{{ patient.first_name }} {{ patient.last_name }}</span>&#8203;
+        <span class="schedule-list__patient-name {{#unless isReduced}}js-patient{{/unless}} {{#if isReduced}}is-reduced{{/if}}">{{ patient.first_name }} {{ patient.last_name }}</span>&#8203;
       </div>
     </td>
     <td class="schedule-list__action-list-cell">
-      <span class="schedule-list__action-name js-action">{{ name }}</span>&#8203;{{~ remove_whitespace ~}}
+      <span class="schedule-list__action-name {{#unless isReduced}}js-action{{/unless}}">{{ name }}</span>&#8203;{{~ remove_whitespace ~}}
```
nmajor25, line 111: `js-action` and `js-patient` classnames are removed if `isReduced === true`. This stops the patient name and action name from being clickable.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -98,11 +98,11 @@ const DayItemView = View.extend({
     <td class="schedule-list__action-list-cell schedule-list__patient">
       <div class="schedule-list__state-patient">
         <span class="schedule-list__action-state action--{{ stateOptions.color }}">{{fa stateOptions.iconType stateOptions.icon}}</span><span class="schedule-list__search-helper">{{ state }}</span>&#8203;{{~ remove_whitespace ~}}
-        <span class="schedule-list__patient-name js-patient">{{ patient.first_name }} {{ patient.last_name }}</span>&#8203;
+        <span class="schedule-list__patient-name {{#unless isReduced}}js-patient{{/unless}} {{#if isReduced}}is-reduced{{/if}}">{{ patient.first_name }} {{ patient.last_name }}</span>&#8203;
```
nmajor25: A `.is-reduced` classname is added here if `isReduced === true`. 

This is used in the `src/js/views/patients/schedule/schedule-list.scss` file to remove the hover styling on the patient name.

src/js/views/patients/schedule/schedule-list.scss
```diff
@@ -79,7 +79,11 @@
     background: $greyed;
   }
 
-  &:hover:not(.is-selected) {
+  &.is-reduced {
+    cursor: default;
```
nmajor25, line 83: I switched the cursor to a `default` version here. Since the row is not clickable, I figured it wouldn't make sense to use the pointer cursor.

Same change was made for the patient name.

test/integration/patients/worklist/schedule.js
```diff
@@ -766,7 +780,81 @@ context('schedule page', function() {
...
+      .visit('/flow/1/action/1')
+      .wait('@routeFlow')
+      .wait('@routePatientByFlow')
+      .wait('@routeFlowActions');
+
+    cy
+      .get('.sidebar')
+      .find('[data-name-region] .action-sidebar__name')
+      .should('contain', 'Test Action');
   });
```
nmajor25, line 823: For the last three sections of this test where I verify that the reduced clinician can load the patient/flow pages properly, do I need to load all this route fixture data? Or can some of it be stripped out?

It seems like a lot of data to load since I'm only tested one small thing on each page.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -116,18 +116,16 @@ const DayItemView = View.extend({
...
-    const isReduced = this.state.get('isReduced');
-
     return {
       isOverdue: this.model.isOverdue(),
       state: state.get('name'),
       stateOptions: state.get('options'),
       patient: this.model.getPatient().attributes,
       form: this.model.getForm(),
-      isSelected: !isReduced && this.state.isSelected(this.model),
+      isSelected: !this.isReduced && this.state.isSelected(this.model),
```
paulfalgout, line 131: can `selected` still be set if the user is reduced now?  I'd assume not.  I'd be this is a missing branch in coverage

src/js/views/patients/schedule/schedule_views.js
paulfalgout: this could probably be `{[#if isReduced}}...{{else}}...{{/if}}`

paulfalgout: I think this class swapping is fine.  However an alternative.. which i don't have a strong opinion on is:
```js
triggers() {
  if (this.isReduced) {
    return {
       'click .js-form': 'click:form',
    };
  }

  return {
    'click .js-form': 'click:form',
    'click': 'click',
    'click .js-select': 'click:select',
  };
}
```

src/js/views/patients/schedule/schedule_views.js
paulfalgout: so you can either move this to `initialize` since it's affecting the `$el` as only the `$el`s content changes on render, or probably better you can make the `className` a function and use `this.getOption('state').get('isReduced')` to decide on the class of the `el`

test/integration/patients/worklist/schedule.js
paulfalgout: yeah you probably need all of that data..  I _think_ you may need to wrap each of the `visit`s in it's own `context` too.  I forget exactly but i think the `visit` call clears the coverage.

paulfalgout (changes requested):
Looking pretty good.

src/js/views/patients/schedule/schedule_views.js
nmajor25: If I remove the `!this.isRedcued` check, a `"this.state.isSelected is not a function"` console error is thrown and the schedule doesn't load any actions.

The normal schedule app uses this custom state model where the `isSelected()` function is used from: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/patients/schedule/schedule_state.js.

Since we don't use that state model for the reduced schedule, it's looking for that function and not finding it.

paulfalgout: ah right.. no state model

src/js/views/patients/schedule/schedule_views.js
nmajor25: Hmm that's an interesting alternative way to do it... I'll play around with both and see which one I like better.

src/js/views/patients/schedule/schedule_views.js
paulfalgout: I forgot this view is in both places

paulfalgout (approved):

test/integration/patients/worklist/schedule.js
```diff
@@ -650,6 +656,20 @@ context('schedule page', function() {
...
+
+        fx.data.relationships.flow = { data: { id: '1' } };
+
+        return fx;
+      })
+      .routePatientActions(_.identity, '2')
+      .routePatientFlows(_.identity, '2')
+      .routeFlow()
+      .routeFlowActions()
+      .routePatientByFlow()
```
paulfalgout: Ah I think `routePatientByAction()` may be needed as well for `cy.navigate('/patient/1/action/1');`

paulfalgout (changes requested):

paulfalgout (approved):

PR #688, opened 2022-04-28
Set federated logout to support SLO

BenCoffeed:
Shortcut Story ID: [sc-28714]

### DIFF
--- a/src/js/auth.js
+++ b/src/js/auth.js
@@ -87,7 +87,7 @@ function login(success) {
 
 function logout() {
   localStorage.removeItem(RWELL_KEY);
-  auth0.logout({ returnTo: location.origin });
+  auth0.logout({ returnTo: location.origin, federated: true });
 }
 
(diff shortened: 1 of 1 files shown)


paulfalgout:
This isn't going to pass the linter.  Also. did it work?

BenCoffeed:
I totally tested this. . .
not sus

BenCoffeed:
I was just trying to help show @nmajor25 what not to do. . .

nmajor25:
With `federated: true` set in my local sandbox, I get completely kicked out of my Google Account (Gmail, Google Calendar, etc.) after signing out in the app.

So this appears to be working.

paulfalgout (approved):

paulfalgout:
Interesting that there's 2 lines lost of coverage.  I _think_ this isn't explicitly tested now that I look at it, but it is probably covered _some_ times if the tests run slower which is why I didn't catch it :-/

PR #696, opened 2022-05-06
Add find in list functionality to clinician admin page

nmajor25:
Shortcut Story ID: [sc-28592]

Add find in list filtering functionality to the clinician admin page.

This entails adding a "Find in List" input element to the clinicians page that will allow users to filter the list by clinician name, group, access type, etc.

Screen Shot 2022-05-06 at 11 20 42 AM


src/js/apps/patients/schedule/reduced_schedule_app.js
```diff
@@ -4,7 +4,7 @@ import App from 'js/base/app';
 
 import { STATE_STATUS } from 'js/static';
 
-import SearchComponent from 'js/views/patients/shared/components/list-search';
+import SearchComponent from 'js/views/shared/components/list-search';
```
nmajor25, line 7: Since the `SearchComponent` is now being used by the both the patients worklist/schedule and clinicians apps, I moved its files to the `/js/views/shared/components` directory.

This import path was also changed for the `schedule_app.js` and `worklist_app.js` files.

src/js/apps/clinicians/clinicians_state.js
```diff
@@ -0,0 +1,9 @@
+import Backbone from 'backbone';
+
+export default Backbone.Model.extend({
+  defaults() {
+    return {
+      searchQuery: '',
+    };
+  },
+});
```
nmajor25: Does this `StateModel` need to be in its own file since it only contains the `searchQuery` value?

I contemplated placing it directly in the `clinicians-all_app.js` file as a `const` like this:

```js
// /src/js/apps/clinicians/clinicians-all_app.js

const StateModel = Backbone.Model.extend({
  defaults: {
    searchQuery: '',
  },
});
```

But I figured I'd follow the pattern we use for the other apps. And it sounds like we'll be adding more filter functionality to the clinicians page in the future, so it will probably need to be in its own file eventually anyways.

src/js/views/clinicians/clinicians-all_views.js
```diff
@@ -116,7 +129,12 @@ const LayoutView = View.extend({
   className: 'flex-region',
   template: hbs`
     <div class="list-page__header">
-      <div class="list-page__title"><span class="list-page__title-icon">{{far "users-cog"}}</span>{{ @intl.clinicians.cliniciansAllViews.layoutView.title }}</div>
+      <div class="flex list-page__title">
+        <div>
```
nmajor25, line 137: This empty `<div>` helps align the title icon and title text elements.

Without the empty `<div>`:

Screen Shot 2022-05-06 at 11 19 53 AM

With the `<div>` added:

Screen Shot 2022-05-06 at 11 20 18 AM

src/js/views/clinicians/clinicians-all_views.js
```diff
@@ -149,13 +168,56 @@ const ListView = CollectionView.extend({
...
+
+    const matchers = this._buildMatchers(searchQuery);
+
+    this.setFilter(function({ searchString }) {
+      return every(matchers, function(matcher) {
+        return matcher.test(searchString);
+      });
+    });
+  },
+  _buildMatchers(searchQuery) {
```
nmajor25, line 216: Possibly move this into a utility function that is imported and used?

This same function is also used in the `src/js/views/patients/schedule/schedule_views.js` and `src/js/views/patients/worklist/worklist_views.js` files.

src/js/apps/patients/schedule/reduced_schedule_app.js
paulfalgout: Yeah good call. I suspect there's probably i18n strings in some of the files that were moved that also need to be moved

src/js/apps/clinicians/clinicians_state.js
paulfalgout: Yeah I probably would have just popped this in the app, but it's fine either way.  _Although_ in this case.. I'm not even sure the default is useful or necessary.  It's more explicit as to what the state model is doing, but I'd suspect if no state model was defined, the default would still work ok.

src/js/views/clinicians/clinicians-all_views.js
paulfalgout: Yeah I think that's a great idea.  Let's do it.

paulfalgout (changes requested):
I think this is great.  One thing I had forgotten to mention previously.. I _think_ you might need to add some non-zero spaces to the items.  The `view.$el.text()` that is being used strips out the html so sometimes it can put the column text together without spacing.  You'll see this being done in the action-item / flow-item templates.  We might even want to add them where they aren't needed so that in case they are needed with a future update, we don't forget to add them.  The character is /u200b which I like to call "YouTube"s although I think the other templates are using the ascii code.

src/js/apps/patients/schedule/reduced_schedule_app.js
nmajor25: Yes you're right, there is a i18n string that needed to be updated. Will be included in the next commit.

src/js/apps/clinicians/clinicians_state.js
nmajor25: I removed the `src/js/apps/clinicians/clinicians_state.js` file completely and everything works the same. So, appears to be no need for it.

src/js/views/clinicians/clinicians-all_views.js
nmajor25: Per our discussion on Slack, we'll leave this as-is for this pull request. And we'll address this change in a separate Shortcut story.

paulfalgout (approved):

PR #697, opened 2022-05-10
Prevents clearing submission on add/edit of editgrid

paulfalgout:
There’s no other information form.io supplies to understand this context.  FWIW the tree component could possibly be fixed with a similar solution.

Shortcut Story ID: [sc-28842]


paulfalgout:
The coverage changes here are actually a race condition I think.  Probably need an explicit test for this but the submission stores to localstorage after 2 seconds..  right now I think the tests are only prefilling localstoage :-/

nmajor25 (approved):

PR #698, opened 2022-05-16
Prefill latest form-response for designated patient-action forms

nmajor25:
Shortcut Story ID: [sc-28739]

A user would like to have a form on a patient-action prefill with the latest form response for that specific `form.id` and `patient.id`.

To implement this, a `tags: ["prefill-latest-response"]` attribute will be attached to patient-actions. If the patient-action has that attribute, we fetch the latest form response for the given `form.id` and `patient.id` and prefill the form if no other responses are available for the patient action.

**Includes an unrelated bug fix**:

When the Cypress browser is hot reloaded while the `npm run dev:coverage` npm script is running, the Cypress environment fails and this error is thrown:

```
Failed to load plugin 'eslint-plugin-mocha' declared in 'test/.eslintrc': Package subpath './src/find' is not defined by "exports" in /Users/nickmajor/projects/rw/care-ops-frontend/node_modules/ramda/package.json
```

To fix this for our app, `eslint-plugin-mocha` was downgraded from `10.0.4` to `9.0.0` (the previous version we were using before we updated the NPM dependencies recently https://github.com/RoundingWell/care-ops-frontend/pull/694).

There's a fix for it merged into the `ramda` repo, but it hasn't been released yet: https://github.com/ramda/ramda/pull/3270. `ramda` is a dependency used by `eslint-plugin-mocha` and where the bug resides.

Once that fix is released, we can update to the newest version of `eslint-plugin-mocha`.

### DIFF
--- a/src/js/entities-service/form-responses.js
+++ b/src/js/entities-service/form-responses.js
@@ -16,8 +16,6 @@ const Entity = BaseEntity.extend({
     return $.ajax(`/api/form-responses/${ responseId }/response`);
   },
   fetchLatestSubmission(patientId, formId) {
-    if (!patientId || !formId) return [{}];
-
     return $.ajax(`/api/patients/${ patientId }/form-responses/latest?filter[form]=${ formId }`);
   },
 });
--- a/src/js/entities-service/form-responses.js
+++ b/src/js/entities-service/form-responses.js
@@ -9,10 +9,16 @@ const Entity = BaseEntity.extend({
     'formResponses:model': 'getModel',
     'formResponses:collection': 'getCollection',
     'fetch:formResponses:submission': 'fetchSubmission',
+    'fetch:formResponses:latestSubmission': 'fetchLatestSubmission',
   },
   fetchSubmission(responseId) {
     if (!responseId) return [{}];
     return $.ajax(`/api/form-responses/${ responseId }/response`);
+  },
+  fetchLatestSubmission(patientId, formId) {
+    if (!patientId || !formId) return [{}];
+
+    return $.ajax(`/api/patients/${ patientId }/form-responses/latest?filter[form]=${ formId }`);
   },
 });
 
--- a/test/integration/forms/form.js
+++ b/test/integration/forms/form.js
@@ -300,6 +300,53 @@ context('Patient Action Form', function() {
       .should('have.value', 'bar');
   });
 
+  specify('prefill a form with latest submission from another form', function() {
+    cy
+      .server()
+      .routeAction(fx => {
+        fx.data.id = '1';
+        fx.data.relationships.form.data = { id: '66666' };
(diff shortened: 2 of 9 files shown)


src/js/entities-service/form-responses.js
```diff
@@ -9,11 +9,17 @@ const Entity = BaseEntity.extend({
...
     'formResponses:collection': 'getCollection',
     'fetch:formResponses:submission': 'fetchSubmission',
+    'fetch:formResponses:latestSubmission': 'fetchLatestSubmission',
   },
   fetchSubmission(responseId) {
     if (!responseId) return [{}];
     return $.ajax(`/api/form-responses/${ responseId }/response`);
   },
+  fetchLatestSubmission(patientId, formId) {
+    if (!patientId || !formId) return [{}];
```
paulfalgout: I'm not sure about this guard.  I know there's one above, but that's almost certainly because that request is called without the responseId.  It might be impossible to test this guard, and if so we should remove it.

paulfalgout (approved):
I think the code is correct

src/js/entities-service/form-responses.js
nmajor25: Okay makes sense, I'll have that removed in the next commit.

test/integration/forms/form.js
```diff
@@ -253,6 +253,100 @@ context('Patient Action Form', function() {
...
+      .routeLatestFormResponseByPatient(fx => {
+        fx.data = {
+          familyHistory: 'Prefilled family history',
+          storyTime: 'Prefilled story time',
+          patient: { fields: { foo: 'bar' } },
+        };
+
+        return fx;
+      })
+      .visit('/patient-action/1/form/11111')
```
paulfalgout: I think this also needs to be `66666`
To test that this test is working after changing this to `66666` if you temporarily remove the `prefill_form_id` from the fixture the test should break.

test/integration/forms/form.js
```diff
@@ -253,6 +253,100 @@ context('Patient Action Form', function() {
...
+      .wait('@routeAction')
+      .wait('@routePatientByAction')
+      .wait('@routeFormDefinition')
+      .wait('@routeLatestFormResponseByPatient');
```
paulfalgout: We should probably test that this endpoint is getting hit correctly.

You'll find this pattern elsewhere but it'd be something like:
```js
cy
  .get('@routeLatestFormResponseByPatient')
  .its('url')
  // filter[form] should equal 11111
```

paulfalgout (changes requested):
Just a small question on the tests

test/integration/forms/form.js
nmajor25: Ah yes, missed that!

paulfalgout (approved):

paulfalgout:
Just the flaky outreach test

PR #699, opened 2022-05-23
Show patient sidebar with widgets via worklist page

nmajor25:
Shortcut Story ID: [sc-28989] [sc-29095] [sc-29115] [sc-29116]

This allows users to open and view a patient sidebar from the worklist page. The sidebar contains the patient's name, a link to the patient's dashboard page, and widgets that contain information about the patient.

Each action/flow item in the worklist is given a button that opens the patient sidebar when clicked.

https://user-images.githubusercontent.com/35355575/170591415-717ffa69-d83e-4cff-baed-d18cc6bc1972.mov

### DIFF
--- a/src/js/apps/patients/sidebar/patient-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/patient-sidebar_app.js
@@ -18,11 +18,11 @@ export default App.extend({
 
     return [patientModel, ...fields];
   },
-  onStart({ patient }, [patientModel]) {
+  onStart({ patient }) {
     const widgets = Radio.request('bootstrap', 'sidebarWidgets');
 
     this.showChildView('widgets', new WidgetCollectionView({
-      model: patientModel,
+      model: patient,
       collection: widgets,
       itemClassName: 'patient-sidebar__section',
     }));
--- a/src/js/apps/patients/sidebar/patient-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/patient-sidebar_app.js
@@ -4,20 +4,28 @@ import Radio from 'backbone.radio';
 import App from 'js/base/app';
 
 import { LayoutView } from 'js/views/patients/sidebar/patient/patient-sidebar_views';
+import { WidgetCollectionView } from 'js/views/patients/widgets/widgets_views';
 
 export default App.extend({
+  onBeforeStart({ patient }) {
+    this.showView(new LayoutView({ model: patient }));
+  },
   beforeStart({ patient }) {
     const patientModel = Radio.request('entities', 'fetch:patients:model', patient.id);
     const fields = map(Radio.request('bootstrap', 'sidebarWidgets:fields'), fieldName => {
       return Radio.request('entities', 'fetch:patientFields:model', patient.id, fieldName);
     });
 
-    return [patientModel, fields];
+    return [patientModel, ...fields];
   },
   onStart({ patient }, [patientModel]) {
     const widgets = Radio.request('bootstrap', 'sidebarWidgets');
 
-    this.showView(new LayoutView({ patient: patientModel, widgets }));
+    this.showChildView('widgets', new WidgetCollectionView({
+      model: patientModel,
+      collection: widgets,
+      itemClassName: 'patient-sidebar__section',
+    }));
   },
   viewEvents: {
     'close': 'stop',
--- a/src/js/apps/patients/sidebar/patient-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/patient-sidebar_app.js
@@ -3,8 +3,7 @@ import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
-import { LayoutView } from 'js/views/patients/sidebar/patient/patient-sidebar_views';
-import { WidgetCollectionView } from 'js/views/patients/widgets/widgets_views';
+import { LayoutView, SidebarWidgetsView } from 'js/views/patients/sidebar/patient/patient-sidebar_views';
 
 export default App.extend({
   onBeforeStart({ patient }) {
@@ -21,10 +20,9 @@ export default App.extend({
   onStart({ patient }) {
     const widgets = Radio.request('bootstrap', 'sidebarWidgets');
 
-    this.showChildView('widgets', new WidgetCollectionView({
+    this.showChildView('widgets', new SidebarWidgetsView({
       model: patient,
       collection: widgets,
-      itemClassName: 'patient-sidebar__section',
     }));
   },
   viewEvents: {
--- a/src/js/apps/patients/sidebar/patient-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/patient-sidebar_app.js
@@ -1,0 +1,28 @@
+import { map } from 'underscore';
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+
+import { LayoutView } from 'js/views/patients/sidebar/patient/patient-sidebar_views';
+
+export default App.extend({
+  beforeStart({ patient }) {
+    const patientModel = Radio.request('entities', 'fetch:patients:model', patient.id);
+    const fields = map(Radio.request('bootstrap', 'sidebarWidgets:fields'), fieldName => {
+      return Radio.request('entities', 'fetch:patientFields:model', patient.id, fieldName);
+    });
+
+    return [patientModel, fields];
+  },
+  onStart({ patient }, [patientModel]) {
+    const widgets = Radio.request('bootstrap', 'sidebarWidgets');
+
+    this.showView(new LayoutView({ patient: patientModel, widgets }));
+  },
+  viewEvents: {
+    'close': 'stop',
+  },
+  onStop() {
+    Radio.request('sidebar', 'close');
+  },
+});
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -58,6 +58,7 @@ export default App.extend({
   },
   onBeforeStart({ worklistId }) {
     if (this.isRestarting()) {
+      Radio.request('sidebar', 'close');
       this.showListTitle();
       this.showTypeToggleView();
       this.showSortDroplist();
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -104,6 +104,10 @@ export default App.extend({
       this.toggleBulkSelect();
     });
 
+    this.listenTo(collectionView, 'click:patientSidebarButton', patient => {
+      Radio.request('sidebar', 'start', 'patient', { patient });
+    });
+
     this.showChildView('list', collectionView);
 
     this.showSearchView();
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -104,7 +104,8 @@ export default App.extend({
       this.toggleBulkSelect();
     });
 
-    this.listenTo(collectionView, 'click:patientSidebarButton', patient => {
+    this.listenTo(collectionView, 'click:patientSidebarButton', ({ model }) => {
+      const patient = model.getPatient();
       Radio.request('sidebar', 'start', 'patient', { patient });
     });
 
--- a/src/js/views/patients/patient/sidebar/sidebar_views.js
+++ b/src/js/views/patients/patient/sidebar/sidebar_views.js
@@ -6,6 +6,7 @@ import hbs from 'handlebars-inline-precompile';
 import intl from 'js/i18n';
 
 import 'sass/modules/widgets.scss';
+import 'sass/domain/patient-sidebar.scss';
 
 import Optionlist from 'js/components/optionlist';
 
--- a/src/js/views/patients/patient/sidebar/sidebar_views.js
+++ b/src/js/views/patients/patient/sidebar/sidebar_views.js
@@ -6,12 +6,12 @@ import hbs from 'handlebars-inline-precompile';
 import intl from 'js/i18n';
 
 import 'sass/modules/widgets.scss';
-import 'sass/domain/patient-sidebar.scss';
 
 import Optionlist from 'js/components/optionlist';
 
 import { WidgetCollectionView } from 'js/views/patients/widgets/widgets_views';
 
+import 'sass/domain/patient-sidebar.scss';
 import './patient-sidebar.scss';
 
 const i18n = intl.patients.patient.sidebar.sidebarViews;
--- a/src/js/views/patients/sidebar/patient/patient-sidebar.hbs
+++ b/src/js/views/patients/sidebar/patient/patient-sidebar.hbs
@@ -7,10 +7,10 @@
       <button class="button--icon js-close">{{fas "times"}}</button>
     </div>
   </div>
-  <div class="worklist-patient-sidebar__patient-info">
+  <div class="worklist-patient-sidebar__patient-info js-patient">
     <div class="worklist-patient-sidebar__patient-name">{{ first_name }} {{ last_name }}</div>
     <div>
-      <button class="button--link js-patient">View Patient Dashboard</button>
+      <button class="button--link">View Patient Dashboard</button>
     </div>
   </div>
   <div data-widgets-region></div>
--- a/src/js/views/patients/sidebar/patient/patient-sidebar.hbs
+++ b/src/js/views/patients/sidebar/patient/patient-sidebar.hbs
@@ -1,0 +1,17 @@
+<div class="flex-grow">
+  <div class="flex">
+    <div class="flex-grow">
+      <span>{{far "address-card"}}</span>
+    </div>
+    <div>
+      <button class="button--icon js-close">{{fas "times"}}</button>
+    </div>
+  </div>
+  <div class="patient-sidebar__patient-info">
+    <div class="name">{{ firstName }} {{ lastName }}</div>
+    <div class="dashboard-link">
+      <span class="js-patient">View Patient Dashboard</span>
+    </div>
+  </div>
+  <div data-widgets-region></div>
+</div>
--- a/src/js/views/patients/sidebar/patient/patient-sidebar.hbs
+++ b/src/js/views/patients/sidebar/patient/patient-sidebar.hbs
@@ -1,16 +1,16 @@
 <div class="flex-grow">
   <div class="flex">
     <div class="flex-grow">
-      <span>{{far "address-card"}}</span>
+      <span class="patient-sidebar__user-icon">{{far "address-card"}}</span>
     </div>
-    <div>
+    <div class="patient-sidebar__close-icon">
       <button class="button--icon js-close">{{fas "times"}}</button>
     </div>
   </div>
   <div class="patient-sidebar__patient-info">
-    <div class="name">{{ firstName }} {{ lastName }}</div>
-    <div class="dashboard-link">
-      <span class="js-patient">View Patient Dashboard</span>
+    <div class="patient-sidebar__patient-name">{{ firstName }} {{ lastName }}</div>
+    <div>
+      <span class="patient-sidebar__dashboard-link js-patient">View Patient Dashboard</span>
     </div>
   </div>
   <div data-widgets-region></div>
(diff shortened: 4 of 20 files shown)


src/js/views/patients/worklist/layout.hbs
```diff
@@ -15,3 +15,4 @@
   <table class="worklist__header w-100 js-list-header" data-table-region></table>
 </div>
 <div class="flex-region list-page__list js-list" data-list-region></div>
+<div data-sidebar-region></div>
```
paulfalgout: I don't think this is needed.  I think the sidebar is one layout up from this

src/js/views/patients/sidebar/patient/patient-sidebar.scss
```diff
@@ -0,0 +1,14 @@
+.patient-sidebar__patient-info {
+  border-bottom: 1px solid $black-90;
+  padding: 16px 0;
+
+  .name {
```
paulfalgout: I think we should go ahead and un-nest these and give them BEM naming.  The styles can be whatever for now, but I don't want to add simple class names.

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -179,10 +179,16 @@ const ListView = CollectionView.extend({
   },
   childViewTriggers: {
     'render': 'listItem:render',
+    'click:patientSidebarButton': 'patientSidebarButton:click',
```
paulfalgout: So this is already bubbling up this event... so either this'd be better as a `childViewEvent` if it s meant to be handled by this view, or the app can do what `onPatientSidebarButtonClick` is doing, which I think in this case is probably preferred.

the app can just listen to the click.  at the app level destructing the view is also nice ie: `onPatientSidebarButtonClick({ model })`

src/js/apps/patients/sidebar/patient-sidebar_app.js
```diff
@@ -0,0 +1,28 @@
+import { map } from 'underscore';
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+
+import { LayoutView } from 'js/views/patients/sidebar/patient/patient-sidebar_views';
+
+export default App.extend({
+  beforeStart({ patient }) {
+    const patientModel = Radio.request('entities', 'fetch:patients:model', patient.id);
```
paulfalgout, line 13: I don't think you really need to fetch the patient model as it'll be in the click event of the row.  We can just pass the model to this app and use it.

src/js/views/patients/sidebar/patient/patient-sidebar_views.js
```diff
@@ -0,0 +1,50 @@
...
+    'click .js-patient': 'click:patient',
+  },
+  templateContext() {
+    return {
+      firstName: this.patient.get('first_name'),
+      lastName: this.patient.get('last_name'),
+    };
+  },
+  initialize({ patient, widgets }) {
+    this.patient = patient;
```
paulfalgout: This'd be a little simple if the patient was passed as the model.  Then you also wouldn't need `templateContext`.  The `onRender` would just pass the model down and although attaching `widgets` here is perfectly fine, since it's single use, you wouldn't need the `initialize` and could just do `collection: this.getOption('widgets'),` in the `onRender`

src/js/views/patients/sidebar/patient/patient-sidebar_views.js
```diff
@@ -0,0 +1,50 @@
+import Radio from 'backbone.radio';
+import { View } from 'marionette';
+
+import 'sass/modules/buttons.scss';
+import 'sass/modules/forms.scss';
+import 'sass/modules/textarea-flex.scss';
```
paulfalgout: I don't think you'll need `textarea-flex` or `forms`.

paulfalgout (changes requested):
Pretty close to done

src/js/apps/patients/sidebar/patient-sidebar_app.js
```diff
@@ -0,0 +1,28 @@
...
+    const fields = map(Radio.request('bootstrap', 'sidebarWidgets:fields'), fieldName => {
+      return Radio.request('entities', 'fetch:patientFields:model', patient.id, fieldName);
+    });
+
+    return [patientModel, fields];
```
paulfalgout: this _probably_ needs to be `...fields`

src/js/apps/patients/sidebar/patient-sidebar_app.js
```diff
@@ -0,0 +1,28 @@
+import { map } from 'underscore';
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+
+import { LayoutView } from 'js/views/patients/sidebar/patient/patient-sidebar_views';
+
+export default App.extend({
+  beforeStart({ patient }) {
```
paulfalgout, line 12: @nmajor25 I was thinking about this a bit more since yesterday.  I think we should do this:
```js
onBeforeStart({ patient }) { // <-- patient model with no relations
  this.showView(new LayoutView({ model: patient });
},
beforeStart({ patient }) {
  // fetch:patients:model' && fields
}
onStart({ patient }) { // <-- At this point the original patient model will contain the relations
  this.showChildView('widgets', new WidgetsCollectionView({ model: patient, collection: widgets ));
}
```

paulfalgout (changes requested):

src/js/views/patients/sidebar/patient/patient-sidebar_views.js
nmajor25: I removed those. Should be included in the next commit.

src/js/views/patients/sidebar/patient/patient-sidebar.scss
nmajor25: Okay, I switched these to BEM naming. Will be included in the next commit.

src/js/views/patients/worklist/layout.hbs
nmajor25: You are right, removed that and everything still worked. Will be included in the next commit.

src/js/apps/patients/sidebar/patient-sidebar_app.js
```diff
@@ -0,0 +1,36 @@
...
+    const fields = map(Radio.request('bootstrap', 'sidebarWidgets:fields'), fieldName => {
+      return Radio.request('entities', 'fetch:patientFields:model', patient.id, fieldName);
+    });
+
+    return [patientModel, ...fields];
+  },
+  onStart({ patient }) {
+    const widgets = Radio.request('bootstrap', 'sidebarWidgets');
+
+    this.showChildView('widgets', new WidgetCollectionView({
```
paulfalgout: so to move the class in the `patient-sidebar_views` file you'd essentially do this:
```js
import { WidgetCollectionView } from 'js/views/patients/widgets/widgets_views';

const SidebarWidgetsView = WidgetCollectionView.extend({
  itemClassName: 'patient-sidebar__section',
});
...
export {
  LayoutView,
  SidebarWidgetsView,
};
```

src/sass/modules/sidebar.scss
```diff
@@ -1,10 +1,18 @@
...
   max-width: 400px;
+  min-width: 400px;
   overflow-y: auto;
   padding: 16px 20px;
 }
 
+.sidebar--small {
+  @extend .sidebar;
+
+  max-width: 256px;
```
paulfalgout: could this just be `width: 256px;` and `width: 400px` above?

src/js/views/patients/worklist/flow-item.hbs
```diff
@@ -1,5 +1,14 @@
 <td class="table-list__cell table-list__item-check js-no-click"><button class="button--checkbox js-select">{{#if isSelected}}{{fas "check-square"}}{{else}}{{fal "square"}}{{/if}}</button></td>
-<td class="table-list__cell w-15 worklist-list__patient-name js-patient">{{ patient.first_name }} {{ patient.last_name }}&#8203;</td>
+<td class="table-list__cell w-15">
+  <span class="worklist-list__patient-sidebar-icon js-patient-sidebar-button">
+    <button class="button-secondary--compact">
+      {{far "address-card"}}&#8203;
+    </button>
+  </span>
+  <div class="worklist-list__patient-name js-patient">
```
paulfalgout, line 8: I think name here could be a `<span>` and `js-patient-sidebar-button` could be put on the button itself without the wrapping span.

I _think_ you can remove `worklist-list__patient-sidebar-icon` and remove `display: table-cell;` from `worklist-list__patient-name`

src/js/views/patients/sidebar/patient/patient-sidebar.hbs
```diff
@@ -0,0 +1,17 @@
...
+      <span class="patient-sidebar__user-icon">{{far "address-card"}}</span>
+    </div>
+    <div class="patient-sidebar__close-icon">
+      <button class="button--icon js-close">{{fas "times"}}</button>
+    </div>
+  </div>
+  <div class="patient-sidebar__patient-info">
+    <div class="patient-sidebar__patient-name">{{ first_name }} {{ last_name }}</div>
+    <div>
+      <span class="patient-sidebar__dashboard-link js-patient">View Patient Dashboard</span>
```
paulfalgout: I think this could be a button with class `button--link`

src/js/views/patients/sidebar/patient/patient-sidebar.scss
```diff
@@ -0,0 +1,21 @@
+.patient-sidebar__user-icon svg,
```
paulfalgout: the svg has the class `icon`.  I think this one of the rare cases where we typically do sass nesting lie:
```css
.patient-sidebar__user-icon,
.patient-sidebar__close-icon {
  .icon {
    font-size: 20px;
 }
}
```

Not using the tag name will preserve the cascade and you won't need the `!important`

src/js/views/patients/sidebar/patient/patient-sidebar.scss
```diff
@@ -0,0 +1,21 @@
...
+  border-bottom: 1px solid $black-90;
+  padding: 16px 0;
+}
+
+.patient-sidebar__patient-name {
+  font-size: 24px;
+  margin-bottom: 8px;
+}
+
+.patient-sidebar__dashboard-link {
```
paulfalgout: as mentioned elsewhere we can remove this class for `button--link` also generally colors are available in a some form of `color(blue)` defined in the `provider-variables.scss`

src/js/views/patients/sidebar/patient/patient-sidebar.scss
```diff
@@ -0,0 +1,21 @@
...
+.patient-sidebar__close-icon svg {
+  font-size: 20px !important;
+}
+
+.patient-sidebar__patient-info {
+  border-bottom: 1px solid $black-90;
+  padding: 16px 0;
+}
+
+.patient-sidebar__patient-name {
```
paulfalgout: we have a naming collision on this file where the block `patient-sidebar` is already defined.  such that this class and `patient-sidebar__name` are a little confusing.  In fact since we are using `patient-sidebar__section` with this technically we are using that other BEM'd sass file.  I think we should make a `sass/domain/patient-sidebar.scss` and move `patient-sidebar__section` to it and include it in both view files.  Then perhaps we can rename these class names slightly to something like `worklist-patient-sidebar__` to avoid any future accidental collision.

paulfalgout (changes requested):
Some style stuff and remember mark asked for sidebar animations.

src/js/apps/patients/sidebar/patient-sidebar_app.js
nmajor25: Awesome, that works nicely!

src/sass/modules/sidebar.scss
nmajor25: Yes, it works but makes me a little nervous that it may effect widths of other sidebars somehow. I'll have to go through each one and verify things weren't changed for them.

src/js/views/patients/worklist/flow-item.hbs
nmajor25: It doesn't work for when the patient's name goes to a second line:

Screen Shot 2022-05-26 at 12 17 26 PM

src/js/views/patients/sidebar/patient/patient-sidebar.hbs
nmajor25: Indeed, that works well!

src/js/views/patients/sidebar/patient/patient-sidebar.scss
nmajor25: Makes sense, that works. I don't like using `!important`, so it's nice to get rid of it.

src/js/views/patients/sidebar/patient/patient-sidebar.scss
nmajor25: 10-4 on the colors. This CSS will be removed in the next commit I push.

src/js/views/patients/sidebar/patient/patient-sidebar.scss
nmajor25: Ya that is confusing, will have that new `sass/domain/patient-sidebar.scss` file created and the class names changed to `worklist-patient-sidebar__`.

src/js/views/patients/worklist/flow-item.hbs
paulfalgout: hmm well there's a `white-space: normal` on that class, but the default is `white-space: nowrap`

nmajor25: It's currently built to do this when the patient name overflows:

Screen Shot 2022-05-26 at 12 56 03 PM

Removing `white-space: normal` will add an overflow ellipsis (`...`) that would make it do this instead:

Screen Shot 2022-05-26 at 12 52 44 PM

We want to do the ellipsis option?

paulfalgout: hmm yeah probably not.  I suppose this is how the action cell works.  There's some sort of cleaner refactor here I think.  A lot of custom classes that aren't doing anything super specific to this context... But I don't know what it is yet, so carry on.  This works for this PR

test/integration/patients/worklist/bulk-edit.js
```diff
@@ -31,10 +31,13 @@ context('Worklist bulk editing', function() {
 
         fx.data[0].relationships.patient = { data: { id: '1' } };
         fx.data[0].relationships.state = { data: { id: '22222' } };
+        fx.data[0].attributes.created_at = testTs();
         fx.data[1].relationships.patient = { data: { id: '2' } };
         fx.data[1].relationships.state = { data: { id: '22222' } };
+        fx.data[1].attributes.created_at = testTsSubtract(1);
         fx.data[2].relationships.patient = { data: { id: '3' } };
         fx.data[2].relationships.state = { data: { id: '22222' } };
+        fx.data[2].attributes.created_at = testTsSubtract(2);
```
nmajor25, line 41: This test was getting a worklist item in the list via the patient name and then using `.prev()` to get the checkbox element. But that breaks the test because the `.prev()` element is now the patient sidebar button.

So, I added test dates to each item in the worklist so they are forced to be in the same order each time.

And then updated the test to retrieve each row item by their index position (e.g. `.first()`, `.eq(1)`, `.last()`) in the list, instead of using the patient name (e.g. `.contains('Patient One')`).

test/integration/patients/worklist/worklist.js
```diff
@@ -2702,6 +2703,204 @@ context('worklist page', function() {
...
+      .find('.table-list__item')
+      .find('.worklist-list__patient-sidebar-icon .js-patient-sidebar-button')
+      .click();
+
+    cy
+      .get('@patientSidebar')
+      .find('.worklist-patient-sidebar__patient-name')
+      .should('contain', 'Test Patient');
+  });
+
```
nmajor25, line 2931: A quick breakdown for what I did in this test:

1. Open the worklist patient sidebar for a flow item.
2. Test the contents of the patient sidebar:
    - Verify the patient's name is shown and. that it links to the correct `/patient/dashboard/:id` page.
    - Verify the `View Patient Dashboard` link goes to the correct `/patient/dashboard/:id` page.
    - Verify that widgets are loaded/displayed. One basic `Sex` widget is shown. And another basic field widget is also shown.
    - Close the sidebar using the `X` icon.
3. Switch to the actions version of the worklist.
    - Open the sidebar and verify the sidebar is loading data by checking that the patient name is there. The other parts of the sidebar were already tested.
    - Verify that the sidebar closes when a worklist filter setting is changed.

src/js/views/patients/worklist/action-item.hbs
```diff
@@ -1,7 +1,7 @@
 <td class="table-list__cell table-list__item-check js-no-click"><button class="button--checkbox js-select">{{#if isSelected}}{{fas "check-square"}}{{else}}{{fal "square"}}{{/if}}</button></td>
 <td class="table-list__cell w-15">
-  <span class="worklist-list__patient-sidebar-icon js-patient-sidebar-button">
-    <button class="button-secondary--compact">
+  <span class="worklist-list__patient-sidebar-icon">
+    <button class="button-secondary--compact js-patient-sidebar-button">
```
nmajor25, line 4: I moved the `.js-patient-sidebar-button` to the button here. Seems to make more sense that the trigger class name is on the button instead of the span.

Did the same in the `flow-item.hbs` file too.

src/js/views/patients/sidebar/patient/patient-sidebar.hbs
```diff
@@ -0,0 +1,17 @@
+<div class="flex-grow">
+  <div class="flex">
+    <div class="flex-grow">
+      <span class="worklist-patient-sidebar__user-icon">{{far "address-card"}}</span>
+    </div>
+    <div class="worklist-patient-sidebar__close-icon">
+      <button class="button--icon js-close">{{fas "times"}}</button>
+    </div>
+  </div>
+  <div class="worklist-patient-sidebar__patient-info">
```
paulfalgout: I think https://app.shortcut.com/roundingwell/story/29116/bug-patient-name-in-the-patient-sidebar-should-be-clickable-as-per-marvel is solved by moving `js-patient` here?

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -104,6 +104,11 @@ export default App.extend({
       this.toggleBulkSelect();
     });
 
+    this.listenTo(collectionView, 'click:patientSidebarButton', ({ model }) => {
```
paulfalgout, line 108: https://app.shortcut.com/roundingwell/story/29115/bug-patient-sidebar-sidebar-does-not-close-if-you-change-any-of-the-worklist-filter-settings is fixed by adding above:
```js
  onBeforeStart({ worklistId }) {
    if (this.isRestarting()) {
```

paulfalgout (changes requested):
I'm going to pull it down and go through the test coverage, but I think we can go ahead and fix the two small bugs and finish out this epic.

src/js/views/patients/sidebar/patient/patient-sidebar.hbs
nmajor25: That works. 

Included in the last commit. And added a `cursor: pointer` CSS style to it as well. Cypress test updated too.

src/js/apps/patients/worklist/worklist_app.js
nmajor25: This did the trick. Included in the last commit. 

I also added a check to the Cypress test that updates a worklist filter while the sidebar is open. And verifies that the sidebar closes when that happens.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -6,6 +6,7 @@ import hbs from 'handlebars-inline-precompile';
 import intl from 'js/i18n';
 
 import 'sass/modules/widgets.scss';
+import 'sass/domain/patient-sidebar.scss';
```
paulfalgout: I think the `domain` imports should go right above the local scss imports to insure they occur after all general modules.  FWIW I don't think this is documented adequately in our readme.

paulfalgout (changes requested):
I think this is done.  just one tiny adjustment I caught in a re-review

src/js/views/patients/patient/sidebar/sidebar_views.js
nmajor25: Updated in the latest commit.

paulfalgout (approved):

paulfalgout:
coveralls report issue is flake

PR #703, opened 2022-05-31
Pull prefill form response data from attributes

nmajor25:
Shortcut Story ID: [sc-29156]

Update the prefill form response data to pull from `response.data.attributes.response.data`. And also update the Cypress tests to accommodate that change when loading fixture route data.

This is a bug fix related to when we added the prefill form response functionality: https://github.com/RoundingWell/care-ops-frontend/pull/698.

### DIFF
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -98,7 +98,7 @@ export default App.extend({
       channel.request('send', 'fetch:form:data', {
         definition,
         formData: fields.data.attributes || {},
-        formSubmission: response && response.data || {},
+        formSubmission: response && response.data.attributes.response.data || {},
         contextScripts: this.form.getContextScripts(),
         reducers: this.form.getReducers(),
         changeReducers: this.form.getChangeReducers(),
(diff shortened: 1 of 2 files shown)


src/js/services/forms.js
```diff
@@ -98,7 +98,7 @@ export default App.extend({
       channel.request('send', 'fetch:form:data', {
         definition,
         formData: fields.data.attributes || {},
-        formSubmission: response && response.data || {},
+        formSubmission: response && response.data.attributes.response.data || {},
```
paulfalgout, line 101: If there's not a form response does this endpoint return nothing?  If it returns `{}` this'd break.

paulfalgout (commented):
question

src/js/services/forms.js
nmajor25: I believe it returns an empty `204` response.

But it should probably be built to not break either way.

paulfalgout: I think this'd work if it's a 204.  Not sure there's a possible other use case

paulfalgout (approved):
you're call if you want to change it, but this looks like it'd handle a 204 or a response.

PR #707, opened 2022-06-06
Add a preload spinner to the worklist patient sidebar

nmajor25:
Shortcut Story ID: [sc-29213]

Add a preload spinner for the widgets section of the worklist patient sidebar. Video of the preloader is included below.

When the worklist patient sidebar is opened, there is often a delay for the widgets data to load into the sidebar. During that delay, we want to show a loading icon instead of just a blank section.

https://user-images.githubusercontent.com/35355575/172242653-66fe178e-c073-4f61-aa53-dc2bb0a0206f.mov

### DIFF
--- a/src/js/views/patients/sidebar/patient/patient-sidebar.scss
+++ b/src/js/views/patients/sidebar/patient/patient-sidebar.scss
@@ -15,3 +15,9 @@
   font-size: 24px;
   margin-bottom: 8px;
 }
+
+.worklist-patient-sidebar__widgets {
+  .spinner {
+    margin-top: 60px;
+  }
+}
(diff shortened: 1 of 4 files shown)


src/js/views/patients/sidebar/patient/patient-sidebar.scss
```diff
@@ -15,3 +15,9 @@
   font-size: 24px;
   margin-bottom: 8px;
 }
+
+.worklist-patient-sidebar__widgets {
+  .spinner {
+    margin-top: 60px;
+  }
+}
```
nmajor25, line 23: Added `60px` of top margin for the spinner. Matches the margin used for the activity section of the action/flow sidebar: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/sass/modules/sidebar.scss#L57.

Let me know if we want to use a different margin for this. I put a video of the sidebar with the spinner in the pull request description so you can see what it looks like.

PR #713, opened 2022-06-07
Close sidebars via the ESC keyboard shortcut

nmajor25:
Shortcut Story ID: [sc-29153]

Close the sidebar when a user presses the ESC keyboard shortcut.

Picklists also close via the ESC keyboard shortcut. And should take precedent over sidebars closing via the ESC keypress. In other words, if a sidebar and picklist are open at the same time, an ESC keypress should close the picklist and keep the sidebar open.

### DIFF
--- a/src/js/behaviors/sidebar.js
+++ b/src/js/behaviors/sidebar.js
@@ -1,0 +1,26 @@
+import $ from 'jquery';
+import Radio from 'backbone.radio';
+import { Behavior } from 'marionette';
+
+import keyCodes from 'js/utils/formatting/key-codes';
+
+import KeyListenerBehavior from './key-listener';
+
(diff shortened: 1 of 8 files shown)


nmajor25:
@paulfalgout got this working and tested it extensively in my local sandbox. But not sure if it's the best way.

I set up a new sidebar behavior that sends the `Radio.request('sidebar', 'close')` request when the `ESC` keyboard shortcut is pressed. But only if there are no picklists currently open elsewhere in the app.

And then applied that behavior to each sidebar.

I need to still add Cypress tests for this. But wanted to get your eyes on it before I jumped into that.

nmajor25:
We decided not to move forward with this feature. Therefore, closing the pull request.

PR #715, opened 2022-06-09
Add a user icon & left border to the patient dashboard sidebar

nmajor25:
Shortcut Story ID: [sc-29273]

Add the Font Awesome [address-card](https://fontawesome.com/icons/address-card?s=regular) icon to the patient dashboard sidebar. And also add a left-border to the sidebar.

To accommodate these changes, the styles for both the sidebar and overall patient dashboard page were updated.

Screenshot of the patient dashboard sidebar after these changes are made.

Screen Shot 2022-06-09 at 2 50 48 PM

### DIFF
--- a/src/js/views/patients/patient/flow/patient-flow.scss
+++ b/src/js/views/patients/patient/flow/patient-flow.scss
@@ -26,7 +26,7 @@
   background: $white;
   color: $black-20;
   line-height: 1;
-  padding: 16px 0 8px 0;
+  padding: 16px 0 16px;
   position: sticky;
   top: 0;
   z-index: 1;
--- a/src/js/views/patients/patient/patient.scss
+++ b/src/js/views/patients/patient/patient.scss
@@ -10,16 +10,15 @@
   flex: 1;
   flex-direction: column;
   min-width: 744px;
-  padding-left: 24px;
+  padding: 0 24px;
   width: 100%;
 }
 
 .patient__sidebar {
   display: flex;
   flex: initial;
-  max-width: 400px;
   min-height: 0;
-  min-width: 256px;
+  width: 256px;
 }
 
 .patient__content {
@@ -30,7 +29,7 @@
   background: $white;
   color: $black-20;
   line-height: 1;
-  padding: 16px 0;
+  padding: 16px 0 24px;
   position: sticky;
   top: 0;
   z-index: 1;
--- a/src/js/views/patients/patient/patient.scss
+++ b/src/js/views/patients/patient/patient.scss
@@ -29,7 +29,9 @@
   background: $white;
   color: $black-20;
   line-height: 1;
-  padding: 16px 0 24px;
+  // NOTE: Margin added for alignment with sidebar icon & patient name.
+  margin-bottom: 6px;
+  padding: 16px 0;
   position: sticky;
   top: 0;
   z-index: 1;
@@ -57,7 +59,7 @@
 .patient__tab--selected {
   display: inline-block;
   font-size: 24px;
-  line-height: 0.9;
+  line-height: 1;
   margin-right: 4px;
   padding: 0 16px 16px;
 }
--- a/src/js/views/patients/patient/patient.scss
+++ b/src/js/views/patients/patient/patient.scss
@@ -57,7 +57,7 @@
 .patient__tab--selected {
   display: inline-block;
   font-size: 24px;
-  line-height: 1;
+  line-height: 0.9;
   margin-right: 4px;
   padding: 0 16px 16px;
 }
--- a/src/js/views/patients/patient/sidebar/patient-sidebar.scss
+++ b/src/js/views/patients/patient/sidebar/patient-sidebar.scss
@@ -1,6 +1,6 @@
 .patient-sidebar {
-  border-left: 1px solid #e6e6e6;
-  padding: 54px 24px 0;
+  border-left: 1px solid $black-90;
+  padding: 52px 24px 0;
   position: relative;
 }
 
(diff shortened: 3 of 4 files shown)


src/js/views/patients/patient/patient.scss
```diff
@@ -10,16 +10,15 @@
   flex: 1;
   flex-direction: column;
   min-width: 744px;
-  padding-left: 24px;
```
nmajor25, line 13: Adds `24px` padding to the right side of the main content area as well.

src/js/views/patients/patient/patient.scss
```diff
@@ -10,16 +10,15 @@
...
   width: 100%;
 }
 
 .patient__sidebar {
   display: flex;
   flex: initial;
-  max-width: 400px;
   min-height: 0;
-  min-width: 256px;
+  width: 256px;
```
nmajor25, line 21: Sets width of the sidebar to be `256px` no matter what.

Before the sidebar had a min width of `256px` and would extend to a max of `400px`. This would cause some UI alignment/visual issues when the patient's name is really long. As the long patient name would stretch the width of sidebar beyond `256px`.

src/js/views/patients/patient/patient.scss
```diff
@@ -30,7 +29,7 @@
   background: $white;
   color: $black-20;
   line-height: 1;
-  padding: 16px 0;
+  padding: 16px 0 24px;
```
nmajor25: Changes the bottom padding for the context trail from `16px` to `24px`.

This is done to align the bottom border for the Dashboard/Data & Events with the bottom border on the patient's name in the sidebar.

src/js/views/patients/patient/sidebar/patient-sidebar.scss
```diff
@@ -1,15 +1,27 @@
 .patient-sidebar {
-  padding: 40px 24px 0;
+  border-left: 1px solid #e6e6e6;
```
nmajor25: Realized just now that I should use `$black-90` here instead of `#e6e6e6`.

src/js/views/patients/patient/sidebar/patient-sidebar.scss
```diff
@@ -1,15 +1,27 @@
...
+  padding: 54px 24px 0;
```
nmajor25: Padding here is adjusted her to get the desired spacing between the icon/ellipsis and patient name below it.

The padding is kind of crazy here because the icon and ellipsis are using `position: absolute`. 

Would it be better to not use `position: absolute`? And position them as a normal `flex` element like we did in the worklist patient sidebar?

I think it would make the CSS styling for it a little simpler.

src/js/views/patients/patient/sidebar/patient-sidebar.scss
```diff
@@ -1,15 +1,27 @@
...
+  .icon {
+    font-size: 20px;
+  }
 }
 
 .patient-sidebar__menu {
   position: absolute;
   right: 24px;
-  top: 8px;
+  top: 16px;
```
nmajor25, line 26: Adds more padding space above the icon/ellipsis, makes them align with the context trail text, and matches the amount of padding we use for the worklist patient sidebar.

src/js/views/patients/patient/sidebar/patient-sidebar.scss
paulfalgout: Hmm yeah this is problematic.  Any pixel size that isn't divisible by 4px should be incorrect or commented as to why it is specifically needed.  If 52 or 56 cannot work it probably indicates some other styling issue somewhere.

paulfalgout (changes requested):

src/js/views/patients/patient/sidebar/patient-sidebar.scss
nmajor25: Okay, I'll investigate this.

src/js/views/patients/patient/flow/patient-flow.scss
```diff
@@ -26,7 +26,7 @@
   background: $white;
   color: $black-20;
   line-height: 1;
-  padding: 16px 0 8px 0;
+  padding: 16px 0 16px;
```
nmajor25, line 29: This update is to align the flow title and patient name (in the patient sidebar) on the flow page.

align-patient-name-with-flow-title

src/js/views/patients/patient/patient.scss
```diff
@@ -57,7 +57,7 @@
 .patient__tab--selected {
   display: inline-block;
   font-size: 24px;
-  line-height: 1;
+  line-height: 0.9;
   margin-right: 4px;
   padding: 0 16px 16px;
 }
```
nmajor25: On the patient dashboard page, the bottom border of the "Dashboard" / "Data & Events" and patient name in the sidebar don't align correctly. And is off by `2px`. This is why I was previously using the padding unit not divisible by `4`.

There is something in the "Dashboard" / "Data & Events" elements that is causing this. Possibly the relationship between the `::before` blue line under the active tab and the "Dashboard" / "Data & Events" text content. I'm not entirely sure.

I spent some time playing around with the CSS for those. And the only two solutions I found to fix the alignment:

1. Switch `line-height: 1` to `line-height: 0.9` instead.
2. Switch padding to `padding: 0 16px 14px`.

If doing `line-height: 0.9` doesn't work, suggestions on this would be much appreciated!

dashboard-sidebar-name-alignment

paulfalgout: I think cleaner than this would be to modify `patient__context-trail`
We can make the padding `16px 0;` and then add something like
```css
// NOTE: Margin added to account for sidebar icon.
margin-bottom: 6px;
```

paulfalgout (changes requested):

src/js/views/patients/patient/patient.scss
nmajor25: Hmm ya, that would work!

paulfalgout (approved):

PR #724, opened 2022-06-17
Build prototype of the form print app

nmajor25:
Shortcut Story ID: [sc-29317]

When a URL route of `/print/:patientId/:formId(/:responseId)` is loaded in the app, it should load just the iframe content for a given form. No other elements should exist on the page.

The `responseId` is an optional URL parameter.

This page will be used by the backend team to convert forms to a PDF version when designated patient actions are completed.

### DIFF
--- a/src/js/formprint/index.js
+++ b/src/js/formprint/index.js
@@ -1,0 +1,98 @@
+import 'js/base/setup';
+import 'js/i18n';
+
+import Backbone from 'backbone';
+import Radio from 'backbone.radio';
+import { View, CollectionView } from 'marionette';
+import hbs from 'handlebars-inline-precompile';
+
+import 'sass/provider-core.scss';
+import 'sass/app-root.scss';
+
+import './formprint.scss';
+
+import initPlatform from 'js/utils/platform';
+
+import App from 'js/base/app';
+
+import FormsService from 'js/services/forms';
+
+import 'js/entities-service';
+
+import BootstrapService from 'js/services/bootstrap';
+
+import IframeFormBehavior from 'js/behaviors/iframe-form';
+
+const RootView = CollectionView.extend({
+  viewComparator: false,
+  el: 'body',
+  template: hbs``,
+  initialize({ model }) {
+    this.render();
+
+    this.addChildView(new FormIframeView({ model }));
+  },
+});
+
+const FormIframeView = View.extend({
+  behaviors: [IframeFormBehavior],
+  className: 'form__frame',
+  template: hbs`
+    <div class="app-frame__content flex-region">
+      <div class="form__layout">
+        <div class="form__content">
+          <iframe src="/formapp/"></iframe>
+        </div>
+      </div>
+    </div>
+  `,
+});
+
+const FormPrintApp = App.extend({
+  channelName: 'app',
+
+  initialize() {
+    initPlatform();
+  },
+
+  onBeforeStart({ name }) {
+    new BootstrapService({ name });
+  },
+
+  beforeStart({ patientId, formId }) {
+    return [
+      Radio.request('bootstrap', 'fetch'),
+      Radio.request('entities', 'fetch:patients:model', patientId),
+      Radio.request('entities', 'forms:model', formId),
+    ];
+  },
+
+  onStart(options, currentUser, [patient], form) {
+    this.addChildApp('formsService', FormsService, {
+      patient: patient,
+      form: form,
+    });
+
+    this.setView(new RootView({ model: form }));
+  },
+});
+
+const Router = Backbone.Router.extend({
+  routes: {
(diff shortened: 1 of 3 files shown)


src/js/formprint/index.js
```diff
@@ -0,0 +1,98 @@
+import 'js/base/setup';
+import 'js/i18n';
```
paulfalgout, line 2: Is this needed?

src/js/formprint/index.js
```diff
@@ -0,0 +1,98 @@
...
+    initPlatform();
+  },
+
+  onBeforeStart({ name }) {
+    new BootstrapService({ name });
+  },
+
+  beforeStart({ patientId, formId }) {
+    return [
+      Radio.request('bootstrap', 'fetch'),
```
paulfalgout, line 64: Ideally we do not have to bootstrap for this app

src/js/formprint/index.js
```diff
@@ -0,0 +1,98 @@
...
+      </div>
+    </div>
+  `,
+});
+
+const FormPrintApp = App.extend({
+  channelName: 'app',
+
+  initialize() {
+    initPlatform();
```
paulfalgout, line 55: I think we won't need this

src/js/formprint/index.js
```diff
@@ -0,0 +1,98 @@
...
+  el: 'body',
+  template: hbs``,
+  initialize({ model }) {
+    this.render();
+
+    this.addChildView(new FormIframeView({ model }));
+  },
+});
+
+const FormIframeView = View.extend({
```
paulfalgout, line 37: what if this view's `el` was `body` and it was the root view.  I think that's be enough

src/js/formprint/index.js
```diff
@@ -0,0 +1,98 @@
...
+      <div class="form__layout">
+        <div class="form__content">
+          <iframe src="/formapp/"></iframe>
+        </div>
+      </div>
+    </div>
+  `,
+});
+
+const FormPrintApp = App.extend({
```
paulfalgout, line 51: My guess is that in order to pull this off with the response, we'll need to pull what we need from the form service and add it to this app, removing the form service.  We have all of the requests we need from the three ids, but they don't cleanly align with what the current form service is doing.

paulfalgout (changes requested):

src/js/formprint/index.js
nmajor25: `import 'js/i18n';` doesn't seem to be needed. Removed it without issue.

Removing `import 'js/base/setup';` breaks the page and throws this browser console error:

```
Uncaught TypeError: dayjs__WEBPACK_IMPORTED_MODULE_1___default(...).utc is not a function
```

Looks like we need the `dayjs` global window variable that is set in the `js/base/setup` file: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/base/setup.js#L43

Replacing the `import 'js/base/setup` import with this fixes the error:

```js
import dayjs from 'dayjs';
import '../base/dayjs';

window.dayjs = dayjs;
```

But it is messier.

src/js/formprint/index.js
nmajor25: Removing this breaks the page and causes errors in the forms service. I'm not really sure why.

Error occurs here: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/services/forms.js#L42

You mentioned in the Shortcut story that there may need to be a modification to the forms service. Is this where the modification would be?

Update the forms service to not need the current user. And only use it if it exists?

src/js/formprint/index.js
nmajor25: Yup, removed this without issue. Good call.

src/js/formprint/index.js
nmajor25: That would be cleaner/simpler. Will play around with that, see if I can get it to work. Seems like it should.

src/js/formprint/index.js
nmajor25: Hmm okay, I'll investigate getting this to work.

src/js/formprint/index.js
paulfalgout: Yep I'm ok with base

src/js/formprint/index.js
paulfalgout: I think it's this https://github.com/RoundingWell/care-ops-frontend/pull/724#discussion_r900739115. We just won't use the form service.  need to put the form service responder stuff on the app here.

nmajor25:
This implementation ended up not working with Puppeteer. Which is how the page would be loaded in its final form.

Therefore, this was closed and replaced by https://github.com/RoundingWell/care-ops-frontend/pull/727.

PR #727, opened 2022-06-22
Create new form service for saving to PDF

nmajor25:
Shortcut Story ID: [sc-29317]

Creates a new form service that handles this URL route: `formservice/:formId/:patientId(/:responseId)` and loads only a form iframe on the page. `responseId` is an optional URL parameter.

This page will be used by the backend team to convert forms to a PDF version when designated patient actions are completed.

### DIFF
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -35,7 +35,7 @@ function startFormService() {
     });
 }
 
-function startAuth(isFormService) {
+function startAuth() {
   import(/* webpackPrefetch: true, webpackChunkName: "auth" */ './auth')
     .then(({ login, logout }) => {
       login(startApp);
@@ -66,7 +66,10 @@ document.addEventListener('DOMContentLoaded', () => {
   const isOutreach = /^\/outreach\//.test(location.pathname);
   const isFormService = /^\/formservice\//.test(location.pathname);
 
-  if (isFormService) startFormService();
+  if (isFormService) {
+    startFormService();
+    return;
+  }
 
   if ((_DEVELOP_ || _E2E_) && sessionStorage.getItem('cypress')) {
     versions.frontend = 'cypress';
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -38,7 +38,7 @@ function startFormService() {
 function startAuth(isFormService) {
   import(/* webpackPrefetch: true, webpackChunkName: "auth" */ './auth')
     .then(({ login, logout }) => {
-      login(isFormService ? startFormService : startApp);
+      login(startApp);
       Radio.reply('auth', {
         logout() {
           logout();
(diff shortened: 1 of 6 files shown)


src/js/index.js
```diff
@@ -57,6 +64,7 @@ const ajaxSetup = {
 document.addEventListener('DOMContentLoaded', () => {
   const isForm = /^\/formapp\//.test(location.pathname);
   const isOutreach = /^\/outreach\//.test(location.pathname);
+  const isFormService = /^\/formservice\//.test(location.pathname);
```
paulfalgout, line 67: OK I found out that puppeteer can just intercept the header and inject the token, so we probably don't need to handle auth with this app at all.

I think we can actually put this right on line 68 even above the cypress check:
```js
if (isFormService) startFormService();
```

No other auth stuff is needed here.

This will mean we can't test this on our sandboxes without some form of puppeteer, but Cypress should likely still work as long as we're properly stubbing all of the requested endpoints.

paulfalgout (changes requested):

src/js/index.js
nmajor25: Okay, just updated and pushed those changes.

src/js/index.js
```diff
@@ -35,7 +35,7 @@ function startFormService() {
     });
 }
 
-function startAuth(isFormService) {
+function startAuth() {
```
nmajor25, line 38: Just removed an unused function parameter. Forgot to remove it in the last commit.

test/integration/formservice/formservice.js
```diff
@@ -0,0 +1,155 @@
...
+      .should('have.value', 'Test Insurance Name');
+
+    cy
+      .get('[name="data[patient.fields.pcp][name]"]')
+      .should('have.value', 'Test PCP Name');
+
+    cy
+      .get('[name="data[patient.fields.address][state]"]')
+      .should('have.value', 'Test State Name');
+  });
```
nmajor25, line 60: Do we need this test form to have all three of these input fields?

We could just trim it down to just the one input field and the test would serve the same purpose, right? Would be cleaner and less lines of code.

paulfalgout: Nope.  In fact you can probably just grab a form from the fixtures for the definition.

It'd be something like: 
```js
  cy
    .fixture('test/form-definition').as('fxTestFormDefinition');

win.postMessage({ message: 'form:pdf', args: {
  definition: this.fxTestFormDefinition,
  ...
```
But it might be simpler to just simplify the form to one input.

paulfalgout (changes requested):
Pretty much done

paulfalgout (approved):

test/integration/formservice/formservice.js
nmajor25: Just went with reducing it to one input.

PR #731, opened 2022-06-28
Load the form model along with the form definition in the form service

nmajor25:
Shortcut Story ID: [sc-29502]

Update the form service so that the form model is always retrieved alongside the form definition. This will ensure that all form data is synced without having to do a hard refresh of the application.

This code change broke several form related tests. So this pull request also includes fixes to those.

### DIFF
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -76,11 +76,9 @@ export default App.extend({
     const channel = this.getChannel();
 
     return $.when(
-      Radio.request(
-        'entities', 'fetch:forms:definition', this.form.id,
-        this.form.fetch(),
-      ),
-    ).then(definition => {
+      Radio.request('entities', 'fetch:forms:definition', this.form.id),
+      this.form.fetch(),
+    ).then(([definition]) => {
       channel.request('send', 'fetch:form:data', {
         definition,
         storedSubmission: submission,
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -64,10 +64,7 @@ export default App.extend({
   fetchForm() {
     const channel = this.getChannel();
 
-    return $.when(
-      Radio.request('entities', 'fetch:forms:definition', this.form.id),
-      this.form.fetch(),
-    )
+    return $.when(Radio.request('entities', 'fetch:forms:definition', this.form.id))
       .then(definition => {
         channel.request('send', 'fetch:form', {
           definition,
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -64,7 +64,10 @@ export default App.extend({
   fetchForm() {
     const channel = this.getChannel();
 
-    return $.when(Radio.request('entities', 'fetch:forms:definition', this.form.id))
+    return $.when(
+      Radio.request('entities', 'fetch:forms:definition', this.form.id),
+      this.form.fetch(),
+    )
       .then(definition => {
         channel.request('send', 'fetch:form', {
           definition,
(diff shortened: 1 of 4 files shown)


test/support/api/forms.js
```diff
@@ -18,6 +18,19 @@ Cypress.Commands.add('routeForms', (mutator = _.identity) => {
     .as('routeForms');
 });
 
+Cypress.Commands.add('routeForm', (mutator = _.identity) => {
+  cy
+    .fixture('test/forms').as('fxTestForms');
+
+  cy.route({
+    url: '/api/forms/*',
```
nmajor25, line 26: Does this work?

Or should we be doing `url: '/api/forms/11111'` instead?

test/integration/patients/patient/sidebar.js
```diff
@@ -91,6 +91,7 @@ context('patient sidebar', function() {
     cy
       .server()
       .routePatientActions(_.identity, '2')
+      .routeForm()
```
nmajor25: Since we don't interact with the form contents in this test, it wasn't broken by adding `this.form.fetch()` in the forms service.

But for the form to display in the snapshots and to remove the `400` request error, this route needed to be added. Since the `/api/forms/11111` request is being made by Cypress when this test runs.

test/support/api/forms.js
paulfalgout: hmm I think the response needs to be a single form.  Tests aren't currently breaking because it can't read this response.

paulfalgout: honestly for now lets just change this to `return mutator({});` without a form model and the `*` becomes ok.  all this is doing is cache busting so it's an unnecessary amount of test setup for little payoff.. however I'll probably regret this :-D

nmajor25: Just pushed that update. With `return mutator({});`.

test/support/api/forms.js
```diff
@@ -18,6 +18,19 @@ Cypress.Commands.add('routeForms', (mutator = _.identity) => {
     .as('routeForms');
 });
 
+Cypress.Commands.add('routeForm', (mutator = _.identity) => {
+  cy
```
paulfalgout, line 22: don't need the fixture now

test/integration/patients/patient/sidebar.js
nmajor25: Correction, not including `routeForm()` does indeed cause a broken test. This test loads and interacts with a form inside a modal window.

paulfalgout (changes requested):
one little clean up and I think this can go in

test/support/api/forms.js
nmajor25: Whoops, good catch. Updated just now.

src/js/services/forms.js
```diff
@@ -75,7 +75,12 @@ export default App.extend({
   fetchFormStoreSubmission({ submission }) {
     const channel = this.getChannel();
 
-    return $.when(Radio.request('entities', 'fetch:forms:definition', this.form.id)).then(definition => {
+    return $.when(
+      Radio.request(
+        'entities', 'fetch:forms:definition', this.form.id,
+        this.form.fetch(),
+      ),
+    ).then(definition => {
```
paulfalgout: Ah this is broken.  Not 100% sure why tests pass but this should be
`$.when(request, request)` not `$.when(Radio.request(request, request))` and the `then` would have `[definition]` because of the multiple requests

paulfalgout (changes requested):

src/js/services/forms.js
nmajor25: :facepalm:

paulfalgout: Ha.. though I know why it works.  because `Radio.request` doesn't accept a 4th parameter.. so it isn't hurting anything.. and because it's not in the `$.when` the `definition` doesn't have to be wrapped as an array.. _but_ the `$.when` isn't waiting on the `form.fetch`

nmajor25: Haha that explains it. Well, it should be fixed now :crossed_fingers:

test/integration/forms/form.js
```diff
@@ -165,7 +165,7 @@ context('Patient Action Form', function() {
       .iframe();
   });
 
-  specify('restoring stored submission', function() {
+  specify.only('restoring stored submission', function() {
```
paulfalgout: only :-) @nmajor25

nmajor25: Haha not my day. Alright, just removed those.

PR #734, opened 2022-06-30
Extend the patient dashboard to be the full height of the page

nmajor25:
Shortcut Story ID: [sc-29467]

The patient sidebar on the patient dashboard and flow page doesn't extend to the full height of the page. This was first noticed on a long version of the flow page, where its left border stopped midway (screenshot below).

This pull request updates the CSS so it extends to the full height of the browser window. Regardless of content length.

image (1)
.

### DIFF
--- a/src/js/views/programs/program/flow/program-flow.scss
+++ b/src/js/views/programs/program/flow/program-flow.scss
@@ -1,8 +1,6 @@
 .program-flow__frame {
   display: flex;
-  height: 100%;
-  overflow-x: hidden;
-  overflow-y: auto;
+  flex: 1;
 }
 
 .program-flow__layout {
--- a/src/js/views/forms/form/form.scss
+++ b/src/js/views/forms/form/form.scss
@@ -1,9 +1,7 @@
 .form__frame {
   background: $black-95;
   display: flex;
-  height: 100%;
-  overflow-x: hidden;
-  overflow-y: auto;
+  flex: 1
 }
 
 .form__layout {
--- a/src/js/views/globals/app-frame.scss
+++ b/src/js/views/globals/app-frame.scss
@@ -23,6 +23,7 @@ $menu-bkg-color: #1F2B33;
 
 .app-frame__content {
   min-width: 440px;
+  overflow-y: auto;
 }
 
 .app-frame__sidebar {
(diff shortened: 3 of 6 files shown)


nmajor25:
@paulfalgout just some background info on how I came to this solution.

Per your suggestion off of GitHub, my initial attempt at this was to add `display: table` to `.patient__layout`, remove the left border from the patient sidebar, and then add a right border to `.patient__layout`.

That made the border extend the full height of the page and fixed the issue for the patient dashboard and flow pages.

But this causes some UI issues for the patient sidebar used on the form page. When the border is added there, it remains visible even when the sidebar is collapsed. Although it wasn't super noticeable, it still didn't feel optimal. So I looked for a different way.

After some additional investigating, I narrowed down the issue to the overflow being applied to the`.patient__frame`. Not entirely sure why, but this was stopping the content and sidebar from extending the full page height.

I replaced those overflow styles with `flex: 1;` (tells the elements to fill the available space) and then added `overflow-y: auto;` to `.app-frame__content`.

This seemed to have fixed the issue. The patient sidebars now take the full height of the page, regardless of content length.

And I went through the other pages in the app to ensure that the change to `.app-frame__content` didn't break the layout or anything else in the UI. From what I can tell, everything still looks/works normally.

src/js/views/programs/program/flow/program-flow.scss
```diff
@@ -1,8 +1,6 @@
 .program-flow__frame {
   display: flex;
-  height: 100%;
-  overflow-x: hidden;
-  overflow-y: auto;
+  flex: 1;
 }
```
nmajor25, line 4: This page doesn't use the patient sidebar. But I updated this to remain consistent with the changes I made for the other pages (i.e. `.patient__frame`, `.patient-flow__frame`, `.form__frame`).

Same goes for `.program__frame` in `src/js/views/programs/program/program.scss`.

paulfalgout (changes requested):
Nice. This is a better solution than I would have imagined it would be.  flex FTW.  
Honestly our current app frame styles have some history in having support old-IE so I'm not at all surprised if there are some flex solutions that would work better at a lower level.

paulfalgout (approved):

PR #735, opened 2022-07-01
Temporarily skip a few flaky forms tests

nmajor25:
Shortcut Story ID: [sc-29504]

There are a few forms related Cypress tests that are flaky. We've gotten used to seeing those failed tests on pull requests and have missed code coverage drops as a result.

We'll remove the `skip()` from these tests once we can fix their flakiness.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,6 +116,7 @@
       "src/js/base/**/*.js",
       "src/js/auth.js",
       "src/js/formapp.js",
+      "src/js/formapp/**/*.js",
       "src/js/formservice.js",
       "src/js/index.js",
       "src/js/datadog.js",
--- a/package.json
+++ b/package.json
@@ -115,6 +115,8 @@
     "exclude": [
       "src/js/base/**/*.js",
       "src/js/auth.js",
+      "src/js/formapp.js",
+      "src/js/formservice.js",
       "src/js/index.js",
       "src/js/datadog.js",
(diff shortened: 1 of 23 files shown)


nmajor25:
@paulfalgout these are the four consistently flaky forms tests.

Looks like skipping the `Outreach > Form` test causes a coverage to drop.

paulfalgout (changes requested):
We should add istanbul coverage comments to ignore lines we aren't covering with these tests like:
```js
/* istanbul ignore next: skipping form test flake */
```
where `next` can be a number of ignore commands for istanbul

paulfalgout (changes requested):
Still seem to be missing quite a bit of coverage.

I also wonder if rather than skipping outreach and preview form if we should just comment out the parts of the tests that are flaky?  Might help with the coverage and still test certain parts that work every time.

nmajor25:
@paulfalgout ya coverage is still down :slightly_frowning_face:. 

I like that idea (commenting out the specific flaky portions of the tests). Will work on that.

nmajor25:
Instead of skipping entire flaky tests, we decided to comment out only the flaky portions inside those tests. This allows us to still run the other portions included in those tests and it also helps with the code coverage issues that that change would cause.

Replaced by this pull request: https://github.com/RoundingWell/care-ops-frontend/pull/740.

PR #736, opened 2022-07-05
Move _buildMatchers function to the utils directory

nmajor25:
Shortcut Story ID: [sc-28864]

We currently use an identical function called `_buildMatchers()` in the worklist, schedule, and clinicians views. This function builds an array of regular expressions that is used for filtering items in the lists on those pages (`"Find in list..."`).

This pull request moves that `_buildMatchers()` function to the `/src/js/utils/formatting` directory and then imports it into the worklist, schedule, and clinicians view files to be used.

This will make things cleaner and more maintainable in the future.

### DIFF
--- a/src/js/utils/formatting/build-matchers-array.js
+++ b/src/js/utils/formatting/build-matchers-array.js
@@ -1,0 +1,13 @@
+import { map } from 'underscore';
+
+import words from 'js/utils/formatting/words';
+
+export default query => {
+  const searchWords = words(query);
+
+  return map(searchWords, function(word) {
+    word = RegExp.escape(word);
+
+    return new RegExp(`\\b${ word }`, 'i');
+  });
+};
--- a/src/js/utils/formatting/build-matchers-array.js
+++ b/src/js/utils/formatting/build-matchers-array.js
@@ -3,11 +3,9 @@ import { map } from 'underscore';
 import words from 'js/utils/formatting/words';
 
 export default query => {
-  const searchWords = words(query);
+  const searchWords = map(words(query), RegExp.escape);
 
   return map(searchWords, function(word) {
-    word = RegExp.escape(word);
-
     return new RegExp(`\\b${ word }`, 'i');
   });
 };
--- a/src/js/views/clinicians/clinicians-all_views.js
+++ b/src/js/views/clinicians/clinicians-all_views.js
@@ -3,7 +3,7 @@ import hbs from 'handlebars-inline-precompile';
 import Radio from 'backbone.radio';
 import { View, CollectionView, Behavior } from 'marionette';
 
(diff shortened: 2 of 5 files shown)


nmajor25:
We currently have the [/js/utils/formatting/build-matcher.js](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/utils/formatting/build-matcher.js) function that is similar to the `_buildMatchers()` function we use for the `Find in list...` functionailty. But it's slightly different.

* **/js/utils/formatting/build-matcher.js**: given a string of `test string`, it creates a regular expression of `/\btest|string/gi`.
* **_buildMatchers()**: given a string of `test string`, it creates an array that contains each word converted to a regular expression: `[/\btest/i, /\bstring/i]`. And then the view is looping through that array to filter out actions/flows, clinicians from the list.

These are different enough that I decided to create a new `/js/utils/formatting/build-matchers-array.js` file and use that instead of the existing `/js/utils/formatting/build-matcher.js` function.

Curious what you think on this, @paulfalgout .

nmajor25:
Side note. If we decide to move forward with this, I'll add a unit test for this new utility function.

paulfalgout (approved):
Let's add that test.

The major difference here is that this is essentially building an `and` search as we always seem to use it with `_.every` vs the `or` `|` in the other regex.  There really isn't a regex and.. so this makes sense to me.

src/js/utils/formatting/build-matchers-array.js
```diff
@@ -3,11 +3,9 @@ import { map } from 'underscore';
...
 export default query => {
-  const searchWords = words(query);
+  const searchWords = map(words(query), RegExp.escape);
 
   return map(searchWords, function(word) {
-    word = RegExp.escape(word);
-
     return new RegExp(`\\b${ word }`, 'i');
   });
 };
```
nmajor25, line 11: Cypress was throwing an error when it tried to use this function (image below). Not really sure why. But updating the function to this fixed the error. And it still works the same.

Screen Shot 2022-07-07 at 10 39 30 AM

paulfalgout: `RegExp.escape` is added in `js/base/setup`. I suppose we could move that to it's own utility to it's clearer.  This helper is also super old.

using `map` is essentially doing this: `map(words(query, undefined))` which returns the array.

It's probably ok in this case as we're certainly not testing the escape function here.

nmajor25: Ah that makes sense now... I was wondering why the unit test for `src/js/utils/formatting/build-matcher.js` was working and this one was not. But `RegExp.escape` is undefined in that one too. It's just not being tested, so it didn't cause issues.

PR #738, opened 2022-07-07
Add formapp and formservice to nyc exclude

paulfalgout:
Shortcut Story ID: [sc-29593]

PR #740, opened 2022-07-08
Comment out flaky portions of form tests

nmajor25:
Shortcut Story ID: [sc-29504]

There are a few forms related Cypress tests that are flaky. We've gotten used to seeing those failed tests on pull requests and have missed code coverage drops as a result.

This pull request comments out the flaky portions of those tests.

We also added the `/js/formapp/` directory to the nyc exclude section of the `package.json` file. This tells Coveralls to not include those files in the report. This is related to https://github.com/RoundingWell/care-ops-frontend/pull/738.

**Note**: The original approach we took was to `skip()` these tests all together (https://github.com/RoundingWell/care-ops-frontend/pull/735). But that caused lots of code coverage report issues. So we switched to this approach instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,6 +116,7 @@
       "src/js/base/**/*.js",
       "src/js/auth.js",
       "src/js/formapp.js",
+      "src/js/formapp/**/*.js",
       "src/js/formservice.js",
       "src/js/index.js",
       "src/js/datadog.js",
--- a/src/js/outreach/index.js
+++ b/src/js/outreach/index.js
@@ -163,7 +163,7 @@ const FormApp = App.extend({
       actionId: this.actionId,
       response,
     })
-      .done(() => {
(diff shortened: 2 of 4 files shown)


nmajor25:
@paulfalgout there is this missing function in the code coverage: https://coveralls.io/builds/50720756/source?filename=src%2Fjs%2Fservices%2Fforms.js#L49.

Is this new? I didn't notice it when I ran the coverage report on my local machine. If it is indeed new, I'm not sure what's causing it...

paulfalgout:
So the issue here with coverage is a race condition I think.  Technically I think it does represent some missing tests.  I think we are testing previous submissions, but not that they're being stored.. However this code gets hit when the tests run slow enough for it to happen which is frequent.  So we probably do need a test here to test the contents of localstorage

PR #744, opened 2022-07-11
Hide tooltip for schedule list actions when their details field is empty

nmajor25:
Shortcut Story ID: [sc-29611]

This configures the tooltip for each action in the schedule list to only show if it has content in its details field.

Currently, the tooltip is shown if it has either a details field or a flow is attached to it. In some scenarios, this means the tooltip is shown even though the details field for that action is empty. This pull request makes it so the tooltip is hidden in those scenarios.

### DIFF
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -177,7 +177,7 @@ const DayItemView = View.extend({
     Radio.trigger('event-router', 'patient:action', this.model.get('_patient'), this.model.id);
   },
   showDetailsTooltip() {
-    if (!this.model.get('details') && !this.model.getFlow()) return;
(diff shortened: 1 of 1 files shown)


nmajor25:
@paulfalgout 

I don’t think the current tests for the schedule need to be changed.

This part verifies that the tooltip should not be shown: https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/integration/patients/worklist/schedule.js#L289.

If the `if (!this.model.get('details')) return;` check is removed, it breaks the test. 

And if you add details content for this fixture action: https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/integration/patients/worklist/schedule.js#L81, the test breaks as well.

Unless I’m missing something…?

PR #745, opened 2022-07-14
Click+shift multi select for worklist items

nmajor25:
Shortcut Story ID: [sc-29507]

When an item is selected in the worklist, allow a user to `shift`+`click` to select multiple list items between the two.

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -295,8 +295,10 @@ export default App.extend({
     });
 
     this.listenTo(typeToggleView, 'toggle:listType', listType => {
-      this.setState('listType', listType);
-      this.setState('lastSelectedId', null);
+      this.setState({
+        listType: listType,
+        lastSelectedId: null,
+      });
     });
 
     this.showChildView('toggle', typeToggleView);
@@ -325,6 +327,7 @@ export default App.extend({
   setSearchState(state, searchQuery) {
     this.setState({
       searchQuery: searchQuery.length > 2 ? searchQuery : '',
+      lastSelectedId: null,
     });
   },
 });
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -55,7 +55,7 @@ export default App.extend({
       // NOTE: Allows for new defaults to get added to stored filters
       storedState.filters = extend({}, filters, storedState.filters);
 
-      storedState.lastSelectedId = null;
+      storedState.lastSelectedIndex = null;
     }
 
     this.setState(extend({ id: this.worklistId }, storedState));
@@ -207,7 +207,7 @@ export default App.extend({
       return;
     }
 
-    this.getState().selectMultiple(this.filteredCollection);
+    this.getState().selectMultiple(this.filteredCollection.map('id'));
   },
   getComparator() {
     const sortId = this.getState().getSort();
@@ -297,7 +297,7 @@ export default App.extend({
     this.listenTo(typeToggleView, 'toggle:listType', listType => {
       this.setState({
         listType: listType,
-        lastSelectedId: null,
+        lastSelectedIndex: null,
       });
     });
 
@@ -327,7 +327,7 @@ export default App.extend({
   setSearchState(state, searchQuery) {
     this.setState({
       searchQuery: searchQuery.length > 2 ? searchQuery : '',
-      lastSelectedId: null,
+      lastSelectedIndex: null,
     });
   },
 });
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -203,7 +203,7 @@ export default App.extend({
       return;
     }
 
-    this.getState().selectAll(this.filteredCollection);
+    this.getState().selectMultiple(this.filteredCollection);
   },
   getComparator() {
     const sortId = this.getState().getSort();
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -51,8 +51,12 @@ export default App.extend({
     const storedState = store.get(`${ this.worklistId }_${ currentUser.id }-v2`);
     const filters = this.getState('filters');
 
-    // NOTE: Allows for new defaults to get added to stored filters
-    if (storedState) storedState.filters = extend({}, filters, storedState.filters);
+    if (storedState) {
+      // NOTE: Allows for new defaults to get added to stored filters
+      storedState.filters = extend({}, filters, storedState.filters);
+
+      storedState.lastSelectedId = null;
+    }
 
     this.setState(extend({ id: this.worklistId }, storedState));
   },
@@ -292,6 +296,7 @@ export default App.extend({
 
     this.listenTo(typeToggleView, 'toggle:listType', listType => {
       this.setState('listType', listType);
+      this.setState('lastSelectedId', null);
     });
 
     this.showChildView('toggle', typeToggleView);
--- a/src/js/apps/patients/worklist/worklist_state.js
+++ b/src/js/apps/patients/worklist/worklist_state.js
@@ -149,7 +149,7 @@ export default Backbone.Model.extend({
   getSelectedList() {
     return this.isFlowType() ? this.get('selectedFlows') : this.get('selectedActions');
   },
-  toggleSelected(model, isSelected, selectedIndex = null) {
+  toggleSelected(model, isSelected, selectedIndex) {
     const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
     const currentList = clone(this.get(listName));
 
@@ -159,7 +159,7 @@ export default Backbone.Model.extend({
 
     this.set({
       [listName]: newList,
-      lastSelectedIndex: selectedIndex,
+      lastSelectedIndex: isSelected ? selectedIndex : null,
     });
   },
   isSelected(model) {
--- a/src/js/apps/patients/worklist/worklist_state.js
+++ b/src/js/apps/patients/worklist/worklist_state.js
@@ -181,8 +181,10 @@ export default Backbone.Model.extend({
   clearSelected() {
     const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
 
-    this.set(listName, {});
-    this.set('lastSelectedId', null);
+    this.set({
+      [listName]: {},
+      lastSelectedId: null,
+    });
 
     this.trigger('select:none');
   },
@@ -196,8 +198,10 @@ export default Backbone.Model.extend({
       return selected;
     }, clone(currentSelectedList));
 
-    this.set(listName, newSelectedList);
-    this.set('lastSelectedId', newLastSelectedId || null);
+    this.set({
+      [listName]: newSelectedList,
+      lastSelectedId: newLastSelectedId,
+    });
 
     this.trigger('select:multiple');
   },
--- a/src/js/apps/patients/worklist/worklist_state.js
+++ b/src/js/apps/patients/worklist/worklist_state.js
@@ -35,7 +35,7 @@ export default Backbone.Model.extend({
         roleId: this.currentClinician.getRole().id,
         noOwner: false,
       },
-      lastSelectedId: null,
+      lastSelectedIndex: null,
       selectedActions: {},
       selectedFlows: {},
       searchQuery: '',
@@ -149,15 +149,18 @@ export default Backbone.Model.extend({
   getSelectedList() {
     return this.isFlowType() ? this.get('selectedFlows') : this.get('selectedActions');
   },
-  toggleSelected(model, isSelected) {
+  toggleSelected(model, isSelected, selectedIndex = null) {
     const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
-    const list = clone(this.get(listName));
+    const currentList = clone(this.get(listName));
 
-    this.set(listName, extend(list, {
+    const newList = extend(currentList, {
       [model.id]: isSelected,
-    }));
+    });
 
-    this.set('lastSelectedId', isSelected ? model.id : null);
+    this.set({
+      [listName]: newList,
+      lastSelectedIndex: selectedIndex,
+    });
   },
   isSelected(model) {
     const list = this.getSelectedList();
@@ -183,24 +186,24 @@ export default Backbone.Model.extend({
 
     this.set({
       [listName]: {},
-      lastSelectedId: null,
+      lastSelectedIndex: null,
     });
 
     this.trigger('select:none');
   },
-  selectMultiple(collection, newLastSelectedId) {
+  selectMultiple(selectedIds, newLastSelectedIndex = null) {
     const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
 
     const currentSelectedList = this.get(listName);
 
-    const newSelectedList = collection.reduce((selected, model) => {
-      selected[model.id] = true;
+    const newSelectedList = selectedIds.reduce((selected, id) => {
+      selected[id] = true;
       return selected;
     }, clone(currentSelectedList));
 
     this.set({
       [listName]: newSelectedList,
-      lastSelectedId: newLastSelectedId,
+      lastSelectedIndex: newLastSelectedIndex,
     });
 
     this.trigger('select:multiple');
--- a/src/js/apps/patients/worklist/worklist_state.js
+++ b/src/js/apps/patients/worklist/worklist_state.js
@@ -36,6 +36,7 @@ export default Backbone.Model.extend({
         noOwner: false,
       },
       selectedActions: {},
+      lastSelectedAction: null,
       selectedFlows: {},
       searchQuery: '',
       listType: 'flows',
@@ -192,5 +193,24 @@ export default Backbone.Model.extend({
     this.set(listName, list);
 
     this.trigger('select:all');
+  },
+  setLastSelectedAction(actionId) {
+    this.set('lastSelectedAction', actionId);
+  },
+  getLastSelected() {
+    return this.get('lastSelectedAction');
+  },
+  selectMultiple(arrayOfActionIds) {
+    const currentSelectedList = this.get('selectedActions');
+
+    const newSelectedList = arrayOfActionIds.reduce((selected, itemId) => {
+      selected[itemId] = true;
+      return selected;
+    }, currentSelectedList);
+
+    this.set('selectedActions', newSelectedList);
+
+    this.trigger('select:multiple');
+    this.trigger('change:selectedActions');
   },
 });
--- a/src/js/apps/patients/worklist/worklist_state.js
+++ b/src/js/apps/patients/worklist/worklist_state.js
@@ -182,6 +182,8 @@ export default Backbone.Model.extend({
     const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
 
     this.set(listName, {});
+    this.set('lastSelectedId', null);
+
     this.trigger('select:none');
   },
   selectMultiple(collection, newLastSelectedId) {
@@ -195,8 +197,7 @@ export default Backbone.Model.extend({
     }, clone(currentSelectedList));
 
     this.set(listName, newSelectedList);
-
-    if (newLastSelectedId) this.set('lastSelectedId', newLastSelectedId);
+    this.set('lastSelectedId', newLastSelectedId || null);
 
     this.trigger('select:multiple');
   },
--- a/src/js/apps/patients/worklist/worklist_state.js
+++ b/src/js/apps/patients/worklist/worklist_state.js
@@ -35,8 +35,8 @@ export default Backbone.Model.extend({
         roleId: this.currentClinician.getRole().id,
         noOwner: false,
       },
+      lastSelectedId: null,
       selectedActions: {},
-      lastSelectedAction: null,
       selectedFlows: {},
       searchQuery: '',
       listType: 'flows',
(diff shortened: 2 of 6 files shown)


nmajor25:
General approach:

1. Whenever a user selects an item in the worklist, store the id of that item. This will allow us to track the user’s last selected item.
2. When a list item is clicked, detect if the user was pressing the shift key (`event.shiftKey`) when they made the click. And that there is a previously selected item stored in the state.
    - If one of those cases are not true, handle the user’s selection normally.
3. Get the `id` of each list item between the currently selected item and the previously selected item.
    - Should support the user selecting both directions (up/down) in the list.
4. Update the state to include those `id`s.
    - New list of selected items should include both the new and previously selected `id`s.

src/js/views/patients/worklist/action_views.js
```diff
@@ -74,9 +75,16 @@ const ActionItemView = View.extend({
...
+
+    if (domEvent.shiftKey && this.state.get('lastSelectedAction') && !isSelected) {
+      this.triggerMethod('click:shiftMultiSelect', this.model.id);
+      return;
+    }
+
     this.state.toggleSelected(this.model, !isSelected);
+    this.state.setLastSelectedAction(!isSelected ? this.model.id : null);
     this.render();
   },
```
nmajor25, line 80: Intercepts the click event on an individual action in the list. If the `shiftKey` event property is `true`, it means the user was holding the shift key when they made the click.

If `shiftKey === true`, a last selection action `id` value exists in the state, and the user wasn’t un-selecting the action, trigger a `click:shiftMultiSelect` method that is passed to the `worklist_views.js` file to handle (where we can access the full list of actions in the list). 

Couple notes:

- `this` and the jQuery click event are passed to the function as parameters. I needed to pass both parameters in to get access to the jQuery event, but didn’t know what to call the first one. So just named it `param` for now.

- In this initial implementation, I only built the `shift`+`click` functionality for selecting items. And not for un-selecting items. So if the user is un-selecting an item, I don’t trigger the `click:shiftMultiSelect` method and I also clear the `lastSelectedAction` state value if that’s the case.

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -213,6 +214,20 @@ const ListView = CollectionView.extend({
...
+    const lastSelectedIndex = actionsList.findIndex(action => action.model.id === lastSelectedActionId);
+
+    const arrayOfActionIdsToSelect = actionsList
+      .slice(Math.min(selectedIndex, lastSelectedIndex), Math.max(selectedIndex, lastSelectedIndex) + 1)
+      .map(action => action.model.id);
+
+    this.state.selectMultiple(arrayOfActionIdsToSelect);
+    this.state.setLastSelectedAction(selectedActionId);
+  },
 });
```
nmajor25: Consumes the selection action `id` passed to it from the action child view.

Using the `lastSelectedAction` value, it gets the index position of the current and previous selected actions. And uses those indexes to slice the list of actions to only include the ones between those two indexes.

Then, it creates an array of action `id`s that are passed to the state. `Math.min()` and `Math.max()` allow the user to select both directions in the list (up or down). 

The array of action `id`s is then passed to the state.

Some notes:

- `this.children_views` is where the currently sorted list items are stored. Not in the app. So my understanding is that this needs to be handled here, in the view.
- Looking at `selectedIndex`, `lastSelectedIndex`, and `arrayOfActionIdsToSelect` now, I should be able to use underscore methods for those.

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -193,4 +194,23 @@ export default Backbone.Model.extend({
...
+      selected[itemId] = true;
+      return selected;
+    }, currentSelectedList);
+
+    this.set('selectedActions', newSelectedList);
+
+    this.trigger('select:multiple');
+    this.trigger('change:selectedActions');
+  },
 });
```
nmajor25, line 211: Takes the array of action `id`s and converts them into an object in the `{‘id’: true, ‘id’: true}` format we use in the state. Then sets that new list in the state.

Make sure to include the `id`s that already existed in the `selectedActions` object.

Some notes:

- I couldn’t get the state change event to trigger. So I’m manually doing `this.trigger(‘change:selectedActions’)` for that reason. Not sure why that isn’t triggering. Any ideas?

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -193,4 +194,23 @@ export default Backbone.Model.extend({
 
     this.trigger('select:all');
   },
+  setLastSelectedAction(actionId) {
```
paulfalgout: I think this function and `getLastSelected` aren't necessary.  Other areas can just use `set` and `get`

src/js/views/patients/worklist/action_views.js
```diff
@@ -74,9 +75,16 @@ const ActionItemView = View.extend({
   onClickPatient() {
     Radio.trigger('event-router', 'patient:dashboard', this.model.get('_patient'));
   },
-  onClickSelect() {
+  onClickSelect(param, domEvent) {
```
paulfalgout: I think `view` would be ok for the 1st parameter

src/js/views/patients/worklist/action_views.js
```diff
@@ -74,9 +75,16 @@ const ActionItemView = View.extend({
...
     const isSelected = this.state.isSelected(this.model);
+
+    if (domEvent.shiftKey && this.state.get('lastSelectedAction') && !isSelected) {
+      this.triggerMethod('click:shiftMultiSelect', this.model.id);
+      return;
+    }
+
     this.state.toggleSelected(this.model, !isSelected);
+    this.state.setLastSelectedAction(!isSelected ? this.model.id : null);
```
paulfalgout: should this happen if anything else were to call `toggleSelected`?  maybe it should move into that function?

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -193,4 +194,23 @@ export default Backbone.Model.extend({
...
+    this.set('lastSelectedAction', actionId);
+  },
+  getLastSelected() {
+    return this.get('lastSelectedAction');
+  },
+  selectMultiple(arrayOfActionIds) {
```
paulfalgout: Actually I don't think this method is necessary.  We should just rename `selectAll` to `selectMultiple` and change the event to `select:multiple`

src/js/apps/patients/worklist/worklist_state.js
paulfalgout: yeah the change not occurring is certainly an issue.. hmm

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -213,6 +214,20 @@ const ListView = CollectionView.extend({
       });
     });
   },
+  onClickShiftMultiSelect(selectedActionId) {
+    const actionsList = this.children._views;
```
paulfalgout: I don't think we should access `_views` directly.

`this.children.findByModel(action)` will work if the event passes the model rather than the id.
Then I think `this.collection.get(lastSelectedActionId)` gets that model.

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -213,6 +214,20 @@ const ListView = CollectionView.extend({
...
+    const lastSelectedActionId = this.state.getLastSelected();
+
+    const selectedIndex = actionsList.findIndex(action => action.model.id === selectedActionId);
+    const lastSelectedIndex = actionsList.findIndex(action => action.model.id === lastSelectedActionId);
+
+    const arrayOfActionIdsToSelect = actionsList
+      .slice(Math.min(selectedIndex, lastSelectedIndex), Math.max(selectedIndex, lastSelectedIndex) + 1)
+      .map(action => action.model.id);
+
+    this.state.selectMultiple(arrayOfActionIdsToSelect);
```
paulfalgout: Well now that I think we should rename `selectAll` we need to pass a collection here.  that'd work like the `this.filteredCollection`.  In fact we could utilize this collection as it is a model dump of the views filtered by find in list.

I think it'd be something like:
```js
const selectedCollection = this.filteredCollection.clone();
selectedCollection.reset(selectedCollection.filter(model => {
  const view = this.children.findByModel(model);
  const index = this.children.findIndexByView(view);
  return  (index > minIndex) && (index < maxIndex);
}));
this.state.selectMultiple(selectedCollection);
```

paulfalgout (changes requested):
I think somehow the `select:all` / `select:none` and probably find in list also needs to be taken into account regarding when to clear the last selected.  Like if the find in list filters out the last selected what do we do?  I kind of think all three of these actions just clear it out always.. otherwise the edge cases get crazy.

src/js/views/patients/worklist/worklist_views.js
paulfalgout: BTW `_` prefixed variables in marionette and the backbone ecosystem are considered "private-like" API.  Meaning there's no guarantee that they're going to work in newer versions, and they're recommending using something else.. _however_ they're exposed because they can be used.  But it typically indicates there's a more preferred way to get at this data.. or using/messing with this directly could have adverse effects etc.

src/js/apps/patients/worklist/worklist_state.js
nmajor25: Using `clone` seems to fix this.

```js
const newSelectedList = arrayOfActionIds.reduce((selected, itemId) => {
  selected[itemId] = true;
  return selected;
}, clone(currentSelectedList));
```

Or, using `Object.assign()` does as well:

```js
Object.assign({}, currentSelectedList, newSelectedList)
```

paulfalgout: ah right.. this was reducing the value already in the state

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -51,8 +51,12 @@ export default App.extend({
     const storedState = store.get(`${ this.worklistId }_${ currentUser.id }-v2`);
     const filters = this.getState('filters');
 
-    // NOTE: Allows for new defaults to get added to stored filters
-    if (storedState) storedState.filters = extend({}, filters, storedState.filters);
+    if (storedState) {
+      // NOTE: Allows for new defaults to get added to stored filters
+      storedState.filters = extend({}, filters, storedState.filters);
+
+      storedState.lastSelectedId = null;
```
nmajor25: Clears the `state.lastSelectedId` value whenever the worklist page is initially loaded.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -292,6 +296,7 @@ export default App.extend({
 
     this.listenTo(typeToggleView, 'toggle:listType', listType => {
       this.setState('listType', listType);
+      this.setState('lastSelectedId', null);
```
nmajor25: Clears `state.lastSelectedId` when the user toggles between the `Actions` and `Flows` versions of the page via the buttons below.
Screen Shot 2022-07-18 at 5 12 55 PM

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -182,6 +182,8 @@ export default Backbone.Model.extend({
     const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
 
     this.set(listName, {});
+    this.set('lastSelectedId', null);
+
```
nmajor25, line 191: Clears the `state.lastSelectedId` value when the multi select option is cancelled.

A user would do that by either clicking the green checkmark or the `Cancel` option in the screenshot below.

Screen Shot 2022-07-18 at 5 19 02 PM

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -195,8 +197,7 @@ export default Backbone.Model.extend({
     }, clone(currentSelectedList));
 
     this.set(listName, newSelectedList);
-
-    if (newLastSelectedId) this.set('lastSelectedId', newLastSelectedId);
+    this.set('lastSelectedId', newLastSelectedId || null);
```
nmajor25: When the `selectMultiple()` function is called for the `click`+`shift` functionality, the `newLastSelectedId` parameter will be included. But when the function is used for the bulk select button (that previously used `selectAll()`), the `newLastSelectedId` function parameter will not be included.

Therefore, the `state.lastSelectedId` value will be set to `newLastSelectedId` if it exists as a function parameter. And set to `null` if not.

As a result, `state.lastSelectedId` will be set when the `click`+`shift` action is taken. And cleared when the user is using the bulk select button in UI.

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -213,6 +213,8 @@ const ListView = CollectionView.extend({
         return matcher.test(searchString);
       });
     });
+
+    this.state.set('lastSelectedId', null);
```
nmajor25: Clears the `state.lastSelectedId` value when the `Find in list...` filtering is applied to the list.

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: should probably be `this.setState({ ... })` instead of two changes

src/js/apps/patients/worklist/worklist_state.js
paulfalgout: `this.set({ ... })` both changes in the same `set` would be ideal.. slightly awkward with the changing key

src/js/apps/patients/worklist/worklist_state.js
paulfalgout: combine `set`s.  I also think you can put `= null` in the parameter more to designate a default for the method.

src/js/views/patients/worklist/worklist_views.js
paulfalgout: I think this can move to the app's `setSearchState` inside that `setState` as this function is getting called when that state change is being made

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -212,6 +213,26 @@ const ListView = CollectionView.extend({
...
+  },
+  onClickShiftMultiSelect(selectedModelId) {
+    const lastSelectedId = this.state.get('lastSelectedId');
+
+    let selectedCollection = this.collection.clone();
+    selectedCollection.reset(this.children.map('model'));
+
+    const selectedIndex = selectedCollection.findIndex(model => model.id === selectedModelId);
```
paulfalgout: I believe these can be a predicate like `findIndex({ id: selectedModelId })` etc without the function

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -179,18 +182,23 @@ export default Backbone.Model.extend({
...
     const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
 
-    const list = collection.reduce((selected, model) => {
+    const currentSelectedList = this.get(listName);
+
+    const newSelectedList = collection.reduce((selected, model) => {
       selected[model.id] = true;
       return selected;
-    }, {});
+    }, clone(currentSelectedList));
```
paulfalgout, line 202: This is _probably_ ok for the select all functionality..  Makes me a little nervous, but I suspect it's always `{}` in the case of select all?

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -212,6 +213,26 @@ const ListView = CollectionView.extend({
...
+    const lastSelectedIndex = selectedCollection.findIndex(model => model.id === lastSelectedId);
+
+    const minIndex = Math.min(selectedIndex, lastSelectedIndex);
+    const maxIndex = Math.max(selectedIndex, lastSelectedIndex);
+
+    selectedCollection = selectedCollection.filter((model, index) => {
```
paulfalgout: actually I think this needs to be `selectedCollection.reset` and keep it a `const` `filter` returns an array of models and not a collection.. which works in this instance because of `Array.prototype.reduce`, but it's not a collection.
Also I think you were using `Array.prototype.slice` previously, and I'm just now realizing we can do that here too `selectedCollection.slice`

paulfalgout (changes requested):
Overall this is getting close but now that I'm seeing it,  this is a lot of data work for a view click event.   I'm writing this last after all of the other comments...

_In fact_ I think maybe we're going to the data too soon.  

What if we changed the event to:
```js
this.triggerMethod('click:shiftMultiSelect', this);
```
and set a `lastSelectedIndex` there instead of the model id in the state's `toggleSelect`
Then I think the handler function would be something like
```js
const lastSelectedIndex = this.state.get('lastSelectedIndex');
const selectedIndex = this.children.findIndexByView(selectedView);

const minIndex = Math.min(selectedIndex, lastSelectedIndex);
const maxIndex = Math.max(selectedIndex, lastSelectedIndex);

const selected = this.children.map('models').slice(minIndex, maxIndex + 1);
this.state.selectMultiple(selected, selectedIndex);
```

_but_ if we do that we should also change it to:
```js
this.getState().selectMultiple(this.filteredCollection.toArray());
```
so that neither is a backbone collection, which seems ok to me.
For that matter it could also be `this.children.map(view => view.model.id)` and `this.filteredCollection.map('id')` instead of passing models.

src/js/apps/patients/worklist/worklist_app.js
nmajor25: Agreed, much cleaner.

src/js/apps/patients/worklist/worklist_state.js
nmajor25: Tried combining these initially, but the linter was force changing it into: `this.set({'listName': newSelectedList, 'lastSelectedId': newLastSelectedId || null});`. Where the `listName` variable key is changed to a string.

But looking at it now, we should be able to use a [computed property name](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#computed_property_names):

```js
this.set({
  [listName]: newSelectedList,
  'lastSelectedId': newLastSelectedId || null,
});
```

paulfalgout: hmm our eslint is doing that?  I can't get it to happen.  FWIW I think the key shouldn't be quoted..  
We definitely do:
```js
this.set({
  [key]: value,
});
```
in other places.

src/js/apps/patients/worklist/worklist_state.js
nmajor25: It doesn't look like `clone(currentSelectedList)` is always `{}` when this function is used by select all. It will contain any items that are currently selected when the select all button is clicked.

I assumed that was okay, since any duplicates are just overridden in that `reduce()` when the new value is set via `selected[model.id] = true;`.

But if there is a better way to do this, I'm open to it 😄

paulfalgout: It's probably not a problem.  If you select something and then find-in-list it out of the views and then select all, does it stay in the list?  I assume it's pulled out, but I'm not sure.

nmajor25: Hmm didn't think of that scenario...

Looks like the selected item stays in the `clone(currentSelectedList)` list when it's find-in-list'd out (and then the select all is clicked afterward). 

When you bulk edit those items, the changes only apply to the ones visible on the page. Not the non-visible one that still exists in `clone(currentSelectedList)`. So that's good.

But now this makes me nervous too...

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -200,6 +200,28 @@ const ListView = CollectionView.extend({
     this.triggerMethod('update:listDom', this);
     this.triggerMethod('filtered', this.children.map('model'));
   },
+  onSelect(selectedView, isShiftKeyPressed) {
+    const isSelected = this.state.isSelected(selectedView.model);
+    const selectedIndex = this.children.findIndexByView(selectedView);
+    const lastSelectedIndex = this.state.get('lastSelectedIndex');
+
+    if (isShiftKeyPressed && lastSelectedIndex !== null && !isSelected) {
```
paulfalgout, line 208: good call on `!== null` here

paulfalgout (approved):
Looks good to me

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -159,7 +159,7 @@ export default Backbone.Model.extend({
 
     this.set({
       [listName]: newList,
-      lastSelectedIndex: selectedIndex,
+      lastSelectedIndex: isSelected ? selectedIndex : null,
```
nmajor25, line 162: I missed this when doing the refactor in the previous commit/s.

This clears the `state.lastSelectedIndex` value (set to `null`) when a list item is un-selected.

test/integration/patients/worklist/worklist.js
```diff
@@ -2703,6 +2703,216 @@ context('worklist page', function() {
...
+
+    cy
+      .go('back')
+      .then(() => {
+        const storage = JSON.parse(localStorage.getItem('owned-by_11111-v2'));
+
+        expect(storage.lastSelectedIndex).to.equal(null);
+      });
+  });
+
```
nmajor25, line 2915: **Overview of these tests**:

1. Load the `/worklist/owned-by` page with the `Flows` version of the list shown and display 3 flows in the list.

2. Test the `click`+`shift` functionality in the list.
    - Multi-select from top to bottom in the list.
    - Multi-select from bottom to top in the list.
    - Verify that the `state.lastSelectedIndex` value is set properly when list items are selected (by examining `localStorage` value).
    
3. Verify that the `state.lastSelectedIndex` value is being properly cleared (set to `null`) in the correct situations.
    - When a list item is un-selected.
    - When the select all option is clicked.
    - When the bulk select cancel option is clicked.
    - When the list is filtered using the `Find in list…` input.
    - When the worklist page initially loads.
      - This is tested by selecting a list item, navigating to a new page, and then going back to the worklist where the `lastSelectedIndex` should equal `null` in `localStorage`.

**Questions I have**:

1. Is this a good place to put these tests in the codebase? Or should they go in the `test/integration/patients/worklist/bulk-edit.js` file?
2. I'm only testing the `Flows` version of the worklist in these tests. The `Flows` and `Actions` lists are almost identical, so I figured that would suffice?

paulfalgout (approved):

PR #752, opened 2022-08-01
Move list item checkbox to component 

paulfalgout:
Shortcut Story ID: [sc-29967]

The checkbox re-rendering the whole item row caused all of the components to re-render.  Not terrible per-row, but in a multiselect situation, it really stacked up.  I _think_ we might still have a memory leak somewhere, but I'm still looking into that, and can measure better once this makes it to QA.

### DIFF
--- a/src/js/apps/patients/sidebar/bulk-edit-actions_app.js
+++ b/src/js/apps/patients/sidebar/bulk-edit-actions_app.js
@@ -147,6 +147,7 @@ export default App.extend({
     });
   },
   onSubmit() {
+    this.modal.disableSubmit();
     this.triggerMethod('save', this.getState().getData());
   },
   onStop() {
--- a/src/js/views/patients/shared/components/check_component.js
+++ b/src/js/views/patients/shared/components/check_component.js
@@ -1,0 +1,33 @@
+import hbs from 'handlebars-inline-precompile';
+import { Component } from 'marionette.toolkit';
+
+import 'sass/modules/buttons.scss';
+
+export default Component.extend({
+  stateEvents: {
+    'change:isSelected': 'onStateChangeIsSelected',
+  },
+  viewEvents: {
+    'click': 'onClick',
+  },
+  viewOptions() {
+    const isSelected = this.getState('isSelected');
+    const template = isSelected ? hbs`{{fas "check-square"}}` : hbs`{{fal "square"}}`;
+    return {
+      tagName: 'button',
+      className: 'button--checkbox js-select',
+      template,
+      triggers: {
+        'click': 'click',
+      },
+    };
+  },
+  onClick(view, domEvent) {
+    this.toggleState('isSelected');
+    this.triggerMethod('select', domEvent);
+  },
+  onStateChangeIsSelected(state, isSelected) {
+    this.show();
+    this.triggerMethod('change:isSelected', isSelected);
+  },
+});
--- a/src/js/views/patients/shared/components/check_component.js
+++ b/src/js/views/patients/shared/components/check_component.js
@@ -13,6 +13,7 @@ export default Component.extend({
   viewOptions() {
     const isSelected = this.getState('isSelected');
     const template = isSelected ? hbs`{{fas "check-square"}}` : hbs`{{fal "square"}}`;
+
     return {
       tagName: 'button',
       className: 'button--checkbox js-select',
--- a/src/js/views/patients/worklist/action_views.js
+++ b/src/js/views/patients/worklist/action_views.js
@@ -53,8 +53,7 @@ const ActionItemView = View.extend({
     });
   },
   modelEvents: {
-    'change:due_date': 'onChangeDueDateTime',
-    'change:due_time': 'onChangeDueDateTime',
+    'change': 'render',
   },
   triggers: {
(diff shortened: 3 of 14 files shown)


src/js/views/patients/worklist/worklist_views.js
```diff
@@ -206,15 +206,13 @@ const ListView = CollectionView.extend({
     const lastSelectedIndex = this.state.get('lastSelectedIndex');
 
     if (isShiftKeyPressed && lastSelectedIndex !== null && !isSelected) {
-      this.handleClickShiftMultiSelect(selectedIndex);
```
nmajor25, line 209: Nice catch. Better to only get `lastSelectedIndex` the one time.

src/js/views/patients/shared/components/check_component.js
```diff
@@ -0,0 +1,33 @@
...
+  stateEvents: {
+    'change:isSelected': 'onStateChangeIsSelected',
+  },
+  viewEvents: {
+    'click': 'onClick',
+  },
+  viewOptions() {
+    const isSelected = this.getState('isSelected');
+    const template = isSelected ? hbs`{{fas "check-square"}}` : hbs`{{fal "square"}}`;
+    return {
```
nmajor25, line 17: _Nitpick alert_.

Add space between  the `const template` and the `return {` lines?

src/js/views/patients/worklist/action_views.js
```diff
@@ -48,8 +48,8 @@ const ActionItemView = View.extend({
     this.flow = this.model.getFlow();
 
     this.listenTo(state, {
-      'select:multiple': this.render,
-      'select:none': this.render,
+      'select:multiple': this.showCheck,
+      'select:none': this.showCheck,
```
nmajor25, line 52: After the bulk edit save is finished, the row items in the UI aren't updated.

To reproduce: Go to the worklist => select one or more list items => make a change (e.g. switch state from `To do` to `Doing` => hit `Save` button. When the bulk edit is finished, those list items aren't updated in the UI. And the list needs to be reloaded to see those changes (via a hard refresh of the app or leaving the worklist page and coming back).

Think it's related to this: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/patients/worklist/worklist_app.js#L166.

Before this PR, `state.clearSelected()` was called when the bulk edit is finished, which triggers `select:none` and a `this.render()` on the list item.

nmajor25: Looks like the date/time are updated in the UI correctly. But the state and owner are not...? Now I'm confused...

paulfalgout: Good catch on this one.  I'm a little surprised that no tests failed with this.  I'm assuming what I just pushed will need line coverage, but either way I plan on adding tests to cover what should have failed here.

src/js/apps/patients/sidebar/bulk-edit-actions_app.js
```diff
@@ -147,6 +147,7 @@ export default App.extend({
     });
   },
   onSubmit() {
+    this.modal.disableSubmit();
```
paulfalgout, line 150: with this, the green save button was staying active while awaiting submission.. particularly noticeable on a patient flow page..and if someone hammered those, because it's parallel calls they could really stack up.  There might be more to do here, but this seems better than it was

src/js/views/patients/worklist/action_views.js
nmajor25: From your last two commits, everything looks to be working well! Also was surprised the tests didn't catch it.

src/js/apps/patients/sidebar/bulk-edit-actions_app.js
nmajor25: Nice, that works well 👍 

One thought I had while testing this in my local sandbox. Should the `Cancel` option be changed to something else like `Clear`, `Clear Changes`, etc. (this sits next to the `Save` button in the sidebar)? Or just hide it once the `Save` button is clicked by the user?

Because once the `Save` button is clicked and the request is sent to the backend (where db updates are made), that request can't be cancelled. At that point, clicking `Cancel` just closes the sidebar, but doesn't really cancel anything.

paulfalgout: Yeah maybe on the cancel thing.  Worth bringing up, but it'll be true everywhere.  The naming of the button is a UX decision.

src/js/views/patients/worklist/action_views.js
paulfalgout: There's a bit of event loop because some of the custom saves don't check for change..  so it works kind of.. that's why there tests are currently failing. :-/

paulfalgout:
@nmajor25 alright, so this was pulling the thread of a sweater..  This PR will fix the the multi-select issues, but it keeps/moves the issue over to bulk edit submitting.  This can be done, but not only does the data need to update, but potentially other state like whether or not the component should be disabled.  But now I have a dev sandbox so I can probably work on tracking down any leak better in the near future.

nmajor25 (approved):

PR #760, opened 2022-08-04
Clear field pojos on form submission

paulfalgout:
```js
_.reduce({ foo: 1, bar: [], baz: {} }, (fields, field, key) => {
  const isEmptyPojo = !_.isArray(field) && _.isObject(field) && _.isEmpty(field);

  if (!isEmptyPojo) fields[key] = field;

  return fields;
}, {});
```
returns `{ foo: 1 bar: Array(0) }`

Shortcut Story ID: [sc-29993]

### DIFF
--- a/test/integration/forms/form.js
+++ b/test/integration/forms/form.js
@@ -483,7 +483,31 @@ context('Patient Action Form', function() {
         return fx;
       })
       .routeForm(_.identity, '11111')
-      .routeFormDefinition()
+      .routeFormDefinition(fx => {
+        fx.components.push(
+          {
+            label: 'Survey',
+            tableView: false,
+            questions: [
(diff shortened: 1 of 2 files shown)


test/integration/forms/form.js
```diff
@@ -483,7 +483,31 @@ context('Patient Action Form', function() {
...
+            ],
+            values: [
+              {
+                label: '',
+                value: '',
+                tooltip: '',
+              },
+            ],
+            key: 'fields.survey',
+            type: 'survey',
```
paulfalgout, line 506: An empty survey component's value is `{}` but it will error if the field value is `[]`

nmajor25 (approved):

PR #761, opened 2022-08-08
Update dependencies

paulfalgout:
Shortcut Story ID: [sc-27813] [sc-28761] [sc-28760]

Most notable change is having to rename the `src/sass` directory to `src/scss` which side steps whatever export issue the sass library added in v1.43.

Then the newer stylelint defaults included a few more useful lints and I add the "fix" option so that it'll fix issues on save if it can.

The other things are mostly straight forward.


nmajor25 (approved):
Looks good to me.

I also navigated through the app in my local sandbox and everything appeared the same as before.

PR #763, opened 2022-08-08
View patient sidebar from the schedule page

nmajor25:
Shortcut Story ID: [sc-29992]

Add the ability to view the patient sidebar from the schedule page.

Similar to what we added for the worklist in this PR: https://github.com/RoundingWell/care-ops-frontend/pull/699.

This also involves moving the `State` to the `Action` column.

https://user-images.githubusercontent.com/35355575/183471560-a110b4e9-dd1c-4d30-8c22-c74b43560c74.mov

### DIFF
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -239,9 +239,9 @@ careOptsFrontend:
           title: Schedule for { owner }
           tooltip: Schedule Worklist displays all Actions with a due date for you (or someone else on your team). States included are To Do and In Progress.
         tableHeaderView:
-          actionHeader: Action
+          actionHeader: State, Action
           dueDateHeader: Due Date
-          statePatientHeader: State, Patient
+          statePatientHeader: Patient
           formheader: Form
     shared:
       addWorkflow:
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -239,9 +239,9 @@ careOptsFrontend:
           title: Schedule for { owner }
           tooltip: Schedule Worklist displays all Actions with a due date for you (or someone else on your team). States included are To Do and In Progress.
         tableHeaderView:
-          actionHeader: State, Action
           dueDateHeader: Due Date
-          statePatientHeader: Patient
+          patientHeader: Patient
+          stateActionHeader: State, Action
           formheader: Form
     shared:
       addWorkflow:
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -239,10 +239,10 @@ careOptsFrontend:
           title: Schedule for { owner }
           tooltip: Schedule Worklist displays all Actions with a due date for you (or someone else on your team). States included are To Do and In Progress.
         tableHeaderView:
+          actionHeader: State, Action
           dueDateHeader: Due Date
-          patientHeader: Patient
-          stateActionHeader: State, Action
           formheader: Form
+          patientHeader: Patient
     shared:
       addWorkflow:
         addWorkflowViews:
--- a/src/js/views/patients/schedule/schedule-list.scss
+++ b/src/js/views/patients/schedule/schedule-list.scss
@@ -83,6 +83,12 @@
     cursor: default;
   }
 
+  &:hover {
+    .js-patient-sidebar-button {
+      color: $black-20;
+    }
+  }
+
   &:hover:not(.is-selected, .is-reduced) {
     background-color: color(light_blue, light);
   }
@@ -90,6 +96,7 @@
 
 .schedule-list__action-list-cell {
   padding: 4px 8px;
+  width: calc(68% - 84px);
 
   &.is-overdue {
     color: color(red);
@@ -116,13 +123,21 @@
 }
 
 .schedule-list__state-patient,
-.schedule-list__action-name {
+.schedule-list__action-list-cell {
   display: inline-block;
   max-width: 100%;
   overflow: hidden;
   text-overflow: ellipsis;
   vertical-align: middle;
   white-space: nowrap;
+}
+
+.schedule-list__patient-sidebar-icon {
+  margin-right: 16px;
+
+  .js-patient-sidebar-button {
+    color: $black-60;
+  }
 }
 
 .schedule-list__action-state {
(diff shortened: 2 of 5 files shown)


src/js/i18n/en-US.yml
```diff
@@ -239,9 +239,9 @@ careOptsFrontend:
           title: Schedule for { owner }
           tooltip: Schedule Worklist displays all Actions with a due date for you (or someone else on your team). States included are To Do and In Progress.
         tableHeaderView:
-          actionHeader: Action
           dueDateHeader: Due Date
-          statePatientHeader: State, Patient
+          patientHeader: Patient
+          stateActionHeader: State, Action
           formheader: Form
```
nmajor25, line 244: Did a few things here besides changing the header label text:

* Renamed `statePatientHeader` to `patientHeader`.
* Renamed `actionHeader` to `stateActionHeader`.
* Re-ordered the variables to match the order they're used in the view file.

But now that I look at it again, does `stateActionHeader` make sense? It could just be `actionHeader`...?

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -277,6 +284,9 @@ const ScheduleListView = CollectionView.extend({
   childViewEvents: {
     'render:children': 'onChildFilter',
   },
+  childViewTriggers: {
+    'click:patientSidebarButton': 'click:patientSidebarButton',
```
nmajor25: Is there a better way to trigger `'click:patientSidebarButton'` than passing it up multiple times like I am here?

I'm handling it three times: `DayItemView` => `DayListView` => `ScheduleListView` (where the `schedule_app` is listening for it).

test/integration/patients/worklist/schedule.js
```diff
@@ -205,6 +205,44 @@ context('schedule page', function() {
...
+      .get('[data-filters-region]')
+      .find('[data-owner-filter-region]')
+      .click();
+
+    cy
+      .get('.picklist')
+      .find('.picklist__group')
+      .contains('Clinician McTester')
+      .click();
+
```
nmajor25, line 245: This test does the following:

* Opens the patient sidebar app from the schedule page and verifies that it contains the correct patient name.
* Verifies that the patient sidebar closes when a filter is applied to the list.

We test the contents of the patient sidebar extensively in the worklist tests, so I figured we don't need to do that again here for the schedule.

src/js/views/patients/schedule/schedule-list.scss
```diff
@@ -115,8 +122,8 @@
   width: 48px;
 }
 
-.schedule-list__state-patient,
-.schedule-list__action-name {
+.schedule-list__patient-details,
+.schedule-list__action-list-cell {
```
nmajor25, line 126: This makes overflowing text for the `State, Action` table cell be hidden with an ellipsis (`...`).

Screen Shot 2022-08-08 at 11 03 56 AM

Before this was only applied to the action name (`.schedule-list__action-name`). But since the state icon is now in that table column, it needs to be applied to the entire table column instead (`.schedule-list__action-list-cell`).

src/js/views/patients/schedule/schedule-list.scss
```diff
@@ -83,13 +83,20 @@
...
+  }
+
   &:hover:not(.is-selected, .is-reduced) {
     background-color: color(light_blue, light);
   }
 }
 
 .schedule-list__action-list-cell {
   padding: 4px 8px;
+  width: calc(68% - 84px);
```
nmajor25, line 99: For the text overflow ellipsis (`...`) styling to work, the `State, Action` column needed a width set to it.

`width: calc(68% - 84px);` matches what's set for the table header column: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/views/patients/schedule/schedule-list.scss#L16.

src/js/i18n/en-US.yml
paulfalgout: ah well it should be alphabetical.   I wonder if we could fix that in the loader somehow..  It doesn't _really_ matter, but if we ever use phraseapp again for translation, it'll alphabetize the keys.  `actionHeader` is fine with me.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -102,12 +102,17 @@ const DayItemView = View.extend({
...
+          <button class="js-patient-sidebar-button">
+            {{far "address-card"}}
+          </button>
+        </span>&#8203;{{~ remove_whitespace ~}}
         <span class="schedule-list__patient-name {{#if isReduced}}is-reduced{{else}}js-patient{{/if}}">{{ patient.first_name }} {{ patient.last_name }}</span>&#8203;
       </div>
     </td>
     <td class="schedule-list__action-list-cell">
+      <span class="schedule-list__action-state action--{{ stateOptions.color }}">{{fa stateOptions.iconType stateOptions.icon}}</span><span class="schedule-list__search-helper">{{ state }}</span>&#8203;{{~ remove_whitespace ~}}
       <span class="schedule-list__action-name {{#unless isReduced}}js-action{{/unless}}">{{ name }}</span>&#8203;{{~ remove_whitespace ~}}
```
paulfalgout: can `schedule-list__action-name` be removed now?

src/js/views/patients/schedule/schedule_views.js
paulfalgout: there's a few ways we could do it..  this is one, but it is a lot of proxying.. could setup a radio channel that gets passed down as an intermediate.  However check out `onClickPatient` we're doing a radio request right there..  probably for this reason..  if there was more data things I'd probably have had that moved to the app, but you can add a `onClick...` here to match the others.  Probably worth it.

paulfalgout (changes requested):

src/js/i18n/en-US.yml
nmajor25: Ah gotcha. I switched it to `actionHeader` and alphabetized them in the commits I just pushed.

src/js/views/patients/schedule/schedule_views.js
nmajor25: Doing the `Radio.request()` in `DayItemView` works well. Updated in the latest commits I just pushed.

src/js/views/patients/schedule/schedule_views.js
nmajor25: Indeed it can, nice catch 😄

paulfalgout (approved):

PR #764, opened 2022-08-09
Schedule - Add a grey background to selected list items

nmajor25:
Shortcut Story ID: [sc-30137]

Adds a `.is-selected` classname to a schedule list item when it's selected, which gives it a grey background.

The `.is-selected` class should be toggled when a user selects/un-selects a single list item. And when a user uses the bulk select controls (select all and un-select all).

Screen Shot 2022-08-08 at 2 27 53 PM

### DIFF
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -160,7 +160,11 @@ const DayItemView = View.extend({
     });
   },
   onRender() {
+    this.toggleSelected(!this.isReduced && this.state.isSelected(this.model));
     this.showDetailsTooltip();
+  },
+  toggleSelected(isSelected) {
+    this.$el.toggleClass('is-selected', isSelected);
   },
   onClickPatientSidebarButton() {
     const patient = this.model.getPatient();
@@ -168,6 +172,7 @@ const DayItemView = View.extend({
   },
   onClickSelect() {
     this.state.toggleSelected(this.model, !this.state.isSelected(this.model));
+    this.toggleSelected(this.state.isSelected(this.model));
     this.render();
   },
   onClickPatient() {
--- a/test/integration/patients/worklist/schedule.js
+++ b/test/integration/patients/worklist/schedule.js
@@ -904,6 +904,19 @@ context('schedule page', function() {
       .should('contain', 'Edit 1 Action');
 
     cy
+      .get('.schedule-list__table')
+      .find('.schedule-list__list-row .is-selected')
+      .should('have.length', 1)
+      .first()
+      .find('.js-select')
+      .click();
+
+    cy
+      .get('.schedule-list__table')
+      .find('.schedule-list__list-row .is-selected')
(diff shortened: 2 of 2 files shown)


src/js/views/patients/schedule/schedule_views.js
```diff
@@ -160,14 +160,19 @@ const DayItemView = View.extend({
     });
   },
   onRender() {
+    this.toggleSelected(!this.isReduced && this.state.isSelected(this.model));
```
nmajor25, line 163: This ensures `.is-selected` is properly added on the initially route load or when the user uses the bulk select controls to select all or un-select all.

The `!this.isReduced` check is added here because the state model is not loaded for the `reduced_schedule_app.js`, which means `this.state.isSelected()` doesn't exist and throws an error when the reduced app is loaded.

test/integration/patients/worklist/schedule.js
```diff
@@ -903,6 +903,19 @@ context('schedule page', function() {
       .find('.js-bulk-edit')
       .should('contain', 'Edit 1 Action');
 
+    cy
+      .get('.schedule-list__table')
+      .find('.schedule-list__list-row .is-selected')
+      .should('have.length', 1)
+      .first()
+      .find('.js-select')
+      .click();
```
nmajor25, line 912: Verifies that the `.is-selected` class is there when a list item initially loads as selected.

test/integration/patients/worklist/schedule.js
```diff
@@ -903,6 +903,19 @@ context('schedule page', function() {
...
+
+    cy
+      .get('.schedule-list__table')
+      .find('.schedule-list__list-row .is-selected')
+      .should('have.length', 0);
+
```
nmajor25, line 918: Verifies that the `.is-selected` class toggles correctly when a select list item is un-selected.

test/integration/patients/worklist/schedule.js
```diff
@@ -935,6 +948,11 @@ context('schedule page', function() {
       .find('.fa-square-check')
       .should('have.length', 0);
 
+    cy
+      .get('.schedule-list__table')
+      .find('.schedule-list__list-row .is-selected')
+      .should('have.length', 0);
+
```
nmajor25, line 955: Verifies that the `.is-selected` class is removed for list items when the bulk selections are cleared/cancelled by the user.

test/integration/patients/worklist/schedule.js
```diff
@@ -949,6 +967,11 @@ context('schedule page', function() {
       .find('.fa-square-check')
       .should('have.length', 20);
 
+    cy
+      .get('.schedule-list__table')
+      .find('.schedule-list__list-row .is-selected')
+      .should('have.length', 20);
+
```
nmajor25, line 974: Verifies that the `.is-selected` class is added to each list item when the user selects-all via the bulk select controls.

paulfalgout (approved):

PR #765, opened 2022-08-10
Fix styling bug in schedule list

nmajor25:
Shortcut Story ID: [sc-29992]

This PR fixes a CSS styling bug in the schedule list where the form link table column is pushed to a new line (screenshot below).

It was introduced in this pull request: https://github.com/RoundingWell/care-ops-frontend/pull/763.

Screen Shot 2022-08-09 at 3 58 17 PM (1)
l

### DIFF
--- a/src/js/views/patients/schedule/schedule-list.scss
+++ b/src/js/views/patients/schedule/schedule-list.scss
@@ -96,7 +96,6 @@
 
 .schedule-list__action-list-cell {
   padding: 4px 8px;
-  width: calc(68% - 84px);
 
   &.is-overdue {
     color: color(red);
@@ -124,7 +123,7 @@
 }
 
 .schedule-list__patient-details,
-.schedule-list__action-list-cell {
+.schedule-list__action-state-name {
   display: inline-block;
   max-width: 100%;
   overflow: hidden;
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -112,9 +112,11 @@ const DayItemView = View.extend({
       </div>
     </td>
     <td class="schedule-list__action-list-cell">
-      <span class="schedule-list__action-state action--{{ stateOptions.color }}">{{fa stateOptions.iconType stateOptions.icon}}</span><span class="schedule-list__search-helper">{{ state }}</span>&#8203;{{~ remove_whitespace ~}}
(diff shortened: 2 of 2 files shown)


src/js/views/patients/schedule/schedule-list.scss
```diff
@@ -124,7 +123,7 @@
 }
 
 .schedule-list__patient-details,
-.schedule-list__action-list-cell {
```
nmajor25, line 127: Thought I was just applying this to the `State, Action` column. But I was mistakenly applying this styling to each table cell.

So I wrapped that `State, Action` cell in a `<div class="schedule-list__action-state-name`></div>` and applied these styles to only that instead.

Still not 100% sure why that only caused issues in Firefox though...

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -112,9 +112,11 @@ const DayItemView = View.extend({
       </div>
     </td>
     <td class="schedule-list__action-list-cell">
-      <span class="schedule-list__action-state action--{{ stateOptions.color }}">{{fa stateOptions.iconType stateOptions.icon}}</span><span class="schedule-list__search-helper">{{ state }}</span>&#8203;{{~ remove_whitespace ~}}
-      <span class="{{#unless isReduced}}js-action{{/unless}}">{{ name }}</span>&#8203;{{~ remove_whitespace ~}}
-      <span class="schedule-list__search-helper">{{ flow }}</span>&#8203;{{~ remove_whitespace ~}}
+      <div class="schedule-list__action-state-name">
```
nmajor25, line 115: Just created a new `<div class="schedule-list__action-state-name">` here to apply those styles to.

src/js/views/patients/schedule/schedule-list.scss
```diff
@@ -96,7 +96,6 @@
 
 .schedule-list__action-list-cell {
   padding: 4px 8px;
-  width: calc(68% - 84px);
```
nmajor25, line 99: After making the other changes, turns out I didn't need to set a width.

And, also, I was mistakenly applying this width to each cell in the table. Not just the `State, Action` cell I was intending to apply it to 🤦

So I removed it.

nmajor25:
@paulfalgout In my local sandbox, I've tested this in Chrome, Firefox, and Safari. And I've tested both the normal and reduced versions of the schedule.

I no longer see the bug Sofia discovered.

paulfalgout:
circleci is borked, and there's no js changes or failed tests.. so coveralls is reporting wrong.

PR #767, opened 2022-08-11
Fix form reducer error flake

paulfalgout:
Possible v10 speed improvements has surfaced this flake I think

### DIFF
--- a/test/integration/forms/form.js
+++ b/test/integration/forms/form.js
@@ -1509,7 +1509,11 @@ context('Patient Form', function() {
         .routeForm(_.identity, '11111')
         .routeFormDefinition()
         .routeFormFields()
-        .visit('/patient/1/form/44444');
+        .visit('/patient/1/form/44444')
+        .wait('@routePatient')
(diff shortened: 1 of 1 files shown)


test/integration/forms/form.js
```diff
@@ -1509,7 +1509,10 @@ context('Patient Form', function() {
         .routeForm(_.identity, '11111')
         .routeFormDefinition()
         .routeFormFields()
-        .visit('/patient/1/form/44444');
+        .visit('/patient/1/form/44444')
+        .wait('@routeForm')
+        .wait('@routeFormFields')
+        .wait('@routeFormDefinition');
```
nmajor25, line 1516: Out of curiosity, should we wait for the patient route data as well? Via `.wait('@routePatient')`?

Or is that not needed in this case?

paulfalgout: oh yeah might as well. it wasn't in the flake screenshot so I missed it.

nmajor25 (approved):

PR #769, opened 2022-08-11
Change fontawesome v6 implementation to symbols

paulfalgout:
https://fontawesome.com/docs/web/add-icons/svg-symbols

Shortcut Story ID: [sc-30187]


package.json
```diff
@@ -28,82 +28,82 @@
...
-      "faGripVertical",
-      "faInfoCircle",
-      "faKeyboard",
-      "faMinusSquare",
-      "faSignInAlt",
-      "faSignOutAlt",
-      "faTimes",
-      "faTimesCircle",
-      "faToggleOn"
+      "arrow-right-arrow-left",
```
paulfalgout, line 34: @nmajor25 now when we add font names they'll match the class we use in the template and what is copy/pasteable from fontawesome.com

config/fontawesome.js
```diff
@@ -0,0 +1,25 @@
...
+const farIcons = require('@fortawesome/pro-regular-svg-icons');
+const falIcons = require('@fortawesome/pro-light-svg-icons');
+
+config.replacementClass = '';
+
+function getIconHtml(lib, fonts) {
+  return _.map(fonts, iconClass => {
+    const iconName = camelCase(`fa-${ iconClass }`);
+    if (!lib[iconName]) console.error('Missing icon:', iconName);
+    return icon(lib[iconName], { symbol: true }).html;
```
paulfalgout, line 15: @nmajor25 here's the biggest part of the change.  This renders the icon as a symbol during the webpack build rather than for each instance of a template.  The fontawesome libraries are actually not loaded in our app any more other than styles

src/js/base/fontawesome.js
```diff
@@ -1,22 +1,14 @@
...
-config.autoAddCss = false;
-config.replacementClass = 'svg-inline--fa icon';
-
 function faHelper(prefix, iconName, { hash = {} }) {
-  const faDefinition = findIconDefinition({ prefix, iconName });
-  const faIcon = icon(faDefinition, hash);
+  const svgId = `${ prefix }-fa-${ iconName }`;
+  const svgClass = `fa-${ iconName }`;
+  const svgClasses = hash.classes || '';
+  const svg = `<svg class="icon svg-inline--fa ${ svgClass } ${ svgClasses }"><use xlink:href="#${ svgId }"></use></svg>`;
```
paulfalgout, line 9: @nmajor25 Here's how the icon symbols are referenced in a template.  Their ids are based on the naming and those are generate by the FA library

src/js/views/globals/root.hbs
```diff
@@ -13,5 +13,8 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=1200, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, user-scalable=no">
   </head>
-  <body><div id="root" class="app-root"></div></body>
+  <body>
+    <div style="display:none">{{{ htmlWebpackPlugin.options.faIconSymbols }}}</div>
```
paulfalgout, line 17: @nmajor25 The icons are hidden in a div above the app so that there's no rendering after the fact.
This also means that the app is now rendered in `#root` instead of `body`

test/fixtures/test/states.json
```diff
@@ -39,8 +39,19 @@
...
     }
+  },
+  {
+    "id": "77777",
+    "name": "THMG Transfered",
+    "sequence": 4,
+    "status": "done",
+    "options": {
+      "iconType": "fas",
+      "icon": "arrow-right-arrow-left",
```
paulfalgout, line 53: @nmajor25 I added this in the tests as this icon is used on THMG and maybe elsewhere but not in our code.  I almost removed it from the build because of this.  This should prevent accidentally pulling the icon

webpack.config.js
```diff
@@ -58,6 +57,7 @@ module.exports = {
     static: {
       directory: path.join(__dirname, 'dist'),
     },
+    watchFiles: ['package.json'],
```
paulfalgout, line 60: @nmajor25 this will auto-refresh the build when a font is added.

paulfalgout:
I split up the commits to make the reviewing a bit easier

src/js/views/forms/form/form_views.js
```diff
@@ -44,9 +44,9 @@ const FormActionsView = View.extend({
   className: 'form__actions',
   template: hbs`
     {{#if hasHistory}}<button class="js-history-button form__actions-icon{{#if shouldShowHistory}} is-selected{{/if}}">{{far "history"}}</button>{{/if}}
```
nmajor25: Looks like the `"history"` icon name was also changed to `"clock-rotate-left"` in v6 (via the [Icons Renamed In Version 6](https://fontawesome.com/docs/web/setup/upgrade/whats-changed#icons-renamed-in-version-6) list.).

There's one more instance of this on line 292 of this file.

Didn't find any other places it's used elsewhere in the app.

paulfalgout: nice catch

nmajor25 (approved):

paulfalgout:
Depends on https://github.com/RoundingWell/care-ops-backend/pull/3922#pullrequestreview-1071104914

PR #772, opened 2022-08-12
Use checkbox component for each schedule list item

nmajor25:
Shortcut Story ID: [sc-30254]

Use our checkbox component for each schedule list item.

The checkbox component was built in this PR: https://github.com/RoundingWell/care-ops-frontend/pull/752. And was originally implemented in the worklist only.

### DIFF
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -164,14 +164,16 @@ const DayItemView = View.extend({
     });
   },
   onRender() {
-    if (!this.isReduced) this.showCheck();
+    this.showCheck();
     this.showDetailsTooltip();
   },
   toggleSelected(isSelected) {
     this.$el.toggleClass('is-selected', isSelected);
   },
   showCheck() {
-    const isSelected = !this.isReduced && this.state.isSelected(this.model);
+    if (this.isReduced) return;
+
+    const isSelected = this.state.isSelected(this.model);
     this.toggleSelected(isSelected);
     const checkComponent = new CheckComponent({ state: { isSelected } });
 
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -15,7 +15,7 @@ import PreloadRegion from 'js/regions/preload_region';
 
 import Tooltip from 'js/components/tooltip';
 
-import { DetailsTooltip } from 'js/views/patients/shared/actions_views';
+import { CheckComponent, DetailsTooltip } from 'js/views/patients/shared/actions_views';
 
 import LayoutTemplate from './layout.hbs';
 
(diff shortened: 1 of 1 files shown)


src/js/views/patients/schedule/schedule_views.js
```diff
@@ -149,34 +149,45 @@ const DayItemView = View.extend({
...
     this.listenTo(state, {
-      'select:all': this.render,
-      'select:none': this.render,
+      'select:all': this.showCheck,
+      'select:none': this.showCheck,
     });
   },
   onRender() {
-    this.toggleSelected(!this.isReduced && this.state.isSelected(this.model));
+    if (!this.isReduced) this.showCheck();
```
nmajor25: Without `if (!this.isReduced)`, the reduced version of the schedule loads an empty list (without showing any errors). So something breaks for the reduced schedule if `showCheck()` is called.

paulfalgout: I think it might be better to throw a guard in the `showCheck`
Then you also wouldn't need the check to set `isSelected`

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -149,34 +149,45 @@ const DayItemView = View.extend({
     'click .js-patient-sidebar-button': 'click:patientSidebarButton',
     'click .js-patient': 'click:patient',
     'click': 'click',
-    'click .js-select': 'click:select',
+  },
+  modelEvents: {
+    'change': 'render',
```
paulfalgout, line 154: So we needed this on the other list to handle bulk editing.  But this app doesn't seem to depend on that.. might need to implement the other lists however schedule is working, but I don't think this event is needed

paulfalgout (changes requested):

src/js/views/patients/schedule/schedule_views.js
nmajor25: Yes, good call 👍

src/js/views/patients/schedule/schedule_views.js
nmajor25: Without this event, the list items don't update in the UI after saving a bulk edit change (update status, date, etc.).

On successful bulk select, `select:none` is fired off. Which the schedule view listens for and calls`this.render()`. But now that doesn't happen since we call `this.showCheck()` instead of `this.render()`.

I believe the schedule and worklist/patient flow lists work the same in this regard.

nmajor25:
@paulfalgout We just got a coverage drop from that last commit I pushed. In the [src/js/views/patients/patient/flow/flow_views.js](https://coveralls.io/builds/51670908/source?filename=src%2Fjs%2Fviews%2Fpatients%2Fpatient%2Fflow%2Fflow_views.js#L206) file.

It seems not connected to any changes in this PR....? Or not that I understand anyway.

Is this from a previous pull request that is just showing now for some reason?

Possibly one of these commits?

* https://github.com/RoundingWell/care-ops-frontend/commit/01548b6ead63f3b6a17e4280ea6ff68f8d3ab11d
* https://github.com/RoundingWell/care-ops-frontend/commit/5fdb37f65a37408f8d457605244addb30c39d80c

paulfalgout:
It's odd because it's covered in the PR for those commits https://coveralls.io/builds/51360699/source?filename=src%2Fjs%2Fviews%2Fpatients%2Fpatient%2Fflow%2Fflow_views.js#L206

makes me think we have some sort of race condition somewhere.  going to rerun these

paulfalgout:
Well rerunning fixed it.. also it looks to me like we explicitly test this.. so I don't know what the deal is with that.

paulfalgout (approved):

PR #773, opened 2022-08-15
Cypress intercept

paulfalgout:
Shortcut Story ID: [sc-30273]

This is the first step towards migrating to `cy.intercept`

It adds a temporary `cy.route` overwrite that uses `cy.intercept` under the hood.  We should work to remove this

This does allow us to remove the `cy.server()` everywhere and it provides a path to upgrade.

The biggest issue to fully refactoring is removing `cy.fixture` putting fixture data in context and rather `import`ing the fixture files directly and utilizing them with an `intercept function setup.

This also adds an `itsUrl` command which decodes the urls to normalize tests which is how `route` was working.  Without this, we'd have to encode for testing which can get hairy quick.

### DIFF
--- a/test/integration/clinicians/clinician-sidebar.js
+++ b/test/integration/clinicians/clinician-sidebar.js
@@ -34,7 +34,6 @@ context('clinician sidebar', function() {
     };
 
     cy
-      .server()
       .routeGroupsBootstrap(_.identity, [
         {
           id: '1',
@@ -327,7 +326,6 @@ context('clinician sidebar', function() {
     };
 
     cy
-      .server()
       .routeGroupsBootstrap(_.identity, [
         {
           id: '1',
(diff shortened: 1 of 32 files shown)


nmajor25 (commented):
Pretty slick, looks good to me 😄 

There are two `.server()` calls that remain:

* [Line 1487: /test/integration/forms/form.js](https://github.com/RoundingWell/care-ops-frontend/blob/359d88a34072d67b34377b92bd97b0f21647f085/test/integration/forms/form.js#L1487)
* [Line 23: /test/support/defaults.js](https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/support/defaults.js#L23)

Can those be removed too?

nmajor25 (approved):

paulfalgout:
Yeah those can be removed.. I know why I missed the support one.. but I have no idea how my search/replace missed the form.js one 🤯

PR #774, opened 2022-08-15
Clear out localstorage on error for stored submissions

paulfalgout:
Shortcut Story ID: [sc-30286]

Should clear up the error logs we’ve been seeing and reinstate stored submissions for the user with the full cache.

### DIFF
--- a/test/integration/forms/form.js
+++ b/test/integration/forms/form.js
@@ -166,9 +166,6 @@ context('Patient Action Form', function() {
   });
 
   specify('storing stored submission', function() {
-    localStorage.clear();
-    // Just short of 5MB to cause quota error
-    localStorage.setItem('form-subm-too-big', _.times(5242800, _.constant('0')).join(''));
     cy
       .routeAction(fx => {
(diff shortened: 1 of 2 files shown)


test/integration/forms/form.js
```diff
@@ -166,6 +166,9 @@ context('Patient Action Form', function() {
   });
 
   specify('storing stored submission', function() {
+    localStorage.clear();
+    // Just short of 5MB to cause quota error
+    localStorage.setItem('form-subm-too-big', _.times(5242800, _.constant('0')).join(''));
```
paulfalgout: If I set `_.times(100` here
Screen Shot 2022-08-16 at 5 22 23 AM

paulfalgout: whelp.. this test may not run on circleci.. it gets here and the test doesn't load..

nmajor25: Ah dang ☹️

nmajor25 (approved):

PR #775, opened 2022-08-16
Remove PreloadRegion from formapp for simpler html spinner

nmajor25:
Shortcut Story ID: [sc-29468]

In the formapp, we currently import and use a `PreloadRegion` view. It does some fading in and waiting for other use cases that aren't needed in the formapp iframe. And it also adds more dependencies to the formapp than are necessary.

This PR removes the `PreloadRegion` import from the formapp and uses a pure HTML version of the spinner instead. The same CSS is used to style the spinner (`js/regions/preload.scss`).

The formapp loading spinner should behave and look the same as before.

### DIFF
--- a/src/js/formapp.js
+++ b/src/js/formapp.js
@@ -24,13 +24,13 @@ import {
 
 import 'js/formapp/components';
 
-import PreloadRegion from 'js/regions/preload_region';
-
 import 'scss/formapp/comment.scss';
 import 'scss/formapp/form.scss';
 import 'scss/formapp/formio-overrides.scss';
 import 'scss/formapp/print.scss';
 import 'scss/formapp/pdf.scss';
+
+import 'js/regions/preload.scss';
 
 let router;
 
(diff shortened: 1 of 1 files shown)


src/js/formapp.js
```diff
@@ -24,14 +24,14 @@ import {
...
-import PreloadRegion from 'js/regions/preload_region';
-
 import 'scss/formapp/comment.scss';
 import 'scss/formapp/form.scss';
 import 'scss/formapp/formio-overrides.scss';
 import 'scss/formapp/print.scss';
 import 'scss/formapp/pdf.scss';
 
+import 'js/regions/preload.scss';
+
```
nmajor25, line 34: Correct spot in the file for this import? Wasn't 100% sure where to place it.

paulfalgout: fine for me

paulfalgout (approved):

PR #778, opened 2022-08-17
Add "up-right-and-down-left-from-center" fontawesome icon to the package.json file

nmajor25:
Shortcut Story ID: [sc-30314] [sc-30308]

When updating the fontawesome icons in a previous PR (https://github.com/RoundingWell/care-ops-frontend/pull/769), we missed adding the `up-right-and-down-left-from-center` icon to the `"far": []` section of the `package.json` file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
       "toggle-off",
       "traffic-cone",
       "trash-can",
+      "up-right-and-down-left-from-center",
       "user",
       "user-plus",
       "users",


package.json
```diff
@@ -100,6 +100,7 @@
       "toggle-off",
       "traffic-cone",
       "trash-can",
+      "up-right-and-down-left-from-center",
```
shadowhand, line 103: Such a weird icon name.

shadowhand (commented):
🎉

package.json
shadowhand: The alias for it, `expand-alt`, makes so much more sense.

paulfalgout: font awesome was naming icons based on an assumed function like expand/contract.. then they had a shift in thinking and now are just describing the icon absent function.

PR #781, opened 2022-08-19
Add grey background to selected actions on patient flow page

nmajor25:
Shortcut Story ID: [sc-30382]

When a list item is selected on the patient flow page, we want to add a grey background to it via the existing `.is-selected` class name. This will follow what we currently do on the worklist and schedule page lists.

Functionality before this PR:

Screen Shot 2022-08-18 at 9 52 44 AM (1)

Functionality after this PR:

Screen Shot 2022-08-18 at 9 54 24 AM (2)

**Additional note**: we need to ensure that the grey background is preserved for actions that are actively being edited (i.e. edit action sidebar is open for the action). So the `.is-selected` class should remain if the action is selected or being edited.

### DIFF
--- a/src/js/views/patients/patient/flow/flow_views.js
+++ b/src/js/views/patients/patient/flow/flow_views.js
@@ -161,6 +161,10 @@ const ActionItemView = View.extend({
     Radio.trigger('event-router', 'flow:action', this.model.get('_flow'), this.model.id);
   },
   onEditing(isEditing) {
+    const isSelected = this.state.isSelected(this.model);
+
+    if (isSelected) return;
+
     this.$el.toggleClass('is-selected', isEditing);
   },
   onRender() {
@@ -171,14 +175,23 @@ const ActionItemView = View.extend({
     this.showDueTime();
     this.showForm();
   },
+  toggleSelected(isSelected) {
+    const isBeingEdited = this.state.isBeingEdited(this.model);
+
+    if (isBeingEdited) return;
+
+    this.$el.toggleClass('is-selected', isSelected);
+  },
   showCheck() {
     const isSelected = this.state.isSelected(this.model);
+    this.toggleSelected(isSelected);
     const checkComponent = new CheckComponent({ state: { isSelected } });
 
     this.listenTo(checkComponent, {
       'select'(domEvent) {
         this.state.toggleSelected(this.model, !this.state.isSelected(this.model));
       },
+      'change:isSelected': this.toggleSelected,
     });
 
     this.showChildView('check', checkComponent);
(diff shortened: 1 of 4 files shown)


src/js/views/patients/patient/flow/flow_views.js
```diff
@@ -161,6 +161,10 @@ const ActionItemView = View.extend({
     Radio.trigger('event-router', 'flow:action', this.model.get('_flow'), this.model.id);
   },
   onEditing(isEditing) {
+    const isSelected = this.state.isSelected(this.model);
+
+    if (isSelected) return;
+
```
nmajor25, line 167: `onEditing()` adds or removes the `.is-selected` class when the edit action sidebar opens or closes.

I added an `if (isSelected)` check here that `return`s if the action is currently selected via its checkbox.

src/js/views/patients/patient/flow/flow_views.js
```diff
@@ -171,14 +175,23 @@ const ActionItemView = View.extend({
     this.showDueTime();
     this.showForm();
   },
+  toggleSelected(isSelected) {
+    const isBeingEdited = this.state.isBeingEdited(this.model);
+
+    if (isBeingEdited) return;
+
+    this.$el.toggleClass('is-selected', isSelected);
+  },
```
nmajor25, line 184: `toggleSelected()` adds or removes the `.is-selected` class whenever the checkbox for that action is clicked (or the bulk select select-all or un-select controls are used).

If the action is currently being edited, this function will `return` without doing anything.

For this function to know whether or not the action is being edited, I needed to add a `state.actionBeingEdited` value that holds the `id` of the action being edited. And the `state.isBeingEdited()` function checks to see if that `id` matches the action `id` that was selected/un-selected via the checkbox.

test/integration/patients/patient/patient-flow.js
```diff
@@ -1014,6 +1014,60 @@ context('patient flow page', function() {
...
+
+    cy
+      .get('[data-sidebar-region]')
+      .find('.js-close')
+      .click();
+
+    cy
+      .get('@firstRow')
+      .should('have.class', 'is-selected');
+
```
nmajor25, line 1070: This test verifies that the `.is-selected` class is added/removed when an action is selected/un-selected via its checkbox. 

And that the `.is-selected` class remains on the action in the list if its either selected or being edited via the action sidebar.

**Sidenote**: I wasn't 100% sure where in this file to add these tests. When we do the `click`+`shift` multi-select for the patient flow page, I figured we could create a `specify('click+shift multiselect')` and move these changes into there.

paulfalgout (approved):

PR #782, opened 2022-08-22
Add quick search org settings for new UI

paulfalgout:
Shortcut Story ID: [sc-30412] [sc-30413]

This PR does two things.  It removes the frontend filtering from the quick search component and it adds a fork in the ui based on the existence of the `patient_search_settings` org setting.  In the future this setting will include configuration for this search component and when we're done we'll pull out the old one


nmajor25 (approved):
Code looks good to me. And I tested in my local sandbox and everything appears to work the same as before.

PR #784, opened 2022-08-23
Display identifiers in patient quick search results

nmajor25:
Shortcut Story ID: [sc-30415]

Display one or more patient identifier values (e.g. `MRN`, `SSN`, etc.) in the patient quick search results.

These will be based on a `patient_search_settings` tenant organizational setting that contains an array of identifiers to show (e.g. `["MRN", "SSN"]`). 

The identifiers should be displayed in the same order they are listed in the organization setting array. And they won't be shown if the organizational setting doesn't exist or is empty, or if a patient's data doesn't contain the given identifier.

For patient identifiers that don't exist or have an empty value, a `-` will be shown in its place. The same as we do for empty widget values.

Screen Shot 2022-08-22 at 11 19 35 PM

### DIFF
--- a/src/js/apps/globals/search_app.js
+++ b/src/js/apps/globals/search_app.js
@@ -6,13 +6,15 @@ import { PatientSearchModal } from 'js/views/globals/search/patient-search_views
 
 export default App.extend({
   onStart({ prefillText }) {
-    // const settings = Radio.request('bootstrap', 'currentOrg:setting', 'patient_search_settings');
-    this.showSearch(prefillText);
+    const settings = Radio.request('bootstrap', 'currentOrg:setting', 'patient_search_settings');
+
+    this.showSearch(prefillText, settings);
   },
-  showSearch(prefillText) {
+  showSearch(prefillText, settings) {
     const patientSearchModal = new PatientSearchModal({
       collection: Radio.request('entities', 'searchPatients:collection'),
       prefillText,
+      settings,
     });
 
     this.listenTo(patientSearchModal, {
--- a/src/js/views/globals/search/patient-search.scss
+++ b/src/js/views/globals/search/patient-search.scss
@@ -47,6 +47,13 @@
   margin-left: 16px;
 }
 
+.patient-search__picklist-item-meta:empty::before {
+  color: $black-60;
+
+  // &ndash;
+  content: '–';
+}
+
 .is-highlighted {
   .patient-search__picklist-item,
   .patient-search__picklist-item-meta {
--- a/src/js/views/globals/search/patient-search_views.js
+++ b/src/js/views/globals/search/patient-search_views.js
@@ -1,4 +1,4 @@
-import { noop } from 'underscore';
+import { noop, find, get, map } from 'underscore';
 import hbs from 'handlebars-inline-precompile';
 import { View } from 'marionette';
 
@@ -42,11 +42,24 @@ const PatientSearchPicklist = Picklist.extend({
   itemTemplate: hbs`
     {{matchText text search}}{{~ remove_whitespace ~}}
     <span class="patient-search__picklist-item-meta">{{formatDateTime birth_date "MM/DD/YYYY"}}</span>
+    {{#each identifiers}}<span class="patient-search__picklist-item-meta">{{this}}</span>{{/each}}
   `,
   itemTemplateContext() {
+    const settings = this.state.get('settings');
+    const settingsIdentifiers = get(settings, 'result_identifiers');
+
+    const patientIdentifiers = this.model.get('identifiers');
+
+    const identifiers = map(settingsIdentifiers, settingsIdentifier => {
+      const identifierToShow = find(patientIdentifiers, { type: settingsIdentifier });
+
+      return identifierToShow && identifierToShow.value;
+    });
+
     return {
       text: this.getItemSearchText(),
       search: this.state.get('search'),
+      identifiers,
     };
   },
   template: hbs`
@@ -87,10 +100,11 @@ const PatientSearchModal = View.extend({
   onRender() {
     const collection = this.collection;
     const search = this.getOption('prefillText');
+    const settings = this.getOption('settings');
 
     const picklistComponent = new PatientSearchPicklist({
       lists: [{ collection }],
-      state: { search },
+      state: { search, settings },
     });
 
     this.listenTo(picklistComponent.getState(), 'change:search', this.onChangeSearch);
--- a/test/integration/services/patient-quick-search.js
+++ b/test/integration/services/patient-quick-search.js
@@ -8,11 +8,12 @@ context('Patient Quick Search', function() {
       patient.id = `${ index }`;
       patient.first_name = 'Test';
       patient.last_name = `${ index } Patient`;
+      patient.identifiers = [{ type: 'SSN', value: '123-45-6789' }, { type: 'MRN', value: '' }];
       return patient;
     });
 
     const data = _.map(patients, patient => {
-      const { id, first_name, last_name, birth_date } = patient;
+      const { id, first_name, last_name, birth_date, identifiers } = patient;
 
       return {
         id,
@@ -21,6 +22,7 @@ context('Patient Quick Search', function() {
           first_name,
           last_name,
           birth_date,
+          identifiers,
         },
         relationships: {
           patient: {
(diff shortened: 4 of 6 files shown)


src/js/apps/globals/search_app.js
```diff
@@ -6,13 +6,13 @@ import { PatientSearchModal } from 'js/views/globals/search/patient-search_views
...
 export default App.extend({
   onStart({ prefillText }) {
-    // const settings = Radio.request('bootstrap', 'currentOrg:setting', 'patient_search_settings');
     this.showSearch(prefillText);
   },
   showSearch(prefillText) {
     const patientSearchModal = new PatientSearchModal({
       collection: Radio.request('entities', 'searchPatients:collection'),
       prefillText,
+      settings: Radio.request('bootstrap', 'currentOrg:setting', 'patient_search_settings'),
```
nmajor25: Can we retrieve this org setting in the `/views/globals/search/patient_search_views.js` view file instead?

src/js/views/globals/search/patient-search_views.js
```diff
@@ -42,11 +42,34 @@ const PatientSearchPicklist = Picklist.extend({
...
+        return settingsIdentifier.toUpperCase() === patientIdentifier.type.toUpperCase();
+      });
+
+      if (identifierToShow) identifiers.push(identifierToShow.value);
+    });
+
     return {
       text: this.getItemSearchText(),
       search: this.state.get('search'),
+      identifiers,
```
nmajor25, line 62: Breakdown of what I did here:

1. Get the list of identifiers set in the `patient_search_settings` organization setting. And store it in the `settingsIdentifiers` variable. 
2. Get the patient's `identifiers` data and store it in the `patientIdentifiers` variable. My assumption is that that data will be structured like this: `[{ 'type': 'PID', 'value': '0123456789' }, { 'type': 'SSN', 'value': '321-45-6789' }]`.
3. Loop through each identifier specified in the org setting and find it's match in the patient's identifier data. Only push it to the `identifiers` array if a match is found. I'm using `toUpperCase()` because the org setting might contain lowercase versions of the identifier type (e.g. `["mrn", "ssn"]`)?
4. Update the handlebars template to display each identifier value in the `identifiers` array. If the `identifiers` array is empty, no identifiers will be displayed.

I tried to move this code into its own function (e.g. `this.getIdentifiers()`) and do this instead:

```js
return {
  text: this.getItemSearchText(),
  identifiers: this.getIdentifiers(),
}
```

But I was getting `this.getIdentifiers() is not defined` errors and couldn't figure out why... 😕 

So not entirely sure where to put this code.

src/js/apps/globals/search_app.js
paulfalgout: we'll need the settings in this app to determine what search stuff will need to be available in future work.  However I don't see why the view couldn't also retrieve them directly.

src/js/views/globals/search/patient-search_views.js
paulfalgout: Yeah the `item___` function context is odd, but the `this` is the item view of the list and not the parent component.

src/js/views/globals/search/patient-search_views.js
```diff
@@ -42,11 +42,34 @@ const PatientSearchPicklist = Picklist.extend({
...
+    const patientIdentifiers = this.model.get('identifiers');
+
+    const identifiers = [];
+
+    each(settingsIdentifiers, settingsIdentifier => {
+      const identifierToShow = find(patientIdentifiers, patientIdentifier => {
+        return settingsIdentifier.toUpperCase() === patientIdentifier.type.toUpperCase();
+      });
+
+      if (identifierToShow) identifiers.push(identifierToShow.value);
```
paulfalgout: I think we'll likely need to display some form of empty dash if the patient doesn't have the identifier.  Mostly because I don't think we can guarantee that ids are formatted differently.  So we can probably use a map

src/js/views/globals/search/patient-search_views.js
```diff
@@ -42,11 +42,34 @@ const PatientSearchPicklist = Picklist.extend({
...
+
+    const settings = this.state.get('settings');
+    const settingsIdentifiers = settings ? settings.result_identifiers : [];
+
+    const patientIdentifiers = this.model.get('identifiers');
+
+    const identifiers = [];
+
+    each(settingsIdentifiers, settingsIdentifier => {
+      const identifierToShow = find(patientIdentifiers, patientIdentifier => {
```
paulfalgout: I think we can ensure that the type and settings are the same case.  `find(patientIdentifiers, { type: settingsIdentifier })` should be ok.

src/js/views/globals/search/patient-search_views.js
```diff
@@ -42,11 +42,34 @@ const PatientSearchPicklist = Picklist.extend({
...
+    {{#each identifiers}}<span class="patient-search__picklist-item-meta">{{this}}</span>{{/each}}
   `,
   itemTemplateContext() {
+    // REMOVE: for sandbox testing purposes only
+    // this.model.set(
+    //  'identifiers',
+    //  [{ 'type': 'PID', 'value': '0123456789' }, { 'type': 'SSN', 'value': '321-45-6789' }, { 'type': 'XID', 'value': '9876543210' }],
+    // );
+
+    const settings = this.state.get('settings');
```
paulfalgout, line 48: could maybe be `const settingsIdentifiers = get(settings, 'result_identifiers')`;  If it's just going into an underscore function it doesn't need the default `[]`.  Otherwise there's a logic branch that needs testing.  Although..  I'm really tempted to start using `settings?.result_identifiers`

paulfalgout (changes requested):

src/js/views/globals/search/patient-search_views.js
nmajor25: Sounds good, will switch it to a `map()`.

For the dash, do we want to set that in the view (e.g. `identifierToShow.value ? identifierToShow.value : '-'`)? Or set that via CSS like we do for empty widget values [here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/scss/modules/widgets.scss).

My inclination is to do it via CSS since that's what we do elsewhere. But I feel like both would work.

paulfalgout: css seems fine.. otherwise.. probably an `if/else` in the template

src/js/apps/globals/search_app.js
nmajor25: Okay, makes sense! Will keep it in the app for future use.

src/js/views/globals/search/patient-search.scss
```diff
@@ -47,6 +47,13 @@
   margin-left: 16px;
 }
 
+.patient-search__picklist-item-meta:empty::before {
+  color: $black-60;
+
+  // &ndash;
+  content: '–';
+}
+
```
nmajor25, line 56: This will show the `-` if:

* Patient identifier doesn't exist. 
* Patient identifier exists, but its value is an empty string.

Both scenarios are assuming the given identifier is set via the `patient_search_settings` org setting.

Matches what we do for empty widget values [here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/scss/modules/widgets.scss).

paulfalgout (changes requested):
LGTM. Now just tests

test/integration/services/patient-quick-search.js
```diff
@@ -213,6 +215,29 @@ context('Patient Quick Search', function() {
...
+        .find('.patient-search__picklist-item-meta')
+        .last()
+        .then($el => {
+          const before = win.getComputedStyle($el[0], '::before');
+          const beforeContent = before.getPropertyValue('content');
+
+          expect(beforeContent).to.equal('"–"');
+        });
+    });
+
```
nmajor25, line 233: In this test I display two patient identifiers in the search results: 

* `SSN` that contains a value.
* `MRN` that contains an empty string value.

For the empty `MRN` value, I test that the `::before` selector contains the correct `"-"` string for its `content` CSS property. Found a way to do that in Cypress using [this](https://github.com/bahmutov/cypress-examples/blob/master/docs/recipes/pseudo-selectors.md#before-content-css-selector-before).

nmajor25:
Not sure what's going on with Coveralls. But it looks like the lines from this PR are covered? 😒

test/integration/services/patient-quick-search.js
paulfalgout: This is interesting but quite complicated for something we should maybe test for in more places.  I either think we test for empty and let the styles go we should make a custom command for this such that the API ends up being:
```js
      cy
        .get('@searchModal')
        .find('.js-picklist-item')
        .first()
        .find('.patient-search__picklist-item-meta')
        .hasBeforeContent('-');
```
It'd have to use the previous context and maybe `within` like the example you referenced.

paulfalgout (changes requested):

test/integration/services/patient-quick-search.js
nmajor25: For future reference, we decided to add the `hasBeforeContent()` Cypress command via this commit: https://github.com/RoundingWell/care-ops-frontend/pull/784/commits/4d1aac900a0c7f9e6df53faf7cdb01f4a4d89e47.

paulfalgout (approved):
I don't understand the coveralls

PR #786, opened 2022-08-26
Update outreach forms for parity form context

paulfalgout:
Outreach had gotten behind with new form features.  This update moves outreach to use the form entity and moves context gathering logic into it so that they're shared between form contexts.

Also fixes the broken button style

Shortcut Story ID: [sc-30330] [sc-30331]

### DIFF
--- a/src/js/outreach/index.js
+++ b/src/js/outreach/index.js
@@ -1,9 +1,12 @@
+import 'js/base/setup';
 import $ from 'jquery';
-import { get, pluck, extend } from 'underscore';
+import { get, map } from 'underscore';
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 import { v4 as uuid } from 'uuid';
 import 'js/base/fontawesome';
+
+import 'js/entities-service/forms';
 
 import RouterApp from 'js/base/routerapp';
 import App from 'js/base/app';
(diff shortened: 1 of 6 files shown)


src/js/outreach/index.js
```diff
@@ -1,10 +1,13 @@
+import 'js/base/setup';
 import $ from 'jquery';
-import { get, pluck, extend } from 'underscore';
+import { get, map } from 'underscore';
```
paulfalgout, line 3: pluck 🙄

src/js/outreach/index.js
```diff
@@ -1,10 +1,13 @@
+import 'js/base/setup';
```
paulfalgout, line 1: This import caused me a lot of headaches.. couldn't figure out why `dayjs().utc` was erroring when fetching a model.  I'm not sure what better way there is to do this.. I guess we could make a local dayjs import that loads the dayjs plugins rather than importing dayjs directly..  but 🤮

test/integration/outreach/outreach.js
```diff
@@ -42,6 +43,7 @@ context('Outreach', function() {
     cy
       .get('.js-submit')
       .click()
+      .wait('@routeFormAction')
```
paulfalgout, line 46: This was just a missing wait.. Ideally tests shouldn't need to change for this PR

src/js/outreach/outreach.scss
```diff
@@ -100,14 +100,14 @@ html, body, .app-root {
 .dialog__button {
   font-size: 16px;
   margin-top: 40px;
-  padding: 10px 16px;
+  padding: 8px 48px;
```
paulfalgout, line 103: these styles match the login ones now

src/js/outreach/outreach.scss
```diff
@@ -100,14 +100,14 @@ html, body, .app-root {
...
   width: 100%;
 
   .fa-right-to-bracket {
     font-size: 20px;
     margin-left: 8px;
     vertical-align: text-top;
-    width: auto;
+    width: 20px;
```
paulfalgout, line 110: I don't love adding a fixed px width here, but we seem to do it where there's a `font-size: 20px` icon

nmajor25 (approved):

PR #788, opened 2022-09-01
Add search by birth date functionality to patient search

nmajor25:
Shortcut Story ID: [sc-30513]

Add ability to search by date of birth to the patient search modal.

Includes adding `Name` and `Date of Birth` buttons to the patient search modal for users to switch between searching for patients by name or birth date.

### DIFF
--- a/src/js/apps/globals/search_app.js
+++ b/src/js/apps/globals/search_app.js
@@ -1,10 +1,23 @@
+import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
 import { PatientSearchModal } from 'js/views/globals/search/patient-search_views';
 
+const StateModel = Backbone.Model.extend({
+  defaults() {
+    return {
+      searchType: 'name',
+    };
+  },
+});
+
 export default App.extend({
+  StateModel,
+  stateEvents: {
+    'change:searchType': 'restart',
+  },
   onStart({ prefillText }) {
     const settings = Radio.request('bootstrap', 'currentOrg:setting', 'patient_search_settings');
 
@@ -15,12 +28,16 @@ export default App.extend({
       collection: Radio.request('entities', 'searchPatients:collection'),
       prefillText,
       settings,
+      searchType: this.getState('searchType'),
     });
 
     this.listenTo(patientSearchModal, {
       'item:select'({ model }) {
         Radio.trigger('event-router', 'patient:dashboard', model.get('_patient'));
         patientSearchModal.destroy();
+      },
+      'select:search:type'(newSearchType) {
+        this.setState('searchType', newSearchType);
       },
       'destroy'() {
         this.stop();
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -186,11 +186,17 @@ careOptsFrontend:
     search:
       patientSearchViews:
         patientSearchPicklist:
-          placeholderText: Search for patients
+          searchByDOB:
+            placeholderText: MM/DD/YYYY
+          searchByName:
+            placeholderText: Search for patients
         picklistEmptyView:
           noResults: No results match your query
           searching: Searching...
-          searchTip: Search by first name or last name with 3 or more characters.
+          searchByDOB:
+            searchTip: Search for patients by date of birth with MM/DD/YYYY format.
+          searchByName:
+            searchTip: Search by first name or last name with 3 or more characters.
           shortcut: "Keyboard Shortcut: /"
   patients:
     patient:
--- a/src/js/views/globals/search/patient-search.scss
+++ b/src/js/views/globals/search/patient-search.scss
@@ -15,16 +15,41 @@
   }
 }
 
+.patient-search__clear-type-button {
+  background-color: color(light_blue, light);
+  border: none;
+  color: $black-20;
+  font-size: 14px;
+  height: 28px;
+  margin-right: 8px;
+  padding: 4px 12px;
+
+  .patient-search__clear-type-button-label {
+    vertical-align: middle;
+  }
+
+  .icon {
+    font-size: 14px;
+    margin-left: 8px;
+    vertical-align: bottom;
+    width: 1em;
+  }
+}
+
 .patient-search__input {
   border: none;
   display: flex;
+  flex-grow: 1; // tells <input /> to fill width regardless of button size
   font-size: 24px;
   font-weight: normal;
 
   // height = line-height: 1.2
   height: 28px;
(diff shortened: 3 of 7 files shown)


src/js/views/globals/search/patient-search.scss
```diff
@@ -15,16 +15,41 @@
...
   font-weight: normal;
 
   // height = line-height: 1.2
   height: 28px;
   padding-right: 24px;
-  width: 100%;
+
+  &[type='date']::-webkit-calendar-picker-indicator {
+    display: none;
+  }
```
nmajor25, line 52: Can we do this instead of doing the overflow we mentioned on Slack?

This should hide the calendar icon, but still allow the date picker to open on mobile/iOS.

src/js/views/globals/search/patient-search.scss
```diff
@@ -61,11 +100,40 @@
...
+  padding: 8px 0;
+}
+
+.patient-search__shortcut {
+  bottom: 0;
+  padding-bottom: 24px;
+  position: absolute;
+  text-align: center;
+  width: 100%;
+}
```
nmajor25, line 139: Haven't had a chance to go through and clean this file up.

So this is still in a "get things working" state. Open to any suggestions here.

src/js/views/globals/search/patient-search_views.js
```diff
@@ -116,16 +61,24 @@ const PatientSearchModal = View.extend({
       },
     });
 
-    if (search) this.collection.search(search);
+    if (search) this.collection.search(search, searchType);
```
nmajor25, line 64: Pass the `searchType` to the patient search service. 

If `searchType === 'dob'`, an additional step is made to validate the inputted date before sending the api request.

src/js/views/globals/search/search-by-dob_views.js
```diff
@@ -0,0 +1,196 @@
...
+  initialize(options) {
+    this.mergeOptions(options, CLASS_OPTIONS_ITEM);
+  },
+  childViewOptions() {
+    const opts = pick(this, ...CLASS_OPTIONS_ITEM);
+    return extend({ state: this.model }, opts);
+  },
+});
+
+const PicklistsCollectionView = CollectionView.extend({
```
nmajor25, line 50: I trimmed the `PicklistsCollectionView` and `Picklist` components down as much as I knew how here.

Any other ways to trim these? Or re-organize? Open to any suggestions.

src/js/views/globals/search/search-by-dob_views.js
```diff
@@ -0,0 +1,196 @@
...
+  template: hbs`
+    <div class="modal__header patient-search__header">
+      <span class="modal__header-icon">{{far "magnifying-glass"}}</span>
+      <div class="js-clear-search-type button-primary patient-search__clear-type-button">
+        <span class="patient-search__clear-type-button-label">Date of Birth</span>{{far "xmark"}}
+      </div>
+      <input
+        type="date"
+        class="js-input patient-search__input"
+        placeholder="{{ @intl.globals.search.patientSearchViews.patientSearchPicklist.searchByDOB.placeholderText }}"
```
nmajor25, line 64: Tried to set a placeholder of `MM/DD/YYYY` here. But as I learned, I guess you can't change the placeholder used for a `type="date"` input.

The default is `mm/dd/yyyy`:

Screen Shot 2022-09-01 at 12 07 27 AM

src/js/views/globals/search/patient-search.scss
paulfalgout: only works on chrome

src/js/views/globals/search/search-by-dob_views.js
paulfalgout: we're doing something for a placeholder for input date on the outreach login

src/js/apps/globals/search_app.js
```diff
@@ -1,10 +1,23 @@
+import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
 import { PatientSearchModal } from 'js/views/globals/search/patient-search_views';
 
+const StateModel = Backbone.Model.extend({
+  defaults() {
```
paulfalgout, line 9: this could be `defaults: { searchType: 'name' }`

src/js/views/globals/search/patient-search.scss
```diff
@@ -35,6 +60,20 @@
...
+.patient-search__search-by-button {
+  background-color: color(light_blue, light);
+  border: none;
+  color: $black-20;
+  font-size: 14px;
+  height: 28px;
+  margin-right: 16px;
+  padding: 4px 12px;
+
+  .patient-search__search-by-button-label {
```
paulfalgout, line 72: ideally not nested

src/js/views/globals/search/patient-search.scss
```diff
@@ -35,6 +60,20 @@
   border-bottom: 1px solid $black-90;
 }
 
+.patient-search__search-by-button {
+  background-color: color(light_blue, light);
+  border: none;
+  color: $black-20;
+  font-size: 14px;
+  height: 28px;
```
paulfalgout, line 68: maybe we can reach a 4px increment, or avoiding setting the height at all by changing the line-height?

src/js/views/globals/search/patient-search.scss
```diff
@@ -15,16 +15,41 @@
   }
 }
 
+.patient-search__clear-type-button {
```
paulfalgout, line 18: can probably be genericized with `patient-search__search-by-button` and margin could be set in the html with a `u-m...` class

src/js/i18n/en-US.yml
```diff
@@ -186,11 +186,17 @@ careOptsFrontend:
...
-          placeholderText: Search for patients
+          searchByDOB:
+            placeholderText: MM/DD/YYYY
+          searchByName:
+            placeholderText: Search for patients
         picklistEmptyView:
           noResults: No results match your query
           searching: Searching...
-          searchTip: Search by first name or last name with 3 or more characters.
+          searchByDOB:
```
paulfalgout, line 196: `searching` should be below these two I suppose

src/js/views/globals/search/search-by-dob_views.js
paulfalgout: We could probably add named exports to the component and modify that one.  Or possibly group shared content into an exported mixin.  I'm not entirely sure yet.  This is a complicated one..

paulfalgout (commented):
Initial thoughts.. but yeah.. this is complicated.. we can work through it next week

PR #795, opened 2022-09-06
Prevent stored submissions on read only forms

paulfalgout:
Shortcut Story ID: [sc-30716]

This work disables the localstorage form submission prefill on forms marked read only.

New forms are updating the form submission to store user state for what they're viewing.

This however was very hard to test for.. I can recreate the issue on the live form, but can't get it to happen in cypress.

Additionally this unfortunately is not only preventing storing the changed submission, but it's also preventing displaying the form submission since we have some out in the wild and deem this a necessary quick fix.  Otherwise we could presumably just prevent updating the localstorage.


paulfalgout:
broken test is legitimate, so I'm going to have to hunt that down tomorrow

nmajor25 (approved):

PR #798, opened 2022-09-08
Add date of birth & identifiers to patient search help text ('Or search for patients by')

nmajor25:
Shortcut Story ID: [sc-30727]

In the patient search modal, this adds a section to the help text labeled `Or search for patients by` that looks like the screenshot below. It will show users that they can search by a patient's date of birth or one/multiple identifiers.

The `Date of Birth` option will always be shown by default.

The identifiers are based on a `patient_search` organization setting:

```js
{
  "identifiers": [
    {
        "type": "MRN",
        "label": "Medical Record Number",
        "example": "A234567"
    },
    {
        "type": "PID",
        "label": "Health Plan ID",
        "example": "123456789"
    }
  ]
}
```

Where `label` is shown in the blue button and the `example` is shown in the `Example: {{example}}` text.

Screen Shot 2022-09-08 at 10 10 13 AM

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -214,7 +214,7 @@ export default App.extend({
     this.showChildView('navContent', navView);
   },
   showSearch(prefillText) {
-    if (!Radio.request('bootstrap', 'currentOrg:setting', 'patient_search_settings')) {
+    if (!Radio.request('bootstrap', 'currentOrg:setting', 'patient_search')) {
       const navView = this.getChildView('navContent');
 
       const patientSearchModal = new PatientSearchModal({
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -190,7 +190,7 @@ careOptsFrontend:
         picklistEmptyView:
           noResults: No results match your query
           searching: Searching...
-          searchTip: Search by first name or last name with 3 or more characters.
+          searchTip: Search by first name, last name with 3 or more characters.
           shortcut: "Keyboard Shortcut: /"
   patients:
     patient:
--- a/src/js/views/globals/search/patient-search.scss
+++ b/src/js/views/globals/search/patient-search.scss
@@ -6,7 +6,7 @@
 
 .patient-search__close {
   position: absolute;
-  right: 24px;
+  right: 16px;
   top: 24px;
   z-index: 10;
 
@@ -33,12 +33,14 @@
 
 .patient-search__header {
   border-bottom: 1px solid $black-90;
+  margin: 0 8px;
+  padding-left: 16px;
 }
 
 .patient-search__picklist-item {
   color: $black-20;
   font-size: 16px;
-  padding: 4px 56px;
+  padding: 4px 64px;
 }
 
 .patient-search__picklist-item-meta {
@@ -68,4 +70,42 @@
   margin: 0 8px;
   padding: 16px 0;
   text-align: center;
+}
+
+.patient-search__tip {
+  border-bottom: 1px solid $black-90;
+  padding-bottom: 16px;
+  padding-left: 60px;
+  text-align: left;
+}
+
+.patient-search__search-by {
+  padding: 16px 0;
+  padding-left: 60px;
+  text-align: left;
+}
+
+.patient-search__search-by-title {
+  font-size: 12px;
+  font-weight: 600;
+}
+
+.patient-search__search-by-label {
+  background-color: color(light_blue, light);
+  border-radius: 4px;
+  color: $black-20;
+  display: inline-block;
+  font-weight: 600;
+  margin-right: 8px;
+  padding: 4px 12px;
+}
+
+.patient-search__shortcut {
+  bottom: 0;
+  left: 0;
+  margin-bottom: 24px;
+  position: absolute;
+  right: 0;
+  text-align: center;
+  width: 100%;
 }
--- a/src/js/views/globals/search/patient-search_views.js
+++ b/src/js/views/globals/search/patient-search_views.js
@@ -1,4 +1,4 @@
-import { noop, find, get, map } from 'underscore';
+import { noop } from 'underscore';
 import hbs from 'handlebars-inline-precompile';
 import { View } from 'marionette';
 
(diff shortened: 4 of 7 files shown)


src/js/apps/globals/nav_app.js
```diff
@@ -214,7 +214,7 @@ export default App.extend({
     this.showChildView('navContent', navView);
   },
   showSearch(prefillText) {
-    if (!Radio.request('bootstrap', 'currentOrg:setting', 'patient_search_settings')) {
+    if (!Radio.request('bootstrap', 'currentOrg:setting', 'patient_search')) {
```
nmajor25, line 217: Changed this and all other instances of `patient_search_settings` to `patient_search`.

src/js/i18n/en-US.yml
```diff
@@ -190,7 +190,7 @@ careOptsFrontend:
         picklistEmptyView:
           noResults: No results match your query
           searching: Searching...
-          searchTip: Search by first name or last name with 3 or more characters.
+          searchTip: Search by first name, last name with 3 or more characters.
```
nmajor25, line 193: Update this text to match what was in the design document.

src/js/views/globals/search/patient-search.scss
```diff
@@ -6,7 +6,7 @@
 
 .patient-search__close {
   position: absolute;
-  right: 24px;
+  right: 16px;
```
nmajor25, line 9: Updated this to match the styling in the design doc. Moves the `X` close icon closer to the right side of the modal.

Screen Shot 2022-09-08 at 12 02 21 PM

src/js/views/globals/search/patient-search.scss
```diff
@@ -33,12 +33,14 @@
 
 .patient-search__header {
   border-bottom: 1px solid $black-90;
+  margin: 0 8px;
+  padding-left: 16px;
```
nmajor25, line 37: The design doc includes `8px` of spacing between the input bottom border on both sides. This adds that and updates the left padding for aligning with the help text and results content.

Screen Shot 2022-09-08 at 12 03 56 PM

src/js/views/globals/search/patient-search.scss
```diff
@@ -33,12 +33,14 @@
...
 }
 
 .patient-search__picklist-item {
   color: $black-20;
   font-size: 16px;
-  padding: 4px 56px;
+  padding: 4px 64px;
```
nmajor25, line 43: This aligns the search results items with the input text and help text content.

Screen Shot 2022-09-08 at 12 06 38 PM

src/js/views/globals/search/patient-search.scss
```diff
@@ -69,3 +71,48 @@
...
+
+.patient-search__shortcut {
+  bottom: 0;
+  left: 0;
+  margin-bottom: 24px;
+  position: absolute;
+  right: 0;
+  text-align: center;
+  width: 100%;
+}
```
nmajor25, line 111: Moves the keyboard shortcut tip to the bottom of the modal.

Screen Shot 2022-09-08 at 12 08 36 PM

src/js/views/globals/search/patient-search_views.js
```diff
@@ -42,24 +65,11 @@ const PatientSearchPicklist = Picklist.extend({
...
-    const settingsIdentifiers = get(settings, 'result_identifiers');
-
-    const patientIdentifiers = this.model.get('identifiers');
-
-    const identifiers = map(settingsIdentifiers, settingsIdentifier => {
-      const identifierToShow = find(patientIdentifiers, { type: settingsIdentifier });
-
-      return identifierToShow && identifierToShow.value;
-    });
-
```
nmajor25, line 58: Removed code that handles showing identifiers in the search results. Will be handled in a future story.

Also removed those portions of the Cypress tests.

src/js/views/globals/search/patient-search.scss
```diff
@@ -69,3 +71,48 @@
...
+  padding: 16px 0;
+  padding-left: 60px;
+  text-align: left;
+}
+
+.patient-search__search-by-title {
+  font-weight: 600;
+}
+
+.patient-search__search-by-option {
```
paulfalgout: this is probably a case where `u-margin--t-8` would work

src/js/views/globals/search/patient-search_views.js
```diff
@@ -10,8 +10,23 @@ import Picklist from 'js/components/picklist';
...
 
 const TipTemplate = hbs`
-<div>{{ @intl.globals.search.patientSearchViews.picklistEmptyView.searchTip }}</div>
-<div>{{fas "keyboard"}}<strong class="u-margin--l-8">{{ @intl.globals.search.patientSearchViews.picklistEmptyView.shortcut }}</strong></div>
+  <div class="patient-search__tip">{{ @intl.globals.search.patientSearchViews.picklistEmptyView.searchTip }}</div>
+  <div class="patient-search__search-by">
+    <div class="patient-search__search-by-title">Or search for patients by</div>
+    <div class="patient-search__search-by-buttons">
+      <div class="patient-search__search-by-option">
+        <div class="patient-search__search-by-button button-primary">Date of Birth</div>
```
paulfalgout: These are no longer buttons, so let's remove `button-primary` and see if we need to override styles as much.  I also suspect this could be a `span`

paulfalgout (changes requested):
Some style requests

src/js/views/globals/search/patient-search_views.js
```diff
@@ -10,8 +10,21 @@ import Picklist from 'js/components/picklist';
 import './patient-search.scss';
 
 const TipTemplate = hbs`
-<div>{{ @intl.globals.search.patientSearchViews.picklistEmptyView.searchTip }}</div>
-<div>{{fas "keyboard"}}<strong class="u-margin--l-8">{{ @intl.globals.search.patientSearchViews.picklistEmptyView.shortcut }}</strong></div>
+  <div class="patient-search__tip">{{ @intl.globals.search.patientSearchViews.picklistEmptyView.searchTip }}</div>
+  <div class="patient-search__search-by">
+    <div class="patient-search__search-by-title">Or search for patients by</div>
+    <div class="u-margin--t-8 qa-patient-search__search-by-option">
```
nmajor25: I need to target these elements in the tests. I was using the `patient-search__search-by-option` class name before.

Does using a `qa-patient-search__search-by-option` class name work instead (with the `qa-` prefix)? Its sole purpose would be for the Cypress tests.

We discussed that here: https://app.shortcut.com/roundingwell/story/30275/update-test-selectors-for-best-practices. But we haven't implemented it in the app yet.

src/js/views/globals/search/patient-search_views.js
nmajor25: Removed the `button-primary` class name in the last commit. And changed it to a `span`.

I also removed the `<div class="patient-search__search-by-buttons">` element since it wasn't being used for anything.

src/js/views/globals/search/patient-search_views.js
paulfalgout: yeah though I don't think it needs to be BEM'd.  `.qa-something-something` is fine.  it really doesn't even need to be very unique because we should be scoping tests down such that something as simple as `.qa-item` would suffice... similar to `.js-select` etc..  I don't think it should necessarily be _that_ generic, but.. I dunno.. `.qa-search-option` or something would work.

paulfalgout (changes requested):

src/js/views/globals/search/patient-search_views.js
nmajor25: Cool, makes sense. Just updated those to `.qa-search-option`.

src/js/views/globals/search/patient-search.scss
```diff
@@ -69,3 +71,41 @@
...
+  padding-left: 60px;
+  text-align: left;
+}
+
+.patient-search__search-by-title {
+  font-size: 12px;
+  font-weight: 600;
+}
+
+.patient-search__search-by-label {
```
paulfalgout, line 93: keep this style in mind.. if we end up using it somewhere else we should genericize it once we know how it can be described generally @nmajor25

paulfalgout (approved):

paulfalgout:
replaces https://github.com/RoundingWell/care-ops-frontend/pull/788

PR #799, opened 2022-09-09
Set form.io survey component emptyValues to array

paulfalgout:
This seems to avoid the error we’re seeing without consequence

Shortcut Story ID: [sc-30750]

### DIFF
--- a/src/js/formapp/components.js
+++ b/src/js/formapp/components.js
@@ -3,6 +3,15 @@ import { debounce } from 'underscore';
 const NestedComponent = Formio.Components.components.nested;
 const SelectComponent = Formio.Components.components.select;
 
+// Modifies the emptyValue to avoid error on php mutating {} to []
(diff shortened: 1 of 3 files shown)


src/js/formapp/components.js
```diff
@@ -3,6 +3,15 @@ import { debounce } from 'underscore';
 const NestedComponent = Formio.Components.components.nested;
 const SelectComponent = Formio.Components.components.select;
 
+// Modifies the emptyValue to avoid error on php mutating {} to []
+class SurveyComponent extends Formio.Components.components.survey {
+  get emptyValue() {
+    return [];
```
shadowhand, line 9: This works better than `null` ?

paulfalgout: shockingly null still throws the error.

nmajor25 (approved):

PR #801, opened 2022-09-13
Fix Cypress flake

paulfalgout:
Shortcut Story ID: [sc-30333]

Previously it appears that Cypress consistently loaded the `/clinicians/me` API prior to the groups bootstrap which allowed for tests to set the bootstrap and expect the correct data to be on the current clinician.  In v10, either might load first, breaking tests when the order changed.

This work aligns the stubs so that either load order will work.


paulfalgout:
ok still some work to do

paulfalgout:
@nmajor25 This doesn't fix _all_ the flake, but it does seem to solve much of it.

nmajor25 (approved):
Look good.

Hopefully this fixes the flakiness!

PR #802, opened 2022-09-13
Add identifiers to patient search results

nmajor25:
Shortcut Story ID: [sc-30739]

Update the patient search results to include identifiers (if it exists in the search results for a patient). Only one identifier is shown in the results at a time.

Search results with only patient name and date of birth:

Screen Shot 2022-09-13 at 11 12 07 AM

Search results with patient name, date of birth, and identifier:

Screen Shot 2022-09-13 at 11 18 15 AM


src/js/views/globals/search/patient-search_views.js
```diff
@@ -56,18 +56,30 @@ const EmptyView = View.extend({
...
+    <div class="patient-search__picklist-item-name u-text--overflow">
+      <span>{{matchText name search}}{{~ remove_whitespace ~}}</span>
+    </div>
+    <div class="patient-search__picklist-item-meta">
+      <div class="patient-search__picklist-item-dob u-text--overflow">
+        {{formatDateTime birth_date "MM/DD/YYYY"}}
+      </div>
+      {{#if identifier}}
+        <div class="patient-search__picklist-item-identifier u-text--overflow">
+          {{matchText identifier search}}
```
paulfalgout: I think this could be `{{matchText identifiers[0].value search}}`.. not 100% confident..

src/js/views/globals/search/patient-search_picklist.js
```diff
@@ -0,0 +1,261 @@
...
+          <span>DOB</span>
+        </div>
+        {{#if showIdentifierHeader}}
+          <div class="patient-search__picklist-header-identifier">
+            <span>Identifier</span>
+          </div>
+        {{/if}}
+      </div>
+    </div>
+    <ul></ul>
```
nmajor25: To get this to work, I copied over the entire picklist component instead of making updates to the original that is used elsewhere in the app. But this is where the update to it was made to get the header shown above the picklist items.

Still need to clean up this file and reduce it down to only what's needed.

src/js/views/globals/search/patient-search_picklist.js
```diff
@@ -0,0 +1,261 @@
...
+  childViewOptions() {
+    const opts = pick(this, ...CLASS_OPTIONS_ITEM);
+    return extend({ state: this.model }, opts);
+  },
+  templateContext() {
+    const firstPatient = this.collection.first();
+    const identifierArray = firstPatient ? firstPatient.get('identifiers') : [];
+
+    return {
+      showIdentifierHeader: identifierArray.length > 0,
```
nmajor25: Only show identifier header if the first patient result in the collection contains an identifier.

`identifers = []` vs. `identifiers = [{ type: 'PID', value: '123456789 }]`.

test/integration/services/patient-quick-search.js
```diff
@@ -8,11 +8,12 @@ context('Patient Quick Search', function() {
       patient.id = `${ index }`;
       patient.first_name = 'Test';
       patient.last_name = `${ index } Patient`;
+      patient.identifiers = [];
```
nmajor25: Still need to add tests here. But this change makes it so the test passes. Was failing without this.

src/js/views/globals/search/patient-search_views.js
```diff
@@ -56,18 +56,30 @@ const EmptyView = View.extend({
...
   `,
   itemTemplateContext() {
+    const name = `${ this.model.get('first_name') } ${ this.model.get('last_name') }`;
+    const identifiers = this.model.get('identifiers');
+
     return {
-      text: this.getItemSearchText(),
+      name,
       search: this.state.get('search'),
+      identifier: identifiers.length && identifiers[0].value,
```
nmajor25: Only show identifier value if one exists in the patient's `identifiers = []` array. And use the first identifier by default (`identifiers[0]`).

src/js/views/globals/search/patient-search_views.js
```diff
@@ -56,18 +56,30 @@ const EmptyView = View.extend({
...
     };
   },
   template: hbs`
```
paulfalgout, line 171: So... I _think_ we can move the headers up to this template.. perhaps outside of the `ul.js-picklist-scroll` and use the generic picklist.
Then below in the `initialize` that listen to `search` on the collection is the event of the results from a search.  So that can change to something like:
```js
this.listenTo(this.collection, 'search', () => {
  const hasIdentifiers = this.collection......do the thing;
  this.setState({ hasIdentifiers });  // makes it available to the template
  this.renderView(); // this covers the re-render
});
```

src/js/views/globals/search/patient-search_picklist.js
```diff
@@ -0,0 +1,261 @@
...
+  initialize(options) {
+    this.mergeOptions(options, CLASS_OPTIONS_ITEM);
+  },
+  childViewOptions() {
+    const opts = pick(this, ...CLASS_OPTIONS_ITEM);
+    return extend({ state: this.model }, opts);
+  },
+  templateContext() {
+    const firstPatient = this.collection.first();
+    const identifierArray = firstPatient ? firstPatient.get('identifiers') : [];
```
paulfalgout: the tertiary is going to make this tough to test.
also we probably need to check for any identifiers unfortunately..
probably something like:
```js
const hasIdentifiers = this.collection.contains(model => {
  return _.get(model.get('identifiers'), 'length');
});
```

paulfalgout (changes requested):

src/js/views/globals/search/patient-search_views.js
nmajor25: That would definitely be cleaner. Handlebars doesn't like it though.

Screen Shot 2022-09-13 at 12 02 01 PM 1

paulfalgout: try `{matchText identifiers.0.value search}}`

nmajor25: Nice, that works.

src/js/views/globals/search/patient-search_views.js
```diff
@@ -38,52 +42,176 @@ const EmptyView = View.extend({
...
       search: this.state.get('search'),
     };
   },
+});
+
+const ListView = CollectionView.extend({
+  className: 'flex-region',
+  behaviors: [
+    PicklistBehavior,
+  ],
```
nmajor25, line 100: This behavior is already on the `DialogView` view.

Does it need to be set again?

paulfalgout:
Alright.  We can fix things in future PRs, but I think everything that was discussed/discovered has been accounted for and this needs to make the release

PR #804, opened 2022-09-22
Fix async and display issue on quick search

paulfalgout:
Shortcut Story ID: [sc-30964] [sc-30961]

Testing the first story in cypress is a bit complicated due to the async request needing to be timed along with the debounce.  At least it was pretty flaky.. maybe a `cy.clock` somehow would help, but nothing at first glance was specific.  That said a real world test was easy and all of the logic branches should be covered here.

### DIFF
--- a/src/js/base/collection.js
+++ b/src/js/base/collection.js
@@ -20,6 +20,8 @@ export default Backbone.Collection.extend(extend({
     // On success resolves the entity instead of the jqxhr success
     const d = $.Deferred();
 
+    d.abort = xhr.abort;
+
     $.when(xhr)
       .fail(bind(d.reject, d))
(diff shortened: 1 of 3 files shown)


nmajor25 (approved):
Tested in my sandbox, bugs appear to be fixed!

paulfalgout:
@nmajor25 last minute revision due to an issue of header alignment with the scrollbar.  mark didn't mind putting the headers in the scroll area to solve the problem.  It could be done with javascript, but a bit overkill for this use case.  This also changed the esc button solution.

Screen Shot 2022-09-22 at 11 57 15 PM

paulfalgout:
the coveralls issue is a fluke

nmajor25 (approved):

PR #806, opened 2022-09-26
Save + Go Back combo button for forms

nmajor25:
Shortcut Story ID: [sc-30405]

Adds a `Save + Go Back` button to the forms page. When clicked, it saves the form and, when complete, takes the user back to their previous page.

Allows the user to choose between either a `Save` (default) or `Save + Go Back` button. That choice is persisted via local storage.

`Save` button shown by default:

Screen Shot 2022-09-26 at 5 01 43 PM

Dropdown where different button option is chosen:

Screen Shot 2022-09-26 at 5 05 56 PM

`Save + Go Back` button:

Screen Shot 2022-09-26 at 5 02 11 PM


src/js/views/forms/form/form.scss
```diff
@@ -146,3 +146,40 @@
...
+  align-items: center;
+  display: flex;
+}
+
+.form__save-dropdown-picklist-icon {
+  margin-left: auto;
+
+  .icon {
+    font-size: 12px;
+    margin-right: 0 !important;
```
nmajor25: I don't like using `!important`, but it's the only way I could get this property applied to the `.icon` here.

It's being overridden by the Picklist component's styling: [.picklist__item .icon:first-child](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/components/picklist/picklist.scss#L53).

Without `margin-right: 0` here, the icon would not have the correct spacing:

Screen Shot 2022-09-26 at 4 24 22 PM

vs.

Screen Shot 2022-09-26 at 4 24 44 PM

test/integration/forms/form.js
```diff
@@ -1225,6 +1225,119 @@ context('Patient Action Form', function() {
...
+        expect(data.relationships.form.data.id).to.equal('11111');
+        expect(data.attributes.response.data.familyHistory).to.equal('Here is some typing');
+        expect(data.attributes.response.data.storyTime).to.equal('Once upon a time...');
+        expect(data.attributes.response.data.patient.fields.weight).to.equal(192);
+      });
+
+    cy
+      .url()
+      .should('contain', 'worklist/owned-by');
+  });
```
nmajor25, line 1347: Brief overview of what I'm doing in this test:

1. Load form page with a basic form loaded. Preset the local storage value to have a `saveButtonType: 'saveAndGoBack'` key-value.
2. Verify the `Save + Go Back` version of the button is shown initially. Based on what is set in local storage.
3. Use the button droplist to switch to the `Save` button. Verify that `saveButtonType: 'save'` is now set in local storage.
4. Use the droplist to switch back to the `Save + Go Back` button and click it to save the form.
5. Make sure form is saved correctly.
6. Check the page URL to ensure the user is taken back to their previous page after they save the form. In this case, the default `/worklist/owned-by` page is used.

test/integration/forms/form.js
```diff
@@ -1225,6 +1225,119 @@ context('Patient Action Form', function() {
...
+      .wait('@routePatientByAction')
+      .wait('@routeFormDefinition')
+      .wait('@routeFormResponse');
+
+    cy
+      .get('.form__controls')
+      .find('button')
+      .contains('Update')
+      .click()
+      .wait('@routeFormActionFields');
```
nmajor25: Is there a way this test can work without doing the clicking of the `Update` button first? I tried skipping that, but it was causing all sorts of errors where elements on the page were being "not attached to the DOM".

Also looking for any other unnecessary things to trim from these tests I added.

test/integration/forms/form.js
```diff
@@ -1225,6 +1225,119 @@ context('Patient Action Form', function() {
...
+        delay: 100,
+        url: '/api/form-responses',
+        response: { data: { id: '12345' } },
+      })
+      .as('routePostResponse');
+
+    cy
+      .get('.form__controls')
+      .find('.js-save-button')
+      .should($el => expect($el.text().trim()).to.equal('Save + Go Back'));
```
nmajor25: I was looking for a way to test the exact text value for the button element. `Save` vs. `Save + Go Back`. 

For instance, if the button has a text value of `'Save + Go Back'` and we test for `.contains('Save')`, that test could pass even when it shouldn't (if we want to make sure it equals `Save + Go Back` exactly).

This method I used gets the jquery element's text and `trims()` the extra spaces from it. Maybe create a `.text()` command for this? Similar to [here](https://github.com/Lakitna/cypress-commands/blob/develop/docs/text.md).

Open to any better ways of doing this.

src/js/apps/forms/form/form_app.js
```diff
@@ -40,7 +40,12 @@ export default App.extend({
   initFormState() {
     const storedState = store.get(`form-state_${ this.currentUser.id }`);
 
-    this.setState(extend({ isActionSidebar: true, isExpanded: true, shouldShowHistory: false }, storedState));
+    this.setState(extend({
+      isActionSidebar: true,
+      isExpanded: true,
+      shouldShowHistory: false,
+      saveButtonType: 'save',
```
nmajor25, line 47: This state value will either be `'save'` or `'saveAndGoBack'`. And determines whether the `Save Form` or `Save Form + Go Back` button is shown in the views.

src/js/views/forms/form/form_views.js
```diff
@@ -259,21 +260,94 @@ const ReadOnlyView = View.extend({
...
+      {{/if}}
+      {{#if showSaveGoBackButton}}
+        {{ @intl.forms.form.formViews.saveView.saveAndGoBack.buttonText }}
+      {{/if}}
+    </button>
+    <button data-save-type-region></button>
+  `,
   triggers: {
-    'click': 'click',
+    'click .js-save-button': 'click:save',
```
nmajor25, line 325: Changed this from `click` to `click:save`. Thought it made more sense in terms of code readability, since there are now multiple clicks going on in this view. Not just the `Save` button. But the droplist button now too.

src/js/views/forms/form/form_views.js
```diff
@@ -259,21 +260,94 @@ const ReadOnlyView = View.extend({
   template: hbs`{{ @intl.forms.form.formViews.readOnlyView.buttonText }}`,
 });
 
+const SaveButtonTypeDroplist = Droplist.extend({
+  popWidth: 192,
+  align: 'right',
```
nmajor25, line 264: `align: 'right'` not working was because I had `popWidth` set to `popWidth: '192px'`. Instead of `popWidth: 192`.

For some reason that made the droplist miscalculate the correct position to display itself 🤷 

_This is something @paulfalgout  and I discussed off of GitHub_.

src/js/apps/forms/form/form_app.js
```diff
@@ -249,14 +271,26 @@ export default App.extend({
...
-    this.listenTo(saveView, 'click', () => {
-      Radio.request(`form${ this.form.id }`, 'send', 'form:submit');
+    const saveView = this.showChildView('formAction', new SaveView({
+      model: this.getState(),
+    }));
+
+    this.listenTo(saveView, {
+      'click:save'() {
+        Radio.request(`form${ this.form.id }`, 'send', 'form:submit');
+        this.showFormSaveDisabled();
```
nmajor25, line 289: This ensures that both buttons are disabled when the user clicks the `Save` or `Save Form + Go Back` button.

Screen Shot 2022-09-26 at 4 55 06 PM

nmajor25:
@paulfalgout

src/js/views/forms/form/form_views.js
paulfalgout: I don't think popWidth is necessary here

test/integration/forms/form.js
paulfalgout: yeah if you remove the `form-responses` relation it won't be needed

test/integration/forms/form.js
paulfalgout: Can we test for `Go Back` or not?

paulfalgout (approved):
Hey so everything looked good, but I want to look at making the styles more generic... but due to my deficiencies in css, I had to play around to figure out what to suggest.. it became a bit easier to add some commits.

We can rewrite them or whatever, but see what you think.  

If this seems reasonable.. we will need tests for the generic component's new `isCheckable`

Thoughts?

test/integration/forms/form.js
nmajor25: Nice, that did the trick. Thanks!

test/integration/forms/form.js
nmajor25: Hmm... yes we could.

For the `Save + Go Back` button:

```js
.should('contain', 'Save + Go Back')
```

For the `Save` button:

```js
.should('contain', 'Save')
.should('not.contain', 'Go Back');
```

That makes sense to me.

test/integration/forms/form.js
```diff
@@ -1318,11 +1305,31 @@ context('Patient Action Form', function() {
...
 
+    cy
+      .get('.form__controls')
+      .find('.js-save-button')
+      .should('be.disabled');
+
+    cy
+      .get('.form__controls')
+      .find('.button__drop-list-select')
+      .should('be.disabled');
```
nmajor25, line 1331: Tests that both the save button and droplist button are `disabled` after a user saves the form.

Do we need to add more delay on the stubbed `@routePostResponse` route? It's currently set to `delay: 100` earlier on in this test. Or is `100ms` enough time for Cypress to grab those buttons before the page is redirected?

test/integration/forms/form.js
```diff
@@ -1234,35 +1234,19 @@ context('Patient Action Form', function() {
         fx.data.id = '1';
         fx.data.relationships.form.data = { id: '11111' };
         fx.data.relationships['program-action'] = { data: { id: '11111' } };
-        fx.data.relationships['form-responses'].data = [
-          { id: '1', meta: { created_at: testTs() } },
-        ];
```
nmajor25: Removed this and `routeFormResponse()` as they weren't necessary for this test. And allows us to not show the `Update` button initially.

test/integration/forms/form.js
paulfalgout: well the communication between the iframe and the app is `postMessage` which is async.. so in this case the delay time is an unknown, but surely 100ms is enough.

paulfalgout: ah, but I suppose it could be faster than that..

nmajor25:
@paulfalgout I think the CSS updates are slick. The `isCheckable` as well. It seems like something we'd definitely use in the future elsewhere.

I just added a commit that cleans up the save + go back Cypress tests. And I also added a test to those that verifies the buttons show the correct `disabled` state.

Working on the picklist tests next.

paulfalgout:
@nmajor25 ah the two tests that are failing are overly specific.  with my stuff, the picklist added a wrapping `div` and so they're breaking.

nmajor25:
@paulfalgout the droplist doesn't close when a user clicks inside the form iframe. 

It's something I noticed when building the initial version of this. Forgot, and just remembered now. But will look into that as well.

https://user-images.githubusercontent.com/35355575/192867505-b0542bf1-3b5d-42f8-b730-0cd465acf438.mov

paulfalgout:
yep. so this is for sure true for anything in the pop region.

The seamless way to fix this I suppose would be to duplicate
```js
    $document.on('keydown.app', function(evt) {
      Radio.trigger('user-activity', 'document:keydown', evt);
    });

    this.setMouseListeners();
    this.setHotkeyListeners();
```
from `app.js` in `formapp.js` but instead of triggering on a local radio, it'd need to send an event over the `router.request` so that the form service can trigger the user activity.

That's a lot of chatter on that postmessage, but maybe it's ok..  otherwise.. perhaps we could just detect some sort of iframe focus somehow and trigger just the one event... which might be preferrable?  It'd need to handle a click or tab into the frame..   I suppose it doesn't necessarily have to happen with this work.. though it is suddenly more noticeable.

nmajor25:
Interesting... that seems doable. I can create a separate story for that and tackle it in a new PR. 

This PR is getting beefy as it is 🍖 .

_Edit: [Related shortcut story](https://app.shortcut.com/roundingwell/story/31196/trigger-keydown-user-activity-events-from-the-form-app)._

test/integration/components/droplist.js
```diff
@@ -165,4 +165,58 @@ context('Droplist', function() {
...
+      .should('have.class', 'fa-circle-check');
+
+    cy
+      .get('.picklist')
+      .find('.js-picklist-item')
+      .last()
+      .find('.icon')
+      .should('not.exist');
+  });
 });
```
nmajor25, line 222: Simple test that:

1. Opens the droplist.
2. Selects a picklist item in the droplist and verifies the check mark icon is shown for it.
3. Verifies only the one picklist item has the check mark. This ensures it's not applied to multiple picklist items and only the selected one. Possibly overkill.

test/integration/components/droplist.js
```diff
@@ -165,4 +165,58 @@ context('Droplist', function() {
       .get('.picklist')
       .should('not.exist');
   });
+
+  specify('isCheckable', function() {
+    let selectlist;
+    const Droplist = this.Droplist;
+
+    cy
+      .getHook($hook => {
```
paulfalgout, line 174: Unrelated to this PR, but every time I see this I wonder when/if we should devise a Marionette/Cypress component testing setup.  In theory we could do it.. almost just like this.. but with their new component API.  Unfortunately pretty much all of it is documented for react/vue

paulfalgout (approved):
Love it

test/integration/components/droplist.js
nmajor25: Ya, something worth looking into!

PR #807, opened 2022-09-27
Use NIL_UUID to check if a clinician has a team

nmajor25:
Shortcut Story ID: [sc-30997]

When a clinician is created, it's given a `nil_uuid` team `id` by default (i.e. `id: '00000000-0000-0000-0000-000000000000'`).

To determine if a clinician has been put on a team, we were simply checking that a clinician has a team assigned (i.e. that it's not `null`). Not what that assigned team is. Which means in some cases the app wrongly thinks a clinician has a team, even though their assigned team is the default `nil_uuid` instead of an actual team.

This PR updates the app to check whether a clinician's assigned team is a `nil_uuid` or not.

_This caused problems in the app where the clinicians `active` state was being determined. Clinicians with a `nil_uuid` team `id` were showing as `active` they shouldn't have._

### DIFF
--- a/src/js/entities-service/entities/clinicians.js
+++ b/src/js/entities-service/entities/clinicians.js
@@ -36,14 +36,18 @@ const _Model = BaseModel.extend({
     return Radio.request('entities', 'groups:collection', this.get('_groups'));
   },
   getTeam() {
-    const team = this.get('_team');
-    if (!team || team === NIL_UUID) {
+    if (!this.hasTeam()) {
       return Radio.request('entities', 'teams:model', {
         name: intl.patients.sidebar.action.activityViews.systemTeam,
       });
     }
 
     return Radio.request('entities', 'teams:model', this.get('_team'));
+  },
+  hasTeam() {
+    const team = this.get('_team');
+
+    return team && team !== NIL_UUID;
   },
   can(prop) {
     const permissions = this.get('permissions');
@@ -70,13 +74,12 @@ const _Model = BaseModel.extend({
     return !this.get('last_active_at');
   },
   isActive() {
-    const hasTeam = !!this.get('_team');
+    const hasTeam = this.hasTeam();
     const hasGroups = !!size(this.get('_groups'));
     const lastActive = this.get('last_active_at');
 
     return hasTeam && hasGroups && lastActive;
   },
-
 });
 
 const Model = Store(_Model, TYPE);
--- a/test/integration/globals/default-route.js
+++ b/test/integration/globals/default-route.js
@@ -68,15 +68,11 @@ context('patient page', function() {
 
   specify('current clinician has no team', function() {
     cy
-      .routeGroupsBootstrap(_.identity, null, fx => {
-        const currentClinician = _.find(fx.data, clinician => clinician.id === '11111');
-
-        currentClinician.relationships.team = { data: null };
-
-        return fx;
-      })
       .routeCurrentClinician(fx => {
-        fx.data.relationships.team = { data: null };
+        fx.data.relationships.team = {
+          data: { id: '00000000-0000-0000-0000-000000000000' },
+        };
+
         return fx;
       })
       .visit('/');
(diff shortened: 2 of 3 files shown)


src/js/entities-service/entities/clinicians.js
```diff
@@ -70,13 +69,17 @@ const _Model = BaseModel.extend({
...
+    const hasTeam = this.hasTeam();
     const hasGroups = !!size(this.get('_groups'));
     const lastActive = this.get('last_active_at');
 
     return hasTeam && hasGroups && lastActive;
   },
+  hasTeam() {
+    const team = this.get('_team');
 
+    return team && team !== NIL_UUID;
```
nmajor25: Could this just be `return team !== NIL_UUID`? Since a clinician will always have a team assigned (_right?_), even if it's just the `NIL_UUID`.

But this covers any potential scenario where that doesn't happen and `team` is `null` for a clinician.

test/integration/globals/default-route.js
```diff
@@ -68,15 +68,11 @@ context('patient page', function() {
 
   specify('current clinician has no team', function() {
     cy
-      .routeGroupsBootstrap(_.identity, null, fx => {
-        const currentClinician = _.find(fx.data, clinician => clinician.id === '11111');
-
-        currentClinician.relationships.team = { data: null };
-
-        return fx;
-      })
```
nmajor25, line 77: I think `routeGroupsBootstrap()` and `routeCurrentClinician()` are doing the same thing?

Test passes with `routeGroupsBootstrap()` removed.

nmajor25:
@paulfalgout based on my understanding of the bug, this should fix it.

Also, feel free to update the PR description if any of that is inaccurate.

nmajor25:
@paulfalgout seems to be a coverage drop related to [outreach](https://coveralls.io/builds/52807674/source?filename=src%2Fjs%2Foutreach%2Findex.js). 

Not entirely sure if it's legit or a coveralls hiccup.

src/js/entities-service/entities/clinicians.js
shadowhand: I think it should be kept like this, in case we were to switch to `id: null` again.

test/integration/globals/default-route.js
paulfalgout: Ah that probably is true now

src/js/entities-service/entities/clinicians.js
paulfalgout: Typically I would not want to support both cases explicitly as we'd be testing for both branches and we'd have no indication of dead code.  However in this case I suspect removing the `team &&` will be more headaches as I suspect we're not setting it across all of the fixtures, but removing the `team &&` would probably surface where those are..

I think it we want to handle both `null` and `NIL_UUID` in most cases we should think through a generic `hasRelationship` solution for the base model, which probably isn't a bad idea anyhow as we have a number of places where we have unique checks for relationships in the app.  The trick will be differentiating between 1-1 and 1-many.  

Anyhow.. I think this is ok for this bugfix PR.  Can we move this method up by the `getTeam` though?

paulfalgout (changes requested):
tiny nitpick

paulfalgout:
that outreach coverage issue is probably a case that's running in the tests but isn't explicitly tested for.. so if the tests run too fast those line coverages aren't recorded.  I wanna say we did remove the outreach form failure testing because it was super flaky... but I might be misremembering.

src/js/entities-service/entities/clinicians.js
nmajor25: That makes sense.

Ya, I originally had it as just `return team !== NIL_UUID;` but that caused a test to fail since team is set to `null` in the tests. So it was a decision between adding the `nil_uuid` in the tests or doing `return team && team !== NIL_UUID;` in the app.

I moved the `hasTeam()` method up below the `getTeam()` method in that last commit I just pushed.

paulfalgout (approved):

PR #808, opened 2022-09-29
Remove form print button

nmajor25:
Shortcut Story ID: [sc-30764]

Remove print button from the forms page. Along with any related icons, `.scss`, and Cypress tests code.

Before removal:

Screen Shot 2022-09-29 at 11 27 25 AM

After removal:

Screen Shot 2022-09-29 at 11 27 55 AM

### DIFF
--- a/src/js/components/picklist/index.js
+++ b/src/js/components/picklist/index.js
@@ -58,7 +58,7 @@ const PicklistEmpty = View.extend({
 const PicklistItem = View.extend({
   tagName: 'li',
   itemTemplate: hbs`
-    <div class="flex-grow">{{#if icon}}{{fa icon.type icon.icon classes=icon.classes}}{{/if}}<span>{{matchText text query}}</span></div
+    <div class="flex-grow">{{#if icon}}{{fa icon.type icon.icon classes=icon.classes}}{{/if}}<span>{{matchText text query}}</span></div>
     <div>{{#if isChecked}}{{fas "circle-check" classes="u-icon--12 u-margin--l-16"}}{{/if}}</div>
   `,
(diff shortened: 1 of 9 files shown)


src/js/components/picklist/index.js
```diff
@@ -58,7 +58,7 @@ const PicklistEmpty = View.extend({
 const PicklistItem = View.extend({
   tagName: 'li',
   itemTemplate: hbs`
-    <div class="flex-grow">{{#if icon}}{{fa icon.type icon.icon classes=icon.classes}}{{/if}}<span>{{matchText text query}}</span></div
+    <div class="flex-grow">{{#if icon}}{{fa icon.type icon.icon classes=icon.classes}}{{/if}}<span>{{matchText text query}}</span></div>
```
nmajor25, line 61: Not related to removing the print button. But snuck it in on this PR.

Just a missing closing tag from the [Save + Go Back PR](https://github.com/RoundingWell/care-ops-frontend/pull/806).

src/scss/formapp/print.scss
```diff
@@ -1,10 +0,0 @@
-@media print {
```
paulfalgout, line 1: This needs to stay.  It's used by the PDF generator which prints.  Could maybe add a comment reflecting that.

paulfalgout (changes requested):

src/scss/formapp/print.scss
nmajor25: Oops, good call!

Added that file back with the comment.

paulfalgout (approved):

PR #809, opened 2022-09-29
Change directories to use entity

paulfalgout:
The API here is both singular and plural, but the type is plural, so we should rename it here as well.

The API also uses the slug over the id for reading.  As these are currently read-only there’s no current issue.

This PR should also temporarily drop the coverage until the epic is completed.

Shortcut Story ID: [sc-31229]

We don't need to necessarily merge this right away if we want to keep the coverage up.  On approval this can be pulled into the PR for 31235


src/js/entities-service/entities/directories.js
```diff
@@ -0,0 +1,21 @@
+import BaseCollection from 'js/base/collection';
+import BaseModel from 'js/base/model';
+
+const TYPE = 'directories';
+
+const Model = BaseModel.extend({
+  type: TYPE,
+  url() {
+    return `/api/directory/${ this.get('slug') }`;
```
shadowhand, line 9: Oh ffs 💢 we should fix this.

paulfalgout: Well at the moment we do need to be able to request via the slug for form definitions, but it could certainly be a `filter[slug]=` or the like

shadowhand: No, I mean that we should use `/api/directories/{slug}`... or maybe not, since this is a unique case that uses `slug` instead of `uuid`.

paulfalgout: I'm ok either way at this point.   it basically just gets set here and forgotten

nmajor25 (approved):

PR #810, opened 2022-09-30
Trigger keydown user activity events from the form app

nmajor25:
Shortcut Story ID: [sc-31196]

Trigger a `pointerdown` event from the formapp so pop regions close when a user clicks in a form.

An example of the `Save + Go Back` button droplist closing when a user clicks into the form:

https://user-images.githubusercontent.com/35355575/193350089-c93d35fe-7d13-4acd-a9ba-229cc08ce683.mov

### DIFF
--- a/src/js/formapp.js
+++ b/src/js/formapp.js
@@ -172,6 +172,12 @@ const Router = Backbone.Router.extend({
       this.trigger(data.message, data.args);
     }, false);
 
+    const self = this;
+
+    $(document).on('pointerdown', function(evt) {
+      self.request('user-activity', { eventType: 'body:down' });
+    });
+
     this.request('version', versions.frontend);
   },
   request(message, args = {}) {
--- a/src/js/behaviors/top-region.js
+++ b/src/js/behaviors/top-region.js
@@ -29,6 +29,11 @@ export default Behavior.extend({
     this.$el.removeClass(this.className);
   },
   onUserActivity({ target }) {
+    if (!target) {
+      this.region.empty();
+      return;
+    }
+
     if (!this.region.hasView() || topRegionCh.request('contains', this.view, target)) return;
 
     this.region.empty();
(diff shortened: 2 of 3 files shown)


nmajor25:
This PR makes the `Save + Go Back` button droplist close whenever a user clicks or tabs into the form app. I included the video of that in the PR description.

But it breaks forms that are in a modal. Since it forces the modal to close when the user clicks into it. Which is what the failed Cypress test is about.

Not sure how to proceed with it. So figured I'd create a draft PR so we can discuss.

src/js/formapp.js
```diff
@@ -172,6 +172,12 @@ const Router = Backbone.Router.extend({
       this.trigger(data.message, data.args);
     }, false);
 
+    const self = this;
+
+    $(document).on('pointerdown', function(evt) {
+      self.request('user-activity', { eventType: 'body:down' });
+    });
```
nmajor25, line 179: I attempted to do this here: 

```js
$(document).on('pointerdown', function(evt) {
    self.request('user-activity', { eventType: 'body:down', evt });
});
```

But was getting this error:

```js
formapp.js:249 Uncaught (in promise) DOMException: Failed to execute 'postMessage' on 'Window': PointerEvent object could not be cloned.
```

And couldn't figure out why it was happening or how to fix it.

paulfalgout: you can only send serializeable things over postMessage... so things that can become text and get parsed on the other.. pure json objects, but nothing more complicated or that has a reference to window or something.

paulfalgout:
Good findings.  Glad the test failed.  I think focus on the iframe window is probably more what we're looking for now and it should work within the `contains` tests of the top regions.  Otherwise the modal will be trouble.  Check out this commit.  We can pull off this branch if useful.  I _think_ this covers the use cases..  testing it.. I'm not certain of the best place to test this, but I assume it'll be missing coverage without new tests somewhere. 

Thoughts?

https://github.com/RoundingWell/care-ops-frontend/commit/5e47044b4a2587bf1984fe4a8997e8d6b3cec2dc

PR #815, opened 2022-10-07
Move clinician filter to header for worklist & schedule

nmajor25:
Shortcut Story ID: [sc-31233]

Move the clinician filter to the header for all worklist and schedule pages.

Screen Shot 2022-10-07 at 9 45 33 AM

### DIFF
--- a/src/js/apps/patients/schedule/schedule_app.js
+++ b/src/js/apps/patients/schedule/schedule_app.js
@@ -13,6 +13,7 @@ import BulkEditActionsApp from 'js/apps/patients/sidebar/bulk-edit-actions_app';
 
 import DateFilterComponent from 'js/views/patients/shared/components/date-filter';
 import SearchComponent from 'js/views/shared/components/list-search';
+import OwnerDroplist from 'js/views/patients/shared/components/owner_component';
 
 import { LayoutView, ScheduleTitleView, TableHeaderView, SelectAllView, ScheduleListView } from 'js/views/patients/schedule/schedule_views';
 import { BulkEditButtonView, BulkEditActionsSuccessTemplate, BulkDeleteActionsSuccessTemplate } from 'js/views/patients/shared/bulk-edit/bulk-edit_views';
@@ -52,6 +53,11 @@ export default App.extend({
     }
 
     this.initListState();
+
+    const currentClinician = Radio.request('bootstrap', 'currentUser');
+    this.canViewAssignedActions = currentClinician.can('app:schedule:clinician_filter');
+    this.groups = currentClinician.getGroups();
+
     this.showView(new LayoutView({
       state: this.getState(),
     }));
@@ -145,9 +151,32 @@ export default App.extend({
     const filters = this.getState().get('filters');
     const owner = Radio.request('entities', 'clinicians:model', filters.clinicianId);
 
-    this.showChildView('title', new ScheduleTitleView({
+    const scheduleTitleView = this.showChildView('title', new ScheduleTitleView({
       model: owner,
+      showOwnerDroplist: this.canViewAssignedActions,
     }));
+
+    if (this.canViewAssignedActions) {
+      const ownerDroplistView = scheduleTitleView.showChildView('owner', new OwnerDroplist(this.getOwnerFilterOptions(owner)));
+
+      this.listenTo(ownerDroplistView, {
+        'change:owner'({ id }) {
+          this.setState({ filters: { ...filters, clinicianId: id } });
+        },
+      });
+    }
+  },
+  getOwnerFilterOptions(owner) {
+    const options = {
+      owner: owner,
+      groups: this.groups,
+      isTitleFilter: true,
+      headingText: intl.patients.schedule.filtersApp.ownerFilterHeadingText,
+      hasTeams: false,
+      align: 'right',
+    };
+
+    return options;
   },
   showSearchView() {
     const searchComponent = this.showChildView('search', new SearchComponent({
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -17,9 +17,12 @@ import BulkEditFlowsApp from 'js/apps/patients/sidebar/bulk-edit-flows_app';
 
 import DateFilterComponent from 'js/views/patients/shared/components/date-filter';
 import SearchComponent from 'js/views/shared/components/list-search';
+import OwnerDroplist from 'js/views/patients/shared/components/owner_component';
 
 import { ListView, SelectAllView, LayoutView, ListTitleView, TableHeaderView, SortDroplist, TypeToggleView, sortCreatedOptions, sortDueOptions, sortPatientOptions, sortUpdateOptions } from 'js/views/patients/worklist/worklist_views';
 import { BulkEditButtonView, BulkEditFlowsSuccessTemplate, BulkEditActionsSuccessTemplate, BulkDeleteFlowsSuccessTemplate, BulkDeleteActionsSuccessTemplate } from 'js/views/patients/shared/bulk-edit/bulk-edit_views';
+
+const i18n = intl.patients.worklist.filtersApp;
 
 export default App.extend({
   StateModel,
@@ -75,6 +78,12 @@ export default App.extend({
     this.worklistId = worklistId;
     this.initListState();
 
+    const currentClinician = Radio.request('bootstrap', 'currentUser');
+    this.shouldShowClinician = this.getState().id !== 'shared-by';
+    this.shouldShowTeam = this.getState().id !== 'owned-by';
+    this.canViewAssignedActions = currentClinician.can('app:worklist:clinician_filter');
+    this.groups = currentClinician.getGroups();
+
     this.showView(new LayoutView({
       worklistId: this.worklistId,
       state: this.getState(),
@@ -122,9 +131,7 @@ export default App.extend({
   startFiltersApp() {
     const filtersApp = this.startChildApp('filters', {
       state: this.getState().getFilters(),
-      shouldShowClinician: this.getState().id !== 'shared-by',
       shouldShowOwnerToggle: this.getState().id === 'shared-by',
-      shouldShowTeam: this.getState().id !== 'owned-by',
     });
 
     this.listenTo(filtersApp.getState(), 'change', ({ attributes }) => {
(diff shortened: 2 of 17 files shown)


src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -282,12 +289,43 @@ export default App.extend({
     const owner = Radio.request('entities', 'clinicians:model', filters.clinicianId);
     const team = Radio.request('entities', 'teams:model', filters.teamId);
 
-    this.showChildView('title', new ListTitleView({
+    const listTitleView = this.showChildView('title', new ListTitleView({
       owner,
       team,
       worklistId: this.worklistId,
       isFlowList: this.getState().isFlowType(),
+      showOwnerDroplist: this.canViewAssignedActions,
```
nmajor25: The clinician filter is only shown if the user has the `'app:worklist:clinician_filter'` permission.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -32,20 +32,40 @@ const LayoutView = View.extend({
...
+    owner: '[data-owner-filter-region]',
+  },
+  className: 'flex list-page__owner',
   template: hbs`
-    <span class="list-page__title-icon">{{far "calendar-star"}}</span>{{formatMessage (intlGet "patients.schedule.scheduleViews.scheduleTitleView.title") owner=name}}{{~ remove_whitespace ~}}
-    <span class="list-page__header-icon js-title-info">{{fas "circle-info"}}</span>
+    <span class="list-page__title-icon">{{far "calendar-star"}}</span>
+    {{#if showOwnerDroplist}}
+      <div class="list-page__owner-name-label">
+        Schedule for
```
nmajor25: Just realized I should move this to the `/js/i18n/en-US.yml` file.

src/js/views/patients/shared/components/owner_component.js
```diff
@@ -13,7 +13,7 @@ import './owner-component.scss';
 const i18n = intl.patients.shared.components.ownerComponent;
 
 const OwnerItemTemplate = hbs`<div>{{matchText name query}} <span class="owner-component__team">{{matchText short query}}</span></div>`;
-const FilterButtonTemplate = hbs`{{far "circle-user"}}<span>{{ name }}</span>{{far "angle-down"}}`;
+const FilterButtonTemplate = hbs`{{far "circle-user"}}<span class="owner-component__name">{{ name }}</span>{{far "angle-down"}}`;
```
nmajor25: Added a class name here so I can target it for the ellipsis overflow CSS styling.

src/scss/modules/list-pages.scss
```diff
@@ -15,18 +15,64 @@
...
+    line-height: normal;
+    margin-left: 6px;
+    overflow: visible;
+    padding: 0;
+
+    .owner-component__name {
+      max-width: 268px;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
```
nmajor25: This is the code that handles the ellipsis overflow for long clinician/team names.

With that `max-width` set, it would support a max of `21 characters`. If the name is greater than that length, the `...` ellipsis is shown.

test/integration/patients/worklist/worklist.js
```diff
@@ -1880,7 +1880,7 @@ context('worklist page', function() {
 
     cy
       .get('[data-owner-filter-region]')
-      .find('button');
+      .should('not.exist');
```
nmajor25: I was a little confused by this test...

If a restricted employee is viewing the `Shared By` worklist, the clinician filter should not be shown, correct?

nmajor25:
Looks like the coverage drop is from the `/src/js/entities-service/directories.js` file: https://coveralls.io/builds/53098252/source?filename=src%2Fjs%2Fentities-service%2Fdirectories.js

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -52,6 +53,11 @@ export default App.extend({
     }
 
     this.initListState();
+
+    const currentClinician = Radio.request('bootstrap', 'currentUser');
+    this.canViewAssignedActions = currentClinician.can('app:worklist:clinician_filter');
```
paulfalgout: `app:schedule:clinician_filter`

test/integration/patients/worklist/worklist.js
paulfalgout: it was just teams.  nothing assigned to anyone else.  The test should be better.

A non-restricted employee, if they picked nurse would see actions assigned to the nurse team or the nurse clinicians.. and so the no owner button filtered out actions assigned to nurse clinicians.. a restricted employee cannot see actions assigned to a nurse clinician so they cannot see the no owner filter, but they can see actions only assigned to whatever team they pick

src/scss/modules/list-pages.scss
```diff
@@ -15,18 +15,64 @@
...
   margin-right: 16px;
-  vertical-align: bottom;
 
   .icon {
     font-size: 20px;
+    vertical-align: middle;
+  }
+}
+
+.list-page__owner-name-label {
```
paulfalgout: I think `owner` is too specific to the domain.   The modules should be super generic such that if this was listing `foo`s and filtered by `bar`s neither `foo` nor `bar` is in the style naming.  Maybe `list-page__title-filter...`?

src/scss/modules/list-pages.scss
paulfalgout: this could be in the owner component's css.  doesn't seem to need to be nested

src/scss/modules/list-pages.scss
```diff
@@ -15,18 +15,64 @@
...
+  white-space: nowrap;
+}
+
+.list-page__owner-name-filter {
+  button {
```
paulfalgout: does `button` need to be specified?  ideally we're class only

src/scss/modules/list-pages.scss
paulfalgout: FWIW on this particular class there is a `u-text--nowrap`

src/js/views/patients/schedule/schedule_views.js
paulfalgout: :+1:

paulfalgout (changes requested):
I'll look more at this next week, but a few things in a quick scan

src/js/views/patients/schedule/schedule_views.js
nmajor25: Updated in last commit push.

src/scss/modules/list-pages.scss
nmajor25: Good call, updated in the last commit push.

src/js/apps/patients/schedule/schedule_app.js
nmajor25: Updated in the last commit push.

src/scss/modules/list-pages.scss
nmajor25: Got rid of `.list-page__owner-name-label` and replaced that class name with `u-text--nowrap`.

And changed `.list-page__owner-name-filter` to `.list-page__title-filter`.

src/scss/modules/list-pages.scss
nmajor25: We could target `.button-filter` instead? That button is given that class name if it has `isFilter = true` ([here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/views/patients/shared/components/owner_component.js#L71)). Which in this case it does.

Updated to that in the last commit push.

src/js/views/patients/shared/components/owner_component.js
paulfalgout: Can we add a new flag for the owner component similar to `isCompact` and `isFilter` for this use of the component?  This way we can specifically set this overflow and remove the icon with a separate template and we don't have to nest `button-filter` and override the `button-filter` styles.

paulfalgout (changes requested):

test/integration/patients/worklist/worklist.js
nmajor25: Just pushed some changes that I believe fixes this: https://github.com/RoundingWell/care-ops-frontend/compare/2c05b32d715c299f7ed38f793692d21e513887d5..c442143343c8690d5b0372703b83af1d1d3c0bf7

nmajor25:
Just noticed that something is off with the width of the droplist/picklist. So need to investigate that.

Screen Shot 2022-10-10 at 3 49 21 PM

paulfalgout:
@nmajor25 in the `OwnerItemTemplate` there's a wrapping `<div>` for removing the flex from those elements.  I think if we added `.u-text--nowrap` to that it'd fix this last bit.

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -145,9 +151,31 @@ export default App.extend({
...
+
+      this.listenTo(ownerDroplistView, {
+        'change:owner'({ id }) {
+          this.setState({ filters: { ...filters, clinicianId: id } });
+        },
+      });
+    }
+  },
+  getOwnerFilterOptions(owner) {
+    const options = {
```
paulfalgout, line 170: I wonder if this filter would be better as `align: 'right'`... _probably_?

src/js/views/patients/shared/components/owner_component.js
```diff
@@ -43,8 +44,9 @@ export default Droplist.extend({
   popWidth() {
     const isFilter = this.getOption('isFilter');
     const isCompact = this.getOption('isCompact');
+    const isTitleFilter = this.getOption('isTitleFilter');
 
-    return (isFilter || isCompact) ? null : this.getView().$el.outerWidth();
+    return (isFilter || isCompact || isTitleFilter) ? null : this.getView().$el.outerWidth();
```
nmajor25: This seems to have fixed the picklist width issue.

src/js/apps/patients/schedule/schedule_app.js
nmajor25: Added the `align: 'right'` in my most recent changes I just pushed.

src/scss/modules/list-pages.scss
```diff
@@ -15,18 +15,24 @@
...
 
 .list-page__title {
+  align-items: center;
   border-bottom: 1px solid $black-90;
   font-size: 24px;
   margin-bottom: 16px;
   padding-bottom: 16px;
 }
 
+.list-page__owner {
```
paulfalgout: this is another style in a module.. I think "owner" is too domain for this.  needs to be made more generic

paulfalgout (changes requested):
Just one small request I think

src/scss/modules/list-pages.scss
nmajor25: Hmm ya I agree. Not sure what to name it though... 🤔

paulfalgout: maybe `list-page__title-filter`?  I know it's not always a filter though..

nmajor25: Works for me 😄 

Updated that in what I just pushed.

src/scss/modules/list-pages.scss
```diff
@@ -15,18 +15,24 @@
 }
 
 .list-page__title {
```
nmajor25, line 17: Just noticed this.

Adding `align-items: center` here messes with title alignments on other pages where `list-page__title` is used.

For example:

Screen Shot 2022-10-11 at 2 17 36 PM

nmajor25: Fixed this in the latest pushed changes.

paulfalgout:
https://coveralls.io/builds/53187924/source?filename=src%2Fjs%2Fviews%2Fpatients%2Fshared%2Fcomponents%2Fowner_component.js#L73

So I'm glad we switched to the `isTitleFilter` as it pointed out that this isn't going to be used as filter anymore.. I'm ok with keeping it `isTitleFilter` or just using `isFilter`, but we don't need to support the old filter template/classes at all.

PR #817, opened 2022-10-13
Add clinician new modal

paulfalgout:
Also change role from a named attribute to a relationship

This turned out to be quite a bit more work than initially expected

Shortcut Story ID: [sc-31295]

### DIFF
--- a/src/js/views/clinicians/clinician-modal/clinician-modal_views.js
+++ b/src/js/views/clinicians/clinician-modal/clinician-modal_views.js
@@ -1,0 +1,159 @@
+import { extend } from 'underscore';
+import hbs from 'handlebars-inline-precompile';
+import { View } from 'marionette';
+import { mixinState } from 'marionette.toolkit';
+
+import 'scss/modules/modals.scss';
+
+import intl from 'js/i18n';
+import trim from 'js/utils/formatting/trim';
+
+import { GroupsComponent, TeamComponent, RoleComponent } from 'js/views/clinicians/shared/clinicians_views';
+
+import InputFocusBehavior from 'js/behaviors/input-focus';
+import ClinicianModalTemplate from './clinician-modal.hbs';
+
+const i18n = intl.clinicians.clinicianModal.clinicianModalViews;
+
+const InputView = View.extend({
+  behaviors() {
+    if (!this.getOption('shouldFocus')) return;
+
+    return [
+      {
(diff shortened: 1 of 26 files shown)


src/js/views/clinicians/clinician-modal/clinician-modal_views.js
```diff
@@ -0,0 +1,159 @@
+import { extend } from 'underscore';
```
paulfalgout, line 1: FWIW this is very very similar to the add patient modal views

src/js/views/clinicians/shared/clinicians_views.js
```diff
@@ -16,6 +16,12 @@ const i18n = intl.clinicians.shared.clinicianViews;
 
 const GroupsComponent = GroupsManagerComponent.extend({
   removeMemberGroup(group) {
+    if (this.member.isNew()) {
```
paulfalgout, line 19: Previously this component wasn't available to a new clinician, but we don't want the warning below in that situation

src/js/views/globals/root_views.js
```diff
@@ -306,10 +306,10 @@ const RootView = CollectionView.extend({
     // Add lowest layer (z-index) to highest
     this.addChildView(this.appView);
     this.addRegionView('tooltip', new TooltipRegionView({ $body }));
-    this.addRegionView('alert', new TopRegionView());
     this.addRegionView('modalSidebar', new ModalSidebarRegionView({ $body }));
     this.addRegionView('modal', new ModalRegionView({ $body }));
     this.addRegionView('modalSmall', new ModalRegionView({ $body }));
+    this.addRegionView('alert', new TopRegionView());
```
paulfalgout, line 312: alerts were rendering below the modal and darkening.. I can't think of a reason why that should be true.

test/integration/clinicians/clinician-modal.js
```diff
@@ -0,0 +1,217 @@
...
+    name: 'Group One',
+  },
+  {
+    id: '2',
+    name: 'Group Two',
+  },
+];
+
+context('clinicians modal', function() {
+  specify('add clinician', function() {
```
paulfalgout, line 19: this was adapted from the similar tests when we were adding them from the sidebar

nmajor25 (approved):

PR #819, opened 2022-10-13
Create all filters sidebar for worklist and schedule

nmajor25:
Shortcut Story ID: [sc-31234]

Add an "All Filters" link to the worklist and schedule pages that opens a sidebar with the groups filter shown.

### DIFF
--- a/src/js/apps/patients/schedule/filters_app.js
+++ b/src/js/apps/patients/schedule/filters_app.js
@@ -4,9 +4,17 @@ import App from 'js/base/app';
 
 import intl from 'js/i18n';
 
-import { FiltersView, GroupsDropList } from 'js/views/patients/schedule/filters_views';
+import AllFiltersApp from 'js/apps/patients/sidebar/filters-sidebar_app';
+
+import { FiltersView, AllFiltersButtonView, GroupsDropList } from 'js/views/patients/schedule/filters_views';
 
 export default App.extend({
+  childApps: {
+    allFilters: AllFiltersApp,
+  },
+  stateEvents: {
+    'change:groupId': 'showGroupsFilterView',
+  },
   onStart() {
     const currentClinician = Radio.request('bootstrap', 'currentUser');
     this.groups = currentClinician.getGroups();
@@ -15,7 +23,38 @@ export default App.extend({
     this.showFilters();
   },
   showFilters() {
+    this.showAllFiltersButtonView();
     this.showGroupsFilterView();
+  },
+  showAllFiltersButtonView() {
+    if (this.groups.length < 2) return;
+
+    const ownerView = this.showChildView('allFilters', new AllFiltersButtonView());
+
+    this.listenTo(ownerView, 'click', this.showGroupSidebar);
+  },
+  showGroupSidebar() {
+    this.trigger('toggle:filtersSidebar', true);
+
+    const state = this.getState();
+
+    const sidebar = Radio.request('sidebar', 'start', 'filters', {
+      state: state.attributes,
+    });
+
+    const filterState = sidebar.getState();
+
+    this.listenTo(filterState, 'change', () => {
+      this.setState(filterState.attributes);
+    });
+
+    sidebar.listenTo(this.getState(), 'change', () => {
+      sidebar.setState(state.attributes);
+    });
+
+    this.listenTo(sidebar, 'stop', () => {
+      this.trigger('toggle:filtersSidebar', false);
+    });
   },
   showGroupsFilterView() {
     if (this.groups.length < 2) return;
--- a/src/js/apps/patients/sidebar/filters-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/filters-sidebar_app.js
@@ -1,0 +1,46 @@
+import Radio from 'backbone.radio';
+
+import App from 'js/base/app';
+
+import { LayoutView, GroupsFilterView } from 'js/views/patients/sidebar/filters/filters-sidebar_views';
+
+export default App.extend({
+  stateEvents: {
+    'change': 'onChangeFilters',
+  },
+  onStart({ state }) {
+    this.currentClinician = Radio.request('bootstrap', 'currentUser');
+
+    this.showView(new LayoutView());
+
+    this.showGroupsFilterView();
+  },
+  viewEvents: {
+    'close': 'stop',
+    'click:clear:filters': 'onClearFilters',
+  },
+  onChangeFilters() {
+    this.showGroupsFilterView();
+  },
+  showGroupsFilterView() {
+    const groups = this.currentClinician.getGroups();
+
+    const groupsFilterView = new GroupsFilterView({
+      groups,
+      groupId: this.getState('groupId'),
+    });
+
+    this.listenTo(groupsFilterView, 'select:group', this.onSelectGroup);
+
+    this.showChildView('groups', groupsFilterView);
+  },
+  onClearFilters() {
+    this.setState('groupId', null);
+  },
+  onSelectGroup(id) {
+    this.setState('groupId', id);
+  },
+  onStop() {
+    Radio.request('sidebar', 'close');
+  },
+});
--- a/src/js/apps/patients/worklist/filters_app.js
+++ b/src/js/apps/patients/worklist/filters_app.js
@@ -4,11 +4,19 @@ import intl from 'js/i18n';
 
 import App from 'js/base/app';
 
-import { FiltersView, GroupsDropList, NoOwnerToggleView } from 'js/views/patients/worklist/filters_views';
+import AllFiltersApp from 'js/apps/patients/sidebar/filters-sidebar_app';
+
+import { FiltersView, GroupsDropList, NoOwnerToggleView, AllFiltersButtonView } from 'js/views/patients/worklist/filters_views';
 
 const i18n = intl.patients.worklist.filtersApp;
 
 export default App.extend({
+  childApps: {
+    allFilters: AllFiltersApp,
+  },
+  stateEvents: {
+    'change:groupId': 'showGroupsFilterView',
+  },
   onStart({ shouldShowOwnerToggle }) {
     const currentClinician = Radio.request('bootstrap', 'currentUser');
     this.canViewAssignedActions = currentClinician.can('app:worklist:clinician_filter');
@@ -19,8 +27,39 @@ export default App.extend({
     this.showFilters();
   },
   showFilters() {
+    this.showAllFiltersButtonView();
     this.showGroupsFilterView();
     this.showOwnerToggleView();
+  },
+  showAllFiltersButtonView() {
+    if (this.groups.length < 2) return;
+
+    const ownerView = this.showChildView('allFilters', new AllFiltersButtonView());
+
+    this.listenTo(ownerView, 'click', this.showGroupSidebar);
+  },
+  showGroupSidebar() {
+    this.trigger('toggle:filtersSidebar', true);
+
+    const state = this.getState();
+
+    const sidebar = Radio.request('sidebar', 'start', 'filters', {
+      state: state.attributes,
+    });
+
+    const filterState = sidebar.getState();
+
+    this.listenTo(filterState, 'change', () => {
+      this.setState(filterState.attributes);
+    });
+
+    sidebar.listenTo(this.getState(), 'change', () => {
+      sidebar.setState(state.attributes);
+    });
+
+    this.listenTo(sidebar, 'stop', () => {
+      this.trigger('toggle:filtersSidebar', false);
+    });
   },
   showGroupsFilterView() {
     if (this.groups.length < 2) return;
--- a/src/js/views/patients/sidebar/filters/filters-sidebar_views.js
+++ b/src/js/views/patients/sidebar/filters/filters-sidebar_views.js
@@ -1,0 +1,88 @@
+import { View } from 'marionette';
+import hbs from 'handlebars-inline-precompile';
+
+import 'scss/modules/buttons.scss';
+import 'scss/modules/sidebar.scss';
+
+import { animSidebar } from 'js/anim';
+
+import Droplist from 'js/components/droplist';
+
+import FiltersSidebarTemplate from './filters-sidebar.hbs';
+
+import './filters-sidebar.scss';
+
+const SidebarGroupsDropList = Droplist.extend({
+  popWidth() {
+    return this.getView().$el.outerWidth();
+  },
+  viewOptions: {
+    className: 'button-secondary w-100',
+    template: hbs`{{ name }}`,
+  },
+  picklistOptions: {
+    attr: 'name',
+  },
+});
+
+const GroupsFilterView = View.extend({
+  className: 'flex flex-align-center',
+  template: hbs`
+    <h4 class="sidebar__label">{{ @intl.patients.sidebar.filters.groups.droplistLabel }}</h4>
+    <div class="flex-grow" data-filter-button></div>
+  `,
+  regions: {
+    filterButton: '[data-filter-button]',
+  },
+  onRender() {
+    const groups = this.getGroups();
+    const selected = groups.get(this.getOption('groupId')) || groups.at(0);
+
+    const groupsFilter = new SidebarGroupsDropList({
+      collection: groups,
+      state: { selected },
+    });
+
+    this.listenTo(groupsFilter.getState(), 'change:selected', (state, { id }) => {
+      this.triggerMethod('select:group', id);
+    });
+
+    this.showChildView('filterButton', groupsFilter);
+  },
+  getGroups() {
+    const groups = this.getOption('groups').clone();
+
+    groups.unshift({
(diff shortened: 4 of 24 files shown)


src/js/apps/patients/sidebar/filters-sidebar_app.js
```diff
@@ -0,0 +1,29 @@
...
+    }));
+  },
+  viewEvents: {
+    'close': 'stop',
+    'select:new:group': 'onSelectNewGroup',
+  },
+  onSelectNewGroup(id) {
+    this.setState('groupId', id);
+
+    Radio.request('sidebar', 'close');
```
paulfalgout: I don't actually think the sidebar will close when the filter is changed as in the future you can set multiple filters.  they'll have to manually close it _I think_

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -19,9 +24,19 @@ export default App.extend({
...
   showFilters() {
+    this.showAllFiltersButtonView();
     this.showGroupsFilterView();
     this.showOwnerToggleView();
   },
+  showAllFiltersButtonView() {
+    const ownerView = this.showChildView('allFilters', new AllFiltersButtonView());
+
+    this.listenTo(ownerView, 'click', () => {
+      Radio.request('sidebar', 'start', 'filters', {
```
paulfalgout: So.. probably this'd solve the issue.
```js
const state = this.getState();
const sidebar = Radio.request('sidebar', 'start', 'filters', {
  state: state.attributes,
});
const filterState = sidebar.getState();
this.listenTo(filterState, 'change', () => {
  this.setState(filterState.attributes);
});
sidebar.listenTo(this.getState(), 'change', () => {
  sidebar.setState(state.attributes);
});
```

paulfalgout (changes requested):
So there's been a handful of cases in the past where we want to share the exact state between two apps.  We haven't really had a perfect solution for this..  We _could_ do this with `backbone.store` as we do with models.. but that's an odd caching strategy.  We could also change toolkit's `initState` to set the app's state model to a passed in state instance.  _but_ I do wonder if this is an anti-pattern.  one that makes sense until some inevitable difference that causes it to be unhelpful

src/js/apps/patients/sidebar/filters-sidebar_app.js
nmajor25: Ah okay, yup makes sense!

src/js/apps/patients/worklist/filters_app.js
nmajor25: Sweet, that has got things working. Thanks!

Just need to hammer out the details now.

src/js/apps/patients/sidebar/filters-sidebar_app.js
nmajor25: Whenever the worklist restarts, any open sidebar is closed via this line of code: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/patients/worklist/worklist_app.js#L68.

So we'd need to configure that to only close the bulk edit and patient sidebars, but not the filters sidebar we're adding in this PR?

src/js/apps/patients/sidebar/filters-sidebar_app.js
```diff
@@ -0,0 +1,45 @@
...
+
+    this.showGroupsFilterView(state);
+  },
+  viewEvents: {
+    'close': 'stop',
+    'click:clear:filters': 'onClearFilters',
+    'select:new:group': 'onSelectNewGroup',
+  },
+  onChangeFilters(state) {
+    this.showGroupsFilterView(state);
```
paulfalgout: if this was `state.attributes` would the tertiary be needed in the function?

src/js/views/patients/worklist/filters_views.js
```diff
@@ -10,19 +10,39 @@ import 'js/views/patients/worklist/worklist-list.scss';
...
+  triggers: {
+    click: 'click',
+  },
+});
+
 const GroupsDropList = Droplist.extend({
-  viewOptions: {
-    className: 'button-filter',
-    template: hbs`{{ name }}{{far "angle-down"}}`,
+  popWidth() {
```
paulfalgout: this droplist is simple enough we can probably just make a different one for the sidebar rather than using this twice with a flag

src/scss/core/_flex.scss
```diff
@@ -22,3 +22,7 @@
 .flex-shrink {
   flex-shrink: 1;
 }
+
+.flex-align-center {
```
paulfalgout, line 26: :+1:

src/scss/modules/buttons.scss
```diff
@@ -112,6 +112,7 @@
 
   .icon.fa-angle-down {
     font-size: 12px;
+    margin-left: 8px;
```
paulfalgout, line 115: this makes me nervous.. it's in a lot of places.

src/js/views/patients/worklist/filters_views.js
```diff
@@ -10,19 +10,39 @@ import 'js/views/patients/worklist/worklist-list.scss';
...
   regions: {
+    allFilters: '[data-all-filters-region]',
     group: '[data-group-filter-region]',
     ownerToggle: '[data-owner-toggle-region]',
   },
 });
 
+const AllFiltersButtonView = View.extend({
+  template: hbs`
+    <div>{{far "sliders"}}All Filters</div>
```
paulfalgout: wrapping div may not be needed.. should this be a button?  maybe `tagName: 'button'` on the view?  `className: 'button--link'`

src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -104,12 +104,26 @@
   width: 72px;
 }
 
+.worklist-list__all-filters {
```
paulfalgout: I think this may be `button--link`

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -19,9 +27,32 @@ export default App.extend({
...
     this.showGroupsFilterView();
     this.showOwnerToggleView();
   },
+  showAllFiltersButtonView() {
+    if (this.groups.length < 2) return;
+
+    const ownerView = this.showChildView('allFilters', new AllFiltersButtonView());
+
+    this.listenTo(ownerView, 'click', () => {
+      const state = this.getState();
```
paulfalgout: I think maybe this handler should be `showGroupSidebar`

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -4,11 +4,19 @@ import intl from 'js/i18n';
...
+import { FiltersView, GroupsDropList, NoOwnerToggleView, AllFiltersButtonView } from 'js/views/patients/worklist/filters_views';
 
 const i18n = intl.patients.worklist.filtersApp;
 
 export default App.extend({
+  childApps: {
+    allFilters: AllFiltersApp,
+  },
+  stateEvents: {
+    'change': 'showGroupsFilterView',
```
paulfalgout: this'd be `change:groupId` I think?

paulfalgout (changes requested):
Seems about right

src/js/views/patients/worklist/filters_views.js
nmajor25: Agreed. I was split between which way to do it when I was building things out. I'll split it out into two different ones.

src/scss/modules/buttons.scss
nmajor25: Me too... I'll look into alternatives.

I've also noticed a few buttons where the angle down icon doesn't have the correct margin (screenshot below). Which is related to this as well.

Screen Shot 2022-10-17 at 9 18 11 AM

src/js/views/patients/worklist/filters_views.js
nmajor25: It seems like it could be... will look into that.

src/scss/modules/buttons.scss
paulfalgout: yeah this certainly could be the correct solution

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -19,9 +27,39 @@ export default App.extend({
...
     this.showGroupsFilterView();
     this.showOwnerToggleView();
   },
+  showAllFiltersButtonView() {
+    if (this.groups.length < 2) return;
+
+    const ownerView = this.showChildView('allFilters', new AllFiltersButtonView());
+
+    this.listenTo(ownerView, 'click', () => {
+      this.trigger('toggle:filters:sidebar', true);
```
paulfalgout: I think maybe `toggle:filtersSidebar` might be better.  I _think_ `verb:subject` or `context:verb:subject` is preferable.  I'm not totally convinced, but I think the `:` helps to relevant define parts of the event vs being a word separator.

paulfalgout (changes requested):

src/js/apps/patients/schedule/filters_app.js
```diff
@@ -15,8 +23,41 @@ export default App.extend({
...
   showFilters() {
+    this.AllFiltersButtonView();
     this.showGroupsFilterView();
   },
+  AllFiltersButtonView() {
+    if (this.groups.length < 2) return;
+
+    const ownerView = this.showChildView('allFilters', new AllFiltersButtonView());
+
+    this.listenTo(ownerView, 'click', () => {
```
paulfalgout: can just be `this.listenTo(ownerView, 'click', this.showGroupSidebar);`

src/js/apps/patients/sidebar/filters-sidebar_app.js
nmajor25: No it would not. Good call.

Updated in my latest push.

src/js/views/patients/worklist/filters_views.js
nmajor25: I split this out into a different one in my latest push.

src/js/views/patients/worklist/filters_views.js
nmajor25: Updated to use `button--link` in my last push.

src/js/apps/patients/worklist/filters_app.js
nmajor25: Added a `showGroupSidebar()` handler function in my last push.

src/js/apps/patients/worklist/filters_app.js
nmajor25: Yes, good call. Updated that in my last push.

src/js/apps/patients/worklist/filters_app.js
nmajor25: Makes sense to me. Updated in my last push.

src/js/apps/patients/schedule/filters_app.js
```diff
@@ -15,8 +23,41 @@ export default App.extend({
     this.showFilters();
   },
   showFilters() {
+    this.AllFiltersButtonView();
     this.showGroupsFilterView();
   },
+  AllFiltersButtonView() {
```
paulfalgout: Missing the `show` I think

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
```diff
@@ -0,0 +1,88 @@
...
+import FiltersSidebarTemplate from './filters-sidebar.hbs';
+
+import './filters-sidebar.scss';
+
+const SidebarGroupsDropList = Droplist.extend({
+  popWidth() {
+    return this.getView().$el.outerWidth();
+  },
+  viewOptions: {
+    className: 'button-secondary--compact w-100',
```
paulfalgout: I don't think this should be `--compact`

src/js/views/patients/worklist/filters_views.js
```diff
@@ -10,15 +10,26 @@ import 'js/views/patients/worklist/worklist-list.scss';
...
   `,
   regions: {
+    allFilters: '[data-all-filters-region]',
     group: '[data-group-filter-region]',
     ownerToggle: '[data-owner-toggle-region]',
   },
 });
 
+const AllFiltersButtonView = View.extend({
+  className: 'button--link worklist-list__all-filters-button',
```
paulfalgout: I missed the font-size change.  I think maybe we should make a `button--link-large` or something.. the icon styling isn't needed I think if the text is wrapped in a `span`

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -19,9 +27,42 @@ export default App.extend({
     this.showFilters();
   },
   showFilters() {
+    this.showAllFiltersButtonView();
     this.showGroupsFilterView();
     this.showOwnerToggleView();
   },
+  showAllFiltersButtonView() {
+    if (this.groups.length < 2) return;
```
paulfalgout, line 35: when we get to the next step this check should be replaced for one that only shows if there are any custom filters.  If there are no custom filters, we don't need the sidebar whether or not there is a group filter.

src/js/apps/patients/sidebar/filters-sidebar_app.js
```diff
@@ -0,0 +1,45 @@
...
+  },
+  onChangeFilters(state) {
+    this.showGroupsFilterView(state.attributes);
+  },
+  showGroupsFilterView(state) {
+    const groups = this.currentClinician.getGroups();
+
+    const groupsFilterView = new GroupsFilterView({
+      groups,
+      state,
```
paulfalgout: I think maybe this could change to `groupId: this.getState('groupId'),` instead of passing the state around

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
```diff
@@ -0,0 +1,88 @@
...
+  template: hbs`
+    <h4 class="sidebar__label">{{ @intl.patients.sidebar.filters.groups.droplistLabel }}</h4>
+    <div class="flex-grow" data-filter-button></div>
+  `,
+  regions: {
+    filterButton: '[data-filter-button]',
+  },
+  onRender() {
+    const groups = this.getGroups();
+    const selected = groups.get(this.getOption('state').groupId || groups.at(0));
```
paulfalgout: this is probably copied from somewhere, but this can be `get() || at() rather than get(id || at)` as `at` returns the same thing as the `get`

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
```diff
@@ -0,0 +1,88 @@
...
+
+    const groupsFilter = new SidebarGroupsDropList({
+      collection: groups,
+      state: { selected },
+    });
+
+    this.listenTo(groupsFilter.getState(), 'change:selected', (state, { id }) => {
+      this.triggerMethod('select:new:group', id);
```
paulfalgout: maybe just `select:group`

src/js/apps/patients/sidebar/filters-sidebar_app.js
paulfalgout: Also I think rather than pass down the event because the app is instantiating it here this could be
```js
this.listenTo(groupsFilterView, 'select:group', this.onSelectGroup)` or just inlining the handler in this case'd be fine too.

```

paulfalgout (changes requested):
some nitpick things

src/js/apps/patients/schedule/filters_app.js
nmajor25: Yup, that's cleaner. Updated in last push.

src/js/apps/patients/schedule/filters_app.js
nmajor25: Oops, good catch. Updated in last push.

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
nmajor25: You are correct. Removed `--compact` in last push.

src/js/views/patients/worklist/filters_views.js
nmajor25: Updated in last push.

src/js/apps/patients/sidebar/filters-sidebar_app.js
nmajor25: Nice, I like it. Updated in the last push.

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
nmajor25: Makes sense, updated in the last push.

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
nmajor25: Updated to `select:group` in the last push.

paulfalgout (approved):

paulfalgout (approved):

PR #821, opened 2022-10-18
Fix picklist checkmark alignment

nmajor25:
Shortcut Story ID: [sc-31509]

Fixes alignment for the checkmark that's used for some picklist items.

Sets the right alignment of checkmarks on selected `Filter by Date...` picklist items:

Screen Shot 2022-10-11 at 1 37 19 PM (1)

And fixes the vertical alignment for the checkmark used for the `Save + Go Back` form button droplist:

Screen Shot 2022-10-17 at 11 22 22 PM

Also, the fontawesome icon for this button was changed from `circle-check` to `check`.

### DIFF
--- a/src/js/components/picklist/index.js
+++ b/src/js/components/picklist/index.js
@@ -59,7 +59,7 @@ const PicklistItem = View.extend({
   tagName: 'li',
   itemTemplate: hbs`
     <div class="flex-grow">{{#if icon}}{{fa icon.type icon.icon classes=icon.classes}}{{/if}}<span>{{matchText text query}}</span></div>
-    <div>{{#if isChecked}}{{fas "circle-check" classes="u-icon--12 u-margin--l-16"}}{{/if}}</div>
+    {{#if isChecked}}{{fas "check" classes="u-icon--12 u-margin--l-16"}}{{/if}}
   `,
   itemClassName() {
     const classNames = [];
--- a/src/js/components/picklist/picklist.scss
+++ b/src/js/components/picklist/picklist.scss
@@ -54,6 +54,11 @@
   .icon:first-child {
     margin-right: 8px;
   }
+
+  .icon.fa-check {
+    margin-left: auto;
+    margin-right: 0;
+  }
 }
 
 .picklist--no-results {
--- a/src/js/views/patients/shared/components/date-filter/date-filter_views.js
+++ b/src/js/views/patients/shared/components/date-filter/date-filter_views.js
@@ -188,7 +188,7 @@ const LayoutView = View.extend({
 
 const DateRanges = Picklist.extend({
   className: 'date-filter__ranges',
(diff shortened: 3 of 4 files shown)


src/js/components/picklist/index.js
```diff
@@ -59,7 +59,7 @@ const PicklistItem = View.extend({
   tagName: 'li',
   itemTemplate: hbs`
     <div class="flex-grow">{{#if icon}}{{fa icon.type icon.icon classes=icon.classes}}{{/if}}<span>{{matchText text query}}</span></div>
-    <div>{{#if isChecked}}{{fas "circle-check" classes="u-icon--12 u-margin--l-16"}}{{/if}}</div>
+    {{#if isChecked}}{{fas "check" classes="u-icon--12 u-margin--l-16"}}{{/if}}
```
nmajor25, line 62: As you stated on Shortcut, the wrapping `<div>` stopped the `align-items: center` from working. Removing that fixed the vertical alignment issue with the checkmark.

And I also switched the icon for this from `circle-check` to `check`. As requested by Mark.

src/js/components/picklist/picklist.scss
```diff
@@ -54,6 +54,11 @@
   .icon:first-child {
     margin-right: 8px;
   }
+
+  .icon.fa-check {
+    margin-left: auto;
+    margin-right: 0;
+  }
```
nmajor25, line 61: This fixes the right alignment issue with the `Filter by Date…` picklist checkmarks. Should be able to remove this when the `DateRanges` component is configured to use the `isCheckable` option.

src/js/views/patients/shared/components/date-filter/date-filter_views.js
```diff
@@ -188,7 +188,7 @@ const LayoutView = View.extend({
 
 const DateRanges = Picklist.extend({
   className: 'date-filter__ranges',
-  itemTemplate: hbs`{{formatMessage (intlGet "patients.shared.components.dateFilterComponent.relativeDate") relativeTo=id}}{{#if isSelected}}{{fas "check" classes="u-float--right"}}{{/if}}`,
+  itemTemplate: hbs`{{formatMessage (intlGet "patients.shared.components.dateFilterComponent.relativeDate") relativeTo=id}}{{#if isSelected}}{{fas "check"}}{{/if}}`,
```
nmajor25, line 191: I had a hard time switching this over to use `isCheckable`. Couldn’t get the custom `itemTemplate` for this picklist to work, as the picklist component needs to handle that directly for the `isCheckable` checkmark to be shown properly.

Was getting this as a result:

Screen Shot 2022-10-17 at 11 38 25 PM

So I don’t think the collection data being passed to it is in the correct format? It’s seeing each picklist item and which one is selected. But the correct text isn’t being shown.

I created a quick/temporary solution with the CSS I added for the `.icon.fa-check` icon if we need it for the release. I figured we could just remove that once we figure out `isCheckable` for this.

paulfalgout (approved):

PR #827, opened 2022-10-24
Add custom filters to worklist/schedule filters sidebar

nmajor25:
Shortcut Story ID: [sc-31235]

Configure the filters sidebar to show a variable number (1 or more) of custom filters. These are set specifically for each org and are retrieved via the `/api/directories?filter[filterable]=true` api route.

### DIFF
--- a/src/js/apps/patients/sidebar/filters-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/filters-sidebar_app.js
@@ -2,43 +2,44 @@ import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
-import { LayoutView, GroupsFilterView } from 'js/views/patients/sidebar/filters/filters-sidebar_views';
+import { LayoutView, CustomFiltersView, groupLabelView } from 'js/views/patients/sidebar/filters/filters-sidebar_views';
 
 export default App.extend({
-  stateEvents: {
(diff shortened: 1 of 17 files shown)


src/js/apps/patients/sidebar/filters-sidebar_app.js
```diff
@@ -2,43 +2,44 @@ import Radio from 'backbone.radio';
...
+
+    this.showChildView('customFilters', customFiltersView);
   },
   onClearFilters() {
-    this.setState('groupId', null);
-  },
-  onSelectGroup(id) {
-    this.setState('groupId', id);
+    this.getState().clear();
+    this.triggerMethod('reset:filters:state');
```
nmajor25, line 42: If only `this.getState().clear()` is called here, the worklist doesn't restart when the `Clear Filters` button is clicked. And, therefore, the worklist doesn't make a new request to the api with the default filters used. I'm not entirely sure why... it seems to only update the state of the filters in the sidebar, and nothing else.

I added a `this.triggerMethod('reset:filters:state');` here that passes an event up to the `worklist_app.js`, where it resets the `state.filters` to it's default and forces the worklist to restart. Feels overkill, but it was the only way I could get direct access to the state methods.

Same thing was applied to the schedule.

paulfalgout:
Added to #830

PR #835, opened 2022-10-31
Remove federated logout

paulfalgout:
Shortcut Story ID: [sc-31912]


paulfalgout:
We need to keep this for now

PR #841, opened 2022-11-07
Add tags for admin to program action and flow

paulfalgout:
Shortcut Story ID: [sc-30565]

Varies slightly from the marvel in that it does not display "No Results" unless there are no tags to add and the select list input is empty.
To simplify things you can also add a tag that already exists, which doesn't cause any issue.

### DIFF
--- a/src/js/components/picklist/index.js
+++ b/src/js/components/picklist/index.js
@@ -1,4 +1,4 @@
-import { debounce, each, extend, noop, pick, result } from 'underscore';
+import { debounce, each, extend, noop, pick, result, size } from 'underscore';
 import hbs from 'handlebars-inline-precompile';
 import { View, CollectionView } from 'marionette';
 
@@ -190,7 +190,7 @@ const Picklists = CollectionView.extend({
     this.addChildView(picklist);
   },
   viewFilter(childView) {
-    return !!childView.children.length;
+    return !!size(childView.children) || childView.shouldShow;
   },
   childViewEvents: {
     'before:render:children'() {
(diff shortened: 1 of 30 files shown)


src/js/components/picklist/index.js
```diff
@@ -190,7 +190,7 @@ const Picklists = CollectionView.extend({
     this.addChildView(picklist);
   },
   viewFilter(childView) {
-    return !!childView.children.length;
+    return !!size(childView.children) || childView.shouldShow;
```
paulfalgout, line 193: this previously required `children` assuming all views were collection views on a picklist.  This allows a view without children to pass it's view-ability to a flag on the view

src/js/components/picklist/picklist.scss
```diff
@@ -94,7 +94,7 @@
   position: relative;
 }
 
-.picklist__group:not(:last-child) {
+.picklist__group:not(:last-of-type) {
```
paulfalgout, line 97: I don't love this, but mark didn't add a border between the "Add" button and the list

src/js/entities-service/tags.js
```diff
@@ -0,0 +1,23 @@
...
+
+const Entity = BaseEntity.extend({
+  Entity: { _Model, Model, Collection },
+  radioRequests: {
+    'tags:model': 'getModel',
+    'tags:collection': 'getCollection',
+    'fetch:tags:collection': 'fetchTags',
+  },
+  fetchTags() {
+    if (tagsCache) return tagsCache;
```
paulfalgout, line 14: this makes it such that we're not expecting other tags to be made by other users during this user's session.  should be ok for now.. particularly since this is an rwell only feature

nmajor25 (approved):
Looks good! And everything works when testing in my local sandbox.

PR #843, opened 2022-11-08
Add styles for consistent table borders in PDF

zfixler:
[sc-32070]
The PDF output has been displaying inconsistent table borders at different zoom levels:

100%: 
image

300%:
image

After adding this CSS to `@media print`, the table displays like this at 100%:
image


paulfalgout (changes requested):
surprised this isn't complaining but these should be alphabetized.

PR #845, opened 2022-11-09
Change `.button-primary` angle-down icon size from `16px` to `12px`

nmajor25:
Shortcut Story ID: [sc-32148]

This CSS change updates the add workflow button on the patient dashboard page.

Before (`font-size: 16px;`):

Screen Shot 2022-11-09 at 3 13 20 PM

After (`font-size: 12px;`):

Screen Shot 2022-11-09 at 3 12 55 PM

### DIFF
--- a/src/scss/modules/buttons.scss
+++ b/src/scss/modules/buttons.scss
@@ -63,6 +63,11 @@
     border-color: color(light_blue);
     color: $white;
   }
+
+  .icon.fa-angle-down {
+    font-size: 12px;
+    vertical-align: -0.125em;
+  }
 }
 
 .button-secondary {


src/scss/modules/buttons.scss
```diff
@@ -63,6 +63,11 @@
     border-color: color(light_blue);
     color: $white;
   }
+
+  .icon.fa-angle-down {
+    font-size: 12px;
+    vertical-align: -0.125em;
+  }
```
nmajor25, line 70: Attempted to apply this to the `.button` class instead, but seemed less risky to just put it on `.button-primary` instead.

[.button-filter .icon.fa-angle-down](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/scss/modules/buttons.scss#L113) would be effected by that. And that class has some left and top margins that would make things complicated as well... as we'd only want that applied on the `.button-filter` button and not others.

paulfalgout (approved):

PR #852, opened 2022-11-14
Set filters sidebar `isFiltering` value to `false` when `onBeforeStop()` occurs

nmajor25:
Shortcut Story ID: [sc-32171]

When `onBeforeStart()` occurs for the filters app, set the `isFiltering` state value to `false` (via a new `toggleFiltering()` function).

This fixes a bug where the filters sidebar wasn't able to be opened after the bulk edit app was started. The bug occurred in this scenario:

1. Open the `All Filters` sidebar.
2. With it still open, select all items (actions or flow).
3. Close `All Filters` sidebar.
4. Deselect all items.
5. Try to re-open the `All Filters` sidebar, but it'll no longer open.

### DIFF
--- a/test/integration/patients/worklist/worklist.js
+++ b/test/integration/patients/worklist/worklist.js
@@ -2137,6 +2137,38 @@ context('worklist page', function() {
     cy
       .get('@filtersSidebar')
       .should('not.exist');
+
+    cy
+      .get('.list-page__filters')
+      .find('[data-all-filters-region]')
+      .click();
+
+    cy
+      .get('[data-select-all-region]')
+      .click();
+
+    cy
+      .get('@filtersSidebar')
+      .find('.js-close')
+      .click();
+
+    cy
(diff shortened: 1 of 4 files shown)


test/integration/patients/worklist/worklist.js
```diff
@@ -2137,6 +2137,38 @@ context('worklist page', function() {
...
+      .click();
+
+    cy
+      .get('.list-page__filters')
+      .find('[data-all-filters-region]')
+      .click();
+
+    cy
+      .get('@filtersSidebar')
+      .should('exist');
```
nmajor25, line 2171: Added this to the test:

1. Open the filters sidebar.
2. Select all list items (starts bulk select app).
3. Cancel select all by clicking the `Cancel` option.
4. Verify the filters sidebar can be closed and then opened properly after that.

paulfalgout (approved):

PR #856, opened 2022-11-18
Ability to see if an action has attachments on list views

nmajor25:
Shortcut Story ID: [sc-32259]

If an action has one or more attachments, display a paperclip icon in list views that indicates it has attachments.

Icon was added to action list items on the following pages:

1. Worklist
2. Patient Dashboard
3. Patient Archive
4. Patient flow

Screen Shot 2022-11-18 at 3 36 08 PM

### DIFF
--- a/src/js/views/patients/worklist/worklist-list.scss
+++ b/src/js/views/patients/worklist/worklist-list.scss
@@ -23,12 +23,10 @@
   color: color(blue);
 }
 
-.worklist-list__details {
+.worklist-list__display-icon {
   color: #999;
   display: inline-block;
-  margin-right: 16px;
   vertical-align: middle;
-  width: 16px;
 }
 
 .work-list__item:hover {
--- a/src/scss/modules/table-list.scss
+++ b/src/scss/modules/table-list.scss
@@ -48,7 +48,7 @@
   color: $black-60;
 }
 
-.table-list__meta {
+.table-list__meta:not(:empty) {
   margin-right: 8px;
 }
 
--- a/test/integration/patients/patient/patient-flow.js
+++ b/test/integration/patients/patient/patient-flow.js
@@ -149,6 +149,7 @@ context('patient flow page', function() {
           type: 'teams',
         };
         fx.data[0].relationships.form.data = { id: '11111' };
+        fx.data[0].relationships.files = { data: [{ id: '1' }] };
 
         fx.data[1].id = '2';
         fx.data[1].attributes.name = 'Third In List';
@@ -161,7 +162,6 @@ context('patient flow page', function() {
           id: '33333',
           type: 'teams',
         };
-
 
         fx.data[2].id = '3';
         fx.data[2].attributes.name = 'Second In List';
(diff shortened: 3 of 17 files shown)


src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -28,7 +28,13 @@
...
   margin-right: 16px;
   vertical-align: middle;
-  width: 16px;
+}
+
+.worklist-list__attachment {
+  color: #999;
+  display: inline-block;
+  margin-left: 8px;
+  vertical-align: middle;
```
nmajor25, line 29: This `.worklist-list__attachment` and `.worklist-list__details` above it have almost identical properties. 

Maybe we could combine them and make use of `.u-margin--l-8` and `.u-margin--r-16` for the `margin-right`/`margin-left` differences?

test/integration/patients/patient/patient-flow.js
```diff
@@ -814,6 +817,8 @@ context('patient flow page', function() {
         return fx;
       })
       .routeActionActivity()
+      .routePatientField()
+      .routeActionComments()
```
nmajor25, line 821: These are unrelated to the topic of this PR, but were causing some 500 errors in the tests. Noticed a few more that I added throughout this PR.

src/scss/modules/table-list.scss
```diff
@@ -50,6 +50,10 @@
 
 .table-list__meta {
   margin-right: 8px;
+
+  &:empty {
+    margin-right: 0;
+  }
```
nmajor25: There are times when the form button region (`[data-form-region]`) is empty but still has `margin-right: 8px` applied to it. That sometimes adds unnecessary extra margin when the attachments icon is being shown.

This stops that from happening.

src/js/views/patients/worklist/worklist-list.scss
paulfalgout: yep I think that's ideal.  perhaps the class can be something more generic.  `__display-icon` or something?

src/scss/modules/table-list.scss
paulfalgout: could also be `.table-list__meta:not(:empty)`

test/integration/patients/patient/patient-flow.js
```diff
@@ -209,6 +210,7 @@ context('patient flow page', function() {
         expect($action.find('[data-owner-region]')).to.contain('NUR');
         expect($action.find('[data-due-day-region] .is-overdue')).to.exist;
         expect($action.find('[data-form-region]')).not.to.be.empty;
+        expect($action.find('.patient__action-attachment')).to.exist;
```
paulfalgout: I think we should test for the icon directly over the style class

paulfalgout (changes requested):

src/js/views/patients/worklist/worklist-list.scss
nmajor25: That works well. Updated in latest push.

src/scss/modules/table-list.scss
nmajor25: Yup... definitely like that better. Updated in last push.

test/integration/patients/patient/patient-flow.js
nmajor25: Updated in last push.

paulfalgout (approved):

PR #859, opened 2022-11-23
View/download file attachments in the action sidebar

nmajor25:
Shortcut Story ID: [sc-32260]

In the action sidebar, allow a user to view a file attachment in a new browser tab or download it to their device.

Screen Shot 2022-11-22 at 4 07 10 PM


src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -9,32 +9,59 @@ import { ACTION_SHARING } from 'js/static';
...
+        'updated_at': '2019-08-24T14:15:22Z',
+      },
+      {
+        'id': '2d577e83-f435-40c3-b992-e7040f5c58bc',
+        'path': 'https://www.bucket_name.s3.amazonaws.com/HRA Final Version v2.pdf',
+        'created_at': '2019-08-25T14:15:22Z',
+        'updated_at': '2019-08-25T14:15:22Z',
+      },
+    ]);
+    */
```
nmajor25: My assumption for what the file attachments api data will look like.

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -9,32 +9,59 @@ import { ACTION_SHARING } from 'js/static';
...
+      this.getRegion('activity').startPreloader();
+    }
   },
   beforeStart() {
     if (this.action.isNew()) return;
 
     return [
       Radio.request('entities', 'fetch:actionEvents:collection', this.action.id),
       Radio.request('entities', 'fetch:comments:collection:byAction', this.action.id),
+      Radio.request('entities', 'fetch:files:collection:byAction', this.action.id),
```
nmajor25, line 31: Should we only make this request if the action has file attachments? (`this.action.hasAttachments() === true`).

Or is this fine as is? Where the request is made regardless.

src/js/views/patients/sidebar/action/action-sidebar.hbs
```diff
@@ -30,10 +30,18 @@
...
+      <h3 class="sidebar__heading">{{far "paperclip"}}<span class="u-margin--l-8">{{ @intl.patients.sidebar.action.actionSidebarTemplate.attachmentsHeadingText }}</span></h3>
+      <div data-attachments-region></div>
+    </div>
+  </div>
+  {{/if}}
   {{#unless isNew}}
   <div class="u-margin--t-24">
-    <h3 class="sidebar__heading">{{ @intl.patients.sidebar.action.actionSidebarTemplate.activityHeadingText }}</h3>
-    <div class="u-margin--t-8 sidebar__activity" data-activity-region></div>
+    <h3 class="sidebar__heading">{{far "timeline-arrow"}}<span class="u-margin--l-8"></span>{{ @intl.patients.sidebar.action.actionSidebarTemplate.activityHeadingText }}</span></h3>
```
nmajor25: Per the design docs, add an icon to the activity heading text.

Screen Shot 2022-11-22 at 4 08 52 PM

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
```diff
@@ -0,0 +1,36 @@
...
+
+import './action-sidebar.scss';
+
+const AttachmentView = View.extend({
+  className: 'u-margin--t-16',
+  template: hbs`
+    <a class="action-sidebar__attachment-filename" target="_blank" href={{path}}>{{filename}}</a>
+    <div>
+      <a class="action-sidebar__attachment-download" href={{path}} download>
+        {{far "download"}} <span>Download</span>
```
nmajor25: Should add this to the `i18n` file.

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
```diff
@@ -0,0 +1,36 @@
...
+import 'scss/modules/sidebar.scss';
+
+import './action-sidebar.scss';
+
+const AttachmentView = View.extend({
+  className: 'u-margin--t-16',
+  template: hbs`
+    <a class="action-sidebar__attachment-filename" target="_blank" href={{path}}>{{filename}}</a>
+    <div>
+      <a class="action-sidebar__attachment-download" href={{path}} download>
```
nmajor25, line 15: Without having an example s3 link to work with, I don't know if this [download](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download) attribute will work...

For the browser to treat this as a download, it has to be a same origin URL or have special HTTP headers set on it.

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
```diff
@@ -0,0 +1,36 @@
...
+        {{far "download"}} <span>Download</span>
+      </a>
+    </div>
+  `,
+  templateContext() {
+    return {
+      filename: this.model.get('path').split('/').pop(),
```
nmajor25, line 22: A basic way to extract the filename from the URL string.

Any additional edge cases to accommodate here? May be hard to plan for those until we know what the `s3` URLs look like.

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
```diff
@@ -0,0 +1,36 @@
...
+    };
+  },
+});
+
+const AttachmentsView = CollectionView.extend({
+  childView: AttachmentView,
+  viewComparator(viewA, viewB) {
```
nmajor25: If there are more than one attachments, do we want them sorted in a certain way?

Could be by `created_by` or `updated_at`.

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -9,32 +9,59 @@ import { ACTION_SHARING } from 'js/static';
...
     this.showView(new LayoutView({
       action: this.action,
       isShowingForm,
     }));
 
-    if (!this.action.isNew()) this.getRegion('activity').startPreloader();
+    if (!this.action.isNew()) {
+      if (this.hasAttachments) this.getRegion('attachments').startPreloader();
+
+      this.getRegion('activity').startPreloader();
```
paulfalgout: I think maybe the one preloader is ok as they'll always resolve at the same time

src/js/apps/patients/sidebar/action-sidebar_app.js
paulfalgout: seems ok to reload.  we're reloading plenty about the action every sidebar opening and to some degree it's because we don't have push when things change.

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -60,6 +87,11 @@ export default App.extend({
       'cancel:comment': this.onCancelNewComment,
     });
   },
+  showAttachments() {
+    if (!this.hasAttachments) return;
```
paulfalgout: I suspect this could just be the check rather than caching it in the onBeforeStart if it isn't needed for the preloader

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
paulfalgout: ah interesting.  Perhaps it'll need to be: https://stackoverflow.com/a/37473099

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
paulfalgout: seems correct

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
paulfalgout: would `viewComparator: 'created_at'` work here?  I can't remember the default asc/desc.. it's probably desc? so maybe not

paulfalgout: also seems like create_at would be ok.. I dunno if/when we'll even have an update_at that'd be any different

src/js/views/patients/sidebar/action/action-sidebar.hbs
```diff
@@ -30,10 +30,18 @@
     <div data-form-sharing-region></div>
     {{/if}}
   </div>
+  {{#if hasAttachments}}
```
paulfalgout: what if you moved this template stuff in this `if` to a template on the attachments collectionview and this was just a plain region.  then this view is really simple and the collectionview gets shown or not.  could even be off of `attachmentCollection.length` rather than the action check.

src/scss/modules/sidebar.scss
```diff
@@ -50,6 +50,16 @@
   width: 84px;
 }
 
+.sidebar__attachments {
+  border-bottom: 1px solid #E5E5E5;
```
paulfalgout: let's just use `$black-90` here

paulfalgout (changes requested):

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
nmajor25: That looks promising 🤔 

Possibly something they could configure in the `s3` settings too? I've never worked in `s3` before, so don't know for sure.

src/scss/modules/sidebar.scss
nmajor25: Updated in the last push.

src/js/views/patients/sidebar/action/action-sidebar.hbs
nmajor25: Updated in the last push.

To accommodate the loading spinner changes, I also moved the activities template to the `/views/patients/sidebar/action/action-sidebar-activity-views.js` collectionview as well.

src/js/views/patients/sidebar/action/action-sidebar.scss
```diff
@@ -22,6 +22,31 @@
   }
 }
 
+.action-sidebar__activity {
+  .spinner {
+    margin-top: 100px;
+  }
+}
```
nmajor25, line 29: Couldn't figure out how to do center vertical alignment. Will keep looking into it, but this is seems to work pretty well.

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
nmajor25: Looks like `viewComparator: 'created_at'` is ascending. And we want descending (newest to oldest).

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: Switched this to the one preloader in the last push.

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: Updated this in the last push.

src/js/views/patients/sidebar/action/action-sidebar.hbs
```diff
@@ -27,15 +26,11 @@
     <div class="flex{{#if hasOutreach}} u-margin--t-24{{/if}}" data-form-sharing-region></div>
     {{else}}
     <div data-form-region></div>
-    <div data-form-sharing-region></div>
+    <div class="u-margin--t-24" data-form-sharing-region></div>
```
nmajor25, line 29: This fixes a styling issue I noticed when viewing tests in Cypress.

Before:

Screen Shot 2022-11-28 at 9 31 44 PM

After:

Screen Shot 2022-11-28 at 9 33 26 PM

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -742,6 +748,110 @@ context('action sidebar', function() {
...
+      .first()
+      .find('.action-sidebar__attachment-download')
+      .should('have.attr', 'href')
+      .and('contain', 'https://www.bucket_name.s3.amazonaws.com/HRA_v2.pdf');
+
+    cy
+      .get('@attachmentItems')
+      .first()
+      .find('.action-sidebar__attachment-download')
+      .should('have.attr', 'download');
```
nmajor25, line 851: I'm simply testing the attributes/properies of the `<a></a>` for both the filename and download button.

* Filename: that it has the correct `href` and the `target=_blank` to open the file in a new tab.
* Download button: that it has the correct `href` and the `download` attribute to download the file to the users device.

Does this work? Or do we need to actually open the links in the tests?

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
```diff
@@ -0,0 +1,57 @@
...
+    return alphaSort('desc', viewA.model.get('created_at'), viewB.model.get('created_at'));
+  },
+});
+
+const AttachmentsView = View.extend({
+  className: 'u-margin--t-24',
+  template: hbs`
+    <div class="sidebar__attachments">
+      <h3 class="sidebar__heading">
+        {{far "paperclip"}}<span class="u-margin--l-8">{{ @intl.patients.sidebar.action.actionSidebarTemplate.attachmentsHeadingText }}</span>
```
paulfalgout: `actionSidebarTemplate` needs to be updated for the view

src/js/views/patients/sidebar/action/action-sidebar.scss
```diff
@@ -22,6 +22,31 @@
...
+
+.action-sidebar__attachment-filename {
```
paulfalgout, line 31: Let's keep these in mind if we ever see any of these styles reused.  They seem generic-y

paulfalgout (changes requested):

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -60,6 +65,11 @@ export default App.extend({
       'cancel:comment': this.onCancelNewComment,
     });
   },
+  showAttachments() {
+    if (!this.action.hasAttachments()) return;
```
nmajor25: Do we want to add an additional check here for `this.attachmentsCollection.length`? Just remembered that you mentioned this previously.

~~Is it possible `this.action.hasAttachments()` returns `false` (meaning `action.relationships._files` is empty), but the `/actions/{actionId}/relationships/files` api request returns files for the action?~~

Is it possible `this.action.hasAttachments()` returns `true`, but the `/actions/{actionId}/relationships/files` api request returns files for the action?

In that scenario, an empty `Attachments` section would be shown in the sidebar.

Screen Shot 2022-11-29 at 10 45 13 AM

nmajor25: Or I guess we could replace `this.action.hasAttachments()` with `this.attachmentsCollection.length` as well.

paulfalgout: just checking the collection I think makes more sense and will be more correct, even if the attachment icon doesn't get updated in the list

nmajor25: Agreed, updated in the last push.

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
paulfalgout: @nmajor25 I think this is the only updated needed for this PR

nmajor25: Missed that one, just updated in the [latest push](https://github.com/RoundingWell/care-ops-frontend/compare/68e0bad7286fe43133101c713b6dfcf1024cb3d8..94513d9123ff0fc3bb004d21722f66285224fbce).

Updated the `src/js/views/patients/sidebar/action/action-sidebar-activity-views.js` as well, since that was also moved from the `actionSidebarTemplate`.

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -248,6 +249,25 @@ const ActivitiesView = CollectionView.extend({
   },
 });
 
+const ActivitiesView = View.extend({
```
paulfalgout: Actually now that I'm looking at this I think this might as well be a `CollectionView` with a `template`.

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
```diff
@@ -0,0 +1,57 @@
...
+});
+
+const AttachmentsCollectionView = CollectionView.extend({
+  childView: AttachmentView,
+  viewComparator(viewA, viewB) {
+    return alphaSort('desc', viewA.model.get('created_at'), viewB.model.get('created_at'));
+  },
+});
+
+const AttachmentsView = View.extend({
```
paulfalgout: Same here this template can just move to the collectionview using `childViewContainer`

paulfalgout (changes requested):

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
nmajor25: Good call, updated in the [last push](https://github.com/RoundingWell/care-ops-frontend/compare/94513d9123ff0fc3bb004d21722f66285224fbce..d40905e94fa7c8344e44f93bef6cc13d1097e014).

src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js
nmajor25: Updated in the [last push](https://github.com/RoundingWell/care-ops-frontend/compare/94513d9123ff0fc3bb004d21722f66285224fbce..d40905e94fa7c8344e44f93bef6cc13d1097e014).

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -1065,9 +1176,9 @@ context('action sidebar', function() {
 
     cy
       .get('[data-activity-region]')
+      .find('[data-activities-region]')
       .should('contain', 'Clinician McTester (Nurse) added this Action from the Test Program program')
       .children()
-      .children()
```
nmajor25, line 1070: There's no longer an empty `<div>` wrapped around these, so the second `.children()` broke the test.

PR #860, opened 2022-11-23
Change XHR/deferred to fetch/promise

paulfalgout:
I still have some bugs to work out in the forms, but I think this is pretty close.

I need to clean up commits and probably review a bit of it in general, but I wanted to get something up on a branch before the holiday.

After this I think we can try setting up https://github.com/GoogleChrome/workbox

I'll add a card or two somewhere before the next monday meeting related to all of this


test/integration/programs/sidebar/action-sidebar.js
```diff
@@ -794,7 +794,7 @@ context('program action sidebar', function() {
       .intercept({
         method: 'PATCH',
         url: 'api/program-actions/1',
-      }, { statusCode: 201 })
+      }, { statusCode: 204 })
```
paulfalgout, line 797: These are 204s when no content on the backend.  201s are for successful POSTs and return data

src/js/outreach/index.js
```diff
@@ -125,7 +121,7 @@ const FormApp = App.extend({
     dialogView.showChildView('content', new ErrorView());
     this.showView(dialogView);
   },
-  onStart({ actionId }, [form], [definition], [fields]) {
+  onStart({ actionId }, form, definition, fields) {
```
paulfalgout, line 125: @nmajor25 major style change is that `entities` radio requests no longer get returned successfully with the xhr so no need to wrap the arg in an array

src/js/base/fetch.js
```diff
@@ -0,0 +1,64 @@
...
+  return response.text();
+}
+
+export function handleJSON(response) {
+  if (!response.ok) return Promise.reject(response);
+
+  return getData(response, 'json');
+}
+
+export default async(url, opts) => {
```
paulfalgout, line 29: I haven't had extensive experience with `fetch`, but this brings fetch to general parity with ajax with an attached `abort` and sensible defaults for our app along with the global 401/5xx handlers.

src/js/base/fetch.js
```diff
@@ -0,0 +1,64 @@
...
+
+  return fetch(url, options)
+    .then(response => {
+      if (!response.ok) {
+        if (response.status === 401) {
+          Radio.request('auth', 'logout');
+          return Promise.reject(response);
+        }
+
+        // FIXME: Should be >= 500, but missing Cypress stubs cause wide failures
```
paulfalgout, line 57: Small note here that _maybe_ we should handle 5xx, although I don't know that we're returning anything other than 500 from the API currently..

src/js/base/fetch.js
```diff
@@ -0,0 +1,64 @@
...
+  return getData(response, 'json');
+}
+
+export default async(url, opts) => {
+  const token = await getToken();
+
+  const options = extend({}, opts);
+  url = getUrl(url, options);
+
+  if (options.method === 'GET') delete options.data;
```
paulfalgout: This allows for `fetch` to construct the get url by passing the same `{ data }` as you would on `ajax` and otherwise puts it on the body.

paulfalgout:
[sc-32293]

nmajor25 (approved):

PR #863, opened 2022-11-30
Fix broken action file attachments download links

nmajor25:
Shortcut Story ID: [sc-32337]

Use `file.meta._download` instead of `file.attributes.path` for downloading action file attachments. And update the handlebars template to accommodate spaces in the download `href` links.

### DIFF
--- a/test/integration/patients/sidebar/action-sidebar.js
+++ b/test/integration/patients/sidebar/action-sidebar.js
@@ -788,15 +788,21 @@ context('action sidebar', function() {
           {
             id: '1',
             attributes: {
-              path: 'https://www.bucket_name.s3.amazonaws.com/HRA_v1.pdf',
+              path: 'patients/1/HRA.pdf',
               created_at: '2019-08-24T14:15:22Z',
+            },
+            meta: {
+              download: 'https://www.bucket_name.s3.amazonaws.com/patients/1/HRA.pdf',
             },
           },
           {
             id: '2',
             attributes: {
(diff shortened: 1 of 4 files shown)


test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -788,15 +788,18 @@ context('action sidebar', function() {
           {
             id: '1',
             attributes: {
-              path: 'https://www.bucket_name.s3.amazonaws.com/HRA_v1.pdf',
+              path: 'patients/1/HRA.pdf',
+              _download: 'https://www.bucket_name.s3.amazonaws.com/patients/1/HRA.pdf',
```
shadowhand: What's up with the underscore? Why doesn't the fixture have a `meta` object?

paulfalgout: this should be here:
```
attributes: {
  path: '',
},
meta: {
  download: '',
},
```
and will end up as an underscored attribute on the model

paulfalgout (changes requested):

test/integration/patients/sidebar/action-sidebar.js
shadowhand: Thanks @paulfalgout, something felt off about the fixture.

nmajor25: Good call @shadowhand and @paulfalgout , just updated that.

shadowhand (approved):
API calls look good.

PR #865, opened 2022-11-30
Use separate `_view` and `_download` links for action file attachments

nmajor25:
Shortcut Story ID: [sc-32343]

Use the `file.meta._view` URL for the filename link. And use the `file.meta._download` URL for the download button link.

Also, update the API request parameters to `?urls=download,view`.

### DIFF
--- a/test/integration/patients/sidebar/action-sidebar.js
+++ b/test/integration/patients/sidebar/action-sidebar.js
@@ -792,7 +792,8 @@ context('action sidebar', function() {
               created_at: '2019-08-24T14:15:22Z',
             },
             meta: {
-              download: 'https://www.bucket_name.s3.amazonaws.com/patients/1/HRA.pdf',
+              view: 'https://www.bucket_name.s3.amazonaws.com/patients/1/view/HRA.pdf',
+              download: 'https://www.bucket_name.s3.amazonaws.com/patients/1/download/HRA.pdf',
             },
           },
           {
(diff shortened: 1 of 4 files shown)


test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -802,7 +803,8 @@ context('action sidebar', function() {
               created_at: '2019-08-25T14:15:22Z',
             },
             meta: {
-              download: 'https://www.bucket_name.s3.amazonaws.com/patients/1/HRA%20v2.pdf',
+              view: 'https://www.bucket_name.s3.amazonaws.com/patients/1/view/HRA%20v2.pdf',
+              download: 'https://www.bucket_name.s3.amazonaws.com/patients/1/download/HRA%20v2.pdf',
```
nmajor25, line 807: Don't know what the difference between these two links will be in the prod environment.

But this just creates two different links and tests that they are used in the correct spots. Filename vs. Download button.

paulfalgout (approved):

test/integration/patients/sidebar/action-sidebar.js
shadowhand: Technically the path won't be different, but rather the query string will be modified. NBD.

shadowhand (approved):

test/integration/patients/sidebar/action-sidebar.js
nmajor25: Ah gotcha. We can update this test later to be more accurate compared to qa/prod.

paulfalgout: meh it's fine, these strings could be `view-url-goes-here` and it's enough for these integration tests.

shadowhand: Yup, exactly. The important part is just differentiating between download/upload/view.

nmajor25: Works for me!

PR #866, opened 2022-12-01
Preload font files to prevent some font flashing

paulfalgout:
Shortcut Story ID: [sc-32352]
Screen Shot 2022-12-01 at 3 26 01 PM

Only caching woff2 as that should cover everyone.. the other font files are just fallbacks.  Similarly the other webfonts are for forms and should be far less obvious due to form.io render times.


nmajor25 (approved):
Makes sense and things looked good in my sandbox.

PR #867, opened 2022-12-01
Filter latest form response by flow

nmajor25:
Shortcut Story ID: [sc-32297]

If an action has the `prefill-flow-response`, filter the latest form response by flow `id`.

### DIFF
--- a/src/js/entities-service/form-responses.js
+++ b/src/js/entities-service/form-responses.js
@@ -14,7 +14,11 @@ const Entity = BaseEntity.extend({
     if (!responseId) return [{}];
     return fetcher(`/api/form-responses/${ responseId }/response`).then(handleJSON);
   },
-  fetchLatestSubmission(patientId, formId) {
+  fetchLatestSubmission(patientId, formId, flowId) {
+    if (flowId) {
+      return fetcher(`/api/patients/${ patientId }/form-responses/latest?filter[form]=${ formId }&filter[flow]=${ flowId }`).then(handleJSON);
+    }
+
     return fetcher(`/api/patients/${ patientId }/form-responses/latest?filter[form]=${ formId }`).then(handleJSON);
   },
 });
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -97,7 +97,7 @@ export default App.extend({
       });
     });
   },
-  fetchLatestFormSubmission() {
+  fetchLatestFormSubmission(flowId) {
     const channel = this.getChannel();
 
     const prefillFormId = this.form.getPrefillFormId();
@@ -105,7 +105,7 @@ export default App.extend({
     return Promise.all([
       Radio.request('entities', 'fetch:forms:definition', this.form.id),
       Radio.request('entities', 'fetch:forms:fields', get(this.action, 'id'), this.patient.id, this.form.id),
-      Radio.request('entities', 'fetch:formResponses:latestSubmission', this.patient.id, prefillFormId),
+      Radio.request('entities', 'fetch:formResponses:latestSubmission', this.patient.id, prefillFormId, flowId),
       this.form.fetch(),
     ]).then(([definition, fields, response]) => {
       channel.request('send', 'fetch:form:data', {
@@ -126,8 +126,9 @@ export default App.extend({
     const channel = this.getChannel();
     const firstResponse = this.responses && this.responses.first();
 
-    if (!firstResponse && this.action && this.action.hasTag('prefill-latest-response')) {
-      return this.fetchLatestFormSubmission();
+    if (!firstResponse && this.action) {
+      if (this.action.hasTag('prefill-latest-response')) return this.fetchLatestFormSubmission();
+      if (this.action.hasTag('prefill-flow-response')) return this.fetchLatestFormSubmission(this.action.get('_flow'));
     }
 
     return Promise.all([
--- a/test/integration/forms/form.js
+++ b/test/integration/forms/form.js
@@ -350,12 +350,72 @@ context('Patient Action Form', function() {
       .should('have.value', 'bar');
   });
 
+  specify('prefill a form with latest submission by flow', function() {
+    cy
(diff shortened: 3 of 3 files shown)


src/js/entities-service/form-responses.js
```diff
@@ -14,7 +14,11 @@ const Entity = BaseEntity.extend({
     if (!responseId) return [{}];
     return fetcher(`/api/form-responses/${ responseId }/response`).then(handleJSON);
   },
-  fetchLatestSubmission(patientId, formId) {
+  fetchLatestSubmission(patientId, formId, flowId) {
+    if (flowId) {
+      return fetcher(`/api/patients/${ patientId }/form-responses/latest?filter[form]=${ formId }&filter[flow]=${ flowId }`).then(handleJSON);
+    }
```
nmajor25, line 20: Using the data object:

```js
fetchLatestSubmission(patientId, formId, flowId) {
  const data = {
    filter: {
      form: formId,
    },
  };

  if (flowId) data.filter.flow = flowId;

  return fetcher(`/api/patients/${ patientId }/form-responses/latest`, { data }).then(handleJSON);
}
```

Causes this error saying we can't do that for `GET` requests:

Screen Shot 2022-12-01 at 2 28 59 PM

nmajor25: Could just do something like this instead:

```js
fetchLatestSubmission(patientId, formId, flowId) {
  let url = `/api/patients/${ patientId }/form-responses/latest?filter[form]=${ formId }`;

  if (flowId) url += `&filter[flow]=${ flowId }`; // or use concat()

  return fetcher(url).then(handleJSON);
}
```

test/integration/forms/form.js
```diff
@@ -350,6 +350,65 @@ context('Patient Action Form', function() {
...
+    cy
+      .iframe()
+      .find('textarea[name="data[storyTime]"]')
+      .should('have.value', 'Prefilled story time');
+
+    cy
+      .iframe()
+      .find('[name="data[patient.fields.foo]"]')
+      .should('have.value', 'bar');
+  });
```
nmajor25, line 410: Same as the existing `'prefill a form with latest submission from another form'` test, but with the ` ['prefill-flow-response']` tag and an added flow relationship:

test/integration/forms/form.js
```diff
@@ -350,6 +350,65 @@ context('Patient Action Form', function() {
       .should('have.value', 'bar');
   });
 
+  specify('prefill a form with latest submission by flow', function() {
```
nmajor25, line 353: Correct name for this test?

nmajor25:
Confirmed with @zfixler that this works in the wild.

src/js/entities-service/form-responses.js
paulfalgout: ah I know why.. it only converts the data if method is explicitly set like  `{ method: 'GET', data }`. It should probably be changed to look for something like `if (method && method. !== 'GET') options.body = options.data;`

Anyhow there's a separate card for this pattern refactor where we can do it.

src/js/services/forms.js
```diff
@@ -101,11 +101,12 @@ export default App.extend({
     const channel = this.getChannel();
 
     const prefillFormId = this.form.getPrefillFormId();
+    const flowId = this.action.get('_flow');
 
     return Promise.all([
       Radio.request('entities', 'fetch:forms:definition', this.form.id),
       Radio.request('entities', 'fetch:forms:fields', get(this.action, 'id'), this.patient.id, this.form.id),
-      Radio.request('entities', 'fetch:formResponses:latestSubmission', this.patient.id, prefillFormId),
+      Radio.request('entities', 'fetch:formResponses:latestSubmission', this.patient.id, prefillFormId, flowId),
```
paulfalgout, line 108: `flowId` can only be used if `prefill-flow-response` so it should be passed into this function `fetchLatestFormSubmission(formId)`

test/integration/forms/form.js
```diff
@@ -389,7 +448,8 @@ context('Patient Action Form', function() {
       .wait('@routeLatestFormResponseByPatient')
       .itsUrl()
       .its('search')
-      .should('contain', 'filter[form]=11111');
+      .should('contain', 'filter[form]=11111')
+      .should('not.contain', 'filter[flow]');
```
paulfalgout, line 453: we should go ahead and add `fx.data.relationships.flow = { data: { id: '1' } };` to the `routeAction`.  Even if the action has a flow it should not show up in this test.

paulfalgout (changes requested):

src/js/entities-service/form-responses.js
nmajor25: Hmm makes sense. So leave as is for now?

src/js/services/forms.js
nmajor25: Ah ya, good call. Updated in last push.

test/integration/forms/form.js
nmajor25: Updated in last push.

test/integration/forms/form.js
```diff
@@ -516,6 +577,7 @@ context('Patient Action Form', function() {
 
         return fx;
       })
+      .routePatientField()
```
nmajor25, line 580: Unrelated to this PR. But noticed a `500` error in this test because this was missing.

src/js/services/forms.js
```diff
@@ -97,15 +97,15 @@ export default App.extend({
       });
     });
   },
-  fetchLatestFormSubmission() {
+  fetchLatestFormSubmission(flowId) {
```
nmajor25, line 100: Looking at this now, should this be `(flowId = null)`? But I guess it's fine if `flowId` is just left as `undefined`...

paulfalgout: undefined it better in this case I think.  null is `&flow=` and undefined is  ``

paulfalgout (approved):

PR #868, opened 2022-12-05
Form Save + Go Back Animation

nmajor25:
Shortcut Story ID: [sc-32340]

The `Save + Go Back` button for forms was added in this PR: https://github.com/RoundingWell/care-ops-frontend/pull/806.

When a user uses this button and is returned to the flow page, it happens before evaluations has time to finish and, therefore, the user has to reload the page to see the new flow action.

This PR shows a `Saving your work...` loading modal with a 5 second delay. This allows time for the evaluations to finish before the page is redirected.

https://user-images.githubusercontent.com/35355575/206292225-4650446a-3614-4407-a8d3-f243bc037f33.mov

### DIFF
--- a/src/js/apps/forms/form/form_app.js
+++ b/src/js/apps/forms/form/form_app.js
@@ -104,17 +104,27 @@ export default App.extend({
     });
 
     this.listenTo(formService, {
-      'success'(response) {
-        response.set({ _created_at: dayjs().format() });
-        this.responses.unshift(response);
-        this.setState({ responseId: response.id });
+      'submit'() {
+        const saveButtonType = this.getState('saveButtonType');
 
+        if (saveButtonType !== 'saveAndGoBack') return;
+
+        this.loadingModal = Radio.request('modal', 'show:loading');
+      },
+      'success'(response) {
         const saveButtonType = this.getState('saveButtonType');
 
         if (saveButtonType === 'saveAndGoBack') {
-          Radio.request('history', 'go:back');
+          this.listenTo(this.loadingModal, 'destroy', () => {
+            Radio.request('history', 'go:back');
+          });
+
           return;
         }
+
+        response.set({ _created_at: dayjs().format() });
+        this.responses.unshift(response);
+        this.setState({ responseId: response.id });
 
         this.showFormStatus();
         this.showFormUpdate();
@@ -123,6 +133,8 @@ export default App.extend({
         this.showFormSave();
       },
       'error'() {
+        if (this.loadingModal) this.loadingModal.destroy();
+
         this.showFormSave();
       },
     });
--- a/src/js/views/globals/modal/loading/loading_modal.scss
+++ b/src/js/views/globals/modal/loading/loading_modal.scss
@@ -1,0 +1,95 @@
+@keyframes reveal-modal {
+  0% {
+    opacity: 0;
+    transform: scale(0.5) translateY(30%);
+  }
+
+  100% {
+    opacity: 1;
+    transform: scale(1) translateY(0%);
+  }
+}
+
+@keyframes progress-loader {
+  0% {
+    transform: translate3d(0, 0, 0);
+  }
+
+  5% {
+    width: 100px;
+  }
+
+  10% {
+    transform: translate3d(192px, 0, 0);
+    width: 64px;
+  }
+
+  15% {
+    width: 100px;
+  }
+
+  20% {
+    transform: translate3d(0, 0, 0);
+    width: 64px;
+  }
+
+  25% {
+    width: 100px;
+  }
+
+  30% {
+    transform: translate3d(200px, 0, 0);
+    width: 64px;
+  }
+
+  35% {
+    width: 100px;
+  }
+
+  40% {
+    transform: translate3d(0, 0, 0);
+    width: 64px;
+  }
+
+  95% {
+    width: 256px;
+  }
+
+  100% {
+    width: 256px;
+  }
+}
+
+.loading-modal__content {
+  background-color: $white;
+  border-radius: 4px;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  line-height: 1;
+  min-height: 262px;
+  text-align: center;
+  width: 524px;
+
+  &.is-animated {
+    animation: 0.5s cubic-bezier(0.165, 0.84, 0.44, 1) 0.5s reveal-modal forwards;
+    opacity: 0;
+  }
+}
+
+.loading-modal__bar {
+  background: $black-90;
+  border-radius: 8px;
+  height: 16px;
+  margin: 32px auto 0;
+  width: 256px;
+}
+
+.loading-modal__bar-progress {
+  animation: 4s ease-in-out 0.5s progress-loader forwards;
+  background: color(green);
+  border-radius: 8px;
+  height: 16px;
+  transform: translate3d(0, 0, 0) scale(1);
+  width: 64px;
+}
--- a/src/js/views/globals/modal/modal_views.js
+++ b/src/js/views/globals/modal/modal_views.js
@@ -1,9 +1,11 @@
-import { extend } from 'underscore';
+import { extend, delay } from 'underscore';
 import hbs from 'handlebars-inline-precompile';
 import { View, Region } from 'marionette';
 
 import 'scss/modules/buttons.scss';
 import 'scss/modules/modals.scss';
+
+import './loading/loading_modal.scss';
 
 import intl from 'js/i18n';
 
@@ -14,6 +16,7 @@ import PreloadRegion from 'js/regions/preload_region';
 import IframeFormBehavior from 'js/behaviors/iframe-form';
 
 import ModalTemplate from './modal.hbs';
+import LoadingModalTemplate from './loading/loading_modal.hbs';
 
 const i18n = intl.globals.modal.modalViews;
 
@@ -100,9 +103,21 @@ const IframeFormView = View.extend({
   template: hbs`<iframe src="/formapp/"></iframe>`,
 });
 
+
+const LoadingModalView = View.extend({
+  className: 'pos--absolute',
+  template: LoadingModalTemplate,
+  onAttach() {
+    delay(() => {
+      this.destroy();
+    }, 5000);
+  },
+});
+
 export {
   ModalView,
   SidebarModalView,
   SmallModalView,
   IframeFormView,
+  LoadingModalView,
 };
--- a/src/js/views/globals/root_views.js
+++ b/src/js/views/globals/root_views.js
@@ -18,6 +18,10 @@ import './tooltip.scss';
 const userActivityCh = Radio.channel('user-activity');
 const historyCh = Radio.channel('history');
 
+function preventRegionClose() {
+  return this.getRegion('region').hasView();
+}
+
 const AppView = View.extend({
   className: 'app-frame',
   template: hbs`
@@ -61,22 +65,22 @@ const TopRegionView = View.extend({
 
 const PreloaderRegionView = TopRegionView.extend({
   // NOTE: ensures preload can't close on user click
-  contains() {
-    return this.getRegion('region').hasView();
-  },
+  contains: preventRegionClose,
   onRender() {
     this.showChildView('region', new PreloaderView());
   },
 });
 
 const ModalRegionView = TopRegionView.extend({
+  className: 'fill-window--dark',
   behaviors: [{
     behaviorClass: TopRegionBehavior,
-    className: 'fill-window--dark',
+    className: 'is-shown',
   }],
-  initialize({ $body }) {
+  initialize(options) {
+    this.mergeOptions(options, ['$body', 'setLocation', 'contains']);
     this.region = this.getRegion('region');
-    this.$body = $body;
+
     const hotkeyCh = Radio.channel('hotkey');
     this.listenTo(hotkeyCh, 'close', this.empty);
   },
@@ -111,10 +115,6 @@ const ModalRegionView = TopRegionView.extend({
     // testView prevents closing layers below
     if (el === target || this.Dom.hasEl(el, target) || this !== testView) return true;
   },
-});
-
-const ModalSidebarRegionView = ModalRegionView.extend({
-  setLocation: noop,
 });
 
 const popDefaults = {
@@ -306,9 +306,10 @@ const RootView = CollectionView.extend({
     // Add lowest layer (z-index) to highest
     this.addChildView(this.appView);
     this.addRegionView('tooltip', new TooltipRegionView({ $body }));
-    this.addRegionView('modalSidebar', new ModalSidebarRegionView({ $body }));
+    this.addRegionView('modalSidebar', new ModalRegionView({ $body, setLocation: noop }));
     this.addRegionView('modal', new ModalRegionView({ $body }));
     this.addRegionView('modalSmall', new ModalRegionView({ $body }));
+    this.addRegionView('modalLoading', new ModalRegionView({ $body, contains: preventRegionClose }));
     this.addRegionView('alert', new TopRegionView());
     this.addRegionView('pop', new PopRegionView({ $body }));
     this.addRegionView('preloader', new PreloaderRegionView());
--- a/test/integration/forms/form.js
+++ b/test/integration/forms/form.js
@@ -1208,6 +1208,8 @@ context('Patient Action Form', function() {
   specify('save and go back button', function() {
     localStorage.setItem('form-state_11111', JSON.stringify({ saveButtonType: 'saveAndGoBack' }));
 
+    cy.clock();
+
     cy
       .routeAction(fx => {
         fx.data.id = '1';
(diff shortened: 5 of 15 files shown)


src/js/views/globals/root_views.js
```diff
@@ -113,6 +114,12 @@ const ModalRegionView = TopRegionView.extend({
   },
 });
 
+const ModalLoadingRegionView = ModalRegionView.extend({
+  contains() {
+    return true;
+  },
```
nmajor25: This stops user clicks from closing the modal. Or doing anything else in the app while the modal is open.

src/js/views/globals/modal/loading/loading_modal.scss
```diff
@@ -0,0 +1,98 @@
...
+.loading-modal__bar {
+  background: #E6E6E6;
+  border-radius: 8px;
+  height: 16px;
+  margin: 32px auto 0;
+  width: 256px;
+}
+
+.loading-modal__bar-progress {
+  animation: 4.5s ease-in-out 0.5s progress-loader forwards;
```
nmajor25: Set this to `4.5s` instead of `5s` to ensure the animation completely finishes before the page is redirected and the modal is destroyed.

src/js/apps/forms/form/form_app.js
```diff
@@ -105,19 +105,22 @@ export default App.extend({
...
-        this.showFormStatus();
-        this.showFormUpdate();
+        if (saveButtonType !== 'saveAndGoBack') {
+          response.set({ _created_at: dayjs().format() });
+          this.responses.unshift(response);
+          this.setState({ responseId: response.id });
+
+          this.showFormStatus();
+          this.showFormUpdate();
+        }
```
nmajor25: If user is using the `Save + Go Back` button, don't update the form content, status, or save button while the loading modal (showing the `Saving your work...` animation) is open.

src/js/apps/forms/form/form_app.js
```diff
@@ -105,19 +105,22 @@ export default App.extend({
...
-          Radio.request('history', 'go:back');
-          return;
+          this.listenTo(this.loadingModal, 'destroy', () => {
+            Radio.request('history', 'go:back');
+          });
         }
 
-        this.showFormStatus();
-        this.showFormUpdate();
+        if (saveButtonType !== 'saveAndGoBack') {
```
paulfalgout: I think we could still just add a `return` to the above `if` and avoid and `if` here

src/js/apps/forms/form/form_app.js
```diff
@@ -105,19 +105,22 @@ export default App.extend({
 
     this.listenTo(formService, {
       'success'(response) {
-        response.set({ _created_at: dayjs().format() });
```
paulfalgout, line 108: Interestingly this didn't make a ton of sense with `saveAndGoBack` anyhow

src/js/apps/forms/form/form_app.js
```diff
@@ -283,6 +286,12 @@ export default App.extend({
       'click:save'() {
         Radio.request(`form${ this.form.id }`, 'send', 'form:submit');
```
paulfalgout: I realize it's an essentially impossible race condition, but I think this request should be the last this to occur in this function so that the modal exists if/when necessary

src/js/views/globals/modal/loading/loading_modal.scss
```diff
@@ -0,0 +1,98 @@
...
+  background: #00BF69;
```
paulfalgout: `background: color(green);`

src/js/views/globals/modal/loading/loading_modal.scss
```diff
@@ -0,0 +1,98 @@
...
+    opacity: 0;
+  }
+}
+
+.loading-modal__text {
+  line-height: 1;
+}
+
+.loading-modal__bar {
+  background: #E6E6E6;
```
paulfalgout: $black-90

src/js/views/globals/root_views.js
```diff
@@ -113,6 +114,12 @@ const ModalRegionView = TopRegionView.extend({
   },
 });
 
+const ModalLoadingRegionView = ModalRegionView.extend({
```
paulfalgout: I think in this case we should go ahead and remove `ModalSidebarRegionView` and make both `contains` and `setLocation` options that can be passed in on instantiation.

Also I think the `contains` should be the same function that's on the preloader region.  It's a little unorthodox, but I think we could do something like:
```js
function preventRegionClose() {
  return this.getRegion('region').hasView();
}
```
Then `contains: preventRegionClose,` for the preload region and the `modalLoading`

src/js/views/globals/modal/modal_views.js
```diff
@@ -100,9 +103,21 @@ const IframeFormView = View.extend({
   template: hbs`<iframe src="/formapp/"></iframe>`,
 });
 
+
+const LoadingModalView = View.extend({
+  className: 'loading-modal__wrapper',
```
paulfalgout: there's a `pos--absolute` class.. should probably be `u-pos--` but /shrug

src/js/views/globals/modal/loading/loading_modal.scss
```diff
@@ -0,0 +1,98 @@
...
+  text-align: center;
+  width: 524px;
+
+  &.is-animated {
+    animation: blow-up-modal 0.5s cubic-bezier(0.165, 0.84, 0.44, 1) forwards;
+    opacity: 0;
+  }
+}
+
+.loading-modal__text {
```
paulfalgout: I think we can probably dump this class.  Add it to `__content` if needed?

src/js/apps/forms/form/form_app.js
```diff
@@ -283,6 +286,12 @@ export default App.extend({
...
         this.showFormSaveDisabled();
+
+        const saveButtonType = this.getState('saveButtonType');
+
+        if (saveButtonType === 'saveAndGoBack') {
+          this.loadingModal = Radio.request('modal', 'show:loading');
```
paulfalgout: this can't happen in this button because the form can error out prior to even being submitted to the server.

In the form submission inside the form service we are triggering `success` and `error` after the submission to the server.  We need to add a `trigger` there for `submit` where we listen to that event in this app and show the modal.  Then we need to handle the modal for a success or destroy the modal for an error.

src/js/views/globals/modal/loading/loading_modal.scss
```diff
@@ -0,0 +1,98 @@
...
+  border-radius: 4px;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  min-height: 262px;
+  text-align: center;
+  width: 524px;
+
+  &.is-animated {
+    animation: blow-up-modal 0.5s cubic-bezier(0.165, 0.84, 0.44, 1) forwards;
```
paulfalgout: let's change this to `0.5s cubic-bezier(0.165, 0.84, 0.44, 1) 0.5s reveal-modal forwards;` I think I copied "blow-up" from some example.  Then this'll add a 0.5s delay overall and we can reduce the other animation to 4s from 4.5s.  I _think_ we want a little initial delay for form errors not feeling odd.

paulfalgout (changes requested):
I'm going to pull this down and play with it, but I have some requests.

src/js/apps/forms/form/form_app.js
nmajor25: Yes that works. Updated in last push.

src/js/apps/forms/form/form_app.js
nmajor25: The showing of the modal was removed from this function. So no longer an issue.

src/js/views/globals/modal/loading/loading_modal.scss
nmajor25: Updated in last push.

src/js/views/globals/modal/loading/loading_modal.scss
nmajor25: Updated in last push.

src/js/views/globals/modal/modal_views.js
nmajor25: Updated in last push.

src/js/views/globals/modal/loading/loading_modal.scss
nmajor25: Removed `.loading-modal__text` and moved `line-height: 1` to `.loading-modal__content`. Without `line-height: 1`, there's some unnecessary spacing between the text and progress bar.

src/js/apps/forms/form/form_app.js
nmajor25: Ah good call, totally missed this. Updated in the last push.

src/js/views/globals/modal/loading/loading_modal.scss
nmajor25: Updated in the last push.

src/js/apps/forms/form/form_app.js
```diff
@@ -104,25 +104,37 @@ export default App.extend({
...
 
     this.listenTo(formService, {
-      'success'(response) {
-        response.set({ _created_at: dayjs().format() });
-        this.responses.unshift(response);
-        this.setState({ responseId: response.id });
+      'submit'() {
+        const saveButtonType = this.getState('saveButtonType');
 
+        if (saveButtonType === 'saveAndGoBack') {
```
paulfalgout: I think this should be a guard `!==` `return;`

src/js/apps/forms/form/form_app.js
```diff
@@ -104,25 +104,37 @@ export default App.extend({
...
+          this.loadingModal = Radio.request('modal', 'show:loading');
+        }
+      },
+      'success'(response) {
         const saveButtonType = this.getState('saveButtonType');
 
         if (saveButtonType === 'saveAndGoBack') {
-          Radio.request('history', 'go:back');
+          this.listenTo(this.loadingModal, 'destroy', () => {
```
paulfalgout, line 118: now that these are closer together.. I think we can add this handler in the submit.  All we need is a guard in this function and the destroy in the error.

paulfalgout (changes requested):
Couple of small changes. The `preventRegionClose` request will also solve the cypress errors I think

src/js/views/globals/root_views.js
nmajor25: Is this what you had in mind for this: https://github.com/RoundingWell/care-ops-frontend/compare/d2a60b3358e3ffbf471d940bcda967b7ebc4afa8..24c037a16d0ddd932548942b1d6fd3b72919da15?

src/js/views/globals/root_views.js
```diff
@@ -61,22 +65,25 @@ const TopRegionView = View.extend({
...
     behaviorClass: TopRegionBehavior,
-    className: 'fill-window--dark',
+    className: 'is-shown',
   }],
-  initialize({ $body }) {
+  initialize({ $body, setLocation, contains }) {
     this.region = this.getRegion('region');
     this.$body = $body;
+
+    if (setLocation) this.setLocation = setLocation;
```
paulfalgout: this could be `this.mergeOptions(options, ['$body', 'setLocation', 'contains'])`. We probably use that pattern more when there's more than a couple options..

src/js/apps/forms/form/form_app.js
nmajor25: Just updated this: https://github.com/RoundingWell/care-ops-frontend/compare/24c037a16d0ddd932548942b1d6fd3b72919da15..77e62275ddd0c4059624da6b8b9e93a5fb8771f3.

src/js/views/globals/root_views.js
nmajor25: Nice, definitely like that better. Just updated that [here](https://github.com/RoundingWell/care-ops-frontend/compare/77e62275ddd0c4059624da6b8b9e93a5fb8771f3..adfede0cd13accd800beda4e240462bba6f2e5f8).

test/integration/forms/form.js
```diff
@@ -1313,6 +1314,18 @@ context('Patient Action Form', function() {
...
+      .should('contain', 'Saving your work...');
+
+    cy
+      .get('.app-frame__content')
+      .click('left', { force: true });
+
+    cy
+      .get('.fill-window--dark.is-shown')
+      .should('exist');
+
```
nmajor25, line 1329: This makes sure the loading modal is shown after the `Save + Go Back` button is clicked. And that it doesn't close when the user clicks elsewhere on the page.

test/integration/forms/form.js
```diff
@@ -1332,7 +1345,8 @@ context('Patient Action Form', function() {
         expect(data.attributes.response.data.familyHistory).to.equal('Here is some typing');
         expect(data.attributes.response.data.storyTime).to.equal('Once upon a time...');
         expect(data.attributes.response.data.patient.fields.weight).to.equal(192);
-      });
+      })
+      .wait(6000);
```
nmajor25: Wait `6` seconds before verifying that the page was redirected properly (added extra second to be safe).

Looking at this now, maybe it would read better if this was its own `cy.wait(6000)`. Instead of attaching it to this block...

test/integration/services/modal.js
```diff
@@ -194,7 +194,7 @@ context('Modal Service', function() {
       .contains('Custom Footer');
 
     cy
-      .get('.fill-window--dark')
+      .get('.fill-window--dark.is-shown')
```
nmajor25, line 197: These were just grabbing the wrong element, since we changed the class name for these to have `.is-shown`.

test/integration/forms/form.js
```diff
@@ -1294,15 +1294,16 @@ context('Patient Action Form', function() {
...
+
     cy
       .iframe()
       .find('textarea[name="data[familyHistory]"]')
       .type('Here is some typing');
 
-    cy
-      .get('.picklist')
-      .should('not.exist');
-
```
nmajor25: This was failing in the tests for some reason. Its purpose is to make sure the save button picklist closes when a user clicks or tabs into the iframe.

The `.type()` wasn't triggering that for some reason. So I added a `cy.tab()` instead above this and verified that the save button picklist closes as expected.

test/integration/forms/form.js
```diff
@@ -2044,8 +1968,113 @@ context('Preview Form', function() {
...
+      .click();
+
+    cy
+      .url()
+      .should('contain', '/worklist/owned-by');
+
+    cy
+      .go('back');
+  });
+});
```
nmajor25, line 2249: The `'Save and go back button'` test for the patient form was in the wrong `context()` section of the file. 

Was in `context('Preview Form')`. But should be in `context('Patient Form')` instead.

A mistake by me in the previous PR for this: https://github.com/RoundingWell/care-ops-frontend/pull/806.

test/integration/forms/form.js
```diff
@@ -1334,6 +1347,8 @@ context('Patient Action Form', function() {
         expect(data.attributes.response.data.patient.fields.weight).to.equal(192);
       });
 
+    cy.wait(6000);
```
paulfalgout: so a `6000` wait is super long.  there's a `cy.clock()` where we can force that to happen.  If you search for `clock()` I think you can find examples

test/integration/services/modal.js
paulfalgout: seems odd that this is needed.. was `get` getting more than one or something?

paulfalgout (changes requested):

test/integration/services/modal.js
nmajor25: Ya, the `get` is returning other `<div class="fill-window--dark">` elements, that don't have `.is-shown`.

But I also find it odd too...

src/js/views/globals/root_views.js
```diff
@@ -61,22 +65,22 @@ const TopRegionView = View.extend({
...
     this.showChildView('region', new PreloaderView());
   },
 });
 
 const ModalRegionView = TopRegionView.extend({
+  className: 'fill-window--dark',
   behaviors: [{
     behaviorClass: TopRegionBehavior,
-    className: 'fill-window--dark',
+    className: 'is-shown',
```
nmajor25, line 78: These changes causing the modal test oddity?

Now every modal has the `.fill-window--dark` class name, even if it's not being shown.  Before it only had that class if it was being shown. So Cypress is `get()`ing more elements now than it was before.

test/integration/forms/form.js
nmajor25: Added a `cy.tick(5000)` to both the patient action and patient form tests.

For some reason, the `Patient Action Form > Save and go back button` test is hanging/not finishing without any errors thrown in the console. Not sure why that's happening.

The `Patient Form > Save and go back button` passes fine.

test/integration/forms/form.js
paulfalgout: I don't think this is testing the same thing

paulfalgout: When I put this back I can't get it to fail, but the `tab` will always close the picklist.. what was being tested here is that typing within the iframe closes the picklist.

nmajor25: I just removed the `tab()` version in my last push. And added the original one back, that simply does this when typing into the iframe:

```js
cy
  .get('.picklist')
  .should('not.exist');
```

src/js/apps/forms/form/form_app.js
```diff
@@ -104,25 +104,37 @@ export default App.extend({
...
+            Radio.request('history', 'go:back');
+          });
+
```
nmajor25, line 121: Moved this listener from the `'submit'` event to the `'success'` event.

If it's placed in the `'submit'` listener, `Radio.request('history', 'go:back');` will be fired when a form error occurs and the modal is destroyed (via `'error'` event). 

We only want the redirect when the  modal is destroyed after a successful form submission. Not when it's destroyed after a form submission error.

PR #869, opened 2022-12-07
Replace preload spinner with a new loader animation

nmajor25:
Shortcut Story ID: [sc-32401]

Replace each instance of the preload spinner with a new loader animation:

ezgif-4-e64e22bd81

### DIFF
--- a/src/js/formapp.js
+++ b/src/js/formapp.js
@@ -14,6 +14,8 @@ import Backbone from 'backbone';
 import Handlebars from 'handlebars/runtime';
 import parsePhoneNumber from 'libphonenumber-js/min';
 
+import intl from 'js/i18n';
+
 import { versions } from './config';
 
 import {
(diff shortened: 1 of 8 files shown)


src/js/formapp.js
```diff
@@ -208,7 +210,12 @@ const Router = Backbone.Router.extend({
 });
 
 function startFormApp() {
-  $('#root').append(`<div class="spinner"><div class="spinner-circle">${ '<div class="spinner-child"></div>'.repeat(12) }</div></div>`);
+  $('#root').append(`
+    <div class="preloader__bar js-progress-bar">
+      <div class="preloader__bar-progress"></div>
+    </div>
+    <div class="preloader__text js-loading">${ intl.regions.preload.loading }</div>
+  `);
```
nmajor25, line 218: This is for the loader that is shown in the form iframe.

Before this PR, the spinner was shown without the `Loading...` text.

Do we want the `Loading...` text here? Or just the progress bar animation by itself?

paulfalgout (approved):

PR #871, opened 2022-12-08
Click+shift multi select for schedule items

nmajor25:
Shortcut Story ID: [sc-32186]

When an item is selected in the schedule, allow a user to `shift`+`click` to select multiple list items between the two.

### DIFF
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -179,7 +179,7 @@ const DayItemView = View.extend({
     this.isReduced = state.get('isReduced');
 
     this.listenTo(state, {
-      'select:all': this.showCheck,
+      'select:multiple': this.showCheck,
       'select:none': this.showCheck,
     });
   },
@@ -198,8 +198,8 @@ const DayItemView = View.extend({
     const checkComponent = new CheckComponent({ state: { isSelected } });
 
     this.listenTo(checkComponent, {
-      'select'() {
-        this.state.toggleSelected(this.model, !isSelected);
+      'select'(domEvent) {
+        this.triggerMethod('select', this, !!domEvent.shiftKey);
       },
       'change:isSelected': this.toggleSelected,
     });
@@ -273,6 +273,10 @@ const DayListView = CollectionView.extend({
   },
   childViewTriggers: {
     'render': 'listItem:render',
+    'select': 'select',
+  },
+  onSelect(selectedView, isShiftKeyPressed) {
+    this.triggerMethod('select:list:item', selectedView, isShiftKeyPressed);
   },
   onListItemRender(view) {
     const date = dayjs(this.model.get('date'));
@@ -324,6 +328,9 @@ const ScheduleListView = CollectionView.extend({
       state: this.state,
     };
   },
+  childViewTriggers: {
+    'select:list:item': 'select',
+  },
   childViewEvents: {
     'render:children': 'onChildFilter',
   },
@@ -349,8 +356,9 @@ const ScheduleListView = CollectionView.extend({
 
     return true;
   },
-  initialize({ state }) {
+  initialize({ state, originalCollection }) {
     this.state = state;
+    this.originalCollection = originalCollection;
   },
   onRenderChildren() {
     this.setVisibleChildren();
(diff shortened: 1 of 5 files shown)


src/js/views/patients/schedule/schedule_views.js
```diff
@@ -364,6 +372,55 @@ const ScheduleListView = CollectionView.extend({
...
+        return a.due_time.localeCompare(b.due_time);
+      });
+
+      each(sortedCollectionOfActions, action => {
+        sortedArrayOfActions.push(action);
+      });
+    });
+
+    return sortedArrayOfActions;
+  },
```
nmajor25, line 413: To get the correct index position of a selected list item, I needed to create a flattened version of the collection and sort those actions by date.

Convert this:

```js
[
  {
    day: {
      actions: [ {action}, {action} ]
    }
  },
  {
    day: {
      actions: [ {action} ]
     }
   }
]
```

Into this:

```js
[
  {action}, {action}, {action}
]
```

Where actions are sorted by due day and time.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -364,6 +372,55 @@ const ScheduleListView = CollectionView.extend({
...
+  handleClickShiftMultiSelect(sortedArrayOfActions, indexOfSelectedAction, lastSelectedIndex) {
+    const minIndex = Math.min(indexOfSelectedAction, lastSelectedIndex);
+    const maxIndex = Math.max(indexOfSelectedAction, lastSelectedIndex);
+
+    const selectedIds = sortedArrayOfActions.map(action => action.id).slice(minIndex, maxIndex + 1);
+
+    this.state.selectMultiple(selectedIds, indexOfSelectedAction);
+  },
+  getSortedArrayOfActions() {
+    const collectionJSON = this.collection.toJSON();
```
paulfalgout: so the this view is being passed `this.collection.groupByDate()` as the collection.  I wonder if we should also pass it the original collection?  Then we just need to do what's in `sortDueAsc`

paulfalgout (changes requested):
I think this is looking pretty close.

src/js/views/patients/schedule/schedule_views.js
nmajor25: Updated to use the original collection in my latest push. I think I have it simplified as much as possible... open to suggestions though.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -364,6 +372,45 @@ const ScheduleListView = CollectionView.extend({
...
+  },
+  getSortedCollection() {
+    const sortedCollection = this.originalCollection;
+
+    sortedCollection.comparator = function(a, b) {
+      const dueDateA = a.get('due_date');
+      const dueDateB = b.get('due_date');
+
+      if (dueDateA === dueDateB) {
+        return alphaSort('asc', a.get('due_time'), b.get('due_time'), '24');
```
paulfalgout, line 406: why the `24`?  is there a bug in our worklist sort?

nmajor25: I added it to match the `viewComparator` used on the `DayListView` ([here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/views/patients/schedule/schedule_views.js#L266)). To ensure the sorting method is same here vs. on the view.

Not sure about worklist. Will investigate.

nmajor25: On the worklist, actions with `due_time: null` are placed either above or below actions that possess a `due_time`. Depends whether the sort is `asc` or `desc`. 

Do we want `due_time: null` actions to always be on the bottom? Adding `'24'` to the worklist would force those to the bottom always.

Example of `ascending` order:

Screen Shot 2022-12-12 at 3 46 41 PM

_Note: `Due: Newest - Oldest` and `Due: Oldest - Newest` don't seem to be sorting in the correct directions. Think those need to be switched._

Compared to the schedule, where those are on the bottom:

Screen Shot 2022-12-12 at 3 47 27 PM

paulfalgout: I _think_ the sort order is correct, though it is confusing.. but consider over due dates.. those would be oldest and would be at the top.. and the "newest" is the latest due.. But yeah the `24` seems better in this case I suppose..  so does the due date sort match the worklist where "oldest" indicates either overdue or soonest to be due?  seems like yes?

paulfalgout (approved):

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -276,7 +276,7 @@ const sortDueOptions = [
       const dueA = a.model.get('due_date');
       const dueB = b.model.get('due_date');
       if (dueA === dueB) {
-        return alphaSort('asc', a.model.get('due_time'), b.model.get('due_time'));
+        return alphaSort('asc', a.model.get('due_time'), b.model.get('due_time'), '24');
```
nmajor25, line 279: When sorting the row for the `Due: Oldest - Newest` option, this forces actions with a `due_date` to be on top of actions without one (`due_date: null`).

The `'24'` is not needed to achieve this for the `Due: Newest - Oldest` sort option.

PR #872, opened 2022-12-09
Replace NPM package: `handlebars-template-loader` => `handlebars-loader`

nmajor25:
Shortcut Story ID: [sc-32351]

[handlebars-template-loader](https://www.npmjs.com/package/handlebars-template-loader) relies on an older version of [loader-utils](https://www.npmjs.com/package/loader-utils) that was causing a NPM security issue to be shown when doing `npm i`.

This PR switches the app over to use [handlebars-loader](https://www.npmjs.com/package/handlebars-loader) instead. Which updated their dependencies to use the newer version of `loader-utils` that doesn't cause the security issue.

### DIFF
--- a/config/webpack.rules.js
+++ b/config/webpack.rules.js
@@ -15,7 +15,7 @@ const babelLoader = {
 const hbsLoader = {
   test: /\.hbs?$/,
   exclude: /node_modules/,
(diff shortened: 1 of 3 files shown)


paulfalgout:
I'm kind of shocked that it was that easy

PR #874, opened 2022-12-13
Fix worklist actions alignment bug

nmajor25:
Shortcut Story ID: [sc-32560]

Fix a bug where the `State, Owner, Due, Form` section for worklist actions don't align.

Before:

Screen Shot 2022-12-13 at 3 40 47 PM

After:

Screen Shot 2022-12-13 at 3 41 10 PM

### DIFF
--- a/src/js/views/patients/worklist/worklist-list.scss
+++ b/src/js/views/patients/worklist/worklist-list.scss
@@ -27,6 +27,7 @@
   color: #999;
   display: inline-block;
   vertical-align: middle;
+  width: 16px;
 }
 
 .work-list__item:hover {


src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -27,6 +27,7 @@
   color: #999;
   display: inline-block;
   vertical-align: middle;
+  width: 16px;
```
nmajor25, line 30: For the alignment to be correct, the details icon`<div>` needs to have a `width: 16px`. Whether the details icon `<svg>` is being displayed or not.

We removed this from the details icon in PR: https://github.com/RoundingWell/care-ops-frontend/pull/856.

Specifically via this code change: https://github.com/RoundingWell/care-ops-frontend/pull/856/files#r1026919485.

paulfalgout (approved):

PR #875, opened 2022-12-13
Use org settings state status for worklists/schedule

nmajor25:
Shortcut Story ID: [sc-32452]

Worklists/schedule currently use static variables for filtering state status.

This PR updates the following pages to use `filter[state]` instead of `filter[status]` in the API requests:

- Worklist
- Schedule
- Reduced schedule
- Patient dashboard
- Patient archive

### DIFF
--- a/src/js/apps/patients/patient/archive/archive_app.js
+++ b/src/js/apps/patients/patient/archive/archive_app.js
@@ -3,8 +3,6 @@ import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
-import { STATE_STATUS } from 'js/static';
-
 import { LayoutView, ListView } from 'js/views/patients/patient/archive/archive_views';
 
 export default App.extend({
@@ -13,7 +11,9 @@ export default App.extend({
     this.getRegion('content').startPreloader();
   },
   beforeStart({ patient }) {
-    const filter = { status: STATE_STATUS.DONE };
+    const currentOrg = Radio.request('bootstrap', 'currentOrg');
+    const states = currentOrg.getStates();
+    const filter = { state: states.groupByDone().done.getFilterIds() };
 
     return [
       Radio.request('entities', 'fetch:actions:collection:byPatient', { patientId: patient.id, filter }),
--- a/src/js/apps/patients/patient/dashboard/dashboard_app.js
+++ b/src/js/apps/patients/patient/dashboard/dashboard_app.js
@@ -3,8 +3,6 @@ import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
-import { STATE_STATUS } from 'js/static';
-
 import AddWorkflowApp from './add-workflow_app';
 
 import { LayoutView, ListView } from 'js/views/patients/patient/dashboard/dashboard_views';
@@ -21,7 +19,10 @@ export default App.extend({
   },
 
   beforeStart({ patient }) {
-    const filter = { status: [STATE_STATUS.QUEUED, STATE_STATUS.STARTED].join(',') };
+    const currentOrg = Radio.request('bootstrap', 'currentOrg');
+    const states = currentOrg.getStates();
+    const filter = { state: states.groupByDone().notDone.getFilterIds() };
+
     return [
       Radio.request('entities', 'fetch:actions:collection:byPatient', { patientId: patient.id, filter }),
       Radio.request('entities', 'fetch:flows:collection:byPatient', { patientId: patient.id, filter }),
--- a/src/js/apps/patients/schedule/schedule_state.js
+++ b/src/js/apps/patients/schedule/schedule_state.js
@@ -4,7 +4,7 @@ import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 import dayjs from 'dayjs';
 
-import { STATE_STATUS, RELATIVE_DATE_RANGES } from 'js/static';
+import { RELATIVE_DATE_RANGES } from 'js/static';
 
 const relativeRanges = new Backbone.Collection(RELATIVE_DATE_RANGES);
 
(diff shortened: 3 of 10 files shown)


src/js/apps/patients/patient/archive/archive_app.js
```diff
@@ -13,7 +11,9 @@ export default App.extend({
     this.getRegion('content').startPreloader();
   },
   beforeStart({ patient }) {
-    const filter = { status: STATE_STATUS.DONE };
+    const currentOrg = Radio.request('bootstrap', 'currentOrg');
+    const states = currentOrg.getStates();
+    const filter = { status: states.filterByDone(true).pluck('status').join(',') };
```
paulfalgout: Ah so this should be `state: states.filterBy...pluck('id').join...`. We don't want to filter by status any longer.

paulfalgout (changes requested):
So at the end of this, I suspect the only changes to the tests need to be replacing `filter[status]` with `filter[state]`s

src/js/apps/patients/patient/archive/archive_app.js
nmajor25: Oops missed that...

src/js/apps/patients/patient/dashboard/dashboard_app.js
```diff
@@ -21,7 +19,10 @@ export default App.extend({
   },
 
   beforeStart({ patient }) {
-    const filter = { status: [STATE_STATUS.QUEUED, STATE_STATUS.STARTED].join(',') };
+    const currentOrg = Radio.request('bootstrap', 'currentOrg');
+    const states = currentOrg.getStates();
+    const filter = { state: states.filterByDone().pluck('id').join(',') };
+
```
nmajor25, line 25: This doesn't work in my local sandbox.

From what I can tell, both the `/patients/{patientId}/relationships/actions` and `/patients/{patientId}/relationships/flows` API endpoints are still using `filter[status]` instead of `filter[state]`.

shadowhand: @nmajor25 🐞 you found a bug, those endpoints are indeed still using status, and should not be. I will create a PR.

src/js/entities-service/entities/states.js
```diff
@@ -17,6 +18,15 @@ const Model = Store(_Model, TYPE);
...
   model: Model,
+  filterByDone(onlyIncludeDone = false) {
+    const filteredStates = this.filter(state => {
+      if (onlyIncludeDone) return state.isDone();
+
+      return !state.isDone();
+    });
+
+    return new Backbone.Collection(filteredStates);
+  },
```
nmajor25, line 32: Returning a new backbone collection instead of modifying `this`.

In the `worklist_state.js` file, I need to call `states.filterByDone()`  twice. One for all non-done states and another for only done states (used for the `Done < 30 Days` page).

Reseting the collection makes the only done statuses unaccessible, as they're cleared when `states.filterByDone()` is first called.

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -107,29 +108,31 @@ export default Backbone.Model.extend({
...
   getEntityFilter() {
+    const states = this.currentOrg.getStates();
     const filtersState = this.getFilters();
     const clinicianId = this.get('clinicianId');
     const teamId = this.get('teamId');
     const noOwner = this.get('noOwner');
     const customFilters = omit(filtersState, 'groupId');
-    const status = [STATE_STATUS.QUEUED, STATE_STATUS.STARTED].join(',');
+    const state = states.filterByDone().pluck('id').join(',');
+    const doneState = states.filterByDone(true).pluck('id').join(',');
```
nmajor25: For reference, this is the reason why I changed `states.filterByDone()` away from doing `this.reset()`.

`const doneState` would always returns empty because the previous `states.filterByDone()` clears all the done states.

src/js/entities-service/entities/states.js
paulfalgout: I think this could probably `this.clone().reset(this.filter`

It shouldn't just `new` a vanilla collection as it won't be a `BaseCollection` or a `states` collection.  In this case `new Collection` is the States Collection.

However since it is the case that we're calling this twice once `true` once `false` I think maybe we should change this to `groupByDone` and do something like:

```js
const { done, notDone } = this.groupBy(state => {
  return state.isDone() ? 'done' : 'notDone';
});

return {
  done: new Collection(done),
  notDone: new Collection(notDone),
};
```
Not sure if there's a better term for `notDone`

paulfalgout (changes requested):
FWIW I think we only _need_ to modify the worklist that has the sidebar filters and don't necessarily have to touch the patient dashboard/archive, but I suppose there's no harm in doing so.

src/js/apps/patients/schedule/schedule_state.js
```diff
@@ -75,13 +78,13 @@ export default Backbone.Model.extend({
     const filtersState = this.getFilters();
     const clinicianId = this.get('clinicianId');
     const customFilters = omit(filtersState, 'groupId');
-    const status = [STATE_STATUS.QUEUED, STATE_STATUS.STARTED].join(',');
+    const state = this.states.filterByDone().pluck('id').join(',');
```
paulfalgout: I also no it's not a huge deal, but I think we're doing it enough places we should make a:
```js
getFilterIds() {
  return this.map('id').join(',');
}
```
on the state collection..

I dunno why I always default to `pluck`, but it's essentially an alias of `map` with no benefit.

paulfalgout (changes requested):

src/js/entities-service/entities/states.js
nmajor25: Great way to do it. Updated to this in my last push.

src/js/apps/patients/schedule/schedule_state.js
nmajor25: Added this in the last push I just made.

PR #876, opened 2022-12-15
Rename worklist due ordering to `"Due: Sooner - Later"`

nmajor25:
Shortcut Story ID: [sc-32517]

On the worklist, rename the due order sorting label from `Due: Oldest - Newest` to `Due: Sooner - Later`.

### DIFF
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -603,8 +603,8 @@ careOptsFrontend:
         sortDroplist:
           headingText: 'Sort Options'
         sortDueOptions:
-          asc: 'Due: Oldest - Newest'
-          desc: 'Due: Newest - Oldest'
+          asc: 'Due: Sooner - Later'
+          desc: 'Due: Later - Sooner'
         sortPatientOptions:
           asc: 'Patient Last: A - Z'
           desc: 'Patient Last: Z - A'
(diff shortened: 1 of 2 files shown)


test/integration/patients/worklist/worklist.js
```diff
@@ -2598,6 +2598,8 @@ context('worklist page', function() {
       .routeAction()
       .routeActionActivity()
       .routePatientFlows()
+      .routeActionComments()
+      .routeActionFiles()
```
nmajor25, line 2602: Their absence was causing `500` errors in the Cypress tests.

src/js/i18n/en-US.yml
```diff
@@ -603,8 +603,8 @@ careOptsFrontend:
         sortDroplist:
           headingText: 'Sort Options'
         sortDueOptions:
-          asc: 'Due: Oldest - Newest'
-          desc: 'Due: Newest - Oldest'
+          asc: 'Due: Sooner - Later'
+          desc: 'Due: Later - Sooner'
```
nmajor25, line 607: Some screenshots of what these look like after the update.

`Due: Sooner - Later`:

Screen Shot 2022-12-15 at 4 40 36 PM

`Due: Later - Sooner`:

Screen Shot 2022-12-15 at 4 40 49 PM

shadowhand: It's odd to me that the "up/down" icon is the same in both directions.

nmajor25: We use the same icon for all options in that droplist. Not just these two.

shadowhand: Right, I didn't mean specific to this sort. Just in general it's weird to me that we don't switch between two icons, "up" and "down".

cc @markott

PR #878, opened 2022-12-20
Separate sort and add custom sort org setting

paulfalgout:
Shortcut Story ID: [sc-32648]

- [x] Add Tests

### DIFF
--- a/src/js/apps/patients/worklist/worklist_sort.js
+++ b/src/js/apps/patients/worklist/worklist_sort.js
@@ -1,0 +1,140 @@
+import { get, union, map, uniq, compact } from 'underscore';
+import Backbone from 'backbone';
+import Radio from 'backbone.radio';
+
+import { alphaSort } from 'js/utils/sorting';
+import { i18n } from 'js/views/patients/worklist/worklist_views';
+
+function getEntityFieldValue(entity, fieldName, keys) {
+  const patientField = entity.getPatient().getField(fieldName);
+  if (!patientField) return null;
+  return get(patientField.getValue(), keys);
+}
+
+// field_key may include a path for the model_attr.deeply_nested.value
+const SortOption = Backbone.Model.extend({
+  defaults: {
+    direction: 'desc',
+    field_key: 'value',
+    entities: ['actions', 'flows'],
+  },
+  getComparator() {
+    const comparator = this.get('comparator');
+    if (comparator) return comparator.bind(this);
+
+    const fieldName = this.get('field_name');
+    const keys = this.get('field_key').split('.');
+    const direction = this.get('direction');
+
+    return (a, b) => {
+      const aValue = getEntityFieldValue(a.model, fieldName, keys);
+      const bValue = getEntityFieldValue(b.model, fieldName, keys);
+
+      return alphaSort(direction, aValue, bValue);
+    };
+  },
+});
+
+const SortOptions = Backbone.Collection.extend({
+  model: SortOption,
+  getByType(entityType) {
+    const clone = this.clone();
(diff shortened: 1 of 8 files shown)


src/js/entities-service/actions.js
```diff
@@ -18,8 +18,8 @@ const Entity = BaseEntity.extend({
     ].join();
     return this.fetchModel(id, { data: { include } });
   },
-  fetchActions({ filter }) {
-    const data = { filter };
+  fetchActions({ filter, include }) {
+    const data = { filter, include };
```
nmajor25, line 22: What is the purpose of the `include` here? Asking out of curiosity.

shadowhand: Adding `include` is required for (eg) `?include=patient.patient-fields.scoring` to get the necessary data (value `ranking`) for Risk Sort.

shadowhand: https://roundingwell.slack.com/archives/C02U318AW/p1670864213694099

paulfalgout: It was a deprecated API on the endpoint, but it was added back so we can pass `patient.patient-field.foo`

nmajor25: Gotcha, thanks @paulfalgout @shadowhand

nmajor25 (approved):

src/js/apps/patients/worklist/worklist_sort.js
```diff
@@ -0,0 +1,140 @@
...
+    const fieldName = this.get('field_name');
+    const keys = this.get('field_key').split('.');
+    const direction = this.get('direction');
+
+    return (a, b) => {
+      const aValue = getEntityFieldValue(a.model, fieldName, keys);
+      const bValue = getEntityFieldValue(b.model, fieldName, keys);
+
+      return alphaSort(direction, aValue, bValue);
+    };
```
shadowhand, line 34: This needs to be updated to check for a sort option called `type` (or some such) with a default of `"alpha"` that also allows for `"numeric"`.

See [this bug][1].

[1]: https://app.shortcut.com/roundingwell/story/32716/custom-sort-on-a-number-is-using-alpha-sort-rather-than-numeric

src/js/apps/patients/worklist/worklist_sort.js
```diff
@@ -0,0 +1,140 @@
...
+
+    const fieldName = this.get('field_name');
+    const keys = this.get('field_key').split('.');
+    const direction = this.get('direction');
+
+    return (a, b) => {
+      const aValue = getEntityFieldValue(a.model, fieldName, keys);
+      const bValue = getEntityFieldValue(b.model, fieldName, keys);
+
+      return alphaSort(direction, aValue, bValue);
```
shadowhand, line 33: To resolve the issue with #879, this should read:

```js
return alphaSort(direction, aValue ?? '', bValue ?? '');
```

But to support `type: numeric` that would expand to something like:

```js
if (type === 'numeric') {
  return numSort(direction, aValue ?? 0, bValue ?? 0);
}

return alphaSort(direction, aValue ?? '', bValue ?? '');
```

src/js/apps/patients/worklist/worklist_sort.js
```diff
@@ -0,0 +1,140 @@
+import { get, union, map, uniq, compact } from 'underscore';
+import Backbone from 'backbone';
+import Radio from 'backbone.radio';
+
+import { alphaSort } from 'js/utils/sorting';
```
shadowhand, line 5: Need `numSort` here.

PR #830, opened 2022-10-25
Add Custom filters

paulfalgout:
Shortcut Story ID: [sc-31691] [sc-31235] [sc-31828]

Replaces https://github.com/RoundingWell/care-ops-frontend/pull/827, https://github.com/RoundingWell/care-ops-frontend/pull/826

PR #880, opened 2022-12-22
Add state status checklist to filters sidebar

nmajor25:
Shortcut Story ID: [sc-32574]

Ability to filter worklists and schedule by action/flow state status (e.g. `To Do`, `In Progress`, `Done`) via the filters sidebar. In addition and combined with filtering via group and custom filters (https://github.com/RoundingWell/care-ops-frontend/pull/830).

On the `Owned By`, `Shared By`, `New < 24hrs`, `Updated < 3 Days` worklist and the schedule, all available state statuses will be shown (done and not done categories). And the not done state statuses will be selected by default (i.e. on initial load or when the user resets the filters using the `Clear Filters` button).

Screenshot 2023-01-11 at 10 49 49 AM

The `Done < 30 Days` worklist will show only the available state statuses with the `done` attribute. And all `done` statuses will be selected by default.

Screenshot 2023-01-11 at 10 50 11 AM

Additional notes:

- If all state statuses are un-selected, a `nil_uuid` (`filter[state]=00000000-0000-0000-0000-000000000000`) is sent in the request to the backend and the worklist/schedule will be empty.
- When calculating/displaying the total filters count, the state statuses are either changed from the default selected (adds `1`) or not changed from the default selected (adds `0`).
- Local storage version was updated from `v3` => `v4` since a new `state.filters` array was added and would conflict with a user's current local storage settings.


src/js/apps/patients/worklist/filters_app.js
```diff
@@ -36,13 +48,31 @@ export default App.extend({
...
+    const statesFilterState = state.get('states');
+    const defaultStates = this.getDefaultStates().map('id').reduce((states, id) => {
+      states[id] = true;
+      return states;
+    }, {});
+
+    const statesFilterCount = !isEqual(defaultStates, statesFilterState) ? 1 : 0;
+
+    return customFiltersSize + statesFilterCount;
+  },
```
nmajor25: This calculates a total filters count. 

Number of custom filters + whether or not a states filter was selected (either `1` or `0`).

If the states filter is different than the default (user changed it), I count it as `1`. If the states filter has not been changed, I count it as `0`.

src/js/apps/patients/sidebar/filters-sidebar_app.js
```diff
@@ -39,10 +55,49 @@ export default App.extend({
...
+    return customFiltersSize + statesFilterCount;
+  },
+  getDefaultStates() {
+    const notDoneStates = this.states.groupByDone().notDone;
+    const doneStates = this.states.groupByDone().done;
+
+    if (this.listId === 'done-last-thirty-days') return doneStates;
+
+    return notDoneStates;
+  },
```
nmajor25: These two functions (`getFiltersCount()` and `getDefaultStates()`) are identical to what's in the `filters_app.js` file.

Ideally I'd like to just pass down the `filtersCount` value to this app (`filters_app.js` => `filters-sidebar_app.js`), but I was having a heard time getting it to update in the `HeaderView`, since the `filtersCount` isn't changing along with the state.

So need to figure that out yet.

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -32,7 +32,9 @@ export default Backbone.Model.extend({
         relativeDate: null,
       },
       isFiltering: false,
-      filters: {},
+      filters: {
+        states: {},
+      },
```
nmajor25: The structure for the filters state will look like this:

```js
filters: {
  insurance: 'value',
  groupId: 'id',
  states: {
    id: true,
    id: false // will be set to false when un-selected by a user
  }
}
```

`filters.states` matches how the selected actions/flows state is built.

Its initial values are loaded from local storage. If there's no local storage item for the list, the default states will be selected initially (via the `getInitialStatesObject()` function below). Otherwise, a user's previous filter choice will be persisted on reload/route changes.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -63,6 +65,11 @@ export default App.extend({
       storedState.lastSelectedIndex = null;
     }
 
+    if (!storedState) {
+      this.setState({ id: this.worklistId, filters: { states: initialStatesStatusObject } });
+      return;
+    }
```
nmajor25, line 69: If the user has no local storage data for the list, this sets the default states to all be selected.

Ideally, this would be done directly in the `worklist_state.js` file, but I need access to `this.worklistId`. Which is set after the state is initially created.

src/js/apps/patients/sidebar/filters-sidebar_app.js
```diff
@@ -39,10 +55,49 @@ export default App.extend({
...
+    this.showChildView('statesFilters', statesFiltersView);
+  },
   onClearFilters() {
-    this.getState().clear();
+    const defaultStates = this.defaultStates.map('id').reduce((states, id) => {
+      states[id] = true;
+      return states;
+    }, {});
+
+    this.getState().clear().set({ states: defaultStates });
```
nmajor25: When the state is cleared via the `Clear Filters` button, set the default states to be selected.

shadowhand:
Is there anything that the API could do differently to eliminate the need to track so much information?

src/js/services/bootstrap.js
```diff
@@ -26,6 +26,7 @@ export default App.extend({
     'currentOrg:setting': 'getOrgSetting',
     'currentOrg:roles': 'getOrgRoles',
     'currentOrg:directories': 'getOrgDirectories',
+    'currentOrg:states': 'getOrgStates',
```
paulfalgout: i don't think we want to introduce this.
```js
      const states = currentOrg.getStates();
```
is preferred because it prevents modifying the cache.
`getOrgDirectories` should likely be refactored as well for pattern's sake.. though doesn't have the same concerns.

src/js/apps/patients/sidebar/filters-sidebar_app.js
```diff
@@ -39,10 +55,49 @@ export default App.extend({
...
   },
   onStop() {
     Radio.request('sidebar', 'close');
   },
+  getFiltersCount() {
+    const state = this.getState();
+
+    const customFiltersState = omit(state.toJSON(), 'states');
```
paulfalgout: `state.omit('states')`

src/js/apps/patients/sidebar/filters-sidebar_app.js
```diff
@@ -39,10 +55,49 @@ export default App.extend({
...
+      states[id] = true;
+      return states;
+    }, {});
+
+    const statesFilterCount = !isEqual(defaultStates, statesFilterState) ? 1 : 0;
+
+    return customFiltersSize + statesFilterCount;
+  },
+  getDefaultStates() {
+    const notDoneStates = this.states.groupByDone().notDone;
```
paulfalgout: maybe just `const { done, notDone } = `

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
```diff
@@ -90,6 +92,83 @@ const CustomFiltersView = CollectionView.extend({
...
+  childView: StatesFilterView,
+  childViewOptions() {
+    return {
+      state: this.getOption('state'),
+    };
+  },
+  childViewTriggers: {
+    'select': 'select:state',
+  },
+  className: 'filters-sidebar__states-filters',
```
paulfalgout: I think this might just be `sidebar__section` with removing the `-t--16` from the `h3`

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
```diff
@@ -90,6 +92,83 @@ const CustomFiltersView = CollectionView.extend({
...
+  template: hbs`
```
paulfalgout: should probably make a `childViewContainer` for the collectionview template

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
```diff
@@ -90,6 +92,83 @@ const CustomFiltersView = CollectionView.extend({
...
+  `,
+  initialize({ state }) {
+    this.state = state;
+
+    this.listenTo(state, 'change:states', this.render);
+  },
+  onSelectState(model, isSelected) {
+    const currentStatesList = clone(this.state.get('states'));
+
+    const newStatesList = extend(currentStatesList, {
```
paulfalgout: you can avoid the clone by `extend({}, currentStatesList, {`

or keep the clone and `currentStatesList[model.id] = !isSelected`

but `clone(foo)` is essentially just sugar for `extend({}, foo)`

paulfalgout (changes requested):

src/js/views/patients/worklist/filters_views.js
```diff
@@ -32,9 +31,12 @@ const AllFiltersButtonView = View.extend({
   triggers: {
     click: 'click',
   },
+  initialize({ filtersCount }) {
+    this.filtersCount = filtersCount;
```
paulfalgout: this doesn't seem correct as it isn't calculated on `render`, but there's a `modelEvents` for change render

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -36,13 +48,31 @@ export default App.extend({
...
   showAllFiltersButtonView() {
     const directories = Radio.request('bootstrap', 'currentOrg:directories');
 
     if (!directories.length) return;
 
     const ownerView = this.showChildView('allFilters', new AllFiltersButtonView({
-      model: this.getState(),
```
paulfalgout, line 45: I think we should continue passing the state here and move `getFiltersCount` to the state model

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -63,6 +65,11 @@ export default App.extend({
       storedState.lastSelectedIndex = null;
     }
 
+    if (!storedState) {
```
paulfalgout: I think the `if (storedState)` should include a `setState` and `return` instead of adding an `if(!`

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -104,8 +111,13 @@ export default App.extend({
     this.startFiltersApp();
   },
   beforeStart() {
+    const filter = this.getState().getEntityFilter();
+
+    // remove when worklist state is updated to not include empty filter[state] param
+    if (!filter.state) return new Backbone.Collection([]);
+
     return Radio.request('entities', `fetch:${ this.getState().getType() }:collection`, {
```
paulfalgout: I think we should be able to just call this request as long as `filter[state]` doesn't end up `''`

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -107,32 +110,56 @@ export default Backbone.Model.extend({
...
+    const defaultStates = worklistId === 'done-last-thirty-days' ? doneStates : notDoneStates;
+
+    return reduce(defaultStates.map('id'), (newStateObj, stateId) => {
+      newStateObj[stateId] = true;
+      return newStateObj;
+    }, {});
+  },
+  getSelectedStates() {
+    const states = this.states;
+    const selectedStates = this.getFilters().states;
```
paulfalgout: if `states` were an array of ids this'd be `return filter(selectedStates, id => states.has(id));`

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -137,6 +149,7 @@ export default App.extend({
   startFiltersApp() {
     const filtersApp = this.startChildApp('filters', {
       state: this.getState().getFilters(),
+      listId: this.getState().id,
```
paulfalgout: I think here we should do something like `isDoneOnly: this.getState().isDoneOnly()`

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
```diff
@@ -90,6 +92,83 @@ const CustomFiltersView = CollectionView.extend({
...
+  childViewTriggers: {
+    'select': 'select:state',
+  },
+  className: 'filters-sidebar__states-filters',
+  template: hbs`
+    <div>
+      <h3 class="sidebar__heading u-margin--t-16 u-margin--b-8">States</h3>
+    </div>
+  `,
+  initialize({ state }) {
```
paulfalgout: I think it'd help me if we called this `stateModel` to clarify this isn't a `state` from `states`

paulfalgout: or the state could just be passed in as the model itself... then you could use `modelEvents`

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -55,6 +56,7 @@ export default App.extend({
     const currentUser = Radio.request('bootstrap', 'currentUser');
     const storedState = store.get(`${ this.worklistId }_${ currentUser.id }-${ STATE_VERSION }`);
     const filters = this.getState('filters');
+    const initialStatesStatusObject = this.getState().getInitialStatesObject(this.worklistId);
```
paulfalgout: I think we can `this.setState({ id: this.worklistId })` here at the top so that we don't have to pass the `worklistId` in..  then we only need to get this initial state if there's no stored state.

paulfalgout: then I think below the `if(storedState) { ... return; }` we can just call `this.getState().setDefaultFilterStates();` or something like that

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -107,32 +110,56 @@ export default Backbone.Model.extend({
       [dateType]: this.formatDateRange(relativeRange, unit, dateFormat),
     };
   },
+  getInitialStatesObject(worklistId) {
+    const notDoneStates = this.states.groupByDone().notDone;
```
paulfalgout: if `states` were stored as an array this'd be 
```js
const { done, notDone } = this.states.groupByDone();
return this.isDoneOnly() ? done.map('id') : notDone.map('id');`
```

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: actually perhaps all we need to do is pass:
```
{
   state: this.getState().getFilters(),
   defaultStatesFilter: this.getState().getDefaultStatesFilter(),
   states: this.getState().isDoneOnly() ? done : notDone,
}
```
and then we'd pass that down to the sidebar as well.

paulfalgout (changes requested):
Alright.. I think I've pulled enough threads on the sweater..  if we change the `filter.states` to an array of ids I _think_ that cleans things up a bit.. and I think if we pass in the default filter.states and the available states collection at the top level, calculated in the state model it will reduce duplicate logic..  I'm sure I didn't think through this entirely, but lets try that direction and see where it leads.

src/js/services/bootstrap.js
nmajor25: Good call, updated in the last push.

I'll create a separate shortcut story for refactoring the `getOrgDirectories()` as well.

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
nmajor25: That works nicely, updated in latest push.

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
nmajor25: Updated in the last push.

src/js/views/patients/worklist/filters_views.js
nmajor25: I removed the initialize and now do this instead:

```js
templateContext() {
  return {
    filtersCount: this.getOption('filtersCount'),
  };
}
```

src/js/apps/patients/worklist/worklist_app.js
nmajor25: Updated in latest push.

src/js/apps/patients/worklist/worklist_state.js
nmajor25: Updated in latest push.

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
nmajor25: In the last push, updated to pass in the model itself and use `modelEvents`.

src/js/apps/patients/worklist/worklist_app.js
nmajor25: Updated in the last push.

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
```diff
@@ -90,6 +91,83 @@ const CustomFiltersView = CollectionView.extend({
...
+      <div data-check-region class="u-margin--r-16"></div>
+      <span class="action--{{ options.color }}">
+        <span class="u-margin--r-8">{{fa options.iconType options.icon}}</span><span>{{ name }}</span>
+      </span>
+    </div>
+  `,
+  regions: {
+    check: '[data-check-region]',
+  },
+  modelEvents: {
```
paulfalgout: this modelEvents doesn't happen I don't think. at least state models shouldn't change

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -107,35 +110,66 @@ export default Backbone.Model.extend({
...
+      return isNull(value);
+    }));
+
+    const statesFilterSize = size(filtersState.states);
+    const defaultStatesSize = size(defaultStatesFilter.map('id'));
+
+    const statesFilterCount = statesFilterSize !== defaultStatesSize ? 1 : 0;
+
+    return customFiltersSize + statesFilterCount;
+  },
```
nmajor25, line 144: Added this function to the state, but not using it yet.

`filters_app.js` and `filters-sidebar_app.js` don't have access to any of the state methods. Since their state is only `this.state.filters` (not the entire `this.state`).

So in those files, `this.getState().getFiltersCount()` doesn't work and results in an error... 🤔

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -107,35 +110,66 @@ export default Backbone.Model.extend({
...
       },
       'done-last-thirty-days': {
         updated_at: dayjs().startOf('day').subtract(30, 'days').format(),
-        state: doneStates,
       },
     };
 
+    if (selectedStates) {
+      filters[this.id].state = selectedStates;
+    }
```
nmajor25: This makes it so `filter[state]` is either `filter[state]=id,id,id` or not added as a api request url parameter all together.

But this causes an issue when all states are un-selected.

For example:

Screenshot 2023-01-05 at 12 22 56 PM

In this scenario, we want the worklist to be empty. But instead `filter[state]` is not in the url parameter, so the filter is not applied and the worklist shows all actions/flows.

nmajor25: Do `filter[state]=null`?

src/js/apps/patients/worklist/worklist_app.js
nmajor25: Updated this in the last push.

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -107,35 +110,66 @@ export default Backbone.Model.extend({
...
+
+    const customFiltersState = omit(filtersState, 'states');
+    const customFiltersSize = size(omit(customFiltersState, function(value) {
+      return isNull(value);
+    }));
+
+    const statesFilterSize = size(filtersState.states);
+    const defaultStatesSize = size(defaultStatesFilter.map('id'));
+
+    const statesFilterCount = statesFilterSize !== defaultStatesSize ? 1 : 0;
```
paulfalgout: I think this check should probably be `_.isEqual(filtersState.states.sort(), defaultStatesFilter.map('id').sort()) ?...`. seems like size checks would return false positives

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -107,35 +110,66 @@ export default Backbone.Model.extend({
...
+    const selectedStates = this.getFilters().states;
+
+    // remove any invalid state ids (i.e. ids that don't belong to any default states)
+    return filter(selectedStates, id => contains(defaultStatesFilterIds, id)).join();
+  },
+  getFiltersCount() {
+    const filtersState = this.get('filters');
+    const defaultStatesFilter = this.getDefaultStatesFilter();
+
+    const customFiltersState = omit(filtersState, 'states');
```
paulfalgout: I think we should break this function up such that this ends up being
`return this.getCustomFiltersCount() + this.getStatesFilterCount();`

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -107,35 +110,66 @@ export default Backbone.Model.extend({
       [dateType]: this.formatDateRange(relativeRange, unit, dateFormat),
     };
   },
+  getDefaultStatesFilter() {
+    const isDoneOnly = this.id === 'done-last-thirty-days';
+    const { done, notDone } = this.states.groupByDone();
+
```
paulfalgout, line 125: I _think_ we could hard cache this.. like 
```js
if (!this._defaultStatesFilter) {
  ....
  this._defaultStatesFilter = isDoneOnly ? done : notDone;
}
return this._defaultStatesFilter;

File: src/js/apps/patients/worklist/filters_app.js [hunk 27]
```diff
@@ -36,13 +39,28 @@ export default App.extend({
...
+      filtersCount: this.getFiltersCount(),
```
paulfalgout: I think we need to remove that `!directories.length` check now above.

Also the `AllFiltersButtonView` has a `modelEvents` in it which won't be used now.

src/js/apps/patients/worklist/filters_app.js
```diff
@@ -56,6 +74,8 @@ export default App.extend({
 
     const sidebar = Radio.request('sidebar', 'start', 'filters', {
       state: state.attributes,
+      listId: this.listId,
```
paulfalgout: don't think this is used anymore

src/js/apps/patients/sidebar/filters-sidebar_app.js
```diff
@@ -39,10 +55,35 @@ export default App.extend({
...
   onClearFilters() {
-    this.getState().clear();
+    const defaultStates = this.defaultStatesFilter.map('id');
+
+    this.getState().clear().set({ states: defaultStates });
   },
   onStop() {
     Radio.request('sidebar', 'close');
   },
+  getFiltersCount() {
```
paulfalgout: I think the way we can reduce this duplication is to make a shared state model that this function lives on like the `worklist_state.js` that both of these filter_apps use.  Then we could also go back to passing the state model as the model to the button and header views rather than re-rendering them in the parent.

src/js/apps/patients/worklist/worklist_state.js
paulfalgout: Ah actually I realize now this method isn't being used.. yeah it should be on a different state model than this one who's class can be shared by both filter apps.

paulfalgout (changes requested):

src/js/apps/patients/worklist/worklist_state.js
paulfalgout: I mentioned this in my review prior to reading this, but yeah we just need to make a defined state model for those other apps with this function

src/js/apps/patients/worklist/worklist_state.js
paulfalgout: ah good question.  does null work?  maybe we do have to short circuit the request as you had before..  try `null` and otherwise we can confirm with BE I guess if it doesn't.

nmajor25: Neither of these work:

```js
filters[this.id].state = null;
filters[this.id].state = 'null';
```

But this works:

```js
filters[this.id].state = NIL_UUID;

// result: filter[state]=00000000-0000-0000-0000-000000000000
```

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
nmajor25: Removed in last push.

src/js/apps/patients/worklist/worklist_state.js
nmajor25: Updated in last push.

src/js/apps/patients/worklist/worklist_state.js
nmajor25: Couldn't get this to work... `this._defaultStatesFilter` is being reset each time the filter is changed and the worklist app restarts...

src/js/apps/patients/worklist/filters_app.js
nmajor25: In a previous push, the `directories.length` check and `modelEvents` were removed.

src/js/apps/patients/worklist/filters_app.js
nmajor25: Updated in previous push.

src/js/apps/patients/worklist/filters_app.js
nmajor25: Removed in a previous push.

src/js/apps/patients/worklist/worklist_state.js
shadowhand: Sorry, I should have commented earlier. NIL UUID is the right solution.

src/js/apps/patients/schedule/schedule_state.js
```diff
@@ -75,19 +92,24 @@ export default Backbone.Model.extend({
...
     const dateFilter = this.getEntityDateFilter();
 
     const filters = extend({
       clinician: clinicianId,
-      state: notDoneStates,
+      state: selectedStates,
     }, dateFilter);
 
+    if (selectedStates) {
+      filters.state = selectedStates;
```
paulfalgout: I think this could be `filters.state = selectedStates || NIL_UUID;`

paulfalgout (changes requested):
Looks pretty good.. might just replace those if/else blocks with an `||` but otherwise it seems like there's just a few lines for coverage and this is 💯

src/js/apps/patients/schedule/schedule_state.js
nmajor25: Yup, I like that much better. Updated to that in last push.

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -107,35 +110,48 @@ export default Backbone.Model.extend({
...
+    return isDoneOnly ? done : notDone;
+  },
+  setDefaultFilterStates() {
+    const defaultStatesFilter = this.getDefaultStatesFilter();
+    this.set({ filters: { states: defaultStatesFilter.map('id'), worklistId: this.id } });
```
nmajor25: I added the `worklistId: this.id` here. To change the filters state into this:

```js
filters: {
  insurance: 'Medicare',
  states: ['id', 'id'],
  worklistId: 'owned-by'
}
```

Because the `filters_state.js` doesn't have access to the `state.id` worklist id.

Hoping for a better way to do this. Since I don't like that the `state.id` and `state.filters.worklistId` are duplicates.

src/js/apps/patients/worklist/filters_state.js
```diff
@@ -0,0 +1,29 @@
...
+import Backbone from 'backbone';
+import Radio from 'backbone.radio';
+
+export default Backbone.Model.extend({
+  preinitialize() {
+    this.currentOrg = Radio.request('bootstrap', 'currentOrg');
+    this.states = this.currentOrg.getStates();
+  },
+  getDefaultStatesFilter() {
+    const isDoneOnly = this.get('worklistId') === 'done-last-thirty-days';
```
nmajor25: This is where the `state.filters.worklistId` is used. Without it, the incorrect filters count is displayed and the incorrect default state is added after `Clear Filters` for the `Done < 30 Days` list.

This state doesn't have access to the `this.id` worklist id that resides in the `worklist_state.js` file.

src/js/apps/patients/sidebar/filters-sidebar_app.js
```diff
@@ -39,8 +52,21 @@ export default App.extend({
...
+    this.showChildView('statesFilters', statesFiltersView);
+  },
   onClearFilters() {
-    this.getState().clear();
+    const defaultStates = this.defaultStatesFilter.map('id');
+    const worklistId = this.getState('worklistId');
+
+    this.getState().clear().set({ states: defaultStates });
+
+    if (worklistId) this.setState({ worklistId });
```
nmajor25: This is shared by both the worklists and schedule. So this ensures that the `state.filters.worklistId` is only added to worklists. Not to the schedule.

On the schedule, `this.getState('worklistId')` will be `undefined`, so it won't be added to the state.

test/integration/patients/worklist/worklist.js
```diff
@@ -2147,6 +2145,182 @@ context('worklist page', function() {
...
+      .find('.fa-square-check')
+      .should('have.length', 3);
+
+    cy
+      .get('@filtersSidebar')
+      .find('[data-states-filters-region]')
+      .should('contain', 'Done')
+      .should('contain', 'Unable to Complete')
+      .should('contain', 'THMG Transfered');
+
```
nmajor25, line 2418: Quick overview for what I added to this test:

1. Verify default states are shown in the UI based on local storage setting.
2. Unselect one state (1 of 2) and verify that the API request param data, local storage data, and UI total filter counts are correct.
3. Unselect second state (2 of 2) and verify that the API request param data, local storage data, and UI total filter counts are correct. API route should contain `filter[state]=nil_uuid`.
4. Clear filters via the `Clear Filters` button in the UI. Verify API request param data and local storage data is set to the correct default states. And the UI filter counts are back to `0`.
5. Switch to the `Shared By` worklist (no initial local storage data set) and verify that the correct default states filters are added to local storage.
6. Switch to `Done < 30 Days` worklist and verify that only done states are shown in the UI. And that the API request params and local storage data is correct.

test/integration/patients/worklist/schedule.js
```diff
@@ -897,6 +899,123 @@ context('schedule page', function() {
...
+    cy
+      .get('.list-page__filters')
+      .find('[data-all-filters-region]')
+      .should('not.contain', '2');
+
+    cy
+      .get('@filtersSidebar')
+      .find('.sidebar__heading')
+      .should('not.contain', '2');
+
```
nmajor25, line 1020: Same as the worklist test additions, minus the `Shared By` and `Done < 30 Days` list testing.

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -107,35 +110,48 @@ export default Backbone.Model.extend({
...
-    const states = this.currentOrg.getStates();
     const filtersState = this.getFilters();
     const clinicianId = this.get('clinicianId');
     const teamId = this.get('teamId');
     const noOwner = this.get('noOwner');
-    const customFilters = omit(filtersState, 'groupId');
-    const notDoneStates = states.groupByDone().notDone.getFilterIds();
-    const doneStates = states.groupByDone().done.getFilterIds();
-
+    const customFilters = omit(filtersState, 'groupId', 'states', 'worklistId');
```
nmajor25: Added `worklistId` to the omit list here so it's not included in the custom filters when creating the API request URL params.

I'm assuming we'd never want a custom filter with a slug of `worklistId`... because if we did, that would cause issues since it would be omitted here.

paulfalgout (approved):
@nmajor25 👀

src/js/apps/patients/schedule/filters_app.js
```diff
@@ -4,23 +4,26 @@ import App from 'js/base/app';
...
 import { FiltersView, AllFiltersButtonView, GroupsDropList } from 'js/views/patients/schedule/filters_views';
 
 export default App.extend({
+  StateModel,
   childApps: {
     allFilters: AllFiltersApp,
   },
   stateEvents: {
-    'change:groupId': 'showGroupsFilterView',
+    'change:filters': 'showGroupsFilterView',
```
paulfalgout, line 18: I don't like that this refreshes on all filter changes, but we're soon to remove this whole thing with the workspaces change that this'll all get cleaned up

test/integration/patients/worklist/schedule.js
```diff
@@ -22,11 +22,13 @@ const testGroups = [
   },
 ];
 
+const STATE_VERSION = 'v4';
```
nmajor25, line 25: Nice, now we'll only have to change that in one place 😄

src/js/apps/patients/sidebar/filters_state.js
```diff
@@ -0,0 +1,47 @@
...
+    this.set('filters', filters);
+  },
+  getFiltersCount() {
+    const customFilters = this.getFilters();
+    const customFiltersCount = size(omit(customFilters, value => {
+      return isNull(value);
+    }));
+
+    const statesFilter = sortBy(this.getStatesFilters());
+    const defaultStates = this.getDefaultSelectedStates();
```
nmajor25, line 30: Both need the `sortBy()`?

paulfalgout: `defaultStates` is sorted at creation

paulfalgout: I suppose we could sort it anyhow

nmajor25: That's just how I had it before, is why I'm asking. Unsure if it's needed.

paulfalgout: so turns out `sortBy` clones, so I'm going to replace the clones in the `get`s

PR #879, opened 2022-12-20
Cast custom sort fields to strings

paulfalgout:
We’ll want this until we legitimately have a sort that requires something other than casting to alpha

Shortcut Story ID: [sc-32656]

PR #881, opened 2022-12-22
Support custom sorting by numerical order

nmajor25:
Shortcut Story ID: [sc-32716]

If the custom sort org setting has a `type: 'numeric'` property, use `numSort()` instead of `alphaSort()`. `alphaSort()` will continue to be used by default.

### DIFF
--- a/src/js/apps/patients/worklist/worklist_sort.js
+++ b/src/js/apps/patients/worklist/worklist_sort.js
@@ -2,14 +2,14 @@ import { get, union, map, uniq, compact } from 'underscore';
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 
-import { alphaSort } from 'js/utils/sorting';
+import { alphaSort, numSort } from 'js/utils/sorting';
 import { i18n } from 'js/views/patients/worklist/worklist_views';
 
 // Casts values to String for alpha sort
 function getEntityFieldValue(entity, fieldName, keys) {
   const patientField = entity.getPatient().getField(fieldName);
-  if (!patientField) return '';
-  return String(get(patientField.getValue(), keys, ''));
+  if (!patientField) return null;
+  return get(patientField.getValue(), keys);
 }
 
 // field_key may include a path for the model_attr.deeply_nested.value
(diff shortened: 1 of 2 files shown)


test/integration/patients/worklist/worklist.js
```diff
@@ -2555,15 +2555,15 @@ context('worklist page', function() {
...
-            last_name: 'Field 1',
+            last_name: 'Field A',
           },
           relationships: { 'patient-fields': { data: [{ id: '1' }] } },
         });
 
         fx.included.push({
           id: '1',
           type: 'patient-fields',
-          attributes: { value: { value: 1 }, name: 'foo' },
```
nmajor25, line 2566: Reverted these tests back to before: https://github.com/RoundingWell/care-ops-frontend/pull/879.

We'll need to update the tests to get coverage back to 100%. Will do that in separate story.

test/integration/patients/worklist/worklist.js
```diff
@@ -2555,15 +2555,15 @@ context('worklist page', function() {
...
+          attributes: { value: { value: 'A' }, name: 'foo' },
```
shadowhand, line 2566: ~~These two changes shouldn't have been reverted.~~ Never mind, see [here](#discussion_r1055907891).

shadowhand (changes requested):

test/integration/patients/worklist/worklist.js
shadowhand: Sorry missed the comment. I'll approve.

shadowhand (approved):

test/integration/patients/worklist/worklist.js
nmajor25: Because of time constraints, just reverted them back so they'll pass. I'll create a separate story to come back and update it.

src/js/apps/patients/worklist/worklist_sort.js
```diff
@@ -26,12 +26,17 @@ const SortOption = Backbone.Model.extend({
     const fieldName = this.get('field_name');
     const keys = this.get('field_key').split('.');
     const direction = this.get('direction');
+    const type = this.get('type');
```
paulfalgout, line 29: we really shouldn't use `type` as an attribute with jsonapi.  @nmajor25

shadowhand: Oh yeah. Technically it's a reserved word. We could use "kind" or "compare" instead...?

nmajor25: Created a new [shortcut story](https://app.shortcut.com/roundingwell/story/32726) to update this.

wweaver4th: `sort_type`?

PR #887, opened 2023-01-03
Change custom sort org setting to `sort_type` & remove duplicate `intSort()` utility

nmajor25:
Shortcut Story ID: [sc-32726] [sc-32720]

`intSort()` is exactly the same as `numSort()` and is not used in the application. So it was removed (along with its test).

`type` is a reserved word in JSON:Api, so we changed it to `sort_type` instead.

Also updated Cypress to test numerical sorting. Previously we were only testing alphabetical sorting, which caused a Coveralls coverage drop.

### DIFF
--- a/src/js/apps/patients/worklist/worklist_sort.js
+++ b/src/js/apps/patients/worklist/worklist_sort.js
@@ -26,13 +26,13 @@ const SortOption = Backbone.Model.extend({
     const fieldName = this.get('field_name');
     const keys = this.get('field_key').split('.');
     const direction = this.get('direction');
-    const type = this.get('type');
+    const sortType = this.get('sort_type');
 
     return (a, b) => {
       const aValue = getEntityFieldValue(a.model, fieldName, keys);
       const bValue = getEntityFieldValue(b.model, fieldName, keys);
 
-      if (type === 'numeric') {
+      if (sortType === 'numeric') {
         return numSort(direction, aValue ?? 0, bValue ?? 0);
       }
 
@@ -143,4 +143,3 @@ function getSortOptions(listType) {
 export {
   getSortOptions,
 };
-
--- a/src/js/apps/patients/worklist/worklist_sort.js
+++ b/src/js/apps/patients/worklist/worklist_sort.js
@@ -33,10 +33,10 @@ const SortOption = Backbone.Model.extend({
       const bValue = getEntityFieldValue(b.model, fieldName, keys);
 
       if (sortType === 'numeric') {
-        return numSort(direction, aValue ?? 0, bValue ?? 0);
+        return numSort(direction, aValue, bValue);
       }
 
-      return alphaSort(direction, aValue ?? '', bValue ?? '');
+      return alphaSort(direction, aValue, bValue);
     };
   },
 });
--- a/src/js/utils/sorting.js
+++ b/src/js/utils/sorting.js
@@ -30,16 +30,9 @@ function intSortBy(sortDir, val, nullVal) {
   return sortByDir(sortDir, parseInt(int, 10));
 }
 
-function intSort(sort, fieldA, fieldB, nullVal) {
-  const sortVal = getSortNum(fieldA, nullVal) > getSortNum(fieldB, nullVal) ? 1 : -1;
-  return sortByDir(sort, sortVal);
-}
-
-
 export {
   alphaSort,
   intSortBy,
   numSortBy,
-  intSort,
   numSort,
 };
--- a/src/js/utils/sorting.js
+++ b/src/js/utils/sorting.js
@@ -20,9 +20,11 @@ function numSortBy(sortDir, val, nullVal) {
   return sortByDir(sortDir, num);
 }
 
-function numSort(sort, fieldA, fieldB, nullVal) {
-  const sortVal = getSortNum(fieldA, nullVal) > getSortNum(fieldB, nullVal) ? 1 : -1;
-  return sortByDir(sort, sortVal);
+function numSort(sortDir, a, b, nullVal = Number.NEGATIVE_INFINITY) {
+  if (!a) a = nullVal;
+  if (!b) b = nullVal;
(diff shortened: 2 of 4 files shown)


nmajor25:
@paulfalgout failing test is because of that date issue we discussed. Looks like I introduced that bug in PR: https://github.com/RoundingWell/care-ops-frontend/pull/875.

paulfalgout (changes requested):
Works for me.  That failing test doesn't help with knowing coverage unfortunately.  Do we know if that's up?  Either way I'd be comfortable running a release off of this branch if we want and then delaying this merge, possibly even fixing that date test, so that we can fix any coverage issues on this PR.

test/integration/patients/worklist/worklist.js
```diff
@@ -339,7 +339,7 @@ context('worklist page', function() {
       .wait('@routeFlows')
       .itsUrl()
       .its('search')
-      .should('contain', `filter[updated_at]=${ dayjs(testTs()).startOf('day').subtract(1, 'month').format() }`)
+      .should('contain', `filter[updated_at]=${ dayjs().startOf('day').subtract(30, 'days').format() }`)
```
nmajor25, line 342: Tried to get too cute when I added this. Just updated this to match what's done in the worklist's `getEntityFilter()`: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/patients/worklist/worklist_state.js#L134. 

Should pass every time now.

src/js/apps/patients/worklist/worklist_sort.js
```diff
@@ -26,13 +26,13 @@ const SortOption = Backbone.Model.extend({
     const fieldName = this.get('field_name');
     const keys = this.get('field_key').split('.');
     const direction = this.get('direction');
-    const type = this.get('type');
+    const sortType = this.get('sort_type');
```
shadowhand, line 29: Why `sortType` instead of just `sort` ?

nmajor25: It's more descriptive when reading the code, in my opinion. Not a huge deal either way for me though.

shadowhand: That's between you and @paulfalgout. To me it feels overly verbose.

test/integration/patients/worklist/worklist.js
```diff
@@ -2541,7 +2541,7 @@ context('worklist page', function() {
       .should('contain', 'APatient AName');
   });
 
-  specify('flow sorting - patient field', function() {
+  specify('flow sorting alphabetical - patient field', function() {
```
nmajor25, line 2544: Split these off into two separate tests. One for alphabetical sorting and another for numerical.

test/integration/patients/worklist/worklist.js
```diff
@@ -2635,6 +2635,108 @@ context('worklist page', function() {
...
+      .find('.table-list__item')
+      .first()
+      .should('contain', 'Patient Field None');
+
+    cy
+      .get('.app-frame__content')
+      .find('.table-list__item')
+      .last()
+      .should('contain', 'Patient Field 2');
+  });
```
nmajor25, line 2738: To create this test, I copied over the `"flow sorting alphabetical - patient field"` test and changed the custom sorting org setting to have `sort_type: 'numeric'` for both sorting methods (`asc` and `desc`).

Then I updated the patient field data to have numercial values instead of alphabetical (e.g. `1` vs. `'A'`). And changed the patient names to have numbers in them to make the test easier to understand (`Patient Field 1` vs. `Patient Field A`).

nmajor25:
Got one branch not covered: https://coveralls.io/builds/55611317/source?filename=src%2Fjs%2Fapps%2Fpatients%2Fworklist%2Fworklist_sort.js#L39.

Not sure why... but investigating.

src/js/apps/patients/worklist/worklist_sort.js
```diff
@@ -33,10 +33,10 @@ const SortOption = Backbone.Model.extend({
       const bValue = getEntityFieldValue(b.model, fieldName, keys);
 
       if (sortType === 'numeric') {
-        return numSort(direction, aValue ?? 0, bValue ?? 0);
+        return numSort(direction, aValue, bValue);
       }
 
-      return alphaSort(direction, aValue ?? '', bValue ?? '');
+      return alphaSort(direction, aValue, bValue);
```
nmajor25, line 39: `alphaSort` already converts null values to `''` by default (via [this code](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/utils/sorting.js#L8)). So the `??` check isn't needed. Which was causing the coveralls missing branch issue.

Also updated `numSort()` to convert null values to `0`. So we can remove the `??` checks for that too.

src/js/utils/sorting.js
```diff
@@ -20,8 +20,10 @@ function numSortBy(sortDir, val, nullVal) {
   return sortByDir(sortDir, num);
 }
 
-function numSort(sort, fieldA, fieldB, nullVal) {
-  const sortVal = getSortNum(fieldA, nullVal) > getSortNum(fieldB, nullVal) ? 1 : -1;
+function numSort(sort, a, b, nullVal = 0) {
```
paulfalgout: hmm..  I dunno about this.  this means that base-0 values would sort the same as `null`.  when in that case you'd set `nullVal` to `-1` or something.. Really the safest default is _likely_ `Number.NEGATIVE_INFINITY`

paulfalgout (changes requested):
one small request.  doesn't require a new release since it's all edge case.

src/js/utils/sorting.js
nmajor25: Ah ya good call. Just updated it to use `Number.NEGATIVE_INFINITY` as the default in the last push.

shadowhand: TIL about `Number.NEGATIVE_INFINITY` 🤯

nmajor25: Same here @shadowhand

PR #896, opened 2023-01-12
Switch group type to workspaces

paulfalgout:
Shortcut Story ID: [sc-32928]

Turns out some relation updating _may_ have broken or at least be sent incorrectly.. not sure what would have happened.

### DIFF
--- a/src/js/entities-service/entities/clinicians.js
+++ b/src/js/entities-service/entities/clinicians.js
@@ -42,12 +42,12 @@ const _Model = BaseModel.extend({
   addGroup(group) {
     const groups = this.getGroups();
     groups.add(group);
-    this.set('_groups', this.toRelation(groups, 'groups').data);
+    this.set('_groups', this.toRelation(groups, 'workspaces').data);
   },
   removeGroup(group) {
     const groups = this.getGroups();
     groups.remove(group);
-    this.set('_groups', this.toRelation(groups, 'groups').data);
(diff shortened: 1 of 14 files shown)


src/js/entities-service/entities/clinicians.js
```diff
@@ -42,12 +42,12 @@ const _Model = BaseModel.extend({
   addGroup(group) {
     const groups = this.getGroups();
     groups.add(group);
-    this.set('_groups', this.toRelation(groups, 'groups').data);
+    this.set('_groups', this.toRelation(groups, 'workspaces').data);
```
paulfalgout, line 45: The `toRelation`s in the entities and the the `type` change are the big changes.. not a huge deal.. largely FE ignores the types with the exception of the `includes` cache.

nmajor25 (approved):

src/js/entities-service/entities/patients.js
```diff
@@ -48,7 +48,7 @@ const _Model = BaseModel.extend({
     attrs = extend({}, this.attributes, attrs);
 
     const relationships = {
-      'groups': this.toRelation(attrs._groups, 'groups'),
+      'groups': this.toRelation(attrs._groups, 'workspaces'),
```
shadowhand, line 51: Just FYI, the relationships _should_ have been duplicated in the API too, so anything that exists in `relationships.group` or `relationships.groups` would also be `relationships.workspace` (s).

PR #897, opened 2023-01-17
New nav design & minimize functionality

nmajor25:
Shortcut Story ID: [sc-32886] [sc-32887]

This is the first stage of the workspaces epic, which will allow a user to switch between different workspaces in the app.

In this first phase, the nav's UI design is updated in the following ways.

## Move nav menu items

The top droplist menu has the `Workspace`, `Dashboards`, `Programs`, and `Clinicians` links removed:

Screenshot 2023-01-18 at 10 12 49 AM

For now this will only include the `Help & Support` and `Sign Out` links. In a later PR, this droplist is where users will switch between workspaces.

The `Dashboards`, `Programs`, and `Clinicians` nav items are moved to an `Admin Tools` droplist at the bottom of the nav:

Screenshot 2023-01-18 at 10 10 23 AM

Each droplist item is only shown if the user has the correct permission. If the user doesn't have a permission for any of the droplist items, the `Admin Tools` button will not be shown.

## Minimize functionality

Toggle the nav menu between minimized and expanded versions. The minimized version shows only icons for each nav menu item.

The user's choice is persisted via local storage. Expanded is shown by default.

## Other small UI changes to the nav menu:

- Add icons to each nav item. And some existing icons were updated.
- Add a light-blue left-border to some nav link items when they are active and/or hovered. These are only applied to nav items that link to a new page. So the `Search`, `Admin Tools`, and `Add Patient` buttons don't have it.
- Slight changes to droplist positioning (i.e. `top`, `bottom`, `left`, `right` CSS properties).
- Opacity of nav items is set to `opacity: 0.5` when they are not active/hovered. And `opacity: 1` when they are active/hovered.


src/js/apps/globals/nav_app.js
```diff
@@ -53,60 +65,80 @@ const appNavMenu = new Backbone.Collection([
...
+    icons: [
+      {
+        type: 'fas',
+        icon: 'angle-left',
+      },
+      {
+        type: 'fas',
+        icon: '1',
+      },
+    ],
```
nmajor25, line 117: Some of these nav link items need to support having two icons placed together (`<1`, `<3`, `30` icons):

Screenshot 2023-01-13 at 9 52 52 AM

So I changed this collection take one or more `icons` instead of just one icon. In the view, I display them using a `{{#each}}` so that two icon elements are rendered next to each other like `<svg></svg><svg></svg>`. And if only one icon is passed, it renders the same as one `<svg></svg>` element.

src/js/apps/globals/nav_app.js
```diff
@@ -154,22 +186,18 @@ export default App.extend({
   },
   onStart() {
     const currentUser = Radio.request('bootstrap', 'currentUser');
+    this.showAdminTools = currentUser.can('dashboards:view') || currentUser.can('clinicians:manage') || currentUser.can('programs:manage');
```
nmajor25: If the user has none of those permissions, the `Admin Tools` droplist won't be shown.

Screenshot 2023-01-12 at 10 01 54 AM

Thought about doing this instead:

```js
canDoAny(...props) {
  const role = this.getRole();
  const permissions = role.get('permissions');

  return props.some(prop => permissions.includes(prop));
}
```

And then you could do this instead:

```js
this.showAdminTools = currentUser.canDoAny('dashboards:view', 'clinicians:manage', 'programs:manage');
```

But figured that was getting too cute.

src/js/apps/globals/nav_app.js
```diff
@@ -179,19 +207,25 @@ export default App.extend({
     this.setView(new AppNavView());
 
     this.mainNav = new MainNavDroplist({ collection: appNavMenu });
-
     this.showChildView('navMain', this.mainNav);
 
+    if (this.showAdminTools) {
```
paulfalgout: I think you could just wrap this (and the check below) in `if (adminNavMenu.length) {`

src/js/components/picklist/index.js
```diff
@@ -74,6 +75,10 @@ const PicklistItem = View.extend({
 
     if (this.model === this.state.get('selected')) classNames.push('is-selected');
 
+    if (this.getOption('styleForNav')) {
+      classNames[indexOf(classNames, 'picklist__item')] = 'picklist__item-nav';
+    }
```
nmajor25: If the `styleForNav: true` option is passed to a droplist/picklist, the default classname of `picklist__item` is replaced with `picklist__item-nav` for each item in the picklist.

This allows me to add special styling for those droplist items used only in the nav menu.

Default `picklist__item` styling:

Screenshot 2023-01-17 at 1 58 51 PM

`picklist__item-nav` styling used for nav menu droplists:

Screenshot 2023-01-17 at 11 27 59 AM
Screenshot 2023-01-17 at 2 00 11 PM

src/js/views/globals/app-nav/app-nav.scss
```diff
@@ -44,52 +44,117 @@ $menu-hover-color: #375467;
...
+
+    &::before {
+      background-color: color(light_blue);
+      bottom: 0;
+      content: '';
+      left: 0;
+      position: absolute;
+      top: 0;
+      width: 4px;
+    }
```
nmajor25, line 145: This adds the light blue left border for any nav links that are hovered or currently selected.

Screenshot 2023-01-17 at 2 03 43 PM

Doing a simple `border-left: 4px solid color(light_blue)` caused the layout for each nav item to jump when hovering. That's why I went the `::before` route instead.

src/js/components/picklist/index.js
```diff
@@ -74,6 +75,10 @@ const PicklistItem = View.extend({
 
     if (this.model === this.state.get('selected')) classNames.push('is-selected');
 
+    if (this.getOption('styleForNav')) {
```
paulfalgout: What problem is this solving?  Ideally we solve this without modify the generic component

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -118,7 +184,7 @@ const AppNavCollectionView = CollectionView.extend({
 
 const PatientsAppNav = View.extend({
   template: hbs`
-    <h3 class="app-nav__search app-nav__link js-search">{{far "magnifying-glass"}}{{ @intl.globals.appNav.appNavViews.patientsAppNav.searchTitle }}</h3>
+    <h3 class="flex app-nav__search js-search">{{fas "magnifying-glass"}}{{ @intl.globals.appNav.appNavViews.patientsAppNav.searchTitle }}</h3>
```
paulfalgout: Is the far version still in use?  If not we should remove it from the package.json

src/js/views/globals/app-nav/app-nav.scss
```diff
@@ -44,52 +44,117 @@ $menu-hover-color: #375467;
...
+  .icon.fa-1 {
+    font-size: 16px;
+    width: 14px;
   }
 
-  &.is-selected {
-    font-weight: 700;
+  .icon.fa-3 {
+    font-size: 16px;
+    width: 13px;
```
nmajor25, line 180: Getting the icons with two combined `<svg>` elements to align correctly was tricky (`<1`, `<3`, `30`).

Each of the numbered icons look slightly different when given differing widths. So each of the `fa-0`, `fa-1`, and `fa-3` icons need different `width` properties for them to align. Or at least that's the only way I could get them to work.

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -12,6 +12,14 @@ const i18n = intl.globals.appNav.appNavViews;
...
 const MainNavDroplist = Droplist.extend({
   popWidth: '248px',
+  position() {
+    const { outerHeight } = this.getView().getBounds();
+
+    return {
+      top: outerHeight,
+      left: 16,
+    };
+  },
```
nmajor25, line 22: This adds `left: 16px` to the  main nav droplist to match the design docs.

Old:

Screenshot 2023-01-17 at 2 10 14 PM

New: 

Screenshot 2023-01-17 at 2 10 24 PM

paulfalgout (changes requested):
I'll look more in depth tomorrow, but a couple of comments to start

src/js/components/picklist/index.js
paulfalgout: I think you can set an `itemClassName` here which would just override any values on `. picklist__item`. And so you'd probably want to add this new class to the `.app-nav__` block / file

nmajor25: Yup, that's a much better way to do that. Updated to do that in a recent push.

src/js/components/picklist/index.js
nmajor25: Removed this in a last push. Using `itemClassName` instead.

src/js/views/globals/app-nav/app-nav_views.js
nmajor25: It's used in two different places. The patient search modal and `Find In List...` inputs.

src/js/apps/globals/nav_app.js
```diff
@@ -5,21 +5,33 @@ import Radio from 'backbone.radio';
...
+    text: i18n.mainNav.signOut,
+    icon: {
+      type: 'fas',
+      icon: 'right-from-bracket',
     },
-    text: i18n.mainNav.patients,
   },
+]);
+
+const adminNavMenu = new Backbone.Collection([
```
paulfalgout, line 42: I think the i18n here should probably be updated

src/js/apps/globals/nav_app.js
```diff
@@ -53,60 +65,80 @@ const appNavMenu = new Backbone.Collection([
...
   },
   {
     text: i18n.patientsAppNav.schedule,
+    icons: [{
+      type: 'fas',
+      icon: 'calendar-star',
+    }],
     event: 'schedule',
     eventArgs: [],
-    className: 'app-nav__spacer',
```
paulfalgout, line 90: this class still in use anywhere?

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -32,18 +40,57 @@ const MainNavDroplist = Droplist.extend({
...
+const AdminToolsDroplist = Droplist.extend({
+  popWidth: '248px',
+  position() {
+    return {
+      top: window.innerHeight - 12,
+      left: 164,
+    };
+  },
+  viewOptions: {
+    tagName: 'div',
```
paulfalgout, line 83: I believe this is the default

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -32,18 +40,57 @@ const MainNavDroplist = Droplist.extend({
...
+    className: 'flex flex-align-center app-nav__bottom-button',
+    template: hbs`{{far "ellipsis"}}{{ @intl.globals.appNav.appNavViews.adminToolsDroplist.adminTools }}`,
+    templateContext() {
```
paulfalgout, line 86: I don't think this context is used?

paulfalgout (changes requested):

src/js/apps/globals/nav_app.js
nmajor25: That works nicely. Updated to this in the last push.

src/js/apps/globals/nav_app.js
nmajor25: Updated in last push.

src/js/apps/globals/nav_app.js
nmajor25: Nope, not anymore. Removed it from the `/views/globals/app-nav/app-nav.scss` file in the last push.

src/js/views/globals/app-nav/app-nav_views.js
nmajor25: When this is removed, `button` is used by default.

src/js/views/globals/app-nav/app-nav_views.js
nmajor25: Correct, it's not. Forgot to remove it. Removed it in last push.

paulfalgout (approved):

paulfalgout:
Going to wait on merging this in case we actually do the service worker release.

src/js/apps/globals/nav_app.js
```diff
@@ -154,28 +161,45 @@ export default App.extend({
...
+    'click:minimizeMenu': 'onClickMinimizeMenu',
   },
   childApps: {
     search: SearchApp,
   },
   selectNav(appName, event, eventArgs) {
     this.setState('currentApp', appName);
 
-    const navMatch = this.getNavMatch(appName, event, compact(eventArgs));
+    this.navMatch = this.getNavMatch(appName, event, compact(eventArgs));
```
nmajor25, line 176: Switch this to `this.navMatch` so it can be used in the `onChangeIsMinimized()` function.

src/js/apps/globals/nav_app.js
```diff
@@ -154,28 +161,45 @@ export default App.extend({
...
+    this.showBottomNavView();
+    this.showNav();
+
+    if (this.navMatch) {
+      this.getView().removeSelected();
+      this.navMatch.trigger('selected');
+    }
+
+    const currentAppName = this.getState('currentApp');
+    this.setSelectedAdminNavItem(currentAppName);
```
nmajor25, line 199: `this.navMatch.trigger('selected'` preserves the selected nav item (left blue border added) when switching between the minimized and expanded versions of the nav.

`this.setSelectedAdminNavItem(currentAppName)` preserves the selected state in the admin tools droplist menu.

src/js/views/globals/app-nav/app-nav.scss
```diff
@@ -4,6 +4,12 @@ $menu-hover-color: #375467;
   color: $black-90;
   height: 100%;
   position: relative;
+  transition: width 250ms ease-in-out 0s;
+  width: 200px;
```
nmajor25, line 8: Simple transition of `250s` whenever the width of nav is changed. This is what Shortcut uses for their nav menu, so copied it from them. Can change this to something else if we want to.

test/integration/globals/app-nav.js
```diff
@@ -48,8 +48,12 @@ context('App Nav', function() {
       .routeFlows()
       .routePrograms()
       .routeDashboards()
-      .visit();
+      .visit()
+      .then(() => {
+        const storageItem = JSON.parse(localStorage.getItem('isNavMenuMinimized'));
 
+        expect(storageItem).to.be.false;
+      });
```
nmajor25, line 56: Verifies that the `localStorage.isNavMenuMinimized` value is set to `false` by default if the local storage item doesn't exist yet.

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -114,14 +212,28 @@ const NavItemView = View.extend({
...
+      {{fas "magnifying-glass"}}{{#unless isMinimized}}<span class="u-text--overflow">{{ @intl.globals.appNav.appNavViews.patientsAppNav.searchTitle }}</span>{{/unless}}
+    </h3>
+    {{#unless isMinimized}}
+      <h3 class="app-nav__title">{{ @intl.globals.appNav.appNavViews.patientsAppNav.worklistsTitle }}</h3>
+    {{/unless}}
     <div data-worklists-region></div>
   `,
+  templateContext() {
+    return {
+      isMinimized: this.model.get('isMinimized'),
```
paulfalgout: no need for the context.  the model attribute will already be available to the view

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -95,30 +111,23 @@ const AdminToolsDroplist = Droplist.extend({
...
-      {{/if}}
-      <div data-nav-admin-tools-region></div>
+    {{#if canPatientCreate}}
+      <div class="flex flex-align-center app-nav__bottom-button js-add-patient">
+        {{fas "circle-plus"}}{{#unless isMinimized}}<span class="u-text--overflow">{{ @intl.globals.appNav.appNavViews.appNavView.addPatient }}</span>{{/unless}}
+      </div>
+    {{/if}}
+    <div data-nav-admin-tools-region></div>
+    <div class="flex flex-align-center app-nav__bottom-button js-minimize-menu">
+      {{fas "square-caret-left"}}{{#unless isMinimized}}<span class="u-text--overflow">{{ @intl.globals.appNav.appNavViews.appNavView.minimizeMenu }}</span>{{/unless}}
```
paulfalgout: the icon should switch when minimized

paulfalgout (changes requested):

src/js/views/globals/app-nav/app-nav_views.js
nmajor25: Updated in recent push.

src/js/views/globals/app-nav/app-nav_views.js
nmajor25: Good call, missed that. Updated in recent push.

PR #902, opened 2023-01-26
Update auth0 to v2

paulfalgout:
Shortcut Story ID: [sc-31889]

This touched on a few other things:
- Updates dependencies including cypress which final removes `.route`, but we've already replaced that API.
- fixes the settings bootstrap fetch.. for some reason this was oddly fetched to a model and parsed when it is a collection.
- adds `federated_logout` as an org setting.  If it's not set it defaults to off.

We bypass auth in cypress testing, so none of this should require new tests.


src/js/auth.js
```diff
@@ -62,7 +63,12 @@ function login(success) {
     config.connection = rwConnection;
   }
 
-  createAuth0Client(config).then(auth0Client => {
+  const clientConfig = {
```
paulfalgout, line 66: Note that this is where the service worker cache options will need to be added.

src/js/services/bootstrap.js
```diff
@@ -54,15 +54,17 @@ export default App.extend({
     return this.currentOrg;
   },
   getOrgSetting(settingName) {
-    return this.getCurrentOrg().getSetting(settingName);
+    const setting = this.getCurrentOrg().getSetting(settingName);
+    if (!setting) return;
```
paulfalgout, line 58: I am not sure if we have anywhere in the tests that a setting is requested that doesn't exist, so this might need coverage.. or maybe just an istanbul ignore.

package.json
```diff
@@ -180,20 +180,20 @@
     "source-map-loader": "^4.0.0",
     "stylelint": "^14.9.1",
     "stylelint-bare-webpack-plugin": "^2.0.2",
-    "stylelint-config-standard-scss": "^5.0.0",
-    "stylelint-order": "^5.0.0",
+    "stylelint-config-standard-scss": "^6.1.0",
```
paulfalgout, line 183: these 3 updates are largely based off of removing support for node v14

src/js/services/bootstrap.js
nmajor25: Doesn't appear that happens in the tests: https://coveralls.io/builds/56308171/source?filename=src%2Fjs%2Fservices%2Fbootstrap.js#L58

nmajor25 (approved):
Besides the small code coverage drop, it looks good!

PR #907, opened 2023-01-27
Rename group to workspace in the code

nmajor25:
Shortcut Story ID: [sc-33140]

Rename all instances of `group`/`groups` to `workspace`/`workspaces` in the code. Including API requests, variable/method names, relations, and UI regions.

Nothing customer facing in the UI was changed.

Note: `widgets.groups` was deprecated (but still supported) in place of `widgets.workspaces`.


src/js/views/patients/widgets/widgets.js
```diff
@@ -75,12 +75,12 @@ const widgets = {
   },
   groups: {
```
nmajor25: Wasn't sure what to do with this so I left the widget name as `groups`. Since that widget is used in production?

Or build it so both the `groups` and `workspaces` widget ids work the same?

paulfalgout: Yeah I think we should rename this to workspaces, and then below this add:

```js
// Note: Groups naming is deprecated
widgets.groups = widgets.workspaces;
```

paulfalgout (changes requested):
Looks good I think, other than that one caveat.

paulfalgout:
Oh also that last merge made a conflict

src/js/views/patients/widgets/widgets.js
nmajor25: Updated to that in the last push.

PR #911, opened 2023-02-01
Add small style transitions when revealing content

paulfalgout:
Including a preload delay this allows for an offline hard refresh not to flash the preload dialog and the feel less jarring on load.

Also just a bit nicer generally.

Shortcut Story ID: [sc-32572]


src/js/views/globals/app-frame.scss
```diff
@@ -17,9 +17,25 @@
...
+}
+
 .app-frame__content {
   min-width: 440px;
+  opacity: 0;
   overflow-y: auto;
+
+  &:not(:empty) {
+    animation: 0.3s cubic-bezier(0.165, 0.84, 0.44, 1) 0s reveal-content forwards;
+    background-color: $white;
```
paulfalgout, line 37: making the app-root the background color requires the content to have a white background when not empty.  All other elements of the app frame include a background style

src/js/views/globals/prelogin/prelogin.hbs
```diff
@@ -1,20 +1,21 @@
...
-  </svg>
+<div class="prelogin">
```
paulfalgout, line 1: This change is just nesting the template in the `prelogin` class. This simplifies how this class is applied

nmajor25 (approved):

PR #912, opened 2023-02-01
Update worklist icons to match new navigation icons

nmajor25:
Shortcut Story ID: [sc-33239]

New navigation icons were added in PR: https://github.com/RoundingWell/care-ops-frontend/pull/897.

This PR updates the icons used in the worklist title to match what's in the new navigation. Which we forgot to do in the original PR.

### DIFF
--- a/src/js/views/patients/worklist/worklist_views.js
+++ b/src/js/views/patients/worklist/worklist_views.js
@@ -129,6 +129,14 @@ const NoOwnerToggleView = View.extend({
   },
 });
 
+const worklistIcons = {
+  'owned-by': ['list'],
+  'shared-by': ['arrow-right-arrow-left'],
+  'new-past-day': ['angle-left', '1'],
+  'updated-past-three-days': ['angle-left', '3'],
+  'done-last-thirty-days': ['3', '0'],
+};
+
 const ListTitleView = View.extend({
   regions: {
     owner: '[data-owner-filter-region]',
@@ -136,7 +144,11 @@ const ListTitleView = View.extend({
   },
   className: 'flex list-page__title-filter',
   template: hbs`
-    <span class="list-page__title-icon">{{far "list"}}</span>
+    <span class="list-page__title-icon">
+      {{#each icons}}
+        {{far this}}
+      {{/each}}
+    </span>
     {{#if showOwnerDroplist}}
       <div class="u-text--nowrap">
         {{formatMessage (intlGet "patients.worklist.worklistViews.listTitleView.listLabels") title=worklistId}}
@@ -147,17 +159,20 @@ const ListTitleView = View.extend({
     {{/if}}
     <span class="list-page__header-icon js-title-info">{{far "circle-info"}}</span>
     <div class="u-margin--l-24" data-owner-toggle-region></div>
-    `,
+  `,
   ui: {
     tooltip: '.js-title-info',
   },
   templateContext() {
+    const worklistId = this.getOption('worklistId');
+
     return {
       team: this.getOption('team').get('name'),
       owner: this.getOption('owner').get('name'),
-      worklistId: underscored(this.getOption('worklistId')),
+      worklistId: underscored(worklistId),
       isFlowList: this.getOption('isFlowList'),
       showOwnerDroplist: this.getOption('showOwnerDroplist'),
+      icons: worklistIcons[worklistId],
     };
   },
   onRender() {
--- a/package.json
+++ b/package.json
@@ -69,11 +69,15 @@
       "xmark"
     ],
     "far": [
+      "0",
+      "1",
+      "3",
       "address-card",
       "angle-down",
       "angle-left",
       "angle-right",
       "arrow-down-arrow-up",
+      "arrow-right-arrow-left",
       "arrow-up-right-from-square",
       "calendar-days",
       "calendar-star",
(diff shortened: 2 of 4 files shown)


src/js/apps/globals/nav/nav_items.js
```diff
@@ -0,0 +1,150 @@
...
+    event: 'worklist',
+    eventArgs: ['done-last-thirty-days'],
+  },
+]);
+
+export {
+  appNavMenu,
+  adminNavMenu,
+  patientsAppWorkflowsNav,
+};
```
nmajor25: Moved these to their own file and exported them. This way I can import them into both the `nav_app.js` and `worklist_app.js` files.

Better place to put these?? Wasn't 100% sure where to put them.

src/js/apps/globals/nav/nav_items.js
```diff
@@ -0,0 +1,150 @@
...
+    icon: {
+      type: 'far',
+      icon: 'users-gear',
+    },
+  },
+]);
+
+const patientsAppWorkflowsNav = new Backbone.Collection([
+  {
+    id: 'owned-by',
```
nmajor25: The one new thing I added to each model in this `patientsAppWorkflowsNav` collection here. It represents the `worklistId` each of these belongs to.

This `id` is used by `worklist_app.js` to determine which icon/s to use in the `ListTitleView`.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -301,12 +303,17 @@ export default App.extend({
     const showOwnerDroplist = (this.shouldShowClinician && this.canViewAssignedActions) || this.shouldShowTeam;
     const shouldShowOwnerToggle = this.getState().id === 'shared-by' && this.canViewAssignedActions;
 
+    const currentNavItem = patientsAppWorkflowsNav.find(model => {
+      return model.get('id') === this.worklistId;
```
paulfalgout: `model.id`

paulfalgout: well really this is `patientsAppWorkflowsNav.get(this.worklistId)`

src/js/views/globals/app-nav/app-nav.scss
```diff
@@ -167,7 +167,7 @@ $menu-hover-color: #375467;
 
   .icon.fa-0 {
     font-size: 16px;
-    width: 10px;
+    width: 9.5px;
```
paulfalgout: yeesh.. I dunno about this one.  I don't think it's worth it.

src/js/views/programs/programs-all_views.js
```diff
@@ -44,7 +44,7 @@ const LayoutView = View.extend({
     <div class="list-page__header">
       <div class="flex list-page__title">
         <div class="flex list-page__title-filter">
-          <span class="list-page__title-icon">{{far "screwdriver-wrench"}}</span>{{ @intl.programs.programsAllViews.layoutView.title }}
+          <span class="list-page__title-icon flex flex-align-center">{{far "screwdriver-wrench"}}</span>{{ @intl.programs.programsAllViews.layoutView.title }}
```
paulfalgout: I think if every instance of `list-page__title-icon` needs the flex classes we can add it to that class.

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -136,7 +136,11 @@ const ListTitleView = View.extend({
   },
   className: 'flex list-page__title-filter',
   template: hbs`
-    <span class="list-page__title-icon">{{far "list"}}</span>
+    <span class="list-page__title-icon flex flex-align-center">
+      {{#each icons}}
+        {{far this.icon}}
```
paulfalgout: I didn't realize the icons would also be a different font family with different style requirements..  I no longer think we should reuse the nav definitions.  it's not removing enough duplication.. 

This can just be 
```js
const worklistIcons {
  'shared-by': ['arrow-right-arrow-left'],
  'done-last-thirty-days': ['3', '0'],
...
templateContext() {
...
icons: worklistIcons[worklistId]
```

paulfalgout (changes requested):
I think I overcomplicated this thinking it'd be the same icons, but really it's quite different.

I'm leaving all of my comments as I went through, but overall I think maybe we can revert changes to the nav.

src/js/apps/patients/worklist/worklist_app.js
nmajor25: Yup that's a better way. But deleted this variable all together in the last push. Since it's no longer being passed down to the `ListTitleView`.

src/js/views/globals/app-nav/app-nav.scss
nmajor25: Okay, removed in the last push.

src/js/views/programs/programs-all_views.js
nmajor25: Good call, updated it to do that in the last push.

src/js/views/patients/worklist/worklist_views.js
nmajor25: Updated to do this in the last push.

And reverted the `nav_app.js` file and removed the `nav_items.js` file.

paulfalgout (approved):

PR #915, opened 2023-02-02
Update rw auth connection to v2 api

paulfalgout:
`connection` is now part of `authorizationParams`

Shortcut Story ID: [sc-33402]

I have no idea if this works, but seems like it will.  Has to be on a non dev environment to really prove it.


paulfalgout:
Included in https://github.com/RoundingWell/care-ops-frontend/tree/release/20230202

nmajor25 (approved):

PR #917, opened 2023-02-06
Workspace switching

nmajor25:
Shortcut Story ID: [sc-33460][sc-33875]


test/integration/patients/worklist/worklist.js
```diff
@@ -2534,126 +2324,6 @@ context('worklist page', function() {
...
-      .click();
-
-    cy
-      .get('[data-sidebar-region]')
-      .find('[data-filter-button]')
-      .should('have.length', 1)
-      .first()
-      .get('.sidebar__label')
-      .should('contain', 'Insurance Plans');
-  });
```
nmajor25, line 2656: Should these two be removed?? I wasn't 100% sure...

test/integration/patients/worklist/schedule.js
```diff
@@ -1331,41 +1201,10 @@ context('schedule page', function() {
...
-
-    cy
-      .url()
-      .should('contain', 'schedule');
-
-    cy
-      .get('[data-list-region]')
-      .find('.schedule-list__list-row')
-      .should('have.length.above', 0);
-  });
```
nmajor25, line 1363: Good to remove??

test/integration/patients/worklist/schedule.js
```diff
@@ -2103,51 +1942,4 @@ context('schedule page', function() {
...
-      .click();
-
-    cy
-      .get('[data-sidebar-region]')
-      .find('[data-filter-button]')
-      .should('have.length', 1)
-      .first()
-      .get('.sidebar__label')
-      .should('contain', 'Insurance Plans');
-  });
```
nmajor25, line 2154: Good to remove?

Looks like this just tests that the workspace filter is not shown at the top of the list and in the filters sidebar.

src/js/views/patients/schedule/filters_views.js
```diff
@@ -3,19 +3,15 @@ import { View } from 'marionette';
 
 import 'scss/modules/buttons.scss';
 
-import Droplist from 'js/components/droplist';
-
 import './schedule.scss';
 
 const FiltersView = View.extend({
```
paulfalgout: Is this view needed?

paulfalgout: Is this file needed even

paulfalgout (changes requested):

src/js/views/patients/schedule/filters_views.js
nmajor25: I don't think it is. In the last push I removed this `/schedule/filters_views.js` file and the `/worklist/filters_views.js` files. And moved the `AllFiltersButtonView` to the `schedule_views.js` and `worklist_views.js` files.

src/js/views/patients/schedule/schedule.scss
```diff
@@ -1,17 +0,0 @@
...
-  margin-right: 24px;
-}
-
-.schedule__filter:not(:empty) {
-  margin-right: 8px;
-}
-
-.schedule__list-search {
-  margin-left: auto;
-}
```
nmajor25, line 17: I deleted this file and moved these styles over to the `schedule-list.scss` file instead. As it didn't make sense why these were separated.

But I only ended up moving `.schedule__list-search` over. 

The `.schedule__filters` class was removed from the view since it has a wrapper `div` that applies these same exact styles, so it wasn't doing anything.

`.schedule__all-filters:not(:empty)` and `schedule__filter:not(:empty)` were removed since those aren't doing anything anymore with the changes we've made to the views.

src/js/views/patients/worklist/filters_views.js
```diff
@@ -1,55 +0,0 @@
...
-  className: 'worklist-list__filters',
-  template: hbs`
-    <div class="worklist-list__all-filters" data-all-filters-region></div>
-    <div class="worklist-list__filter" data-workspace-filter-region></div>
-    <div class="worklist-list__filter" data-owner-toggle-region></div>
-  `,
-  regions: {
-    allFilters: '[data-all-filters-region]',
-    workspace: '[data-workspace-filter-region]',
-    ownerToggle: '[data-owner-toggle-region]',
```
nmajor25, line 20: This `ownerToggle` region was moved to the `ListTitleView` a while ago, but wasn't removed from the `FiltersView` here. So this region wasn't be used for anything.

So I removed that region when I moved these views over to the `worklist_views.js` file.

test/integration/patients/worklist/schedule.js
```diff
@@ -665,29 +638,23 @@ context('schedule page', function() {
...
-      .get('.list-page__filters')
-      .find('[data-workspace-filter-region]')
-      .should('contain', 'Group One');
+      .find('.qa-all-filters')
+      .should('contain', '1');
 
     cy
       .get('.list-page__filters')
-      .find('[data-all-filters-region]')
+      .find('.qa-all-filters')
```
nmajor25: If we don't want to use a `.qa-` selector here, I could do this instead:

```js
cy
  .get('[data-filters-region]')
  .find('button')
```

paulfalgout: yep better without it

nmajor25: In the last push, I removed `.qa-all-filters` and updated the tests to `.find('button')` instead.

PR #918, opened 2023-02-08
Update `prelogin.css` to a `.scss` file

nmajor25:
Shortcut Story ID: [sc-33392]

This was originally a `.css` file because the form iframe needed authentication. But now that it doesn't need authentication, the `prelogin.css` file can be made `.scss` without breaking the form styles.

The new `prelogin.scss` file is also updated to include sass code patterns we use elsewhere in the app.


src/js/views/globals/prelogin/prelogin.scss
```diff
@@ -1,23 +1,3 @@
...
-
-@font-face {
-  font-family: 'ProximaSoft';
-  font-style: normal;
-  font-weight: bold;
-  src: url('/fonts/ProximaSoft/3955CC_1_0.woff2') format('woff2'),
-    url('/fonts/ProximaSoft/3955CC_1_0.woff') format('woff'),
-    url('/fonts/ProximaSoft/3955CC_1_0.ttf') format('truetype');
-}
-
```
nmajor25: I think this is good to remove...? The fonts look the same without it.

src/js/views/globals/prelogin/prelogin.scss
```diff
@@ -85,13 +65,13 @@ body {
 }
 
 .prelogin__continue-button {
```
paulfalgout, line 50: This looks like it is reproducing our button classes.  Can we reuse the buttons module?

paulfalgout (changes requested):

src/js/views/globals/prelogin/prelogin.scss
nmajor25: Yup, good call. Just pushed an update that uses the buttons module.

paulfalgout: well I'd assume we could replace this class in the html with some `button-class`

src/js/views/globals/prelogin/prelogin.scss
```diff
@@ -0,0 +1,107 @@
+body {
+  margin: 0;
+}
+
+.prelogin * {
```
paulfalgout: I bet this is covered by the core scss as well.

paulfalgout (changes requested):

src/js/views/globals/prelogin/prelogin.scss
nmajor25: After importing `'scss/provider-core.scss'`, I was able to remove this `.prelogin *` along with several other properties in the file that were duplicates from the core scss.

src/js/views/globals/prelogin/prelogin.scss
nmajor25: The closest button class is:

```css
.button--large {
  @extend .button;

  font-size: 16px;
  padding: 8px 40px;
}
```

But it doesn't have the correct `margin-top` or `padding`.

paulfalgout: ah I didn't realize `.button--green` was already in the html.. clearly wasn't doing anything before.

paulfalgout (approved):

src/js/views/globals/prelogin/prelogin.scss
nmajor25: Correct, before `'scss/modules/buttons.scss';` was imported, it wasn't doing anything.

PR #920, opened 2023-02-09
Remove `include=` from `/actions`, `/actions/{id}`, & `/flows` API requests

nmajor25:
Shortcut Story ID: [sc-33614]

`include=` is deprecated by the backend in these API routes:

* GET `/actions`
* GET `/actions/{id}`
* GET `/flows`


shadowhand:
@nmajor25 if FE actually uses these includes, we don't need to deprecate them. There was a period of time when we wanted to get rid of `include` across the app, but that is no longer the case.

paulfalgout:
Yep these are definitely in use.

PR #921, opened 2023-02-13
Hardcode the user’s first workspace for header

paulfalgout:


paulfalgout:
We're gonna get workspace switching working

PR #923, opened 2023-02-14
Remove state colors from `Cypress.env`

nmajor25:
Shortcut Story ID: [sc-30278]

Export them from a `/test/helpers/state-colors.js` file instead.

### DIFF
--- a/test/state-colors.json
+++ b/test/state-colors.json
@@ -1,8 +1,0 @@
-{
-  "error": "#E65140",
-  "info": "#0582DA",
-  "success": "#00BF68",
-  "warning": "#FFBE00",
-  "highlighted": "#FFF8E5",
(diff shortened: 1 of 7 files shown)


test/state-colors.json
```diff
@@ -1,8 +0,0 @@
-{
-  "error": "#E65140",
-  "info": "#0582DA",
-  "success": "#00BF68",
-  "warning": "#FFBE00",
-  "highlighted": "#FFF8E5",
-  "notspecified": "#CCCCCC"
-}
```
nmajor25, line 8: I don't believe this file is being used by anything. Even before this PR.

And the `error` color is incorrect in this file as well.

paulfalgout (changes requested):
just the codeql alert

PR #924, opened 2023-02-14
Update dashboard iframe template to accomodate URL parameters

nmajor25:
Shortcut Story ID: [sc-33693]


shadowhand (approved):

shadowhand (changes requested):
Actually I don't think this is desirable.

paulfalgout:
It doesn't seem like the result we're looking for is: `/test-dashboard?foo&#x3D;bar#p.baz&#x3D;bazing`, but that's what `{{ embed_url }}` produces.

PR #928, opened 2023-02-17
Fetch form in the action form app

paulfalgout:
Shortcut Story ID: [sc-33890][sc-33832]

We do this for patient and preview forms, but were relying on the cache to grab it from the action.

Note: also includes removing the `form.fetch()` in the form service. Which is no longer needed after this PR.


shadowhand (approved):

paulfalgout:
this is going to have some breaking tests because they're expecting the stubbed form endpoint and it's not stubbed..

test/support/api/forms.js
```diff
@@ -18,20 +18,20 @@ Cypress.Commands.add('routeForms', (mutator = _.identity) => {
...
     response() {
       return mutator({
         data: getResource(_.find(this.fxTestForms, { id: formId }), 'forms'),
         included: [],
       });
     },
   })
-    .as('routeForm');
+    .as('routeFormByAction');
 });
```
nmajor25, line 56: Is this correct?

New `routeFormByAction` command is identical to `routeForm` (now removed), apart from the URL?

paulfalgout:
@nmajor25 errors are legit I think.  Seems like a couple more route fixes

nmajor25:
@paulfalgout weirdly, these failing tests don't fail for me locally. But I just pushed a fix to hopefully the last one 🤞.

PR #934, opened 2023-02-24
Add workspaces to the top nav menu

nmajor25:
Shortcut Story ID: [sc-33871]

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -16,29 +16,6 @@ const StateModel = Backbone.Model.extend({
   },
 });
 
-const appNavMenu = new Backbone.Collection([
-  {
-    onSelect() {
-      window.open('https://help.roundingwell.com/');
-    },
-    text: i18n.mainNav.help,
-    icon: {
-      type: 'far',
-      icon: 'life-ring',
-    },
-  },
-  {
-    onSelect() {
-      Radio.request('auth', 'logout');
-    },
-    text: i18n.mainNav.signOut,
-    icon: {
-      type: 'fas',
-      icon: 'right-from-bracket',
-    },
-  },
-]);
-
 const adminNavMenu = new Backbone.Collection([
   {
     onSelect() {
@@ -310,9 +287,32 @@ export default RouterApp.extend({
     }
   },
   showMainNavDroplist() {
+    const currentWorkspace = Radio.request('bootstrap', 'currentWorkspace');
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    const workspaces = currentUser.getWorkspaces();
+
+    const workspacesMenu = new Backbone.Collection(
+      workspaces.map(workspace => {
+        return {
+          id: workspace.id,
+          onSelect() {
+            Radio.trigger('event-router', `workspace:${ workspace.get('slug') }`);
+          },
+          text: workspace.get('name'),
+          icon: {
+            type: 'far',
+            icon: 'window',
+          },
+        };
+      }),
+    );
+
     this.showChildView('navMain', new MainNavDroplist({
-      collection: appNavMenu,
-      state: this.getState(),
+      collection: workspacesMenu,
+      state: {
+        selected: workspacesMenu.get(currentWorkspace.id),
+        isMinimized: this.getState('isMinimized'),
+      },
     }));
   },
   showNav() {
--- a/src/js/components/picklist/index.js
+++ b/src/js/components/picklist/index.js
@@ -31,6 +31,9 @@ const CLASS_OPTIONS = [
   'noResultsText',
   'placeholderText',
   'template',
+  'templateContext',
+  'triggers',
+  'ui',
 ];
 
 const CLASS_OPTIONS_ITEM = [
--- a/src/js/views/globals/app-nav/app-nav.scss
+++ b/src/js/views/globals/app-nav/app-nav.scss
@@ -26,10 +26,38 @@ $menu-hover-color: #375467;
   vertical-align: 0;
 }
 
+.app-nav__picklist-heading {
+  color: $black-60;
+  font-size: 14px;
+  line-height: 16px;
+  margin-bottom: 4px;
+  padding: 0 16px;
+  white-space: nowrap;
+}
+
+.app-nav__picklist-workspace-name {
+  border-bottom: 1px solid $black-90;
+  color: $black-20;
+  font-size: 16px;
+  font-weight: 600;
+  margin-bottom: 8px;
+  padding: 0 16px 12px;
+
+  &.is-highlighted:not(&.is-disabled),
+  &:hover:not(&.is-disabled) {
+    background-color: $white;
+    color: $black-20;
+  }
+}
+
 .app-nav__picklist-item {
   color: $black-20;
   font-weight: 600;
   padding: 8px 16px;
+
+  .icon {
+    color: $black-20;
+  }
 
   &.is-highlighted:not(&.is-disabled),
   &:hover:not(&.is-disabled),
@@ -39,7 +67,7 @@ $menu-hover-color: #375467;
     position: relative;
 
     .icon {
-      color: $black-20;
+      color: color(blue);
     }
 
     &::before {
@@ -52,6 +80,12 @@ $menu-hover-color: #375467;
       width: 4px;
     }
   }
+}
+
+.app-nav__picklist-bottom {
+  border-top: 1px solid $black-90;
+  margin-top: 8px;
+  padding-top: 8px;
 }
 
 .app-nav__header {
--- a/src/js/views/globals/app-nav/app-nav_views.js
+++ b/src/js/views/globals/app-nav/app-nav_views.js
@@ -23,7 +23,7 @@ const MainNavDroplist = Droplist.extend({
   viewOptions: {
     tagName: 'div',
     className() {
-      if (this.getOption('state').get('isMinimized')) return 'app-nav__header minimized';
+      if (this.getOption('state').isMinimized) return 'app-nav__header minimized';
       return 'app-nav__header';
     },
     template: hbs`
@@ -31,7 +31,7 @@ const MainNavDroplist = Droplist.extend({
         <img class="app-nav__header-logo" src="/rwell-logo.svg" />
       {{else}}
         <div class="u-text--overflow">
-          <h2 class="app-nav__header-title u-text--overflow">{{ orgName }}</h2>
+          <h2 class="app-nav__header-title u-text--overflow">{{ workspaceName }}</h2>
           <span class="app-nav__header-arrow">{{far "angle-down"}}</span>
         </div>
         <div class="u-text--overflow">{{ userName }}</div>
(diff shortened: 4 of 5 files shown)


src/js/components/picklist/index.js
```diff
@@ -107,7 +108,9 @@ const Picklist = CollectionView.extend({
   className: 'picklist__group',
   tagName: 'li',
   template: hbs`
-    {{#if headingText}}<div class="picklist__heading">{{ headingText }}</div>{{/if}}
+    {{#if headingText}}
+      <div class={{ headingClassName }}>{{ headingText }}</div>
+    {{/if}}
```
nmajor25: This allows me to apply specific styling to the headings in the top nav menu, without effecting other picklists. For the top nav menu, we want those headings to have `14px` instead of the normal `12px` used on other picklists.

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -31,41 +32,74 @@ const MainNavDroplist = Droplist.extend({
...
-          <span>{{formatMessage text}}</span>
-        `,
-      }],
     };
   },
+  onPicklistSelect({ model }) {
+    if (model.get('disableOnSelect')) return;
+
+    this.popRegion.empty();
+  },
```
nmajor25, line 85: This overrides the droplist's version of `onPicklistSelect()`. It stops the picklist from closing when the organization section of the top nav menu is clicked on. Since that area is just for display purposes.

This also removes `this.setState('selected', model);`, which keeps the selected workspace as the correct `state.selected` item when the help nav item is clicked on.

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -31,41 +32,74 @@ const MainNavDroplist = Droplist.extend({
...
+    this.lists = [];
+
+    this.lists.push({
+      collection: new Backbone.Collection([{
+        text: currentOrg.get('name'),
+        disableOnSelect: true,
+      }]),
+      headingText: i18n.mainNavDroplist.organizationHeading,
+      headingClassName: 'app-nav__picklist-heading',
+      itemClassName: 'app-nav__picklist-workspace-name',
```
nmajor25: Added a separate classname to the organization name, so I could apply different font size and padding properties.

src/js/components/picklist/index.js
```diff
@@ -130,9 +133,12 @@ const Picklist = CollectionView.extend({
     return extend({ state: this.model }, opts);
   },
   templateContext() {
+    const headingClassName = this.getOption('headingClassName') ?? 'picklist__heading';
```
paulfalgout: I think `headingClassName: 'picklist__heading',` should be added up by `className` then only the `getOption` is needed.

src/js/apps/globals/nav_app.js
```diff
@@ -304,8 +304,26 @@ export default RouterApp.extend({
     }
   },
   showMainNavDroplist() {
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    const workspaces = currentUser.getWorkspaces();
+
+    const workspacesMenuItems = workspaces.map(workspace => {
+      return {
+        id: workspace.get('id'),
```
paulfalgout: just `workspace.id`

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -31,41 +32,74 @@ const MainNavDroplist = Droplist.extend({
...
         userName: currentUser.get('name'),
-        orgName: currentOrg.get('name'),
+        workspaceName: currentWorkspace.get('name'),
         isMinimized: this.getOption('state').get('isMinimized'),
       };
     },
   },
   picklistOptions() {
     return {
       className: 'picklist app-nav__picklist',
```
paulfalgout, line 53: I think this can be simplified a bit by passing a template here:
something like:
```js
template: hbs`
    <div class="app-nav__picklist-heading">{{ @intl.globals.appNav.appNavViews.mainNavDroplist.organizationHeading }}</div>
    </div>
    <div class="app-nav__picklist-workspace-name">{{ name }}</div>
    <ul class="flex-region picklist__scroll js-picklist-scroll"></ul>
`
```
Then it seems like `headingClassName`, `itemClassName`, and `itemTemplate` could probably be defined here as well rather than on each list.

Which I think allows for:
```js
const currentOrg...
const workspacesMenu....
this.showChildView('navMain', new MainNavDroplist({
  model: currentOrg,
  lists: [{ collection: workspacesMenu }, { collection: appNavMenu }], 
  state: this.getState(),
}));
```

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -31,41 +32,74 @@ const MainNavDroplist = Droplist.extend({
...
+    this.lists.push({
+      collection: appNavMenuItems,
+      itemTemplate: hbs`
+        {{fa icon.type icon.icon classes=icon.classes~}}
+        <span>{{formatMessage text}}</span>
+      `,
+      itemClassName: 'flex flex-align-center app-nav__picklist-item',
+    });
+
+    this.setState('selected', workspacesList.collection.find({ id: currentWorkspace.get('id') }));
```
paulfalgout: this'll always be the same as `this.setState('selected', currentWorkspace.id)`

src/js/views/globals/app-nav/app-nav_views.js
paulfalgout: Although.. I'm now thinking.. why not include the help and sign out links in the template?
Otherwise we have to deal with some items changing the selected state and some not..  then all we're passing is a collection of workspaces.

paulfalgout (changes requested):

src/js/components/picklist/index.js
nmajor25: Updated in last push.

src/js/apps/globals/nav_app.js
nmajor25: Updated in last push.

src/js/views/globals/app-nav/app-nav_views.js
nmajor25: But the entire model needs to be passed to the picklist correct?

To determine which item is selected, the picklist looks for `model` === `model`. Not `model.id` === `model.id`.

At least that's my understanding of how that works.

Doing `this.setState('selected', currentWorkspace.id)` didn't work.

src/js/views/globals/app-nav/app-nav_views.js
nmajor25: Updated to do this in the last push.

But ran into an issue with the help and sign out event trigger functions. I'll make comments on those portions of the code.

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -31,34 +31,61 @@ const MainNavDroplist = Droplist.extend({
...
+      ui: {
+        help: '.js-help',
+        signOut: '.js-sign-out',
+      },
+      onHelp() {
+        window.open('https://help.roundingwell.com/');
+      },
+      onSignOut() {
+        Radio.request('auth', 'logout');
+      },
```
nmajor25: `onHelp()` and `onSignOut()` are not working... any ideas?

Only way I can get these to call is if I add them directly to the picklist (`/js/components/picklist/index.js` file).

src/js/apps/globals/nav_app.js
```diff
@@ -310,9 +287,36 @@ export default RouterApp.extend({
...
+            icon: 'window',
+          },
+        };
+      }),
+    );
+
+    const workspacesMenuList = {
+      collection: workspacesMenu,
+      headingText: i18n.mainNavDroplist.workspacesHeading,
+      headingClassName: 'app-nav__picklist-heading',
```
nmajor25: I think `headingClassName` needs to stay here...

src/js/views/globals/app-nav/app-nav_views.js
paulfalgout: This is the only list of options you can pass at instantiation: https://github.com/marionettejs/backbone.marionette/blob/master/src/view.js#L12

So anything else'd have to be `this.listenTo(view, 'help'...` elsewhere

But the droplist can have a `picklistEvents` like the `picklistOptions` which'd be something like:
```
picklistEvents: { 
  'help'() {
    window.open('https://help.roundingwell.com/');
  },
  'signOut'() {
    Radio.request('auth', 'logout');
  },
},
```

or the value can be a function name on the droplist

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -67,6 +94,11 @@ const MainNavDroplist = Droplist.extend({
   onSelect({ model }) {
     model.get('onSelect')();
   },
+  initialize({ lists, selected }) {
+    this.lists = lists;
```
paulfalgout: lists is attached by default so it doesn't need to happen here.

Also the `selected` can be passed in on a `state` object as well as in this example below, so the `initialize` isn't necessary

```
    const sortSelect = new SortDroplist({
      collection: this.sortOptions,
      state: { selected: this.sortOptions.get(this.getState().getSort()) },
    });

```

src/js/views/globals/app-nav/app-nav.scss
```diff
@@ -26,11 +26,39 @@ $menu-hover-color: #375467;
...
+  color: $black-60;
+  font-size: 14px;
+  line-height: 16px;
+  margin-bottom: 4px;
+  padding: 0 16px;
+  white-space: nowrap;
+}
+
+.app-nav__picklist-workspace-name {
+  border-bottom: 1px solid #E6E6E6;
```
paulfalgout: $black-90

src/js/views/globals/app-nav/app-nav_views.js
```diff
@@ -31,34 +31,61 @@ const MainNavDroplist = Droplist.extend({
...
-        collection: this.collection,
-        itemTemplate: hbs`
-          {{fa icon.type icon.icon classes=icon.classes~}}
-          <span>{{formatMessage text}}</span>
-        `,
-      }],
+      template: hbs`
+        <div class="app-nav__picklist-heading">{{ @intl.globals.appNav.appNavViews.mainNavDroplist.organizationHeading }}</div>
+        <div class="app-nav__picklist-workspace-name">{{ name }}</div>
+        <ul class="flex-region picklist__scroll js-picklist-scroll"></ul>
```
paulfalgout, line 58: What if we put the heading text for workspaces above the `ul` here and then we could avoid modifying the picklist component.  It might be inevitable at some point, but for now it's an edge case.. I know it'll scroll a little differently, but that may not matter.

src/js/apps/globals/nav_app.js
```diff
@@ -310,9 +287,36 @@ export default RouterApp.extend({
...
+    };
+
     this.showChildView('navMain', new MainNavDroplist({
-      collection: appNavMenu,
+      lists: [workspacesMenuList],
       state: this.getState(),
+      selected: workspacesMenu.find({ id: currentWorkspace.id }),
```
paulfalgout: `myCollection.find({ id: foo.id })` is functionally the same, but more expensive than `workspacesMenu.get(currentWorkspace.id)`

However in this case the following will always be true:
`currentWorkspace === workspacesMenu.find({ id: currentWorkspace.id })`
so `state: { selected: currentWorkspace }` is all that's necessary..

it's actually an issue that the state model is passed as `state` here as the droplist has it's own state.
so this should be:

```
state: {
  selected: currentWorkspace,
  isMinimized: this.getState('isMinimized'),
}
```
then the droplist should use `getState('isMinimized')` instead of `this.getOption('state').get('isMinimized');`

src/js/apps/globals/nav_app.js
```diff
@@ -310,9 +287,36 @@ export default RouterApp.extend({
...
+
+    const workspacesMenuList = {
+      collection: workspacesMenu,
+      headingText: i18n.mainNavDroplist.workspacesHeading,
+      headingClassName: 'app-nav__picklist-heading',
+    };
+
     this.showChildView('navMain', new MainNavDroplist({
-      collection: appNavMenu,
+      lists: [workspacesMenuList],
```
paulfalgout: you can just pass this as `collection: workspacesMenuList`
The droplist only needs a list if there's more than one.

paulfalgout (changes requested):

src/js/views/globals/app-nav/app-nav_views.js
paulfalgout: oops yeah.. remove the `id`

src/js/views/globals/app-nav/app-nav.scss
nmajor25: Updated to use this in the last push.

src/js/views/globals/app-nav/app-nav_views.js
nmajor25: Nice this works. Updated to use this in the last push. So those droplist items work properly now.

PR #935, opened 2023-02-27
User current workspace for owner component

paulfalgout:
Shortcut Story ID: [sc-33869]

We must load the clinicians every time we change workspaces, but workspace only bootstrap at the beginning.  Because they are fetched at different times there's a chance they get outdated so this updates the clinician relations on the workspace to prevent errors.

Then it moves all the workspace/team caching for the owner component into the component and clears the cache when the current workspace changes.

The clinician owner component was following a similar caching strategy, but with no clearing, and there's really no point in doing so


paulfalgout:
[sc-34046] Added this

paulfalgout:
[sc-34047] and this

nmajor25 (approved):

PR #937, opened 2023-02-28
Move dashboards link to the left nav

nmajor25:
Shortcut Story ID: [sc-33794]

Before this link was in the admin tools droplist:

Screenshot 2023-02-28 at 9 47 18 AM

Now it will be in the nav:

Screenshot 2023-02-28 at 9 46 31 AM

Still hidden if the user doesn't have the `dashboards:view` permission.

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -7,7 +7,7 @@ import store from 'store';
 import RouterApp from 'js/base/routerapp';
 
 import SearchApp from './search_app';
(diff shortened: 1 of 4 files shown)


nmajor25:
@paulfalgout this still needs the tests updated.

paulfalgout (changes requested):
This all makes sense to me.  But can you cherry pick this commit in, and then update the height to account for the added item? https://github.com/RoundingWell/care-ops-frontend/pull/917/commits/0328a08304af0f5b90fff3df01dfc649765f96e3

paulfalgout:
alternatively we can just wait until the workspace switching is in, rebase and then update it.. might be easier

PR #942, opened 2023-03-01
Don't trigger event-router event if the user selects the current workspace

nmajor25:
Shortcut Story ID: [sc-34106]


src/js/views/globals/app-nav/app-nav.scss
```diff
@@ -59,6 +59,10 @@ $menu-hover-color: #375467;
     color: $black-20;
   }
 
+  &.is-selected {
+    pointer-events: none;
+  }
+
```
nmajor25: This is causing some weird things with hovering. Where the hover CSS doesn't get removed when the user is no longer hovering over the item.


https://user-images.githubusercontent.com/35355575/222220248-d6f26618-c29b-4f20-8658-56a9c0ca1027.mov

nmajor25: Or maybe that's not the cause. But something is...

PR #943, opened 2023-03-01
Restart the nav app when 'change:workspace' is triggered

nmajor25:
Shortcut Story ID: [sc-34124]

Fixes an issue where the incorrect workspace was being shown in the top nav menu button. The nav app wasn't restarting when the `change:workspace` was triggered, which means it didn't have the correct `currentWorkspace` when a user changes between workspaces.

### DIFF
--- a/src/js/apps/globals/app-frame_app.js
+++ b/src/js/apps/globals/app-frame_app.js
@@ -20,9 +20,10 @@ export default App.extend({
     this.currentWorkspace = Radio.request('bootstrap', 'currentWorkspace');
     this.getRegion('content').empty();
 
+    new NavApp({ region: this.getRegion('nav') });
+
     if (this.isRestarting()) return;
 
-    new NavApp({ region: this.getRegion('nav') });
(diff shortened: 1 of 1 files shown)


src/js/apps/globals/app-frame_app.js
```diff
@@ -20,9 +20,10 @@ export default App.extend({
     this.currentWorkspace = Radio.request('bootstrap', 'currentWorkspace');
     this.getRegion('content').empty();
 
+    new NavApp({ region: this.getRegion('nav') });
+
```
nmajor25, line 24: This fixes the issue with the incorrect workspace being shown in the top nav button (this story: [sc-34124](https://app.shortcut.com/roundingwell/story/34124/bug-showing-incorrect-current-workspace-in-top-nav-menu-button)). Which only occurs after the user switches workspaces.

But I'm wary of unintended consequences of this...

paulfalgout:
Replaced by #946

PR #944, opened 2023-03-02
Fix go back history on forms

paulfalgout:
Shortcut Story ID: [sc-34125]

So this latest history service is working far far better than the previous one, but 🤦 I wasn't removing the current route before going back in this release, so it was going to the current route.

In any case I snuck in an empty history handler which defaults forms to go to the patient's dashboard if there was no previous history recorded @shadowhand   seems to work.

### DIFF
--- a/src/js/services/history.js
+++ b/src/js/services/history.js
@@ -48,7 +48,16 @@ export default App.extend({
   triggerChange() {
     this.getChannel().trigger('change:route', this._prevHistory[0]);
   },
-  goBack() {
+  goBack(handleEmptyHistory) {
+    if (handleEmptyHistory && this._prevHistory.length < 2) {
+      handleEmptyHistory();
+      return;
+    }
+
+    // Remove the current route from the stack
+    this._prevHistory.shift();
+
+    // Navigate to the previous route
     Backbone.history.navigate(this._prevHistory.shift(), { trigger: true });
   },
 });
--- a/src/js/views/forms/form/form_views.js
+++ b/src/js/views/forms/form/form_views.js
@@ -34,9 +34,14 @@ const ContextTrailView = View.extend({
     'click .js-dashboard': 'click:dashboard',
   },
   onClickBack() {
-    Radio.request('history', 'go:back');
+    Radio.request('history', 'go:back', () => {
+      this.routeToPatient();
+    });
   },
   onClickDashboard() {
+    this.routeToPatient();
+  },
+  routeToPatient() {
     Radio.trigger('event-router', 'patient:dashboard', this.patient.id);
   },
 });
--- a/src/js/apps/forms/form/form-patient_app.js
+++ b/src/js/apps/forms/form/form-patient_app.js
@@ -71,39 +71,46 @@ export default App.extend({
       form: this.form,
     });
 
-    this.listenTo(formService, {
-      'submit'() {
-        const saveButtonType = this.getState('saveButtonType');
+    this.bindEvents(formService, this.serviceEvents);
+  },
+  serviceEvents: {
+    'submit': 'onFormServiceSubmit',
+    'success': 'onFormServiceSuccess',
+    'ready': 'onFormServiceReady',
+    'error': 'onFormServiceError',
+  },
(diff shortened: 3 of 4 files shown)


src/js/services/history.js
```diff
@@ -48,7 +48,16 @@ export default App.extend({
...
   },
-  goBack() {
+  goBack(onEmptyHistory) {
+    if (onEmptyHistory && this._prevHistory.length < 2) {
+      onEmptyHistory();
+      return;
+    }
+
+    // Remove the current route from the stack
+    this._prevHistory.shift();
```
shadowhand, line 58: We're sure this is a good idea? Should it happen before the top `if` condition?

src/js/views/forms/form/form_views.js
```diff
@@ -34,9 +34,14 @@ const ContextTrailView = View.extend({
...
   onClickBack() {
-    Radio.request('history', 'go:back');
+    Radio.request('history', 'go:back', () => {
+      this.routeToPatient();
+    });
   },
   onClickDashboard() {
+    this.routeToPatient();
+  },
+  routeToPatient() {
```
shadowhand, line 44:
```suggestion
    Radio.request('history', 'go:back', () => {
      this.onClickDashboard();
    });
  },
  onClickDashboard() {
```

src/js/services/history.js
```diff
@@ -48,7 +48,16 @@ export default App.extend({
   triggerChange() {
     this.getChannel().trigger('change:route', this._prevHistory[0]);
   },
-  goBack() {
+  goBack(onEmptyHistory) {
+    if (onEmptyHistory && this._prevHistory.length < 2) {
+      onEmptyHistory();
+      return;
+    }
```
shadowhand: Clever solution. 👍🏼

src/js/services/history.js
paulfalgout: yeah so it's a _little_ awkward.. if you're directly linking to a form the `_prevHistory` will be just that form url.  so then the go back isn't go back in history.. it's just linking to the patient..  

so in theory if there was a "go back" button on the patient page, if we remove it here, that button wouldn't go back to the form, it would go to the default app route.  I think you'd want it to bounce back and forth between the two routes. 

The `onEmptyHistory` would be clearing the history instead of doing some sort of default redirect, and I don't know if that's the ideal behavior.

src/js/views/forms/form/form_views.js
paulfalgout: yeah.. Ironically despite my naming of the `onEmptyHistory` on this PR an `onFoo` on a class is an indication that it's triggered by an event and we attempt not to call them directly as it's clearer when they need to be cleaned up. _but_ now I'm convinced I should rename `onEmptyHistory` :-D

src/js/views/forms/form/form_views.js
```diff
@@ -34,9 +34,14 @@ const ContextTrailView = View.extend({
     'click .js-dashboard': 'click:dashboard',
   },
   onClickBack() {
-    Radio.request('history', 'go:back');
+    Radio.request('history', 'go:back', () => {
+      this.routeToPatient();
+    });
```
nmajor25, line 39: It's not letting me comment on that line, but does the go back in the `PreviewView` in this file need a `handleEmptyHistory` function as well?

This line here: https://github.com/RoundingWell/care-ops-frontend/pull/944/commits/01d2176d7336ef6d2ce2b1e0772d6f6459ae4b6e#diff-e2f36acf877614e07a231a6f78b71f617ab8c1cc839269088dad7a7469d74a0bR229

paulfalgout: Not sure where we'd go to necessarily.  I guess we could potentially go back to the program, but TBH the form preview is pretty broken.. it can't load context in a way that most forms are probably pretty broken to preview.

nmajor25: Okay, only thing I noticed. Things are working nicely locally.

nmajor25 (approved):

PR #950, opened 2023-03-03
When a new action is created, set its initial state to the first state with `status: queued`

nmajor25:
Shortcut Story ID: [sc-34207]

We currently set `states[0]` to be the initial state when creating a new action. This causes issues if the states data is ordered in a way we don't expect, where `states[0]` is not a state with `status: queued` (e.g. an initial action state is `status: done`).

So we should do this instead: `_state: first(states.filter({ status: 'queued' })).id`.

### DIFF
--- a/src/js/apps/patients/patient/action/action_app.js
+++ b/src/js/apps/patients/patient/action/action_app.js
@@ -1,3 +1,4 @@
+import { first } from 'underscore';
 import Radio from 'backbone.radio';
 import dayjs from 'dayjs';
 
@@ -14,9 +15,11 @@ export default App.extend({
       const currentWorkspace = Radio.request('bootstrap', 'currentWorkspace');
       const states = currentWorkspace.getStates();
 
+      const defaultInitialState = first(states.filter({ status: 'queued' }));
+
       return Radio.request('entities', 'actions:model', {
         _patient: patientId,
-        _state: states.at(0).id,
+        _state: defaultInitialState.id,
         _owner: {
           type: 'clinicians',
           id: currentUser.id,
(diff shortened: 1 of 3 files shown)


paulfalgout (approved):

src/js/entities-service/entities/program-actions.js
```diff
@@ -40,14 +40,11 @@ const _Model = BaseModel.extend({
...
   getAction({ patientId, flowId }) {
     const currentUser = Radio.request('bootstrap', 'currentUser');
-    const currentWorkspace = Radio.request('bootstrap', 'currentWorkspace');
-    const states = currentWorkspace.getStates();
 
     return Radio.request('entities', 'actions:model', {
       name: this.get('name'),
       _flow: flowId,
       _patient: patientId,
-      _state: states.at(0).id,
```
shadowhand, line 50: It looks like this method is converting a program action into an action, so it would need a `state` assigned.

src/js/entities-service/entities/program-flows.js
```diff
@@ -46,13 +46,9 @@ const _Model = BaseModel.extend({
     return Radio.request('entities', 'teams:model', owner.id);
   },
   getFlow(patientId) {
-    const currentWorkspace = Radio.request('bootstrap', 'currentWorkspace');
-    const states = currentWorkspace.getStates();
-
     const flow = Radio.request('entities', 'flows:model', {
       _patient: patientId,
       _program_flow: this.get('id'),
-      _state: states.at(0).id,
```
shadowhand, line 55: Same here, this seems to be creating a flow from a program flow.

src/js/apps/patients/patient/action/action_app.js
```diff
@@ -14,9 +15,11 @@ export default App.extend({
       const currentWorkspace = Radio.request('bootstrap', 'currentWorkspace');
       const states = currentWorkspace.getStates();
 
+      const defaultInitialState = first(states.filter({ status: 'queued' }));
```
shadowhand, line 18: Make make this logic part of `states` or `currentWorkspace`, so that you can do:

```js
const defaultInitialState = currentWorkspace.getDefaultState()
```

PR #951, opened 2023-03-03
Add `_state:` back to program-actions and program-flows entities

nmajor25:
Shortcut Story ID: [sc-34214]

### DIFF
--- a/src/js/entities-service/entities/program-actions.js
+++ b/src/js/entities-service/entities/program-actions.js
@@ -1,4 +1,4 @@
-import { extend } from 'underscore';
+import { extend, first } from 'underscore';
 import Radio from 'backbone.radio';
 import Store from 'backbone.store';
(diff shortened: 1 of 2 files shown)


src/js/entities-service/entities/program-actions.js
```diff
@@ -40,11 +40,16 @@ const _Model = BaseModel.extend({
   },
   getAction({ patientId, flowId }) {
     const currentUser = Radio.request('bootstrap', 'currentUser');
+    const currentWorkspace = Radio.request('bootstrap', 'currentWorkspace');
+    const states = currentWorkspace.getStates();
+
+    const defaultInitialState = first(states.filter({ status: 'queued' }));
```
shadowhand, line 46: This bit could probably be added to `currentWorkspace`, but eh.

shadowhand (approved):

PR #953, opened 2023-03-07
Form submit UI updates

nmajor25:
Shortcut Story ID: [sc-34162] [sc-34069]

Changes made in this PR (applied to both patient and patient action forms):

- Change the verbage on the form submit button from `Save`/`Save + Go Back` => `Submit`/`Submit + Go Back`.
    - Any other places where `Save` verbage existed was also changed to `Submit`.
- Allow for the form submit button to be optionally hidden. 
    - Based on `form.model.options.submit_hidden = BOOLEAN`.
    - Doesn't affect the `Read Only` or `Update` buttons.
- Add a section on the form page that shows the last time the form was updated.
    - Last edit date is based on the `updated: DATE` property stored by the latest submission in local storage ([screenshot](https://user-images.githubusercontent.com/35355575/223540109-f4fcf8e9-352b-4200-9e5d-e4f8e243c4c2.png)).
    - If there's no stored submission, the section is shown without the last edited date ([screenshot](https://user-images.githubusercontent.com/35355575/223540500-637ff1c5-1da1-494a-a0f9-8c8adb32d765.png)).
    - This section is hidden on read only forms or if the form has already been submitted (i.e. when the `Read Only` or `Update` buttons are shown).
    - The last edited date should be removed from the UI after a user discards their stored form submission.
- Show a confirmation modal when a user selects the `Discard My Work` button to remove their stored form submission.
    - [Screenshot](https://user-images.githubusercontent.com/35355575/223541364-1045fb8a-3fe5-40a1-b1a4-3bbc6384891b.png)
- Stop the nav menu from showing when the form page is hard refreshed.

### DIFF
--- a/src/js/apps/forms/form/form_app.js
+++ b/src/js/apps/forms/form/form_app.js
@@ -22,6 +22,7 @@ import {
   SaveView,
   UpdateView,
   HistoryView,
+  LastUpdatedView,
 } from 'js/views/forms/form/form_views';
 
 export default App.extend({
@@ -74,6 +75,7 @@ export default App.extend({
     this.action = action;
     this.responses = action.getFormResponses();
     this.isReadOnly = this.form.isReadOnly();
+    this.isSubmitHidden = this.form.isSubmitHidden();
 
     this.listenTo(action, 'destroy', function() {
       Radio.request('alert', 'show:success', intl.forms.form.formApp.deleteSuccess);
@@ -89,6 +91,9 @@ export default App.extend({
     this.showFormStatus();
     this.showFormAction();
     this.showActions();
+
+    const { updated } = Radio.request(`form${ this.form.id }`, 'get:storedSubmission');
+    this.showLastUpdated(updated);
 
     this.startChildApp('widgetHeader');
 
@@ -110,6 +115,7 @@ export default App.extend({
     'submit': 'onFormServiceSubmit',
     'success': 'onFormServiceSuccess',
     'ready': 'onFormServiceReady',
+    'update:submission': 'onFormServiceUpdateSubmission',
     'error': 'onFormServiceError',
   },
   onFormServiceSubmit() {
@@ -141,6 +147,9 @@ export default App.extend({
   },
   onFormServiceReady() {
     this.showFormSave();
+  },
+  onFormServiceUpdateSubmission(updated) {
+    this.showLastUpdated(updated);
   },
   onFormServiceError() {
     if (this.loadingModal) this.loadingModal.destroy();
@@ -179,6 +188,13 @@ export default App.extend({
       model: this.responses.first(),
       isEditing: !this.getState('responseId'),
     }));
+  },
+  showLastUpdated(updated) {
+    if (this.isReadOnly || this.getState('responseId')) return;
+
+    const lastUpdatedView = new LastUpdatedView({ updated });
+
+    this.showChildView('formUpdated', lastUpdatedView);
   },
   showActions() {
     const formActions = new FormActionsView({
@@ -220,9 +236,10 @@ export default App.extend({
         'submit'() {
           this.showForm();
         },
-        'cancel'() {
+        'discard:submission'() {
           Radio.request(`form${ this.form.id }`, 'clear:storedSubmission');
           this.showForm();
+          this.showLastUpdated();
         },
       });
 
@@ -287,13 +304,15 @@ export default App.extend({
     this.showChildView('formAction', new ReadOnlyView());
   },
   showFormSaveDisabled() {
+    if (this.isSubmitHidden) return;
+
     this.showChildView('formAction', new SaveView({
       isDisabled: true,
       model: this.getState(),
     }));
   },
   showFormSave() {
-    if (this.isReadOnly) return;
+    if (this.isReadOnly || this.isSubmitHidden) return;
 
     const saveView = this.showChildView('formAction', new SaveView({
       model: this.getState(),
--- a/src/js/apps/globals/app-frame_app.js
+++ b/src/js/apps/globals/app-frame_app.js
@@ -63,7 +63,7 @@ export default App.extend({
       }
     });
   },
-  onBeforeStop() {
+  onStop() {
     invoke(this.routers, 'destroy');
     this.routers = [];
   },
@@ -89,6 +89,8 @@ export default App.extend({
     });
   },
   toggleNav(shouldShow) {
+    if (!this.isRunning()) return;
+
     this.getView().toggleNav(!!shouldShow);
   },
 });
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -52,6 +52,7 @@ export default App.extend({
     const updated = dayjs().format();
     try {
       store.set(this.getStoreId(), { submission, updated });
+      this.trigger('update:submission', updated);
     } catch (e) /* istanbul ignore next: Tested locally, test runner borks on CI */ {
       store.each((value, key) => {
         if (String(key).startsWith('form-subm-')) store.remove(key);
--- a/src/js/views/forms/form/form.scss
+++ b/src/js/views/forms/form/form.scss
@@ -39,7 +39,7 @@
 .form__title-icon {
   align-items: center;
   display: flex;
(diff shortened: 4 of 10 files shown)


src/js/services/forms.js
```diff
@@ -52,11 +52,13 @@ export default App.extend({
     const updated = dayjs().format();
     try {
       store.set(this.getStoreId(), { submission, updated });
+      this.trigger('submissionUpdated', updated);
     } catch (e) /* istanbul ignore next: Tested locally, test runner borks on CI */ {
       store.each((value, key) => {
         if (String(key).startsWith('form-subm-')) store.remove(key);
       });
       store.set(this.getStoreId(), { submission, updated });
+      this.trigger('submissionUpdated', updated);
```
nmajor25: Do we need this one?

I don't know how to test that scenario, in Cypress or my sandbox.

src/js/services/forms.js
```diff
@@ -52,11 +52,13 @@ export default App.extend({
     const updated = dayjs().format();
     try {
       store.set(this.getStoreId(), { submission, updated });
+      this.trigger('submissionUpdated', updated);
```
nmajor25: This is used by the `form_app.js` and `form-patient_app.js` to update the last updated time/date. Or it will cause that content to show for the first time if it's the first submission stored in local storage for that form.

src/js/views/forms/form/form_views.js
```diff
@@ -284,8 +299,28 @@ const SaveButtonTypeDroplist = Droplist.extend({
...
+const LastUpdatedView = View.extend({
+  className: 'form__last-updated',
+  template: hbs`
+    <div class="form__last-updated-icon">
+      {{far "shield-check"}}
+    </div>
+    <div class="form__last-updated-text">
+      <div class="u-text--overflow">{{ @intl.forms.form.formViews.lastUpdatedView.storedWork }}</div>
+      {{#if updated}}
+        <div class="u-text--overflow">{{formatHTMLMessage (intlGet "forms.form.formViews.lastUpdatedView.updatedAt") updated=(formatDateTime updated "AT_TIME")}}</div>
```
nmajor25: This displays a date/time like this:

Screenshot 2023-03-06 at 9 25 35 PM

I don't think we have any utility functions that do `X seconds ago` like the design doc shows.

Save that for a next iteration?

src/js/apps/globals/app-frame_app.js
```diff
@@ -63,7 +63,7 @@ export default App.extend({
       }
     });
   },
-  onBeforeStop() {
+  onStop() {
```
nmajor25, line 66: This doesn't appear to fix the issue where the nav menu shows on form page refreshes.

src/js/apps/forms/form/form_app.js
```diff
@@ -180,6 +189,13 @@ export default App.extend({
       isEditing: !this.getState('responseId'),
     }));
   },
+  showLastUpdated(updated) {
+    if (this.isReadOnly || this.getState('responseId')) return;
```
nmajor25, line 193: Last updated content will be hidden if the form is read only or has already been submitted.

src/js/apps/forms/form/form_app.js
```diff
@@ -110,6 +115,7 @@ export default App.extend({
     'submit': 'onFormServiceSubmit',
     'success': 'onFormServiceSuccess',
     'ready': 'onFormServiceReady',
+    'submissionUpdated': 'onFormServiceSubmissionUpdated',
```
paulfalgout: The event shouldn't be past-tense, and should generally be `context:verb:subject`.  So in this case `update:submission` probably.

src/js/apps/globals/app-frame_app.js
paulfalgout: if we keep this `stop` and add `if (this.isRunning()) return` to the top of `toggleNav` that should do it.

src/js/services/forms.js
paulfalgout: nope this section is ignored in tests because cypress can't do it.  we probably don't actually need to do this though.. it's an edge case, and it will fix itself on the next use interaction with the form.  Let's not add this.

src/js/views/forms/form/form_views.js
paulfalgout: it'd be `{formatDateTime updated "TIME_OR_DAY"}`

src/js/views/forms/form/form.scss
```diff
@@ -121,13 +121,14 @@
...
+  line-height: 1.2;
+  padding: 16px 0 16px 16px;
 }
 
 .form__actions {
-  border-right: 1px solid $black-80;
   display: flex;
-  margin-right: 20px;
+
+  /* margin-right: 20px; */
```
paulfalgout: doesn't seem like this class is needed.  just swap it out with the flex utility class

paulfalgout (changes requested):

src/js/services/forms.js
nmajor25: Gotcha, removed in the last push.

src/js/apps/globals/app-frame_app.js
nmajor25: Like this?

```js
  toggleNav(shouldShow) {
    if (this.isRunning()) return;

    this.getView().toggleNav(!!shouldShow);
  },
```

That makes it so the nav menu is always showing.

src/js/apps/forms/form/form_app.js
nmajor25: Updated in last push.

src/js/views/forms/form/form.scss
nmajor25: Good call, missed that one. Updated in last push.

src/js/views/forms/form/form_views.js
nmajor25: This doesn't totally match the design docs, but probably good for now:

Screenshot 2023-03-07 at 9 44 44 AM
Screenshot 2023-03-07 at 9 45 02 AM

And I also updated the unsaved submission data to `"TIME_OR_DAY"`, as that was updated in the design docs too:

Screenshot 2023-03-07 at 9 50 27 AM

src/js/apps/globals/app-frame_app.js
paulfalgout: sorry `!this.isRunning()`

paulfalgout: the issue is that the toggle happens when the nav stops.  well.. the app frame is now destroying thus stopping all child apps when in restarts.. so we need to make sure it doesn't toggle when the app frame is stopped.

src/js/views/forms/form/form_views.js
paulfalgout: Oh right.. well where we define `TIME_OR_DAY` we should add:
```js
  AGO_OR_DAY(date) {
    if (date.isSame(dayjs(), 'day')) {
      return date.fromNow();
    }

    return formats.DATE(date);
  },
```

paulfalgout (changes requested):

src/js/apps/globals/app-frame_app.js
nmajor25: `!this.isRunning()` seems to have fixed the issue

src/js/views/forms/form/form_views.js
nmajor25: Added the `AGO_OR_DAY` in the last push

paulfalgout (approved):
Shuweet.  We'll test it in QA 🎉 🦜

src/js/apps/forms/form/form_app.js
```diff
@@ -220,9 +236,10 @@ export default App.extend({
         'submit'() {
           this.showForm();
         },
-        'cancel'() {
+        'discard:submission'() {
           Radio.request(`form${ this.form.id }`, 'clear:storedSubmission');
           this.showForm();
+          this.showLastUpdated();
```
nmajor25, line 242: Also just added this in the last push.

It removes the `Last edit was ...` text when the user discards their stored submission.

PR #954, opened 2023-03-07
Use bootstrap workspaces for clinician admin

paulfalgout:
Previous we were using the workspaces of the current user

Shortcut Story ID: [sc-34256]

This'll need tests

### DIFF
--- a/src/js/services/bootstrap.js
+++ b/src/js/services/bootstrap.js
@@ -27,6 +27,7 @@ export default App.extend({
     'currentUser': 'getCurrentUser',
     'currentWorkspace': 'getCurrentWorkspace',
     'setCurrentWorkspace': 'setCurrentWorkspace',
+    'workspaces': 'getWorkspaces',
     'organization': 'getOrganization',
     'directories': 'getDirectories',
     'setting': 'getSetting',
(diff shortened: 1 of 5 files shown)


src/js/services/bootstrap.js
```diff
@@ -72,7 +73,10 @@ export default App.extend({
     return this.organization;
   },
   getDirectories() {
-    return this.directories;
+    return this.directories.clone();
```
paulfalgout, line 76: not really part of the PR, but any collection from bootstrap should really be cloned..

src/js/views/globals/patient-modal/patient-modal_views.js
```diff
@@ -212,8 +213,10 @@ const PatientModal = View.extend({
     });
   },
   showWorkspacesComponent() {
+    const currentUser = Radio.request('bootstrap', 'currentUser');
     const workspacesManager = this.showChildView('workspaces', new WorkspacesManagerComponent({
       member: this.model,
+      workspaces: currentUser.getWorkspaces(),
```
paulfalgout, line 219: patient modal is currently still just the workspaces the user is in..

nmajor25 (approved):

PR #958, opened 2023-03-07
Remove `Your edits are not submitted` message from the form page

nmajor25:
Shortcut Story ID: [sc-34286]

No longer needed after adding `Last edit was X ago` section in #953.

With these updates, only the `Last submitted <date>` status will show if the form has been submitted.

Screenshot 2023-03-07 at 4 39 19 PM

### DIFF
--- a/src/js/apps/forms/form/form_app.js
+++ b/src/js/apps/forms/form/form_app.js
@@ -255,6 +255,8 @@ export default App.extend({
       return;
     }
 
+    this.getRegion('formUpdated').empty();
+
     // If there's a submission this always shows
     this.showFormStatus();
 
--- a/src/js/apps/forms/form/form_app.js
+++ b/src/js/apps/forms/form/form_app.js
@@ -183,10 +183,10 @@ export default App.extend({
     });
   },
   showFormStatus() {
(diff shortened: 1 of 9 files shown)


src/js/apps/forms/form/form_app.js
```diff
@@ -141,75 +137,46 @@ export default App.extend({
...
-  showLastUpdated(updated) {
-    if (this.isReadOnly || this.getState('responseId')) return;
-
-    const lastUpdatedView = new LastUpdatedView({ updated });
-
-    this.showChildView('formUpdated', lastUpdatedView);
   },
-  showActions() {
-    const formActions = new FormActionsView({
+  showStateActions() {
```
paulfalgout, line 166: I renamed this to state actions to differentiate from the other actions as these things modify stuff on the state model

src/js/apps/forms/form/form_app.js
```diff
@@ -282,27 +249,66 @@ export default App.extend({
       this.setState('isActionSidebar', false);
     });
   },
-  showFormAction() {
+  showFormActions() {
```
paulfalgout, line 252: this function now handles all of the various states of the button and text

PR #966, opened 2023-03-09
Add cypress tests for new dashboards nav menu item

nmajor25:
Shortcut Story ID: [sc-34331]

The `Dashboards` link was originally located in the Admin Tools droplist, but me moved it in #937.  Due to time constraints, tests weren't added at the time.

This PR also updates a few unrelated portions of the `/integration/globals/app-nav.js` tests to get them passing.

### DIFF
--- a/test/integration/globals/app-nav.js
+++ b/test/integration/globals/app-nav.js
@@ -16,15 +16,10 @@ context('App Nav', function() {
       .visit();
 
     cy
-      .get('.app-nav__header')
-      .click();
-
-    cy
-      .get('.picklist')
-      .should('contain', 'Sign Out')
-      .contains('Help & Support')
-      .should('have.attr', 'href')
-      .and('contain', 'help.roundingwell.com');
+      .get('.app-nav')
+      .find('.app-nav__bottom')
+      .contains('Dashboards')
+      .should('not.exist');
 
     cy
       .get('.app-nav')
@@ -34,7 +29,6 @@ context('App Nav', function() {
   });
 
   specify('display nav', function() {
-    let logoutStub;
     cy
       .routeActions()
       .routePrograms()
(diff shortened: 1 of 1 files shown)


test/integration/globals/app-nav.js
```diff
@@ -16,15 +16,10 @@ context('App Nav', function() {
...
     cy
-      .get('.app-nav__header')
-      .click();
-
-    cy
-      .get('.picklist')
-      .should('contain', 'Sign Out')
-      .contains('Help & Support')
-      .should('have.attr', 'href')
-      .and('contain', 'help.roundingwell.com');
```
nmajor25, line 27: I don't think this is relevant for this specific test anymore. 

I believe this is testing that the nav droplist has only the `Help & Support` and `Sign Out` links in it for non-manager users. And that it doesn't include the admin tools links (`Dashboards`, `Programs`, `Clinicians`), which no longer reside there.

test/integration/globals/app-nav.js
```diff
@@ -177,9 +175,8 @@ context('App Nav', function() {
       .get('.picklist')
       .contains('Sign Out')
       .click()
-      .then(() => {
-        expect(logoutStub).to.have.been.calledOnce;
-      });
+      .url()
+      .should('contain', 'logout');
```
nmajor25: The `Sign Out` item was recently changed to `<a href="/logout">` instead of `Radio.request('auth', 'logout');` (https://github.com/RoundingWell/care-ops-frontend/commit/c7e1d45562c3af934f3b3b869a48ebf5a5f7977c).

So the `logoutStub` we were using in this test didn't work anymore.

nmajor25:
Failing test in `app-nav.js` that should be passing: https://cloud.cypress.io/projects/ep9zr6/runs/e60ea8ed-8461-482f-bfac-edbf5d3a0ae9/test-results/51c56489-3b79-4f14-aa32-b56451417fb3

This test is hitting a `/404` page and failing because of that. So that's pretty weird 😕 

Edit: related to this: https://github.com/RoundingWell/care-ops-frontend/commit/7bd17ad5b8a07d8741c41f25ee90fba26823dfed?

test/integration/globals/app-nav.js
paulfalgout: is anywhere else checking the contents of the `href`?  Because this is a feature we'll have to test re: if the help setting is `false` or some other string

test/integration/globals/app-nav.js
paulfalgout: I _think_ we should just check the `href` here and not click it.    I don't know if we handled the logout exactly for Cypress which is why we were previously stubbing that radio request. I think it might mess up the coverage reports if i remember corrrectly.

paulfalgout (changes requested):
The issue with the 404'ing test is that the overwrite for `visit` assumes the default workspace `one/` https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/support/commands.js#L69. but that test is replacing the workspaces.  So it either needs the first workspace to be `one` or it has to handle the visit by sending `isRoot` to the `visit`

test/integration/globals/app-nav.js
nmajor25: Okay makes sense, I switched that over to check the `href` value.

test/integration/globals/app-nav.js
nmajor25: No, that was the only place testing the `Help & Support` `href`. But in the last push I added a check for it in the `display nav` test.

I figured we could do the hidden/custom help link tests in a different story.

test/integration/globals/app-nav.js
```diff
@@ -46,18 +40,55 @@ context('App Nav', function() {
...
+      .contains('Help & Support')
+      .should('have.attr', 'href')
+      .and('contain', 'help.roundingwell.com');
+
+    cy
+      .get('.picklist')
+      .find('.app-nav__picklist-bottom')
+      .contains('Sign Out')
+      .should('have.attr', 'href')
+      .and('contain', '/logout');
```
nmajor25, line 82: Moved these up to the top of the test. I think the `Sign Out` code was originally placed at the bottom because it was navigating the test to a new page...?

And I also added some checks to ensure the organization name, workspaces, and help items are displayed.

test/integration/globals/app-nav.js
```diff
@@ -46,18 +40,55 @@ context('App Nav', function() {
...
+
+    // NOTE: this closes the main nav droplist so it doesn't cover other nav links
+    cy
+      .get('.picklist')
+      .find('.picklist__group')
+      .find('.picklist__item')
+      .first()
+      .click();
```
nmajor25, line 90: Needed to close the main nav droplist somehow because it was overlapping other elements in the nav and causing the test to fail.

This just clicks on the currently active workspace picklist item, which does nothing other than close the droplist.

test/integration/globals/app-nav.js
```diff
@@ -431,7 +472,7 @@ context('App Nav', function() {
       })
       .routeActions()
       .routePatient()
-      .visit();
+      .visit('/', { isRoot: true });
```
nmajor25, line 475: This seems to have fixed the `404`ing issue. At least it did on my local machine 🤞

PR #967, opened 2023-03-10
Fix broken shift+click multiselect tests

nmajor25:
Shortcut Story ID: [sc-34350]

We're no longer storing the `lastSelectedIndex` value in local storage (changed in https://github.com/RoundingWell/care-ops-frontend/commit/1bad202df027bff6434657f5dde126417d8784ab). So tests that checked for that value broke as a result.

### DIFF
--- a/test/integration/patients/worklist/schedule.js
+++ b/test/integration/patients/worklist/schedule.js
@@ -1752,12 +1752,7 @@ context('schedule page', function() {
       .first()
       .as('firstActionRow')
       .find('.js-select')
-      .click()
-      .then(() => {
-        const storage = JSON.parse(localStorage.getItem(`schedule_11111_11111-${ STATE_VERSION }`));
-
-        expect(storage.lastSelectedIndex).to.equal(0);
-      });
+      .click();
 
     cy
       .get('@scheduleList')
@@ -1765,13 +1760,9 @@ context('schedule page', function() {
       .eq(2)
       .find('.schedule-list__day-list-row')
       .first()
+      .as('sixthActionRow')
       .find('.js-select')
-      .click({ shiftKey: true })
-      .then(() => {
-        const storage = JSON.parse(localStorage.getItem(`schedule_11111_11111-${ STATE_VERSION }`));
-
-        expect(storage.lastSelectedIndex).to.equal(5);
-      });
+      .click({ shiftKey: true });
 
     cy
       .get('@scheduleList')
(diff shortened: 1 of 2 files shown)


test/integration/patients/worklist/schedule.js
```diff
@@ -1895,11 +1901,20 @@ context('schedule page', function() {
...
+      .eq(2)
+      .find('.js-select')
+      .click({ shiftKey: true });
+
+    cy
+      .get('@scheduleList')
+      .find('.schedule-list__day-list-row.is-selected')
+      .should('have.length', 2);
   });
 });
```
nmajor25, line 1920: In this test I removed all the checks for the `localstorage.lastSelectedIndex` we were doing and updated/added portions of it to ensure the `state.lastSelectedIndex` values were being cleared properly.

The test now does the following:

1. Shift+click multi-select from the top of the list to the bottom.
2. Shift+click multi-select the bottom of the list to the top.
3. Verify `state.lastSelectedIndex` is cleared when a user un-selects a list item.
4. Verify `state.lastSelectedIndex` is cleared when the `Clear` button is clicked.
5. Verify `state.lastSelectedIndex` is cleared when the `Find in list...` input is used to filter the list.
6. Verify `state.lastSelectedIndex` is cleared when a user navigates away from the page and comes back.

test/integration/patients/worklist/worklist.js
```diff
@@ -3488,67 +3472,78 @@ context('worklist page', function() {
...
+      .find('.js-select')
+      .click();
+
+    cy
+      .get('.app-frame__content')
+      .find('.table-list__item.is-selected')
+      .should('have.length', 2);
   });
 
   specify('patient sidebar', function() {
```
nmajor25, line 3549: Same updates as I made for the schedule test.

But there's an additional test to make sure `state.lastSelectedIndex` is cleared when the list is switched between actions and flows (via the `Actions`/`Flows` toggle buttons)

nmajor25, line 3549: Same updates as I made for the schedule test.

But there's an additional test to make sure `state.lastSelectedIndex` is cleared when the list is switched between actions and flows (via the `Actions`/`Flows` toggle buttons).

paulfalgout (approved):
Yeah.. I noticed we cleared it from localstorage when we read in the localstorage, so it looked like we were storing it for no particular reason.

PR #969, opened 2023-03-13
Update program action/flow sidebar tests to not check for the default state droplist element

nmajor25:
Shortcut Story ID: [sc-34374]

These tests are failing because we're checking for the default state droplist element. But those were removed in #949 ([shortcut story](https://app.shortcut.com/roundingwell/story/34154/design-qa-state-on-program-action-and-flow-is-not-correct)).

Due to time constraints, those tests weren't updated at the time.

### DIFF
--- a/test/integration/programs/sidebar/action-sidebar.js
+++ b/test/integration/programs/sidebar/action-sidebar.js
@@ -39,12 +39,6 @@ context('program action sidebar', function() {
 
     cy
       .get('.sidebar')
-      .find('[data-state-region]')
-      .find('.is-disabled')
-      .contains('To Do');
-
-    cy
-      .get('.sidebar')
       .find('[data-owner-region]')
       .contains('Select Team...')
       .should('be.disabled');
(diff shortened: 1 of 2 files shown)


nmajor25:
This test is failing: [programs => sidebar => action sidebar => display action sidebar with no org forms](https://cloud.cypress.io/projects/ep9zr6/runs/4629/test-results/86c95ec5-97a4-4e51-af34-a0358375bc4e). But it's unrelated to the default state droplist removal.

I think it's because of this change: https://github.com/RoundingWell/care-ops-frontend/commit/07496672dd2f5f9d44dec5e54d23f1d68284ab3c#diff-0f1072df55db91cb736db8b101dc92244094f48038f4bf28beec3f4a48e99146. The test needs to be updated, not a bug in the app code (I believe).

paulfalgout (approved):
Yep you're correct.  Since forms are no longer just attached to the fakey `organization` the form dropdown is loading off of the `workspace` `_forms` relationship.  In fact, _only_ that relationship needs be modified for the tests.  We don't have to clear the `routeForms`

PR #971, opened 2023-03-13
Fix broken tests caused by the change to the default worklist toggle

nmajor25:
Shortcut Story ID: [sc-34380]

The default toggle for worklists was recently changed from `Flows` => `Actions` (https://github.com/RoundingWell/care-ops-frontend/pull/935/commits/0aa5481ea0fe9ebf333258508452b22bc27eb784).

Due to time constraints, we didn't update the tests at that time.

### DIFF
--- a/test/integration/patients/patient/patient-flow.js
+++ b/test/integration/patients/patient/patient-flow.js
@@ -53,7 +53,14 @@ context('patient flow page', function() {
 
         return fx;
       })
+      .routeActions()
       .visit('/worklist/owned-by')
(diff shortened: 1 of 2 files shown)


nmajor25:
The remaining failing tests are being caused by either changes to the owner/clinician filter droplist or something with `filters.states` not being set properly for some reason ([here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/patients/worklist/worklist_state.js#L172)).

paulfalgout (approved):

PR #973, opened 2023-03-14
In the worklist tests, load local storage with an `id`

nmajor25:
Shortcut Story ID: [sc-34396]

In the worklist tests, this error was being thrown: [screenshot](https://user-images.githubusercontent.com/35355575/225112662-af8fe16f-3a81-49c1-9f0d-239185b11c98.png).

Related to this update we made in a previous commit: https://github.com/RoundingWell/care-ops-frontend/commit/1bad202df027bff6434657f5dde126417d8784ab#diff-d986239fbc7422a7a336caea629491ff4a66c6f998558bcccca984a46b29ab6a. Which changes how the `onChange()` worklist state function is called on the initial load.

Updating the worklist tests to load `localStorage` with an `id` fixed these errors.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13859,9 +13859,9 @@
       }
     },
     "node_modules/webpack": {
-      "version": "5.75.0",
-      "resolved": "https://registry.npmjs.org/webpack/-/webpack-5.75.0.tgz",
-      "integrity": "sha512-piaIaoVJlqMsPtX/+3KTTO6jfvrSYgauFVdt8cr9LTHKmcq/AMd4mhzsiP7ZF/PGRNPGA8336jldh9l2Kt2ogQ==",
+      "version": "5.76.1",
+      "resolved": "https://registry.npmjs.org/webpack/-/webpack-5.76.1.tgz",
+      "integrity": "sha512-4+YIK4Abzv8172/SGqObnUjaIHjLEuUasz9EwQj/9xmPPkYJy2Mh03Q/lJfSD3YLzbxy5FeTq5Uw0323Oh6SJQ==",
       "dev": true,
       "dependencies": {
         "@types/eslint-scope": "^3.7.3",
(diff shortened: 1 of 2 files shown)


test/integration/patients/worklist/worklist.js
```diff
@@ -2081,16 +2086,21 @@ context('worklist page', function() {
...
       .routeFlow()
       .routeFlowActions()
       .routePatientByFlow()
       .routeDirectories()
       .visit('/worklist/done-last-thirty-days')
-      .wait('@routeFlows')
+      .wait('@routeActions')
       .then(() => {
-        const storage = JSON.parse(localStorage.getItem(`done-last-thirty-days_11111-${ STATE_VERSION }`));
+        const storage = JSON.parse(localStorage.getItem(`done-last-thirty-days_11111_11111-${ STATE_VERSION }`));
```
paulfalgout: ah interesting.. this was a result of the state being triggered twice I think that it didn't need to be prefilled.  Honestly though, we're not testing the defaults here, we're setting them..  so I think this check should be removed as well as the localstorage set on 2089 so that the defaults are tested here.

test/integration/patients/worklist/worklist.js
```diff
@@ -3086,6 +3097,7 @@ context('worklist page', function() {
       },
       selectedActions: {},
       selectedFlows: {},
+      listType: 'flows',
```
paulfalgout, line 3090: :+1:

paulfalgout (changes requested):

package-lock.json
```diff
@@ -13859,9 +13859,9 @@
       }
     },
     "node_modules/webpack": {
-      "version": "5.75.0",
-      "resolved": "https://registry.npmjs.org/webpack/-/webpack-5.75.0.tgz",
-      "integrity": "sha512-piaIaoVJlqMsPtX/+3KTTO6jfvrSYgauFVdt8cr9LTHKmcq/AMd4mhzsiP7ZF/PGRNPGA8336jldh9l2Kt2ogQ==",
+      "version": "5.76.1",
+      "resolved": "https://registry.npmjs.org/webpack/-/webpack-5.76.1.tgz",
+      "integrity": "sha512-4+YIK4Abzv8172/SGqObnUjaIHjLEuUasz9EwQj/9xmPPkYJy2Mh03Q/lJfSD3YLzbxy5FeTq5Uw0323Oh6SJQ==",
```
nmajor25, line 13864: `npm audit fix` only updated the `package-lock.json` file. And not the `package.json`. 

Is that correct?

test/integration/patients/worklist/worklist.js
nmajor25: Good call, I've got that removed in the last push.

package-lock.json
paulfalgout: yep

PR #976, opened 2023-03-15
Fix broken tests caused by changes to the owner droplist

nmajor25:
Shortcut Story ID: [sc-34393]

Owner droplist was updated during the worklist switching feature work: https://github.com/RoundingWell/care-ops-frontend/pull/935.

Due to time constraints, tests weren't updated at the time.

### DIFF
--- a/test/integration/patients/patient/patient-flow.js
+++ b/test/integration/patients/patient/patient-flow.js
@@ -1227,6 +1227,8 @@ context('patient flow page', function() {
       .first()
       .should('contain', 'Clinician McTester')
       .next()
+      .should('contain', 'Workspace One')
+      .next()
       .find('.js-picklist-item')
       .contains('Nurse')
       .click();
--- a/test/integration/patients/worklist/bulk-edit.js
+++ b/test/integration/patients/worklist/bulk-edit.js
@@ -395,6 +395,8 @@ context('Worklist bulk editing', function() {
       .first()
       .should('contain', 'Clinician McTester')
       .next()
+      .should('contain', 'Workspace One')
+      .next()
       .find('.js-picklist-item')
       .contains('Nurse')
       .click();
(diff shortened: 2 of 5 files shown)


test/integration/patients/patient/patient-flow.js
nmajor25, line 1: All these tests should be passing now.

test/integration/patients/worklist/bulk-edit.js
nmajor25, line 1: All these tests should be passing now.

test/integration/patients/worklist/worklist.js
nmajor25, line 1: Both the `worklist => owner filtering` and `worklist => clinician filtering` tests are failing intermittently.

Something weird seems to be happening with the `.routeWorkspaceClinicians()` command that I haven't figured out yet.

It's almost like it's being called twice... And the mutator (edits made to `fx`) is being overridden by the second time it's called... 🤷 

Same thing is happening with the `schedule => filter schedule` test.

paulfalgout: Yeah I suspect we need to relook at any test using `routeWorkspaceClinicians` a lot of that likely covered things like clinicians in various groups or dealt with adding specific data to what was once dynamically generated clinician fixtures.  They're test fixtures now, so we're probably overdoing the test setup when we could just be using the fixture data and/or testing stuff that can't not work (because you can only see one workspace at a time now)

Probably need to look strip out the owner / clinician fixture stuff and run coverage to see what we're not getting.

nmajor25: I think the issue with the intermittently breaking tests were from using `_.sample()` [here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/support/api/clinicians.js#L9). Those tests need clinicians to have specific role and team relationships. And the `_.sample()` was sometimes setting the correct team/role and other times it wasn't.

I agree on the coverage stuff. But wanted to get all tests passing before looking at the coverage.

PR #980, opened 2023-03-17
On clinicians page, disable 'Find in list...' input until clinicians are loaded

nmajor25:
Shortcut Story ID: [sc-34469]

When the clinicians page is initially loaded, the `Find in list...` should be disabled until the clinicians data is done being retrieved and displayed in the list. And then it should be enabled so users can filter the list.

On the first page load, this works properly. But on subsequent visits, the `Find in list...` input is not disabled before the clinicians data is loaded.

### DIFF
--- a/src/js/apps/clinicians/clinicians-all_app.js
+++ b/src/js/apps/clinicians/clinicians-all_app.js
@@ -18,6 +18,9 @@ export default SubRouterApp.extend({
   viewEvents: {
     'click:addClinician': 'onClickAddClinician',
   },
+  onBeforeStop() {
+    this.clinicians = null;
+  },
   onBeforeStart() {
     this.showView(new LayoutView());
     this.getRegion('list').startPreloader();
--- a/test/integration/clinicians/clinicians-all.js
+++ b/test/integration/clinicians/clinicians-all.js
@@ -13,8 +13,8 @@ context('clinicians list', function() {
         fx.data[0].attributes.name = 'Aaron Aaronson';
         fx.data[0].attributes.enabled = true;
         fx.data[0].attributes.last_active_at = testTs();
-        fx.data[0].relationships.role = { data: { id: '33333' } };
-        fx.data[0].relationships.team = { data: { id: '11111' } };
+        fx.data[0].relationships.role.data.id = '33333';
+        fx.data[0].relationships.team.data.id = '11111';
 
         fx.data[1].attributes.name = 'Baron Baronson';
         fx.data[1].attributes.enabled = true;
@@ -209,12 +209,12 @@ context('clinicians list', function() {
         fx.data[0].attributes.name = 'Aaron Aaronson';
         fx.data[0].attributes.enabled = true;
         fx.data[0].relationships.workspaces = { data: [{ type: 'workspaces', id: '11111' }] };
-        fx.data[0].relationships.role = { data: { id: '33333' } };
+        fx.data[0].relationships.role.data.id = '33333';
 
         fx.data[1].attributes.name = 'Baron Baronson';
         fx.data[1].attributes.enabled = true;
         fx.data[1].relationships.workspaces = { data: [{ type: 'workspaces', id: '22222' }] };
-        fx.data[1].relationships.role = { data: { id: '22222' } };
+        fx.data[1].relationships.role.data.id = '22222';
 
         return fx;
       })
--- a/test/integration/clinicians/clinicians-all.js
+++ b/test/integration/clinicians/clinicians-all.js
@@ -202,6 +202,10 @@ context('clinicians list', function() {
 
   specify('find in list', function() {
     cy
+      .routeActions()
+      .visit();
+
+    cy
       .routeClinicians(fx => {
         fx.data = _.sample(fx.data, 2);
 
@@ -218,7 +222,7 @@ context('clinicians list', function() {
 
         return fx;
       })
-      .visit('/clinicians');
+      .navigate('/clinicians');
 
     cy
       .get('.list-page__header')
@@ -293,12 +297,10 @@ context('clinicians list', function() {
       .should('contain', 'Employee');
 
     cy
-      .get('[data-nav-content-region]')
-      .find('[data-worklists-region]')
-      .find('.app-nav__link')
-      .first()
-      .click()
       .go('back');
+
+    cy
+      .navigate('/clinicians');
 
     cy
       .get('.list-page__header')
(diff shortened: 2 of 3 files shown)


test/support/api/clinicians.js
```diff
@@ -54,6 +54,7 @@ Cypress.Commands.add('routeCurrentClinician', (mutator = _.identity) => {
 Cypress.Commands.add('routeClinicians', (mutator = _.identity) => {
   cy.route({
     url: '/api/clinicians',
+    delay: 100,
```
nmajor25: This should fix the flaky `clinicians list > find in list` test we were seeing.

This request was finishing before the test could check that the `Find in list...` input was disabled.

test/integration/clinicians/clinicians-all.js
```diff
@@ -13,8 +13,8 @@ context('clinicians list', function() {
         fx.data[0].attributes.name = 'Aaron Aaronson';
         fx.data[0].attributes.enabled = true;
         fx.data[0].attributes.last_active_at = testTs();
-        fx.data[0].relationships.role.data.id = '33333';
-        fx.data[0].relationships.team.data.id = '11111';
+        fx.data[0].relationships.role = { data: { id: '33333' } };
+        fx.data[0].relationships.team = { data: { id: '11111' } };
```
nmajor25: These were sometimes throwing errors when ran locally. Not entirely sure why.

test/integration/clinicians/clinicians-all.js
```diff
@@ -190,6 +190,11 @@ context('clinicians list', function() {
       .visit('/clinicians')
       .wait('@routeClinicians');
 
+    cy
+      .get('.list-page__header')
+      .find('[data-search-region] .js-input:disabled')
+      .should('have.attr', 'placeholder', 'Find in List...');
+
```
nmajor25, line 197: This just verifies that the `Find in list...` input is disabled when the clinicians data is done being loaded, but empty.

test/integration/clinicians/clinicians-all.js
```diff
@@ -286,5 +291,25 @@ context('clinicians list', function() {
...
+    cy
+      .wait('@routeClinicians');
+
+    cy
+      .get('.list-page__header')
+      .find('[data-search-region] .js-input:not([disabled])');
```
nmajor25, line 315: This addition to the test simply navigates to a new page, comes back to the clinicians page, and verifies that the `Find in list...` input is disabled/enabled properly.

src/js/apps/clinicians/clinicians-all_app.js
```diff
@@ -18,6 +18,9 @@ export default SubRouterApp.extend({
   viewEvents: {
     'click:addClinician': 'onClickAddClinician',
   },
+  onBeforeStop() {
+    this.clinicians = null;
+  },
```
nmajor25, line 23: I believe this issue was caused by `this.clinicians` being cached when a user visits the clinicians page more than once.

As a result, the `Find in list...` input would always have `state.isDisabled = false` when the page is loaded after the first visit. Which means it would never show as disabled.

Another solution I came up with was to update the `showSearchView()` function to this:

```js
showSearchView(setAsDisabled = false) {
  const isDisabled = setAsDisabled || !this.clinicians || !this.clinicians.length;

  const searchComponent = this.showChildView('search', new SearchComponent({
    state: {
      query: this.getState('searchQuery'),
      isDisabled,
    },
  }));

  this.listenTo(searchComponent.getState(), 'change:query', this.setSearchState);
},
```

And then do `this.showSearchView(true)` in the `onBeforeStart()`.

But clearing `this.clinicians` in the `onBeforeStop()` seemed more simple.

nmajor25:
This same thing is happening on the worklist and schedule pages.

`Find in list...` is disabled/enabled correctly on the initial page visit. But shows as always enabled on subsequent visits.

And on the worklist, the disabling/enabling is also not working when `Actions`/`Flows` is toggled.

If this solution makes sense, we can implement something similar for those pages as well.

test/support/api/clinicians.js
paulfalgout: I think the test just needs to be before the `wait` instead of after it.  I don't think we should introduce `delay`s unless it's css animation or completely outside of our control (form.io)

paulfalgout (changes requested):

test/support/api/clinicians.js
nmajor25: That still results in the test failing about 1/2 the time 😞 

It's weird because this test was never flaky before (at least not that I noticed). So it must be something we changed recently with the `routeClinicians()` command.

paulfalgout:
https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting

There must be some order of things or additional call that is causing the
flake. A delay is almost never the solution.

nmajor25:
Looks like we were doing the following to get around this before: https://github.com/RoundingWell/care-ops-frontend/commit/86183ae60a1db5fd7168fe6ab6a7eee90cfb7c8e#diff-8ee3b9163a66a1b57592a8f8737bca706da038eaf50aaae7973050863171c21bL237.

Do a `visit()` => then do `routeClinicians()` => then do `visit('/clinicians')` 🤷‍♂️ 

It looks like that's fixing the flake. But makes the stuff I added at the end of the test difficult (https://github.com/RoundingWell/care-ops-frontend/pull/980#discussion_r1140608327). Where I'm navigating away and then going back to the clinicians page.

nmajor25:
In the app, the FIL input is disabled while the app waits for the api request to come back with the clinicians list. If the request was instantaneous like it is in the tests, it would never be disabled. Without some kind of delay in the request, we can't test that the form is disabled initially before the request is done.

`delay` is used in several places in the intercept documentation: https://docs.cypress.io/api/commands/intercept. I don't fully understand why it'd be considered bad practice to use it. We're only adding `100ms`, which seems insignificant.

But anyways, I can't get the test to consistently work without the `delay: 100`. Not sure what else to try, so I could remove the delay from this PR, create a separate card for dealing with this flaky test and wait until you get back to explore it further?

test/support/api/clinicians.js
```diff
@@ -55,7 +55,8 @@ Cypress.Commands.add('routeClinicians', (mutator = _.identity) => {
   cy.route({
     url: '/api/clinicians',
     response() {
-      const clinicians = getResource(fxClinicians, 'clinicians');
+      // Remove current clinician
```
paulfalgout, line 58: @nmajor25 this was the major issue with this stub flake.  Fixing this let me rollback a few of the other changes.  See what you think.

nmajor25: Nice, looks good 👍 And that test appears to be running smoothly now, without flake.

nmajor25:
@paulfalgout this same thing is also happening with the schedule and worklist pages. Where the find in list is not always disabled properly. I'll create a separate shortcut story for that.

PR #982, opened 2023-03-20
Clear `state.searchQuery` value when the schedule and worklists are restarted

nmajor25:
Shortcut Story ID: [sc-34056]

### DIFF
--- a/src/js/apps/patients/schedule/schedule_app.js
+++ b/src/js/apps/patients/schedule/schedule_app.js
@@ -281,4 +281,11 @@ export default App.extend({
       lastSelectedIndex: null,
     });
   },
+  onStop() {
+    if (this.isRestarting()) {
+      this.setState('searchQuery', '');
+
+      this.showSearchView();
+    }
+  },
 });
--- a/test/integration/patients/worklist/schedule.js
+++ b/test/integration/patients/worklist/schedule.js
@@ -1705,6 +1705,32 @@ context('schedule page', function() {
       .find('.schedule-list__day-list-row')
       .should('contain', 'Test Patient')
       .should('contain', 'Second Action');
+
+    cy
+      .get('@listSearch')
+      .next()
+      .click();
+
+    cy
+      .get('@listSearch')
+      .type('abc');
+
+    cy
+      .get('[data-date-filter-region]')
+      .find('.js-prev')
+      .click();
+
+    cy
+      .get('@listSearch')
+      .should('have.attr', 'placeholder', 'Find in List...');
+
+    cy
+      .wait('@routeActions');
+
+    cy
+      .get('@scheduleList')
+      .find('.schedule-list__day-list-row')
+      .should('have.length', 20);
   });
 
   specify('click+shift multiselect', function() {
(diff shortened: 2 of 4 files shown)


nmajor25:
I believe the bug  on the schedule was being caused by `state.searchQuery` still having a value after a filter is changed and the app is restarted. Clearing the `searchQuery` before the `this.restart()` seems to fix it.

I applied the change to both the schedule and worklist.

Still need to figure out how to test this...

test/integration/patients/worklist/schedule.js
```diff
@@ -1705,6 +1705,32 @@ context('schedule page', function() {
...
+      .should('have.attr', 'placeholder', 'Find in List...');
+
+    cy
+      .wait('@routeActions');
+
+    cy
+      .get('@scheduleList')
+      .find('.schedule-list__day-list-row')
+      .should('have.length', 20);
   });
```
nmajor25, line 1734: Test update:

1. Add `abc` input, which will filter out all results and show `No results match your Find in List search`.
2. Click on the previous month filter button.
3. Verify that the input was cleared in the UI and that it doesn't filter results for the new list of actions.

Added the same test to the worklist.

Sidenote: both these find in list tests are being flaky when ran locally. Doesn't appear to be caused by these changes, since the flakiness stays when I remove the code changes from this PR.

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -87,6 +87,13 @@ export default App.extend({
     this.showSearchView();
     this.toggleBulkSelect();
   },
+  onChangeFilters() {
+    this.getState().clearSearchQuery();
```
paulfalgout: I wonder if this cleanup should be in an `onStop`. could even be wrapped in an `isRestarting` I think...

paulfalgout (changes requested):

src/js/apps/patients/schedule/schedule_app.js
nmajor25: That's a better way to do it. Updated to that in my last push.

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -281,4 +281,11 @@ export default App.extend({
       lastSelectedIndex: null,
     });
   },
+  onStop() {
+    if (this.isRestarting()) {
+      this.getState().clearSearchQuery();
```
paulfalgout: After looking over this again I think `this.setState('searchQuery', '')` here would work just as well with a bit less code.

paulfalgout (changes requested):

src/js/apps/patients/schedule/schedule_app.js
nmajor25: Works for me, just pushed that change.

test/integration/patients/worklist/schedule.js
paulfalgout: I think the flake mentioned here is: https://github.com/RoundingWell/care-ops-frontend/pull/980#discussion_r1147967181

paulfalgout (approved):

PR #985, opened 2023-03-20
Show ellipsis overflow on the form title

nmajor25:
Shortcut Story ID: [sc-34474]

Was added a while ago, but stopped working after a recent, unrelated code change.

This PR also includes some small UI CSS alignment fixes for the form title and form action elements.

### DIFF
--- a/src/js/views/forms/form/form.scss
+++ b/src/js/views/forms/form/form.scss
@@ -32,7 +32,7 @@
 .form__title {
   display: flex;
   font-size: 24px;
-  line-height: 1.2;
+  line-height: 1.4;
   padding: 16px 16px 16px 0;
 }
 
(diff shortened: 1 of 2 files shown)


src/js/views/forms/form/form.scss
```diff
@@ -137,6 +137,7 @@
   .icon {
     color: $black-60;
     font-size: 20px;
+    vertical-align: middle;
```
nmajor25, line 140: Better aligns the action buttons (e.g expand, show sidebar, etc.) with the other elements in that section.

Screenshot 2023-03-20 at 3 53 41 PM

src/js/views/forms/form/form.scss
```diff
@@ -32,7 +32,7 @@
 .form__title {
   display: flex;
   font-size: 24px;
-  line-height: 1.2;
+  line-height: 1.4;
```
nmajor25, line 35: Better aligns the title text with the form controls section.

Screenshot 2023-03-20 at 3 34 19 PM

PR #986, opened 2023-03-21
Add tests for forms with a hidden submit button

nmajor25:
Shortcut Story ID: [sc-34330]

This feature was added in https://github.com/RoundingWell/care-ops-frontend/pull/953. But due to time constraints, tests weren't added at the time.

### DIFF
--- a/test/integration/forms/form.js
+++ b/test/integration/forms/form.js
@@ -1597,6 +1597,72 @@ context('Patient Action Form', function() {
       .find('.alert')
       .contains('Insufficient permissions');
   });
+
+  specify('hidden submit button', function() {
+    cy
+      .routeAction()
+      .routeFormByAction(_.identity, '88888')
+      .routeFormDefinition()
+      .routeFormActionFields()
+      .routePatientByAction()
+      .visit('/patient-action/1/form/88888')
+      .wait('@routeAction')
+      .wait('@routeFormByAction')
+      .wait('@routePatientByAction')
+      .wait('@routeFormDefinition');
+
+    cy
+      .iframe();
+
+    cy
+      .get('.form__controls')
+      .find('button')
+      .contains('Submit')
+      .should('not.exist');
+  });
+
+  specify('hidden submit button - update form', function() {
+    cy
+      .routeAction(fx => {
+        fx.data.id = '1';
+        fx.data.relationships['form-responses'].data = [
+          { id: '1', meta: { created_at: testTs() } },
+        ];
+
+        return fx;
+      })
+      .routeFormByAction(_.identity, '88888')
+      .routeFormDefinition()
+      .routeFormActionFields()
+      .routeFormResponse(fx => {
+        fx.data = {};
+
+        return fx;
+      })
+      .routeActionActivity()
+      .routePatientByAction()
+      .visit('/patient-action/1/form/88888')
+      .wait('@routeFormByAction')
+      .wait('@routeAction')
+      .wait('@routePatientByAction')
+      .wait('@routeFormDefinition')
+      .wait('@routeFormResponse');
+
+    cy
+      .get('.form__controls')
+      .find('button')
+      .contains('Update')
+      .click();
+
+    cy
+      .iframe();
+
+    cy
+      .get('.form__controls')
+      .find('button')
+      .contains('Submit')
+      .should('not.exist');
+  });
 });
 
 context('Patient Form', function() {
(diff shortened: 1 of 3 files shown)


test/integration/forms/form.js
nmajor25, line 1: If a form has `form.model.options.submit_hidden = true`, these two tests verify:

1. The submit button is hidden on form that hasn't been submitted yet.
2. The submit button is hidden after it's updated.

test/integration/forms/form.js
```diff
@@ -1597,6 +1597,79 @@ context('Patient Action Form', function() {
...
+
+    cy
+      .get('.form__controls')
+      .find('button')
+      .contains('Update')
+      .click()
+      .wait('@routeFormActionFields');
+
+    cy
+      .iframe();
```
nmajor25, line 1658: Added `cy.iframe()` throughout because the coverage was getting messed up without them.

test/integration/forms/form.js
```diff
@@ -1597,6 +1597,79 @@ context('Patient Action Form', function() {
...
+      .contains('Submit')
+      .should('not.exist');
+  });
+
+  specify('hidden submit button - update form', function() {
+    cy
+      .routeAction(fx => {
+        fx.data.id = '1';
+        fx.data.relationships.form.data = { id: '88888' };
+        fx.data.relationships['program-action'] = { data: { id: '88888' } };
```
paulfalgout: I don't think this is doing anything?

test/integration/forms/form.js
```diff
@@ -1597,6 +1597,79 @@ context('Patient Action Form', function() {
...
+        fx.data.relationships['form-responses'].data = [
+          { id: '1', meta: { created_at: testTs() } },
+        ];
+
+        return fx;
+      })
+      .routeFormByAction(_.identity, '88888')
+      .routeFormDefinition()
+      .routeFormActionFields(fx => {
```
paulfalgout: do we need to clear this and below?

test/integration/programs/sidebar/action-sidebar.js
```diff
@@ -558,6 +558,8 @@ context('program action sidebar', function() {
       .next()
       .should('contain', 'E Form')
       .next()
+      .should('contain', 'Submit Hidden Form')
```
paulfalgout: Let's add `fx.data = _.first(fx.data, 6)` to the `routeForms` rather than update this with new form names

test/integration/forms/form.js
paulfalgout: How was it getting messed up?  Like badly, or a line or two somewhere?

paulfalgout (changes requested):

test/integration/forms/form.js
nmajor25: A few lines. 

The `if (this.isSubmitHidden)` branches weren't being covered in the `form_app.js` and `form-patient_app.js` files.

For example: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/forms/form/form_app.js#L326.

paulfalgout: Ah yep ok.. so form.io tells the outer app when it is done rendering so the app can show the submit button so this makes sense.

test/integration/forms/form.js
nmajor25: Correct, I just removed it along with the `fx.data.relationships.form.data = { id: '88888' };` line. That didn't appear to do anything either.

test/integration/forms/form.js
nmajor25: Nope, good catch. I just removed that.

test/integration/programs/sidebar/action-sidebar.js
```diff
@@ -324,6 +329,12 @@ context('program action sidebar', function() {
         fx.data[2].attributes.published_at = testTs();
         fx.data[3].attributes.name = 'E Form';
         fx.data[3].attributes.published_at = testTs();
+
+        return fx;
+      })
+      .routeWorkspaces(fx => {
+        fx.data[0].relationships.forms = { data: _.first(getRelationship(fxForms, 'forms'), 6) };
+
```
nmajor25, line 334: Without this, this shows in the UI of the test:

Screenshot 2023-03-27 at 2 25 02 PM

The test will still pass without this, but it seemed incorrect for it to be configured like that.

test/integration/programs/sidebar/action-sidebar.js
nmajor25: I just updated to that but with a caveat that I commented on [below](https://github.com/RoundingWell/care-ops-frontend/pull/986#discussion_r1149707681).

test/integration/programs/sidebar/action-sidebar.js
paulfalgout: could just be `fx.data[0].relationships.forms.data = _.first(fx.data[0].relationships.forms.data), 6);`

alternatively use `fxForms` here and in `routeForms` and do the `_.first` prior to the `getRelationship`/ `getResource`

paulfalgout (changes requested):

test/integration/programs/sidebar/action-sidebar.js
nmajor25: I updated it to the first option: `fx.data[0].relationships.forms.data = _.first(fx.data[0].relationships.forms.data), 6);`. Seems the simplest. And I like not having to import anything.

paulfalgout (approved):

PR #990, opened 2023-03-23
Add tests for workspace switching

nmajor25:
Shortcut Story ID: [sc-34470]

Feature was added here: https://github.com/RoundingWell/care-ops-frontend/pull/917.

Due to time constraints, tests weren't added/updated at the time.

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -366,9 +366,6 @@ export default RouterApp.extend({
 
     navView.triggerMethod('search:active', true);
   },
-  onClickDashboards() {
-    Radio.trigger('event-router', 'dashboards:all');
-  },
   onClickAddPatient() {
     this.showPatientModal();
   },
--- a/test/integration/globals/app-nav.js
+++ b/test/integration/globals/app-nav.js
@@ -210,6 +210,98 @@ context('App Nav', function() {
       .should('not.have.class', 'is-selected');
   });
 
+  specify('switch workspaces', function() {
+    cy
+      .routeActions()
+      .visit()
+      .wait('@routeActions')
+      .its('request.headers')
+      .should('have.property', 'workspace', '11111')
+      .then(() => {
+        const storage = JSON.parse(localStorage.getItem('currentWorkspace'));
+
+        expect(storage).to.equal('11111');
+      });
+
+    cy
+      .url()
+      .should('contain', '/one/worklist/owned-by');
+
+    cy
+      .get('.app-nav__header')
+      .as('mainNav')
+      .click();
+
+    cy
+      .get('.picklist')
+      .find('.picklist__group')
+      .find('.picklist__item')
+      .first()
+      .should('have.class', 'is-selected')
+      .next()
+      .should('not.have.class', 'is-selected')
+      .click();
+
+    cy
+      .wait('@routeActions')
+      .its('request.headers')
+      .should('have.property', 'workspace', '22222')
+      .then(() => {
+        const storage = JSON.parse(localStorage.getItem('currentWorkspace'));
+
+        expect(storage).to.equal('22222');
+      });
+
+    cy
+      .url()
+      .should('contain', '/two/worklist/owned-by');
+
+    cy
+      .get('.app-nav__header')
+      .as('mainNav')
+      .click();
+
+    cy
+      .get('.picklist')
+      .find('.picklist__group')
+      .find('.picklist__item')
+      .first()
+      .should('not.have.class', 'is-selected')
+      .next()
+      .should('have.class', 'is-selected');
+
+    cy
+      .go('back');
+
+    cy
+      .wait('@routeActions')
+      .its('request.headers')
+      .should('have.property', 'workspace', '11111')
+      .then(() => {
+        const storage = JSON.parse(localStorage.getItem('currentWorkspace'));
+
+        expect(storage).to.equal('11111');
+      });
+
+    cy
+      .url()
+      .should('contain', '/one/worklist/owned-by');
+
+    cy
+      .get('.app-nav__header')
+      .as('mainNav')
+      .click();
+
+    cy
+      .get('.picklist')
+      .find('.picklist__group')
+      .find('.picklist__item')
+      .first()
+      .should('have.class', 'is-selected')
+      .next()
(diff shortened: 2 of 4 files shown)


nmajor25:
Wanted to get this pushed so I could see the coverage run.

Still need to add:

- [x] If a user doesn't belong to any workspaces.
- [x] Handling of legacy routes ([here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/globals/error_app.js#L52-L56)).
- [x] Add patient modal where clinician is in only one workspace ([cover these lines of code](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/globals/nav_app.js#L382-L386)).

src/js/apps/globals/nav_app.js
```diff
@@ -366,9 +366,6 @@ export default RouterApp.extend({
 
     navView.triggerMethod('search:active', true);
   },
-  onClickDashboards() {
-    Radio.trigger('event-router', 'dashboards:all');
-  },
```
nmajor25, line 371: I don't believe this is used anymore. Noticed it in the coverage report.

test/integration/globals/app-nav.js
```diff
@@ -210,6 +210,98 @@ context('App Nav', function() {
...
+    cy
+      .get('.picklist')
+      .find('.picklist__group')
+      .find('.picklist__item')
+      .first()
+      .should('have.class', 'is-selected')
+      .next()
+      .should('not.have.class', 'is-selected');
+  });
+
```
nmajor25, line 304: Checks that the page URL, workspace request header, workspace local storage value, and CSS active states of the droplist items are set properly when switching between workspaces via the app nav menu.

test/integration/globals/app-nav.js
```diff
@@ -866,6 +958,60 @@ context('App Nav', function() {
...
+      .get('.modal')
+      .find('[data-workspaces-region] .list-manager__item')
+      .should('contain', 'Workspace One');
+
+    cy
+      .get('.modal')
+      .find('[data-workspaces-region] [data-droplist-region] button')
+      .should('be.disabled');
+  });
+
```
nmajor25, line 1014: Adds coverage for this: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/globals/nav_app.js#L382-L386.

test/integration/globals/error.js
```diff
@@ -18,6 +18,40 @@ context('Global Error Page', function() {
...
+
+    cy
+      .url()
+      .should('contain', 'one/worklist/owned-by');
+
+    cy
+      .get('.error-page')
+      .should('not.exist');
+  });
+
```
nmajor25, line 54: I created these to cover this: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/globals/error_app.js#L50-L60.

test/integration/globals/error.js
```diff
@@ -18,6 +18,40 @@ context('Global Error Page', function() {
       .should('not.exist');
   });
 
+  // NOTE: 'legacy' means the workspace name isn't included in the URL
+  specify('404 not found - legacy routes', function() {
+    cy
+      .visit('/404', { isRoot: true });
```
paulfalgout: really I think this and the above test should be combined and it should by something like `visit('/route-does-not-exist')` instead of `/404`.

test/integration/globals/default-route.js
```diff
@@ -23,7 +23,21 @@ context('patient page', function() {
 
     cy
       .url()
-      .should('contain', 'worklist/owned-by');
+      .should('contain', 'one/worklist/owned-by');
+  });
+
+  specify('current clinician has no workspaces', function() {
```
paulfalgout, line 29: This behavior is happening, but I _think_ it's throwing an error doing so.  We might actually want to attempt to force this to be true rather than depend on the error.  (and also clean up datadog)

test/integration/globals/app-nav.js
```diff
@@ -866,6 +958,60 @@ context('App Nav', function() {
     cy.clock().invoke('restore');
   });
 
+  specify('add patient - clinician in one workspace', function() {
```
paulfalgout, line 961: I'm surprised this wasn't tested before..  I would have thought this worked the same when it was groups..

paulfalgout (changes requested):

test/integration/globals/error.js
nmajor25: I attempted that, but couldn't get coverage for these lines: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/globals/error_app.js#L50-L60.

That code wouldn't run if the workspace name is included in the url on the initial load (i.e. `/one/` vs. `/`).

But maybe I missed something, so I'll look at these tests again.

test/integration/globals/default-route.js
nmajor25: So remove this test from this PR and create a separate shortcut story to update the app code?

test/integration/globals/app-nav.js
nmajor25: Me too. It was getting covered before somehow... but I couldn't find where or how.

test/integration/globals/error.js
paulfalgout: oh right.. yeah so it would be need to be `visit` with `{ isRoot: true }`.. are we testing legacy routes somewhere else and this is testing for a non-legacy route (the else)?  regardless we should not use `/404` for this as it's confusing.  We're answering "can I go to the route or is it a 404?"  So "can I go to 404 or is it a 404?" is 😬

test/integration/globals/default-route.js
paulfalgout: this is ok to me I think for now

nmajor25: Okay I'll leave this test as-is for now.

test/integration/globals/error.js
nmajor25: I just updated those 404 tests to load an initial route of `/route-does-not-exist` instead of `/404`.

My understanding is the `'404 not found'` test in this file checks the non-legacy route. And the `'404 not found - legacy routes'` test I created in this PR checks the legacy route, where `{isRoot: true}` is added.

PR #994, opened 2023-03-28
Fix flakey `Find in list...` worklist test

nmajor25:
Shortcut Story ID: [sc-34501]

### DIFF
--- a/test/integration/patients/worklist/schedule.js
+++ b/test/integration/patients/worklist/schedule.js
@@ -1456,7 +1456,7 @@ context('schedule page', function() {
       .should('be.disabled');
   });
 
-  specify('filter in list', function() {
+  specify('find in list', function() {
     cy
       .routeActions(fx => {
         fx.data[0].attributes = {
--- a/test/integration/patients/worklist/worklist.js
+++ b/test/integration/patients/worklist/worklist.js
@@ -3127,7 +3127,7 @@ context('worklist page', function() {
         fx.data[0].attributes.updated_at = `${ currentYear }-01-05`;
         fx.data[0].relationships.patient.data.id = '1';
         fx.data[0].relationships.owner.data = {
-          id: '55555',
+          id: '11111',
           type: 'clinicians',
         };
 
@@ -3159,14 +3159,6 @@ context('worklist page', function() {
           attributes: {
             first_name: 'Test',
             last_name: 'Patient',
-          },
-        },
-        {
-          id: '55555',
-          type: 'clinicians',
-          attributes: {
-            role: 'employee',
-            name: 'Test Clinician',
           },
         });
         return fx;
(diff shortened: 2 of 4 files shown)


paulfalgout (approved):

test/integration/patients/worklist/schedule.js
```diff
@@ -1456,7 +1456,7 @@ context('schedule page', function() {
       .should('be.disabled');
   });
 
-  specify('filter in list', function() {
+  specify('find in list', function() {
```
nmajor25, line 1459: Changed this to `find in list` to match what the worklist one is called.

test/integration/patients/worklist/worklist.js
```diff
@@ -3127,7 +3127,7 @@ context('worklist page', function() {
         fx.data[0].attributes.updated_at = `${ currentYear }-01-05`;
         fx.data[0].relationships.patient.data.id = '1';
         fx.data[0].relationships.owner.data = {
-          id: '55555',
+          id: '11111',
```
nmajor25, line 3130: Use the existing `Clinician McTester` clinician fixture instead of `Test Clinician` that we create for only this test.

test/integration/patients/worklist/worklist.js
```diff
@@ -3160,14 +3160,6 @@ context('worklist page', function() {
...
             last_name: 'Patient',
           },
-        },
-        {
-          id: '55555',
-          type: 'clinicians',
-          attributes: {
-            role: 'employee',
-            name: 'Test Clinician',
-          },
```
nmajor25, line 3170: After the move to use `/test/clinicians.json` in #952, the `Test Clinician` clinician was sometimes replacing the `55555` fixture clinician that already exists [here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/fixtures/test/clinicians.json). Which would cause the find in list test to sometimes fail because additional flows were showing in the results that it wasn't expecting.

test/support/api/actions.js
```diff
@@ -45,7 +45,7 @@ function generateData(patients = _.sample(this.fxPatients, 1)) {
 
   return {
     data,
-    included: [],
```
nmajor25, line 48: I don't think this was breaking any tests, but was causing the patient name to be empty on actions:

Screenshot 2023-03-28 at 1 27 30 PM

I made the same change in the `test/support/api/flows.js` file, which was causing the same issue but for flows.

test/support/api/flows.js
```diff
@@ -45,7 +45,7 @@ function generateData(patients = _.sample(this.fxPatients, 1)) {
     };
 
     if (_.random(1)) {
-      const clinician = _.sample(this.fxClinicians);
+      const clinician = _.sample(_.rest(this.fxClinicians));
```
nmajor25, line 48: Exclude the `Clinician McTester` (`id: '11111'`) from being auto-assigned as the owner to a flow. This was causing some flakiness with the find in list test specifically.

@paulfalgout should this being added to the `routeActions` command too ([here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/support/api/actions.js#L34))?

PR #995, opened 2023-03-29
Fix custom sort issues

paulfalgout:
Shortcut Story ID: [sc-34744]

Indirectly related, if a sort option is in localstorage but no longer sort option we'd get this same error.

This also contains a fix for swallowed `onStart` errors.

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -42,6 +42,8 @@ export default App.extend({
     'change:selectedActions': 'onChangeSelected',
   },
   onChangeStateSort() {
+    if (!this.isRunning()) return;
+
     this.getChildView('list').setComparator(this.getComparator());
   },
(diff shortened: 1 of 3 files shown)


paulfalgout:
I have a fix for the failing test on a separate branch.. it isn't related to this.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -42,6 +42,8 @@ export default App.extend({
     'change:selectedActions': 'onChangeSelected',
   },
   onChangeStateSort() {
+    if (!this.isRunning()) return;
```
nmajor25, line 45: Nice, this fixes it. Not seeing any console errors in my local sandbox.

And even if this function is returned without setting the comparator, the sort will still be applied when the the list view is done loading and rendered. Since the sort option was set in the state.

nmajor25 (approved):

PR #996, opened 2023-03-29
Fix some test coverage misses

nmajor25:
Shortcut Story ID: [sc-34796]

### DIFF
--- a/test/integration/patients/worklist/worklist.js
+++ b/test/integration/patients/worklist/worklist.js
@@ -102,6 +102,10 @@ context('worklist page', function() {
 
         return fx;
       })
+      .routePatient(fx => {
+        fx.data.id = '1';
+        return fx;
+      })
       .routeActions()
       .routeFlow()
       .routeFlowActions()
@@ -266,6 +270,20 @@ context('worklist page', function() {
 
     cy
       .go('back');
+
+    cy
+      .get('@firstRow')
+      .contains('Test Patient')
+      .click();
+
+    cy
+      .url()
+      .should('contain', 'patient/dashboard/1')
+      .wait('@routePatient');
+
+    cy
+      .go('back')
+      .wait('@routeFlows');
 
     cy
       .get('.worklist-list__toggle')
@@ -1734,7 +1752,22 @@ context('worklist page', function() {
               ],
             },
           },
+          {
+            attributes: {
+              name: 'ACO',
+              slug: 'aco',
+              value: [
+                'Basic',
+                'Premier',
+              ],
+            },
+          },
         ];
+
+        return fx;
+      })
+      .routeSettings(fx => {
+        fx.data.push({ id: 'custom_filters', attributes: { value: ['team', 'insurance'] } });
 
         return fx;
       })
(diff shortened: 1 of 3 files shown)


src/js/views/forms/form/form_views.js
```diff
@@ -34,9 +34,7 @@ const ContextTrailView = View.extend({
     'click .js-dashboard': 'click:dashboard',
   },
   onClickBack() {
-    Radio.request('history', 'go:back', () => {
-      this.routeToPatient();
-    });
```
nmajor25: Pertaining to this coverage miss: https://coveralls.io/builds/58331475/source?filename=src%2Fjs%2Fviews%2Fforms%2Fform%2Fform_views.js#L38.

Does `this.routeToPatient()` do anything here? Testing in my local sandbox, this behaves the same without it.

If that can be removed here, we should be able to change `onClickDashboard()` below to this:

```js
onClickDashboard() {
  Radio.trigger('event-router', 'patient:dashboard', this.patient.id);
},
```

And get rid of the `routeToPatient()` function.

test/integration/patients/worklist/worklist.js
nmajor25, line 1: To add coverage for this: https://coveralls.io/builds/58331475/source?filename=src%2Fjs%2Fapps%2Fpatients%2Fsidebar%2Ffilters-sidebar_app.js#L40. 

That code/feature was added here: https://github.com/RoundingWell/care-ops-frontend/pull/935/commits/b82e7ab24919f0afd2bd5b89268bb86426b9b886.

I also added the same changes to the schedule version of this test.

test/integration/patients/worklist/worklist.js
```diff
@@ -267,6 +271,20 @@ context('worklist page', function() {
...
+
+    cy
+      .url()
+      .should('contain', 'patient/dashboard/1')
+      .wait('@routePatient');
+
+    cy
+      .go('back')
+      .wait('@routeFlows');
+
```
nmajor25, line 287: To add coverage for this: https://coveralls.io/builds/58331475/source?filename=src%2Fjs%2Fviews%2Fpatients%2Fworklist%2Fflow_views.js#L76.

Don't really understand how this line was missing coverage. But I'm guessing it was from when the default actions vs. flows toggle was changed for the worklist tests.

src/js/views/forms/form/form_views.js
paulfalgout: This was added so if you hard refreshed on a form and did not have any previous history, the back button would go to the patient.  It's needed.. I dunno how we'd actually test it though.

paulfalgout (changes requested):

src/js/views/forms/form/form_views.js
nmajor25: Ah okay, I just added that back.

`istanbul ignore`?

paulfalgout: actually now that I think about it, it's testable.. just `visit` a patient form and push go back.. it should go to that patient's dashboard even though that was never a route.

nmajor25: Just added that in my last push.

paulfalgout (approved):

PR #997, opened 2023-03-30
Fix test flake and :100: coverage

paulfalgout:
Shortcut Story ID: [sc-33548]

This touches a number of things and removes some dead code so I'll need to add some tests.. kind of hijacking that sc card a bit this which is a bit broader.


test/integration/forms/action.js
```diff
@@ -672,6 +674,7 @@ context('Patient Action Form', function() {
 
   specify('submitting the form', function() {
     cy
+      .routesForPatientAction()
       .routeAction(fx => {
```
nmajor25, line 678: It shouldn't cause any issues if `.routeAction()` is added twice here, right? Since `.routesForPatientAction()` includes `.routeAction()` as well.

Running the test locally, it doesn't appear to cause any issues.

test/integration/forms/patient.js
```diff
@@ -0,0 +1,806 @@
...
+      .wait('@routeFormFields')
+      .wait('@routeFormDefinition');
+
+    cy
+      .iframe()
+      .find('textarea[name="data[storyTime]"]')
+      .should('contain', 'foo');
+  });
+
+  if (Cypress.browser.name !== 'firefox') {
```
nmajor25, line 350: Do we still need this? Can't remember why we added it in the first place...

nmajor25 (approved):

test/integration/forms/action.js
paulfalgout: yeah the 2nd will overwrite it.. for this reason I contemplated if `routesForPatientAction` shouldn't include `routeAction` at all or if I should add all of the rest of the routes from that command here.. but the simplest was just to duplicate it..  I _think_ that's reasonable?  thoughts?

test/integration/forms/patient.js
paulfalgout: I don't think you can stub `console.foo` in firefox

test/integration/forms/action.js
nmajor25: That seems reasonable to me. Similar scenarios will probably happen when using these commands in the future. So just wanted to make sure that doesn't cause any issues..

PR #999, opened 2023-03-30
Disable the `Find in list...` input when the worklist/schedule list is loading

nmajor25:
Shortcut Story ID: [sc-34746]

This was only happening on the initial page load. It should now be properly disabled on subsequent page loads or when filters are applied to the list (which causes just the list of actions/flows to reload).

Applies to the worklists, schedule, and reduced schedule.

### DIFF
--- a/test/integration/patients/worklist/worklist.js
+++ b/test/integration/patients/worklist/worklist.js
@@ -3221,6 +3221,7 @@ context('worklist page', function() {
         return fx;
       })
       .routePatientByAction()
+      .routeActions()
       .visit('/worklist/owned-by');
 
     cy
@@ -3441,6 +3442,24 @@ context('worklist page', function() {
       .get('@flowList')
       .find('.work-list__item')
       .should('have.length', 10);
+
+    cy
+      .get('[data-nav-content-region]')
+      .find('[data-worklists-region]')
+      .find('.app-nav__link')
+      .contains('Shared By')
+      .click()
+      .wait('@routeActions');
+
+    cy
+      .navigate('/worklist/owned-by');
+
+    cy
+      .intercept('GET', '/api/flows?*', { delay: 100, body: { data: [] } });
+
+    cy
+      .get('.list-page__header')
+      .find('[data-search-region] .js-input:disabled');
   });
 
   specify('click+shift multiselect', function() {
--- a/src/js/apps/patients/schedule/reduced_schedule_app.js
+++ b/src/js/apps/patients/schedule/reduced_schedule_app.js
@@ -7,6 +7,9 @@ import SearchComponent from 'js/views/shared/components/list-search';
 import { LayoutView, ScheduleTitleView, TableHeaderView, ScheduleListView } from 'js/views/patients/schedule/schedule_views';
 
 export default App.extend({
+  onBeforeStop() {
+    this.collection = null;
+  },
   onBeforeStart() {
     this.setState({ isReduced: true });
 
--- a/src/js/apps/patients/schedule/schedule_app.js
+++ b/src/js/apps/patients/schedule/schedule_app.js
@@ -46,6 +46,9 @@ export default App.extend({
 
     this.getState().setDefaultFilterStates();
   },
+  onBeforeStop() {
+    this.collection = null;
+  },
   onBeforeStart() {
     if (this.isRestarting()) {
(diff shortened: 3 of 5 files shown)


test/integration/patients/worklist/worklist.js
```diff
@@ -3461,6 +3462,28 @@ context('worklist page', function() {
...
+      .get('.list-page__header')
+      .find('[data-search-region] .js-input:disabled');
+
+    cy
+      .wait('@routeFlows');
+
+    cy
+      .get('.list-page__header')
+      .find('[data-search-region] .js-input:not([disabled])');
   });
```
nmajor25, line 3463: I see these having the potential to be flaky...

paulfalgout: yeah hmm..
since these are right at the end and you're not testing the content you could add a
```js
cy
  .intercept('GET', '/api/actions?*', { delay: 100, body: { data: [] } })
  .as('@routeDelayActions');
```
or something in that ballpark to ensure the load doesn't occur before the check, but something that doesn't add a delay to every instance of the route

Though I bet the better way to do it is using the intercept request thing.. like:
```js
cy
  .intercept('GET', '/api/actions?*', req => {
    cy.get('.list-page__header).....
    req.reply({ data: [] });
  });
```
But I'm not sure.. just had that thought.. makes me want to go back and look at the couple places we have delays.

paulfalgout (changes requested):
had a thought

test/integration/patients/worklist/worklist.js
nmajor25: That looks interesting... I'll give that a try.

nmajor25: This doesn't appear to work:

```js
cy
  .intercept('GET', '/api/actions?*', req => {
    cy.get('.list-page__header).....
    req.reply({ data: [] });
  });
```

I don't think you can make dom/app state queries inside the intercept. It only seems to have access to the `req` object. I scoured the intercept documentation without luck.

nmajor25: Doing this works:

```js
cy
  .intercept('GET', '/api/actions?*', { delay: 100, body: { data: [] } })
  .as('@routeDelayActions');
```

But in the scenario needed for this test, the data can't be an empty array (`data: []`). Because the `Find in list...` input will remain disabled if there are not flows/actions in the list. And in the test we need it to become enabled when the list is done loading.

nmajor25: This works:

```js
cy
.intercept('GET', 'api/flows?*', {
  delay: 100,
  body: {
    data: [
      {
        id: '1',
        type: 'flows',
        attributes: {
          name: 'Test Flow',
        },
        relationships: {
          owner: {
            data: {
              id: '11111',
              type: 'clinicians',
            },
          },
        },
      },
    ],
  },
})
.as('routeDelayFlows');
```

But I'm looking for a better/cleaner way to set that flow data. 

Without the owner relationship, the worklist breaks, which makes the test fail.

Guess I could just grab a flow from the fixtures and then add an owner relationship to it.

paulfalgout: hmm yeah bummer.  Does this test need to resolve the request for coverage?  We'd already be testing the disable/enable of that field in the earlier tests.  Perhaps we can just not resolve the `intercept`?

nmajor25: For coverage, it shouldn't need to. We could just verify the input is disabled on the initial load and then not check that it becomes enabled again.

paulfalgout (changes requested):
one question

paulfalgout:
also needs a rebase

paulfalgout (approved):

test/integration/patients/worklist/worklist.js
paulfalgout: A cypress guy got back to me on this saying "ok, so I don't like doing this, but you could return a promise from the handler and then resolve it from your Cypress chain outside. But you also need to think what happens when multiple intercepts happen and it checks more commands. If I add an example of this, it will be part of my paid course "  He added an example behind his paywall https://cypress.tips/courses/network-testing/lessons/bonus97. Might look into getting Will to get this for us

PR #1000, opened 2023-03-31
Update dependencies

paulfalgout:
These are all of the breaking changes dependency updates. We were previously unable to update `babel-loader` in an older version of Cypress. Similarly one of the css libraries prevented us from updating others, but it seems like everything works now.

If tests pass here I'll merge it.


paulfalgout:
@nmajor25 I don't think this PR will be controversial, but lemme know

test/integration/forms/patient.js
```diff
@@ -326,7 +326,7 @@ context('Patient Form', function() {
       .should('have.value', 'bar');
   });
 
-  specify('form scripts and reducers', function() {
+  specify('form scripts and reducers', { retries: 4 }, function() {
```
nmajor25, line 329: Nice, didn't even know you could do this...

paulfalgout:
1000th PR

PR #1001, opened 2023-04-02
Fix firefox and other flake

paulfalgout:
The relative dates flake is for everything.

### DIFF
--- a/test/integration/patients/patient/archive.js
+++ b/test/integration/patients/patient/archive.js
@@ -2,11 +2,11 @@ import _ from 'underscore';
 import dayjs from 'dayjs';
 
 import { testTs, testTsSubtract } from 'helpers/test-timestamp';
-import { testDateSubtract } from 'helpers/test-date';
+import { testDate, testDateSubtract } from 'helpers/test-date';
 
 context('patient archive page', function() {
   specify('action, flow and events list', function() {
-    const testTime = dayjs().hour(10).utc().valueOf();
+    const testTime = dayjs(testDate()).hour(12).valueOf();
 
     cy.clock(testTime, ['Date']);
 
--- a/test/integration/patients/worklist/worklist.js
+++ b/test/integration/patients/worklist/worklist.js
@@ -3,7 +3,7 @@ import dayjs from 'dayjs';
 import { NIL as NIL_UUID } from 'uuid';
 
 import formatDate from 'helpers/format-date';
-import { testTs, testTsAdd, testTsSubtract } from 'helpers/test-timestamp';
+import { testTs, testTsSubtract } from 'helpers/test-timestamp';
 import { testDate, testDateAdd, testDateSubtract } from 'helpers/test-date';
 import { getResource } from 'helpers/json-api';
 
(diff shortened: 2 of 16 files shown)


test/support/api/flows.js
```diff
@@ -77,16 +80,14 @@ Cypress.Commands.add('routeFlow', (mutator = _.identity) => {
     .as('routeFlow');
 });
 
-Cypress.Commands.add('routePatientFlows', (mutator = _.identity, patientId = '1') => {
```
paulfalgout, line 80: This default to id `1` was very problematic.  Created many patient models that were included with the same id.  It made some tests flaky when this or the action version resolved last.

test/support/api/flows.js
```diff
@@ -14,24 +14,27 @@ function flowFixtures() {
...
 
-function generateData(patients = _.sample(this.fxPatients, 1)) {
+function generateData(patients = _.sample(this.fxPatients, 5)) {
   const data = getResource(_.sample(this.fxFlows, 10), 'flows');
+  const programFlows = _.sample(this.fxProgramFlows, 5);
+  const programActionsSample = _.sample(this.fxProgramActions, 20);
+  const programs = _.sample(this.fxPrograms, 1);
   let included = [];
 
+  included = getIncluded(included, patients, 'patients');
```
paulfalgout, line 24: Moving the included to outside of the each reduces the includes from ~180 to 31.  This is at least a perf improvement

test/integration/patients/patient/archive.js
```diff
@@ -2,11 +2,11 @@ import _ from 'underscore';
 import dayjs from 'dayjs';
 
 import { testTs, testTsSubtract } from 'helpers/test-timestamp';
-import { testDateSubtract } from 'helpers/test-date';
+import { testDate, testDateSubtract } from 'helpers/test-date';
 
 context('patient archive page', function() {
   specify('action, flow and events list', function() {
-    const testTime = dayjs().hour(10).utc().valueOf();
+    const testTime = dayjs(testDate()).hour(12).valueOf();
```
paulfalgout, line 9: not utilizing `testDate` for these situations made the tests potentially slightly different at odd times.

test/integration/patients/worklist/worklist.js
```diff
@@ -1273,8 +1265,8 @@ context('worklist page', function() {
 
     cy
       .get('.tooltip')
-      .should('contain', formatDate(dayjs(filterDate).startOf('week'), 'MM/DD/YYYY'))
-      .should('contain', formatDate(dayjs(filterDate).endOf('week'), 'MM/DD/YYYY'));
+      .should('contain', formatDate(dayjs(testDate()).startOf('week'), 'MM/DD/YYYY'))
```
paulfalgout, line 1268: this was a big issue.  `filterDate` is not relative to the current time, it's the selected time.  But the display in the tooltip _is_ relative to the current time

test/integration/patients/worklist/worklist.js
```diff
@@ -3806,7 +3783,9 @@ context('worklist page', function() {
       .as('patientSidebar')
       .find('.worklist-patient-sidebar__patient-name')
       .should('contain', 'Test Patient')
-      .click();
+      .click()
+      .wait('@routePatientField')
+      .wait('@routePrograms');
```
paulfalgout, line 3788: there are a few instances where we need to wait on routes loaded by a `routesFor*` command.  I don't love this, but not doing so makes these tests flaky.

paulfalgout:
There certainly could be other flake.  If we see a flaky test and it has a 504 or a 401 in the screenshot or locally, chances are this is a missing endpoint stub.

paulfalgout:
The console error test is by far the most flaky test.  I added `retries: 4` recently to it.  It has also been improved.  I suspect this is very unlikely to cause failure at this point.

paulfalgout:
[sc-34953]

nmajor25 (approved):

PR #1003, opened 2023-04-03
Feature/update field form api

paulfalgout:
Shortcut Story ID: [sc-34956]

From a form:
```js
const updatedFieldValue = await updateField(fieldName, { newValue: true });
```

- [x] Todo: Add a test for a PATCH error response

### DIFF
--- a/src/js/entities-service/entities/patient-fields.js
+++ b/src/js/entities-service/entities/patient-fields.js
@@ -1,4 +1,4 @@
-import { isObject, isEmpty } from 'underscore';
+import { isObject, isEmpty, extend } from 'underscore';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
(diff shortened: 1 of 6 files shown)


test/integration/forms/form.js
```diff
@@ -0,0 +1,345 @@
...
+        return {
+          display: 'form',
+          components: [
+            {
+              label: 'Select Foo',
+              widget: 'choicesjs',
+              tableView: true,
+              dataSrc: 'custom',
+              data: {
+                custom: 'values = updateField(\'foo\', [\'one\', \'two\'])',
```
paulfalgout, line 153: this would not be a likely place within a form to use this API, but it does work to hit the patch and test that values are returned.

paulfalgout:
I need to investigate the coverage drop, but I think it's probably something I missed consolidating the form specific tests from patient/action

zfixler (approved):

nmajor25 (approved):

src/js/entities-service/entities/patient-fields.js
```diff
@@ -7,10 +7,27 @@ const TYPE = 'patient-fields';
...
+    // NOTE: This will treat the PATCH like a PUT
+    // We won't always have an ID, but never need to POST
+    return false;
+  },
   getValue() {
     const value = this.get('value');
     return isObject(value) && isEmpty(value) ? null : value;
   },
+  saveAll(attrs) {
+    attrs = extend({}, this.attributes, attrs);
```
zfixler, line 23: @paulfalgout , @nmajor25 , I think we might need to add a `uniqueId` here in `attrs`.  See my comments here: https://app.shortcut.com/roundingwell/story/34920/hra-appointment-mark-as-seen

PR #1004, opened 2023-04-04
Make form read-only on "done" actions 

paulfalgout:
Shortcut Story ID: [sc-34969]

Not doing much here.  This is a policy and not a permission, so it just prompts the form how to render based on the status of the action at render time, but it is not enforced beyond that.

### DIFF
--- a/test/integration/forms/action.js
+++ b/test/integration/forms/action.js
@@ -946,6 +946,7 @@ context('Patient Action Form', function() {
       .iframe()
       .find('[name="data[patient.fields.foo]"]')
       .should('have.value', 'bar')
+      .should('be.disabled');
   });
 
   specify('read only form', function() {
@@ -990,7 +991,8 @@ context('Patient Action Form', function() {
     cy
       .iframe()
       .find('[name="data[patient.fields.foo]"]')
-      .should('have.value', 'bar');
+      .should('have.value', 'bar')
+      .should('be.disabled');
   });
 
   specify('routing to form-response', function() {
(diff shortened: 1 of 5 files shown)


test/support/api/actions.js
```diff
@@ -57,6 +57,7 @@ Cypress.Commands.add('routeAction', (mutator = _.identity) => {
     response() {
       const apiData = generateData.call(this);
       apiData.data = _.sample(apiData.data);
+      apiData.data.relationships.state.data.id = '33333';
```
paulfalgout, line 60: This is by far the bigger change on this PR. I don't _think_ it is a problem, though we'll se a lot less to-do actions randomly around.

test/integration/forms/action.js
```diff
@@ -1030,21 +1030,58 @@ context('Patient Action Form', function() {
...
+        fx.data.id = '1';
+        fx.data.relationships.form.data = { id: '11111' };
+        fx.data.relationships.state.data = { id: '55555' };
+        fx.data.relationships['form-responses'].data = [];
+        return fx;
+      })
+      .routePatient()
+      .routeFormByAction()
+      .routeFormDefinition()
+      .routeActionActivity()
```
nmajor25: Do we need `routePatient()` or `routeActionActivity()` here?

nmajor25 (approved):

test/integration/forms/action.js
paulfalgout: ah right this should be a `routesFor` this branch is actually really old.. will update

nmajor25 (approved):

PR #1005, opened 2023-04-04
Handle workspace specific settings

nmajor25:
Shortcut Story ID: [sc-35002]

Initial use case was adding patient sidebar widgets that are specific to a workspace. And that take precedent over that org setting.

### DIFF
--- a/test/integration/patients/patient/sidebar.js
+++ b/test/integration/patients/patient/sidebar.js
@@ -821,6 +821,34 @@ context('patient sidebar', function() {
       .should('not.contain', 'Workspace Two');
   });
 
+  specify('workspace specific widgets setting', function() {
+    cy
+      .routesForPatientDashboard()
+      .routeWorkspaces(fx => {
+        fx.data[0].attributes.settings = {
+          widgets_patient_sidebar: {
+            widgets: [
+              'divider',
+            ],
+          },
+        };
+
+        return fx;
+      });
+
+    cy
+      .visit('/patient/dashboard/1')
+      .wait('@routePatient');
+
+    cy
+      .get('.patient-sidebar')
+      .as('patientSidebar')
+      .find('.patient-sidebar__section')
+      .should('have.length', 1)
+      .first()
+      .find('.widgets__divider');
+  });
+
   specify('edit patient modal', function() {
     cy
       .routesForPatientDashboard()
--- a/src/js/services/bootstrap.js
+++ b/src/js/services/bootstrap.js
@@ -78,8 +78,12 @@ export default App.extend({
   },
   getSetting(settingName) {
     const workspaceSettings = this.currentWorkspace.get('settings');
-    const setting = get(workspaceSettings, settingName) || this.settings.get(settingName);
+    const workspaceSetting = get(workspaceSettings, settingName);
+    if (workspaceSetting) return workspaceSetting;
+
+    const setting = this.settings.get(settingName);
     if (!setting) return;
+
     return setting.get('value');
   },
   // Returns roles that the current user can manage


test/integration/patients/patient/sidebar.js
```diff
@@ -821,6 +821,120 @@ context('patient sidebar', function() {
...
+
+    cy
+      .visit('/patient/dashboard/1')
+      .wait('@routePatient')
+      .wait('@routePrograms')
+      .wait('@routeWidgets');
+
+    _.each(_.keys(fields), fieldName => {
+      cy.wait(`@routePatientField${ fieldName }`);
+    });
```
nmajor25: Still trying to get rid of these unnecessary `_.each`s since I'm only using the one patient field.

I created this test by copying and modifying the existing `'display patient data'` test, where a big list of patient fields are used.

test/integration/patients/patient/sidebar.js
```diff
@@ -821,6 +821,120 @@ context('patient sidebar', function() {
       .should('not.contain', 'Workspace Two');
   });
 
+  specify('workspace specific widgets setting', function() {
```
nmajor25, line 824: Better name for this test?

test/integration/patients/patient/sidebar.js
paulfalgout: I don't think you need to test a patient field here at all.  Just add the simplest widget.. for that matter it could just be a divider.  Just needs to prove it overrides the default,

paulfalgout (changes requested):

test/integration/patients/patient/sidebar.js
nmajor25: My assumption was that we needed to ~cover~ test the [getSidebarWidgets()](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/services/bootstrap.js#L104-L106) pulling from the workspace setting instead of the org setting.

But that would make the test a lot simpler without the field.

paulfalgout: I think we can rely on the default of `routeSettings` to set the sidebar we're testing to overwrite.

test/integration/patients/patient/sidebar.js
paulfalgout: seems ok

test/integration/patients/patient/sidebar.js
```diff
@@ -821,6 +821,120 @@ context('patient sidebar', function() {
...
+      'test-field': {
+        id: '1',
+        name: 'test-field',
+        value: '1',
+      },
+    };
+
+    cy
+      .routesForPatientDashboard()
+      .routeFormDefinition()
```
paulfalgout: I don't think any of these forms routes are in use

paulfalgout: I think you can get away with just  `routesForPatientDashboard` and `routeWorkspaces`

paulfalgout (changes requested):

test/integration/patients/patient/sidebar.js
nmajor25: Updated to that in my last push.

test/integration/patients/patient/sidebar.js
```diff
@@ -821,6 +821,35 @@ context('patient sidebar', function() {
...
+  specify('workspace specific widgets setting', function() {
+    cy
+      .routesForPatientDashboard()
+      .routeWorkspaces(fx => {
+        fx.data[0].attributes.settings = {
+          widgets_patient_sidebar: {
+            widgets: [
+              'divider',
+            ],
+            fields: [],
```
nmajor25: I don't think this is needed... now that I look at it again

nmajor25: I just removed it.

paulfalgout (approved):

paulfalgout:
failed test was the infamous console error :-/

nmajor25:
> failed test was the infamous console error :-/

😞

PR #1007, opened 2023-04-07
Add capped results messaging & action/flow counts to lists

nmajor25:
Shortcut Story ID: [sc-34741]

The backend recently capped list results to a maximum of 500 flows/actons. This PR is for adding messaging to lists on the frontend that indicates when a user's results are being capped.

The list count messaging can be in 4 different states:

1. The list count is below the capped maximum. ([screenshot](https://user-images.githubusercontent.com/35355575/230633166-9bc679b5-4dd3-4a38-bb72-aaf0149cdfff.png))
2. The capped maximum has been reached. ([screenshot](https://user-images.githubusercontent.com/35355575/230633296-faa7687b-80d0-4087-9e29-9a48074ef567.png))
3. The capped maximum has been reached, and the user narrows the list via the `Find in list...` input. ([screenshot](https://user-images.githubusercontent.com/35355575/230633351-df6688ad-a4eb-43df-a1b5-3337a25e712c.png))
4. The action/flow list is empty, so no count messaging is shown. ([screenshot](https://user-images.githubusercontent.com/35355575/230633239-20c180bb-1710-4c04-add8-b574856bc56d.png))

These changes apply to the worklists, schedule, and reduced schedule.

### DIFF
--- a/src/js/apps/patients/schedule/reduced_schedule_app.js
+++ b/src/js/apps/patients/schedule/reduced_schedule_app.js
@@ -4,14 +4,14 @@ import App from 'js/base/app';
 
 import SearchComponent from 'js/views/shared/components/list-search';
 
-import { LayoutView, ScheduleTitleView, TableHeaderView, ScheduleListView } from 'js/views/patients/schedule/schedule_views';
+import { LayoutView, ScheduleTitleView, TableHeaderView, ScheduleListView, CountView } from 'js/views/patients/schedule/schedule_views';
 
 export default App.extend({
   onBeforeStop() {
     this.collection = null;
   },
   onBeforeStart() {
-    this.setState({ isReduced: true });
+    this.setState({ isReduced: true, searchQuery: '' });
 
     this.showView(new LayoutView({
       state: this.getState(),
@@ -38,14 +38,21 @@ export default App.extend({
   },
   onStart(options, collection) {
     this.collection = collection;
+    this.filteredCollection = collection.clone();
 
     this.showScheduleList();
     this.showSearchView();
+    this.showCountView();
   },
   showScheduleList() {
     const collectionView = new ScheduleListView({
       collection: this.collection.groupByDate(),
       state: this.getState(),
+    });
+
+    this.listenTo(collectionView, 'filtered', filtered => {
+      this.filteredCollection.reset(filtered);
+      this.showCountView();
     });
 
     this.showChildView('list', collectionView);
@@ -72,6 +79,14 @@ export default App.extend({
     const tableHeadersView = new TableHeaderView();
 
     this.showChildView('table', tableHeadersView);
+  },
+  showCountView() {
+    const countView = new CountView({
+      collection: this.collection,
+      filteredCollection: this.filteredCollection,
+    });
+
+    this.showChildView('count', countView);
   },
   setSearchState(state, searchQuery) {
     this.setState({
--- a/src/js/apps/patients/schedule/schedule_app.js
+++ b/src/js/apps/patients/schedule/schedule_app.js
@@ -14,7 +14,7 @@ import DateFilterComponent from 'js/views/patients/shared/components/date-filter
 import SearchComponent from 'js/views/shared/components/list-search';
 import OwnerDroplist from 'js/views/patients/shared/components/owner_component';
 
-import { LayoutView, ScheduleTitleView, TableHeaderView, SelectAllView, ScheduleListView } from 'js/views/patients/schedule/schedule_views';
+import { LayoutView, ScheduleTitleView, TableHeaderView, SelectAllView, ScheduleListView, CountView } from 'js/views/patients/schedule/schedule_views';
 import { BulkEditButtonView, BulkEditActionsSuccessTemplate, BulkDeleteActionsSuccessTemplate } from 'js/views/patients/shared/bulk-edit/bulk-edit_views';
 
 export default App.extend({
@@ -54,9 +54,12 @@ export default App.extend({
       const isFiltersSidebarOpen = this.getState('isFiltering');
 
       if (!isFiltersSidebarOpen) Radio.request('sidebar', 'close');
+
       this.showScheduleTitle();
+      this.showCountView();
       this.showDateFilter();
       this.getRegion('list').startPreloader();
+
       return;
     }
 
@@ -89,6 +92,7 @@ export default App.extend({
     this.showScheduleList();
     this.showSearchView();
     this.toggleBulkSelect();
+    this.showCountView();
   },
   onChangeSelected() {
     this.toggleBulkSelect();
@@ -151,6 +155,7 @@ export default App.extend({
     this.listenTo(collectionView, 'filtered', filtered => {
       this.filteredCollection.reset(filtered);
       this.toggleBulkSelect();
+      this.showCountView();
     });
 
     this.showChildView('list', collectionView);
@@ -213,6 +218,14 @@ export default App.extend({
     this.listenTo(filtersApp, 'toggle:filtersSidebar', isSidebarOpen => {
       this.setState('isFiltering', isSidebarOpen);
     });
+  },
+  showCountView() {
+    const countView = new CountView({
+      collection: this.collection,
+      filteredCollection: this.filteredCollection,
+    });
+
+    this.showChildView('count', countView);
   },
   showDateFilter() {
     const dateTypes = ['due_date'];
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -16,7 +16,7 @@ import OwnerDroplist from 'js/views/patients/shared/components/owner_component';
 
 import { getSortOptions } from './worklist_sort';
 
-import { ListView, SelectAllView, LayoutView, ListTitleView, TableHeaderView, SortDroplist, TypeToggleView, NoOwnerToggleView } from 'js/views/patients/worklist/worklist_views';
+import { ListView, SelectAllView, LayoutView, ListTitleView, TableHeaderView, SortDroplist, TypeToggleView, NoOwnerToggleView, CountView } from 'js/views/patients/worklist/worklist_views';
 import { BulkEditButtonView, BulkEditFlowsSuccessTemplate, BulkEditActionsSuccessTemplate, BulkDeleteFlowsSuccessTemplate, BulkDeleteActionsSuccessTemplate } from 'js/views/patients/shared/bulk-edit/bulk-edit_views';
 
 const i18n = intl.patients.worklist.filtersApp;
@@ -69,12 +69,15 @@ export default App.extend({
 
     if (this.isRestarting()) {
       if (!isFiltersSidebarOpen) Radio.request('sidebar', 'close');
+
       this.showListTitle();
       this.showTypeToggleView();
       this.showSortDroplist();
       this.showTableHeaders();
       this.showDateFilter();
+      this.showCountView();
       this.getRegion('list').startPreloader();
+
       return;
     }
 
@@ -120,6 +123,7 @@ export default App.extend({
     this.listenTo(collectionView, 'filtered', filtered => {
       this.filteredCollection.reset(filtered);
       this.toggleBulkSelect();
+      this.showCountView();
     });
 
     this.listenTo(collectionView, 'click:patientSidebarButton', ({ model }) => {
@@ -131,6 +135,7 @@ export default App.extend({
 
     this.showSearchView();
     this.toggleBulkSelect();
+    this.showCountView();
   },
   startFiltersApp() {
     const state = this.getState();
@@ -257,6 +262,15 @@ export default App.extend({
     }
 
     Radio.request('alert', 'show:success', renderTemplate(BulkEditActionsSuccessTemplate, { itemCount }));
+  },
+  showCountView() {
+    const countView = new CountView({
+      isFlowList: this.getState().isFlowType(),
+      collection: this.collection,
+      filteredCollection: this.filteredCollection,
+    });
+
+    this.showChildView('count', countView);
   },
   showDateFilter() {
     if (this.getState().id !== 'shared-by' && this.getState().id !== 'owned-by') return;
--- a/src/js/static.js
+++ b/src/js/static.js
@@ -1,3 +1,5 @@
+const MAXIMUM_LIST_COUNT = /* istanbul ignore next */ _TEST_ ? 50 : 500;
+
 const ACTION_OUTREACH = {
   DISABLED: 'disabled',
   PATIENT: 'patient',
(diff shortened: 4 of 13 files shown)


src/js/views/patients/worklist/worklist_views.js
```diff
@@ -222,6 +223,43 @@ const EmptyFindInListView = View.extend({
...
+          {{formatMessage (intlGet "patients.worklist.worklistViews.countView.actionsCount") itemCount=count}}
+        {{/if}}
+      </span>
+    {{/if}}
+  `,
+  templateContext() {
+    return {
+      maximumCount: this.getOption('maximumCount'),
+      hasReachedMaximum: this.getOption('hasReachedMaximum'),
+      count: this.getOption('count'),
```
nmajor25: Do we want to add thousand separator formatting to the `count` and `maximumCount` (i.e. `1000` => `1,000`? 

It's not a problem now. But in the future, the capped results could be more than 1,000?

Could just throw a `.toLocaleString()` on it.

Just a thought as I was reviewing this.

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -222,6 +223,43 @@ const EmptyFindInListView = View.extend({
...
+      <div>{{ @intl.patients.worklist.worklistViews.countView.narrowFilters }}</div>
+    {{else}}
+      <span class="worklist-list__count-total">
+        {{#if isFlowList}}
+          {{formatMessage (intlGet "patients.worklist.worklistViews.countView.flowsCount") itemCount=count}}
+        {{else}}
+          {{formatMessage (intlGet "patients.worklist.worklistViews.countView.actionsCount") itemCount=count}}
+        {{/if}}
+      </span>
+    {{/if}}
```
nmajor25, line 234: This got a little complicated because the messaging needs to accommodate plural vs. non-plural wording (i.e. `"1 Action"` vs. `"1 Actions"`). Along with the different versions of `Flows` vs. `Actions`.

src/js/static.js
```diff
@@ -1,3 +1,5 @@
+const MAXIMUM_LIST_COUNT = /* istanbul ignore next */ _TEST_ ? 30 : 500;
```
nmajor25: Set to a maximum of `500` in production. And `30` in Cypress.

Loading a list of `500` actions/flows in Cypress was making me nervous that it would cause issues. 

The largest list I could find in the tests was `20`, so I figured `30` was safe. But maybe we want to raise it to `50` or something a little higher to be sure other tests don't unknowingly reach that max in the future... 🤔

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -258,6 +266,25 @@ export default App.extend({
...
+      this.hideCountView();
+      return;
+    }
+
+    const countView = new CountView({
+      isFlowList: this.getState().isFlowType(),
+      count: this.filteredCollection.length,
+      maximumCount: MAXIMUM_LIST_COUNT,
+      isFindInListApplied: this.getState('searchQuery').length > 2,
+      hasReachedMaximum: this.collection.length === MAXIMUM_LIST_COUNT,
```
nmajor25: The capped messaging will only show if the list count is exactly equal to the maximum count (i.e. `500`). Which is based on the `MAXIMUM_LIST_COUNT` static variable.

We'll have to keep the maximum list count in-sync with what the backend is using.

Would it be a good idea to have the actions/flows API requests include a `maxCount: 500` item or something similar?

Than we wouldn't have to update `MAXIMUM_LIST_COUNT` manually if the backend changes the maximum count on their end.

src/js/apps/patients/schedule/reduced_schedule_app.js
```diff
@@ -38,16 +40,23 @@ export default App.extend({
...
   showScheduleList() {
     const collectionView = new ScheduleListView({
       collection: this.collection.groupByDate(),
       state: this.getState(),
     });
 
+    this.listenTo(collectionView, 'filtered', filtered => {
+      this.filteredCollection.reset(filtered);
+      this.showCountView();
+    });
```
nmajor25, line 56: In order for the count messaging to be updated with the `Find in list...`, I needed to add this `'filtered'` listener to the list and the `this.filteredCollection` that is used in the other schedule and worklist apps.

nmajor25:
A few things I noticed when working on this that need to be explored further:

1. On the schedule, actions without a due date are removed from the list (via `groupByDate()` [here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/patients/schedule/schedule_app.js#L146)). But that happens after the API request is retrieved.
    - So there could be a scenario where the schedule retrieves results that are capped at `500`. But then the `groupByDate()` removes say `100` actions without a due date. So the schedule results would end up only showing `400` actions.
    - Can we update the API request to only retrieve actions with a due date that is not `null`? 
2. When the count messaging is shown/hidden on the reduced schedule, there's a page layout jump ([screengrab](https://user-images.githubusercontent.com/35355575/230641807-cc6c8e7f-c88e-4dca-af91-cec7c8508874.mov)). We could set a fixed height to that to avoid the page jump... or maybe there's a better way. But still need to look into that.

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -214,6 +222,24 @@ export default App.extend({
...
+    if (!this.collection || !this.filteredCollection.length) {
+      this.hideCountView();
+      return;
+    }
+
+    const countView = new CountView({
+      count: this.filteredCollection.length,
+      maximumCount: MAXIMUM_LIST_COUNT,
+      isFindInListApplied: this.getState('searchQuery').length > 2,
+      hasReachedMaximum: this.collection.length === MAXIMUM_LIST_COUNT,
```
paulfalgout: these same calculations are in a few places.  I think this view could be passed just `this.collection` and `this.filteredCollection` 

Then the view can import `MAXIMUM_LIST_COUNT` and find in list can be implied if `this.filteredCollection.length < this.collection.length` which might be better anyhow as 500 of 500 isn't entirely useful.

src/js/views/patients/worklist/worklist_views.js
paulfalgout: I think in the case of lots of `if` `else` templates, we can clean it up a bit by using that logic in a `getTemplate` and returning smaller simpler templates for each (or  maybe just some) logic branch.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -111,6 +112,33 @@ const SelectAllView = View.extend({
...
+      <div>
+        {{#if isFindInListApplied}}
+          {{formatMessage (intlGet "patients.schedule.scheduleViews.countView.maximumListCountNarrowed") itemCount=count maximumCount=maximumCount}}
+        {{else}}
+          {{formatMessage (intlGet "patients.schedule.scheduleViews.countView.maximumListCount") maximumCount=maximumCount}}
+        {{/if}}
+      </div>
+      <div>{{ @intl.patients.worklist.worklistViews.countView.narrowFilters }}</div>
+    {{else}}
+      <span class="schedule__count-total">
```
paulfalgout: I think we could probably use `<strong>` here

test/integration/patients/worklist/schedule.js
```diff
@@ -1709,6 +1929,12 @@ context('schedule page', function() {
       .get('@listSearch')
       .type('Second Action');
 
+    cy
+      .get('[data-count-region] .schedule__count-total')
+      .should(el => {
+        expect(el.text().trim()).to.equal('1 Action');
```
paulfalgout: why is the trim necessary?  `11 Actions`?
could maybe be `should('contain', '1 Action').should('not.contain', 'Actions')`

src/js/views/patients/worklist/worklist_views.js
paulfalgout: I don't think we need to worry about the separator for now.

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -222,6 +223,43 @@ const EmptyFindInListView = View.extend({
   `,
 });
 
+const CountView = View.extend({
```
paulfalgout: could see a world where this view is in some `patient/shared` directory.  I know the flow doesn't apply everywhere, but if that flag isn't set, it isn't set..  otherwise it is all identical correct?

src/js/static.js
paulfalgout: not a bad idea.  if this number changes in the future, let's request it get put on the meta maybe?  I'm not 100% sure how we're going to get back that result or handle it, but for now I think your static option works well.

paulfalgout (changes requested):

src/js/views/patients/worklist/worklist_views.js
nmajor25: Updated these to use a `getTemplate()`. Should be simpler now, but open to any other ways to simplify them more.

src/js/static.js
nmajor25: Sounds good, we'll wait for it to change in the future.

I also updated the maximum list count for Cypress from `30` => `50`. Should reduce the chance of accidentally hitting that limit in future tests.

src/js/apps/patients/schedule/schedule_app.js
nmajor25: That seems simpler. Updated to that in my last push.

src/js/views/patients/schedule/schedule_views.js
nmajor25: Yup, good call. Switched to that in the last push.

test/integration/patients/worklist/schedule.js
nmajor25: I'm not sure why the `trim()` is/was necessary, it had a bunch of spaces added when Cypress was comparing them.

But `should('contain', '1 Action').should('not.contain', 'Actions')` is a better way to do it. So I switched it over to that instead.

src/js/views/patients/worklist/worklist_views.js
nmajor25: Ya, it should all be identical besides the flow wording. I pushed the other changes up, but I'm still working on this.

nmajor25:
Reason for this commit: https://github.com/RoundingWell/care-ops-frontend/compare/ed8533d52a03a5886a9ef3199849087d42445934..12e03119648a547f1c6dd5f22c5b84c70bc474b2.

In the worklist test I added, I was using the `Action 50`/`Flow 50` and `Action 5`/`Flow 5` find in list queries to test the count messaging being shown if it's narrowed to show `Showing 1 Action of 50.` or `Showing 2 Actions of 50.`.

But this will cause flakiness if the tests run at the wrong time because of the `Added`/`Updated` list columns being included in the FIL filter. 

For example, the test would fail if the test ran at 9:50am, because each action would be included in the `Action 50` FIL query when it's only supposed to include one.

To fix this, I updated the test so the first action/flow has a name of `First Action`/`First Flow` (used to test `Showing 1 Action of 50.` messaging). And I made it so every other action/flow has `Test Patient`/`Other Patient` (used to test `Showing 25 Actions of 50.` messaging with the `Test Patient` FIL query).

This shouldn't be a problem on the schedule, but I updated those tests to match the worklist one.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -258,6 +264,23 @@ export default App.extend({
 
     Radio.request('alert', 'show:success', renderTemplate(BulkEditActionsSuccessTemplate, { itemCount }));
   },
+  showCountView() {
+    if (!this.collection || !this.filteredCollection.length) {
```
paulfalgout: you know what else would work here?  Since the `CountView` is pretty simple if could do this `!this.collection || !this.filteredCollection.length` and then just render an empty template?  I think a `<div></div>` isn't going to hurt anything?

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -222,6 +225,62 @@ const EmptyFindInListView = View.extend({
...
+    {{#if isFlowList}}
+      {{formatMessage (intlGet "patients.worklist.worklistViews.countView.flowsCount") itemCount=count}}
+    {{else}}
+      {{formatMessage (intlGet "patients.worklist.worklistViews.countView.actionsCount") itemCount=count}}
+    {{/if}}
+  </strong>
+`;
+
+const MaximumCountTemplate = hbs`
+  <div>{{formatMessage (intlGet "patients.worklist.worklistViews.countView.maximumListCount") maximumCount=maximumCount listTypeLabel=listTypeLabel}}</div>
```
paulfalgout: these don't need the wrapping `<div>`s

paulfalgout (changes requested):
Yeah looks good.. FWIW it doesn't have to be a component to be shared.  We have also have `_views` files in some shared folders.

src/js/views/patients/worklist/worklist_views.js
nmajor25: The `<div>`s are what make the two lines stack.

Without `<div>`s:

Screenshot 2023-04-11 at 9 49 36 AM

Or just remove the `<div>` around the `{{ formatMessage ... }}` portion?

paulfalgout: oh no you're right.. the line wrapping just made it look like the whole template was wrapped in a div.

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -222,6 +225,64 @@ const EmptyFindInListView = View.extend({
...
+  },
+  templateContext() {
+    const filteredCollection = this.getOption('filteredCollection');
+    const isFlowList = this.getOption('isFlowList');
+
+    return {
+      maximumCount: MAXIMUM_LIST_COUNT,
+      count: filteredCollection.length,
+      isFlowList,
+      listTypeLabel: isFlowList ? 'Flows' : 'Actions',
```
paulfalgout: can't do this.  needs to be in the i18n.. but the i18n could have a `{listTypeLabel, select, true "Flows" false "Actions"}` or something like that?  I dunno if it works with boolean..

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -111,6 +114,51 @@ const SelectAllView = View.extend({
   },
 });
 
+const ListCountTemplate = hbs`
```
paulfalgout, line 122: can we get away with not using any of these views/templates and moving the worklist ones to a shared location?

paulfalgout (changes requested):

src/js/views/patients/worklist/worklist_views.js
nmajor25: Good catch, just pushed an update for that.

paulfalgout (approved):

PR #1008, opened 2023-04-07
Change `filter[action.tag]` => `filter[action.tags]` in patient form response API request

nmajor25:
Shortcut Story ID: [sc-35176]

Feature for requesting a form response by a `prefill_action_tag` was added in #972.

But the `filter[action.tag]` should have been `filter[action.tags]`.


shadowhand:
FYI this needs to be a hotfix.

paulfalgout (approved):

PR #1010, opened 2023-04-11
Refactor worklists/schedule `CountView`s to use one shared view

nmajor25:
Shortcut Story ID: [sc-35368]

In #1007, we added a separate `CountView` for both the worklist and schedule. But they are almost completely identical. The only difference being that the worklist version supports messaging for `Flows` via the `isFlowList` variable.

This PR creates a shared `CountView` in a `/js/views/patients/shared/list_views.js` file that is used by the worklist, schedule, and reduced schedule.

### DIFF
--- a/src/js/apps/patients/schedule/reduced_schedule_app.js
+++ b/src/js/apps/patients/schedule/reduced_schedule_app.js
@@ -4,7 +4,9 @@ import App from 'js/base/app';
 
 import SearchComponent from 'js/views/shared/components/list-search';
 
-import { LayoutView, ScheduleTitleView, TableHeaderView, ScheduleListView, CountView } from 'js/views/patients/schedule/schedule_views';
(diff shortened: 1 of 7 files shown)


paulfalgout (changes requested):
The flaky tests are legit I think.  What we're seeing is `this.showCountView();` is called in the `onBeforeStart` this can occur when there's no `filteredCollection` if the app is restarted during its initial load.  So this should be replaced with clearing the region.

paulfalgout:
I assume this is a bug in the release as well that we should probably ship

nmajor25:
@paulfalgout nice catch, I've got the fix pushed. Not an issue for the reduced schedule right now, but we'll need to remember to include this when we add the filters sidebar.

paulfalgout (approved):
just waiting on the tests and I'll merge

paulfalgout:
Alright need to update the release

PR #1012, opened 2023-04-12
Update capped list + find in list filtered messaging

nmajor25:
Shortcut Story ID: [sc-35407]

This version of the messaging is shown when the list has reached its maximum length and then the user further filters the list via the `Find in list...` input.

Current messaging:

```
Showing 167 Actions of 500.
Try narrowing your filters.
```

New messaging:

```
Showing 167 of 500 Actions.
Try narrowing your filters.
```

### DIFF
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -424,9 +424,8 @@ careOptsFrontend:
         countView:
           actionsCount: "{itemCount, plural, one {# Action} other {# Actions}}"
           flowsCount: "{itemCount, plural, one {# Flow} other {# Flows}}"
-          maximumActionsCountNarrowed: Showing {itemCount, plural, one {# Action} other {# Actions}} of {maximumCount}.
-          maximumFlowsCountNarrowed: Showing {itemCount, plural, one {# Flow} other {# Flows}} of {maximumCount}.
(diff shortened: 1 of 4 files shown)


src/js/i18n/en-US.yml
```diff
@@ -424,9 +424,8 @@ careOptsFrontend:
         countView:
           actionsCount: "{itemCount, plural, one {# Action} other {# Actions}}"
           flowsCount: "{itemCount, plural, one {# Flow} other {# Flows}}"
-          maximumActionsCountNarrowed: Showing {itemCount, plural, one {# Action} other {# Actions}} of {maximumCount}.
-          maximumFlowsCountNarrowed: Showing {itemCount, plural, one {# Flow} other {# Flows}} of {maximumCount}.
           maximumListCount: Showing {maximumCount} of many {isFlowList, select, true {Flows} false {Actions}}.
+          maximumListCountNarrowed: Showing {itemCount} of {maximumCount} {isFlowList, select, true {Flows} false {Actions}}.
```
nmajor25, line 428: We no longer need to display a plural vs. non-plural version of `"Flow"`/`"Flows"` and `"Action"`/`"Actions"`. So we can now use the `isFlowList` variable to determine if `"Actions"` or `"Flows"` is displayed.

paulfalgout (approved):

PR #1013, opened 2023-04-14
Feature/form get field

paulfalgout:
Shortcut Story ID: [sc-35533]

I'd been looking into fixing [sc-35282] and then @zfixler, we had talked about adding a `getField`

In order to successfully test a patient field the form had loaded and a patient field the form hadn't loaded regarding patching it's `id` for coverage, the easiest way was to finish the `getField`

### DIFF
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -27,6 +27,7 @@ export default App.extend({
     'update:storedSubmission': 'updateStoredSubmission',
     'get:storedSubmission': 'getStoredSubmission',
     'clear:storedSubmission': 'clearStoredSubmission',
+    'fetch:field': 'fetchField',
     'update:field': 'updateField',
     'version': 'checkVersion',
   },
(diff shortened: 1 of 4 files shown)


src/js/services/forms.js
```diff
@@ -89,6 +105,9 @@ export default App.extend({
     return Promise.resolve(Radio.request('entities', 'fetch:directories:model', directoryName, query))
       .then(directory => {
         channel.request('send', 'fetch:directory', { value: directory.get('value'), requestId });
+      })
+      .catch(error => {
```
paulfalgout, line 109: @zfixler also I added a error to `getDirectory` to match the others so your `try/catch`es on `getDirectory` should reflect the API error

test/integration/forms/form.js
```diff
@@ -145,96 +166,186 @@ context('Noncontext Form', function() {
...
       .find('.choices__item--selectable')
+      .should('contain', 'error')
       .first()
-      .should('contain', 'Error: Bad Request');
+      .click();
   });
 
   specify('form scripts and reducers', { retries: 4 }, function() {
     cy
+      .route({
```
paulfalgout, line 344: this tests a form load check, but isn't worth it's own specify.  I moved it to a more generic test than the directory one

zfixler (approved):

shadowhand (approved):
Looks good as far as the UUID generation goes.

nmajor25 (approved):

PR #1014, opened 2023-04-14
Support automation program status

paulfalgout:
Shortcut Story ID: [sc-34988]

This updates flow/action status to the published/behavior pair

### DIFF
--- a/src/js/views/programs/shared/components/published_component.js
+++ b/src/js/views/programs/shared/components/published_component.js
@@ -52,12 +52,12 @@ export default Droplist.extend({
   isCompact: false,
   initialize(options) {
     const { published, behavior } = options;
-    this.mergeOptions(options, ['isPublishedDisabled', 'isConditionalAvailable']);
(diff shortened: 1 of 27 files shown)


src/js/views/programs/shared/components/published_component.js
```diff
@@ -19,41 +19,61 @@ const PublishedTemplate = hbs`<span class="{{ className }}">{{far icon}}<span>{{
...
+    behavior: PROGRAM_BEHAVIORS.AUTOMATED,
   },
 ];
 
 export default Droplist.extend({
   isCompact: false,
   initialize(options) {
-    this.mergeOptions(options, ['status', 'isPublishedDisabled', 'isConditionalAvailable']);
+    const { published, behavior } = options;
+    this.mergeOptions(options, ['isPublishDisabled', 'isConditionalAvailable']);
```
paulfalgout, line 55: this used the option `isPublishedDisabled` which wasn't in use because it's `isPublishDisabled`, but it is consistent with the `isCond...` and doesn't cause coverage issues so I kept it

nmajor25 (approved):

PR #1017, opened 2023-04-17
Test with firefox

paulfalgout:
CI should run in Firefox here


paulfalgout:
@zfixler @nmajor25 So it appears updating the firefox version fixed that dataclone issue

PR #1018, opened 2023-04-17
Add the `All Filters` sidebar to the reduced schedule

nmajor25:
Shortcut Story ID: [sc-34917]

### DIFF
--- a/src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
+++ b/src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
@@ -2,6 +2,10 @@ import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
+import StateModel from './reduced_schedule_state';
+
+import FiltersApp from 'js/apps/patients/schedule/filters_app';
+
 import SearchComponent from 'js/views/shared/components/list-search';
 
 import { CountView } from 'js/views/patients/shared/list_views';
@@ -9,11 +13,45 @@ import { CountView } from 'js/views/patients/shared/list_views';
 import { LayoutView, ScheduleTitleView, TableHeaderView, ScheduleListView } from 'js/views/patients/schedule/schedule_views';
 
 export default App.extend({
+  StateModel,
+  childApps: {
+    filters: {
+      AppClass: FiltersApp,
+      regionName: 'filters',
+      restartWithParent: false,
+      getOptions: ['currentClinician'],
+    },
+  },
+  stateEvents: {
+    'change:filters change:states': 'restart',
+  },
+  initListState() {
+    const storedState = this.getState().getStore();
+
+    if (storedState) {
+      this.setState(storedState);
+
+      return;
+    }
+
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    this.setState({ id: `reduced-schedule_${ currentUser.id }` });
+
+    this.getState().setDefaultFilterStates();
+  },
   onBeforeStop() {
     this.collection = null;
   },
   onBeforeStart() {
-    this.setState({ isReduced: true, searchQuery: '' });
+    if (this.isRestarting()) {
+      this.getRegion('list').startPreloader();
+
+      this.getRegion('count').empty();
+
+      return;
+    }
+
+    this.initListState();
 
     this.showView(new LayoutView({
       state: this.getState(),
@@ -23,19 +61,11 @@ export default App.extend({
 
     this.showScheduleTitle();
     this.showSearchView();
+    this.startFiltersApp();
     this.showTableHeaders();
   },
   beforeStart() {
-    const currentClinician = Radio.request('bootstrap', 'currentUser');
-
-    const currentWorkspace = Radio.request('bootstrap', 'currentWorkspace');
-    const states = currentWorkspace.getStates();
-
-    const filter = {
-      clinician: currentClinician.id,
-      state: states.groupByDone().notDone.getFilterIds(),
-    };
-
+    const filter = this.getState().getEntityFilter();
     return Radio.request('entities', 'fetch:actions:collection', { filter });
   },
   onStart(options, collection) {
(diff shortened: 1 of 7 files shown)


src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
```diff
@@ -2,18 +2,60 @@ import Radio from 'backbone.radio';
...
+  },
   onBeforeStop() {
     this.collection = null;
   },
   onBeforeStart() {
-    this.setState({ isReduced: true, searchQuery: '' });
+    if (this.isRestarting()) {
+      const isFiltersSidebarOpen = this.getState('isFiltering');
+
+      if (!isFiltersSidebarOpen) Radio.request('sidebar', 'close');
```
nmajor25: I don't *think* this is needed. But I wanted to make sure before I removed it.

My understanding is that this is needed on the worklist and schedule because we want the patient sidebar to close when the app is restarted via the date filter or owner filter. And we want the filters sidebar to not close in that scenario.

But neither of those filters exist on the reduced schedule. So this check shouldn't be necessary?

nmajor25: It should account for the only coverage drop from this PR.

Unless I missed something else.

src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
```diff
@@ -2,18 +2,60 @@ import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
+import StateModel from './reduced_schedule_state';
+
+import FiltersApp from 'js/apps/patients/schedule/filters_app';
```
nmajor25, line 7: I've split the `reduced_schedule_app.js` and `reduced_schedule_state.js` into their own directory (`/patients/schedule/reduced-schedule`).

But I'm still using the same `FiltersApp` shared by the schedule. I didn't want to create a new `reduced_filters_app` because it'd be identical.

src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_state.js
nmajor25, line 1: This is a copy of the `patients/schedule/schedule_state.js` file with everything stripped out that's not needed for the reduced schedule.

src/js/views/patients/schedule/layout.hbs
```diff
@@ -4,7 +4,9 @@
     <div class="schedule__list-search" data-search-region></div>
   </div>
   <div class="list-page__filters flex">
-    <div class="flex u-margin--r-24" data-select-all-region></div>
+    {{#unless isReduced}}
+      <div class="flex u-margin--r-24" data-select-all-region></div>
+    {{/unless}}
```
nmajor25, line 9: Hide this region completely if it's the reduced schedule. 

To move the `All Filters` button to the left without there being an empty space because of this element's margin.

nmajor25: ~~But maybe I could remove `u-margin--r-24` from this `data-select-all-region` element. And add `u-margin--l-24` to the `data-filters-region` element instead... 🤔~~

Nevermind, that won't work. It would still leave a margin-left of `24px` when there should be none on the reduced schedule.

src/js/views/patients/schedule/schedule-list.scss
```diff
@@ -12,6 +12,7 @@
   white-space: nowrap;
 
   &.is-reduced {
+    margin-right: 0;
     min-height: 36px;
```
nmajor25, line 16: `margin-right: 0` pushes the count view all the way to the right.

Screenshot 2023-04-17 at 1 37 16 PM

vs. 

Screenshot 2023-04-17 at 1 37 23 PM

I missed that when I initially built that out last week (#1007).

nmajor25: Also, we still need to keep the `min-height: 36px`.

The `All Filters` button is only `22px` in height. So if we removed it, there would be a slight layout jump when the capped messaging is shown/hidden.

test/integration/patients/worklist/schedule.js
```diff
@@ -1441,6 +1457,359 @@ context('schedule page', function() {
...
+    cy
+      .get('@filtersSidebar')
+      .find('.js-close')
+      .click();
+
+    cy
+      .get('@filtersSidebar')
+      .should('not.exist');
+  });
+
```
nmajor25: Should we split the reduced schedule tests into it's own context or a different file?

src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
paulfalgout: yeah so the sidebar close I think is for multi-select and the additional `if (!isFiltersSidebarOpen)` was needed to prevent that happening to the filter sidebar.  Definitely not needed since the other cases is impossible.

test/integration/patients/worklist/schedule.js
paulfalgout: yep. let's do it.

paulfalgout (changes requested):
cool.. just that one close sidebar thing and splitting the tests I think

test/integration/patients/worklist/schedule.js
nmajor25: Okay I just moved the reduced schedule tests into it's own context.

src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
nmajor25: Okay I just removed that, along with the `state.isFiltering`.

nmajor25: Well, we may need the `state.isFiltering`... it's used by the `filters_app`. Let me check that quick.

nmajor25: I think it's okay to remove `state.isFiltering` here.

The `filters_app` doesn't use it. It uses a `this.isFiltering` that threw me off ([here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/patients/schedule/filters_app.js#L27-L32)).

paulfalgout:
that's that coverage flake.. need to look into that.  not related to this

paulfalgout (approved):

PR #1021, opened 2023-04-18
Add an `All Time` date filter to the worklist & schedule

nmajor25:
Shortcut Story ID: [sc-35043]

When the `All Time` filter is selected, the API request will have the date filter removed entirely (i.e. `filter[created_at]`, `filter[update_at]`, or `filter[due_date]`).

In the UI, it looks like this:

Clipboard 2023-04-04 at 2 25 58 PM

### DIFF
--- a/src/js/apps/patients/worklist/worklist_state.js
+++ b/src/js/apps/patients/worklist/worklist_state.js
@@ -110,6 +110,10 @@ export default Backbone.Model.extend({
       };
     }
 
+    if (relativeDate === 'alltime') {
+      return {};
+    }
+
     const { prev, unit } = relativeRanges.get(relativeDate || 'thismonth').pick('prev', 'unit');
     const relativeRange = dayjs().subtract(prev, unit);
 
--- a/test/integration/patients/worklist/worklist.js
+++ b/test/integration/patients/worklist/worklist.js
@@ -1845,7 +1845,8 @@ context('worklist page', function() {
     cy
       .get('[data-date-filter-region]')
       .find('.js-next')
-      .click();
+      .click()
+      .wait('@routeActions');
 
     cy
       .get('[data-date-filter-region]')
@@ -1909,7 +1910,8 @@ context('worklist page', function() {
     cy
       .get('.datepicker')
       .find('.is-today')
-      .click();
+      .click()
+      .wait('@routeActions');
 
     cy
       .get('.datepicker')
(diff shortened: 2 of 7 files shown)


test/integration/patients/worklist/worklist.js
```diff
@@ -1920,6 +1920,29 @@ context('worklist page', function() {
...
+        expect(storage.actionsDateFilters.selectedMonth).to.be.null;
+        expect(storage.actionsDateFilters.dateType).to.equal('created_at');
+      });
+
+    cy
+      .wait('@routeActions')
+      .itsUrl()
+      .its('search')
+      .should('not.contain', 'filter[created_at]');
+
```
nmajor25: The `.should('not.contain', 'filter[created_at]')` portion of this is failing and I'm not sure why...

When the `All Time` picklist option is clicked, the `routeActions()` request's query parameters are correct (i.e. `filter[created_at]` is absent).

But when I do the `.wait('@routeActions')` and check the URL, the `filter[created_at]=` is now in the URL request query somehow??

Screenshot 2023-04-18 at 5 09 39 PM

If I move this block of code up at the beginning of the test, it passes.

It has me perplexed... 🤔

paulfalgout: all of the `routeAction` requests above weren't waited on.  need to add a `wait` for every route hit in the tests above so this one know which route hit you're waiting on.

paulfalgout: It looks like `routeActions` was called 23 times by this point, so this should be the 23rd `wait('@routeActions')` by this point in the test.

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -110,6 +110,10 @@ export default Backbone.Model.extend({
       };
     }
 
+    if (relativeDate === 'alltime') {
+      return null;
```
paulfalgout: Just return `{}` here then the `|| {}` aren't necessary.

paulfalgout (changes requested):
Looks pretty good.  just need the same state changes on schedule

test/integration/patients/worklist/worklist.js
nmajor25: Ahhh interesting, well that explains it... nice catch

src/js/apps/patients/worklist/worklist_state.js
nmajor25: Yup that's a better way to do it. Just pushed that update.

test/integration/patients/worklist/worklist.js
nmajor25, line 1: For the worklist I'm testing the `All Time` option for each tab in the picklist (`Added`, `Updated`, & `Due`).

Possibly overkill?

paulfalgout: oh yeah, you can just do it once

nmajor25: Okay, I've got those removed.

paulfalgout (approved):

PR #1022, opened 2023-04-19
Replace $.params with rwell specific GET param handling

paulfalgout:
Shortcut Story ID: [sc-32354]

This replaces the magic of `$.param` and reduces our dependency on jQuery but I am uncertain if it handles every edge case.  We may run into issues in the future where it isn't enough, but we'll be able to modify it as necessary.

This also allows for `fetcher(fooUrl, { data: { filter: { foo: 1 } }).then...` so we aren't building url params all over the place.

_However_ this doesn't handle everything like the file upload which is a `POST` to a url with params, but those seem like edge cases beyond the scope of this work.

### DIFF
--- a/src/js/base/fetch.js
+++ b/src/js/base/fetch.js
@@ -1,15 +1,47 @@
 //  Similar to https://github.com/akre54/Backbone.Fetch
-import $ from 'jquery';
-import { isObject, defaults, extend } from 'underscore';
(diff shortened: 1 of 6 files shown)


paulfalgout:
The coverage drop is coverage flake from https://github.com/RoundingWell/care-ops-frontend/pull/1020. I need to add a wait so the second check is always hit

nmajor25 (approved):
A definite improvement 👍

PR #1023, opened 2023-04-19
Allow alwaysEnable on Button Components

zfixler:
For use in `Read Only` forms where buttons are used for toggling show/hide conditionals.
Shortcut Story ID: [sc-35714]

### DIFF
--- a/src/js/formapp/components.js
+++ b/src/js/formapp/components.js
@@ -91,9 +91,26 @@ class DirectoryComponent extends SelectComponent {
   }
 }
 
+class CustomButtonComponent extends Formio.Components.components.button {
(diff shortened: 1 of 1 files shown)


src/js/formapp/components.js
```diff
@@ -91,9 +92,35 @@ class DirectoryComponent extends SelectComponent {
   }
 }
 
+class ButtonComponent extends Button {
```
paulfalgout: I think `extends Formio.Components.components.button` here is ok like on line 8

paulfalgout (changes requested):
one nit-pick then all good.

paulfalgout (approved):

PR #1024, opened 2023-04-19
Remove some test/coverage flake

paulfalgout:
No story here.. 3 small test-only changes.. related to that might fix test/coverage flake

### DIFF
--- a/test/integration/forms/action.js
+++ b/test/integration/forms/action.js
@@ -603,6 +603,7 @@ context('Patient Action Form', function() {
       })
       .routeActionActivity()
       .routePatientByAction(fx => {
+        fx.data.id = '1';
         fx.data.attributes.first_name = 'Testin';
         fx.data.attributes.last_name = 'Mctester';
 
--- a/test/integration/patients/patient/dashboard.js
+++ b/test/integration/patients/patient/dashboard.js
@@ -853,7 +853,9 @@ context('patient dashboard page', function() {
       })
       .visit('/patient/dashboard/1')
       .wait('@routePatient')
-      .wait('@routePatientFlows');
+      .wait('@routePatientField')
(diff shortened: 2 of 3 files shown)


test/integration/forms/action.js
```diff
@@ -603,6 +603,7 @@ context('Patient Action Form', function() {
       })
       .routeActionActivity()
       .routePatientByAction(fx => {
+        fx.data.id = '1';
```
paulfalgout, line 606: this was using an id out of the fixtures.. patients are also loaded along with the actions list, so if you got unlucky and the same fixture was pulled, it would get overridden

test/integration/patients/patient/dashboard.js
```diff
@@ -853,7 +853,8 @@ context('patient dashboard page', function() {
       })
       .visit('/patient/dashboard/1')
       .wait('@routePatient')
-      .wait('@routePatientFlows');
+      .wait('@routePatientFlows')
+      .wait('@routeAllProgramActions');
```
paulfalgout: not waiting for `routeAll` sometimes had the tests ending prior to the 2nd check this is testing in the `onStart`

test/integration/patients/patient/patient-flow.js
```diff
@@ -291,7 +291,8 @@ context('patient flow page', function() {
       .get('.picklist')
       .find('.js-picklist-item')
       .contains('Nurse')
-      .click();
+      .click()
+      .wait('@routePatchAction');
```
paulfalgout, line 295: I'm not sure on this one.. it's a coverage check for the owner changing on an action on a flow..    we see it _very_ rarely..  I have a hard time believing this will fix it, but this route does get hit, and it's not waiting for it..

nmajor25 (approved):

PR #1025, opened 2023-04-19
Fix some css styling for the form widget button

nmajor25:
Shortcut Story ID: [sc-35401]

Fix icon/text alignment and add an ellipsis text overflow.

Before:

Screenshot 2023-04-19 at 2 46 47 PM

After: 

Screenshot 2023-04-19 at 3 16 50 PM

### DIFF
--- a/src/js/views/patients/patient/patient.scss
+++ b/src/js/views/patients/patient/patient.scss
@@ -16,7 +16,6 @@
   display: flex;
   flex: initial;
   min-height: 0;
-  width: 256px;
 }
 
 .patient__content {
--- a/src/js/views/patients/widgets/widgets.scss
+++ b/src/js/views/patients/widgets/widgets.scss
@@ -7,14 +7,20 @@
 }
 
 .widgets__form-widget {
+  align-items: center;
+  display: flex;
+  min-width: 0;
+  padding-right: 8px;
   width: 100%;
 
-  .fa-up-right-and-down-left-from-center {
-    font-size: 12px;
-    margin-right: 0;
-    position: absolute;
-    right: 8px;
-    top: 8px;
+  .widgets__form-widget-expand-icon {
+    margin-left: auto;
+    padding-left: 8px;
+
+    .icon {
+      font-size: 12px;
+      vertical-align: -0.125em;
+    }
   }
 }
 
--- a/src/js/views/patients/patient/sidebar/patient-sidebar.scss
+++ b/src/js/views/patients/patient/sidebar/patient-sidebar.scss
@@ -2,6 +2,7 @@
   border-left: 1px solid $black-90;
   padding: 52px 24px 0;
   position: relative;
+  width: 256px;
 }
 
 .patient-sidebar__name {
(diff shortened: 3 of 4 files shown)


src/js/views/patients/widgets/widgets.scss
```diff
@@ -7,14 +7,17 @@
...
+  padding: 8px 8px 8px 16px;
   width: 100%;
 
   .fa-up-right-and-down-left-from-center {
     font-size: 12px;
-    margin-right: 0;
-    position: absolute;
-    right: 8px;
-    top: 8px;
+    margin-left: auto !important;
```
nmajor25: I really don't like using `!important`, but this is being overriden by the `buttons.scss` module styles for this ([here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/scss/modules/buttons.scss#L36-L38)).

Maybe there's another way to avoid using `!important`. Not sure though.

paulfalgout: I think you could wrap the svg in a span to avoid this

src/js/views/patients/widgets/widgets.scss
```diff
@@ -7,14 +7,17 @@
 }
 
 .widgets__form-widget {
+  align-items: center;
+  display: flex;
+  max-width: 204px;
+  min-width: 0;
+  padding: 8px 8px 8px 16px;
```
paulfalgout: I think this should just be `padding-right: 8px;` leave the rest of the padding to the button

src/js/views/patients/widgets/widgets.scss
```diff
@@ -7,14 +7,17 @@
 }
 
 .widgets__form-widget {
+  align-items: center;
+  display: flex;
+  max-width: 204px;
```
paulfalgout: So I dislike `max-width` generally, _but_ the `div` this is in has a `max-width` of `256px`.  That said I think we probably have a problem there.  If something does have max-width of 256 we have inner horizontal scrolling.. we need to figure out where that went wrong before we can apply another max width.. which hopefully we don't need.

paulfalgout (changes requested):

src/js/views/patients/widgets/widgets.scss
nmajor25: Yup that removed the need for the `!important`, but had to move some things around since some styles now need to be applied to that `span` instead of the `svg`.

src/js/views/patients/widgets/widgets.scss
nmajor25: That worked well. That update is pushed.

src/js/views/patients/patient/patient.scss
```diff
@@ -16,7 +16,6 @@
   display: flex;
   flex: initial;
   min-height: 0;
-  width: 256px;
```
nmajor25, line 19: Removing this and adding it to `.patient-sidebar` instead seems to fix the issue. I think the padding on `.patient-sidebar` was pushing the width.

My understanding is that this `.patient__sidebar` class is only used for the patient sidebar on the patient dashboard. The other patient sidebars use different CSS classes. So it shouldn't effect any other sidebars.

src/js/views/patients/widgets/widgets.scss
nmajor25: Just pushed what I believe to be a fix for this.

PR #1028, opened 2023-04-24
Persist 'Find in list...' query on the worklist/schedule

nmajor25:
Shortcut Story ID: [sc-34500]

A user's `Find in list...` search query will now be stored in local storage. 

Which means it will persist when the worklist/schedule is restarted (i.e. due to a new date, clinician, etc. filter being selected) and on route changes.

These changes apply to the worklist, schedule, and reduced schedule.

### DIFF
--- a/src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_state.js
+++ b/src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_state.js
@@ -1,4 +1,4 @@
-import { clone, omit, each, filter, contains } from 'underscore';
+import { clone, each, filter, contains } from 'underscore';
 import store from 'store';
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
@@ -31,7 +31,7 @@ export default Backbone.Model.extend({
     return store.get(this.getStoreKey());
   },
   onChange() {
-    store.set(this.getStoreKey(), omit(this.attributes, 'searchQuery'));
+    store.set(this.getStoreKey(), this.attributes);
   },
   getFilters() {
     return clone(this.get('filters'));
--- a/src/js/views/patients/schedule/schedule_views.js
+++ b/src/js/views/patients/schedule/schedule_views.js
@@ -295,6 +295,9 @@ const DayListView = CollectionView.extend({
 
     this.listenTo(state, 'change:searchQuery', this.searchList);
   },
+  onAttach() {
+    this.searchList(null, this.state.get('searchQuery'));
+  },
   childViewTriggers: {
     'render': 'listItem:render',
     'select': 'select',
@@ -359,7 +362,7 @@ const ScheduleListView = CollectionView.extend({
     'render:children': 'onChildFilter',
   },
   emptyView() {
-    if (this.state.get('searchQuery')) {
+    if (this.collection.length && this.state.get('searchQuery')) {
       return EmptyFindInListView;
     }
 
@@ -369,7 +372,7 @@ const ScheduleListView = CollectionView.extend({
     return alphaSort('asc', viewA.model.get('date'), viewB.model.get('date'));
   },
   viewFilter(view) {
-    if (this.state.get('searchQuery')) {
+    if (this.isAttached() && this.state.get('searchQuery')) {
       return !view.isEmpty();
     }
 
--- a/src/js/views/shared/components/list-search/index.js
+++ b/src/js/views/shared/components/list-search/index.js
@@ -4,7 +4,6 @@ import Component from 'js/base/component';
 
 import { SearchView } from './list-search_views';
 
-
 const StateModel = Backbone.Model.extend({
   defaults: {
     query: '',
--- a/src/js/views/shared/components/list-search/list-search_views.js
+++ b/src/js/views/shared/components/list-search/list-search_views.js
@@ -9,8 +9,13 @@ import './list-search-component.scss';
 
 const InputTemplate = hbs`
   <span class="list-search__search-icon">{{far "magnifying-glass"}}</span>
-  <input class="list-search__input input-primary--small js-input w-100" type="text" {{#if isDisabled}}disabled{{/if}} placeholder="{{@intl.shared.components.listSearch.listSearchViews.placeholder}}"/>
-  <span class="list-search__clear-icon js-clear is-hidden">{{fas "circle-xmark"}}</span>
+  <input
+    class="list-search__input input-primary--small js-input w-100"
+    type="text"
+    placeholder="{{@intl.shared.components.listSearch.listSearchViews.placeholder}}"
+    value="{{query}}"
+  />
+  <span class="list-search__clear-icon js-clear {{#unless query}}is-hidden{{/unless}}">{{fas "circle-xmark"}}</span>
 `;
 
 const SearchView = View.extend({
@@ -21,7 +26,6 @@ const SearchView = View.extend({
   template: InputTemplate,
   templateContext() {
     return {
-      isDisabled: this.getOption('state').isDisabled,
       query: this.getOption('state').query,
     };
   },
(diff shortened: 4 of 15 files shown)


src/js/views/patients/schedule/schedule_views.js
```diff
@@ -369,6 +372,7 @@ const ScheduleListView = CollectionView.extend({
     return alphaSort('asc', viewA.model.get('date'), viewB.model.get('date'));
   },
   viewFilter(view) {
+    // potential source of issues
     if (this.state.get('searchQuery')) {
       return !view.isEmpty();
     }
```
nmajor25, line 377: When the FIL query is persisted on an app restart, this is deeming all views to be `view.isEmpty() === true` and, therefore, the list is shown empty even when it shouldn't be.

Which I think is because the view filter is being applied before the views are rendered.

It's almost like we need a `if (isRestarting) return` at the top of this `viewFilter()` function. And then apply a the filter once the app is done restarting and the list items are loaded. But I couldn't get that to work yet.

nmajor25: For reference, if this code is removed:

```js
if (this.state.get('searchQuery')) {
  return !view.isEmpty();
}
```

The list doesn't filter out empty days ([screenshot](https://user-images.githubusercontent.com/35355575/234058815-e890baf7-02d5-4e3d-960e-2afc2fd82653.png)).

nmajor25: Or maybe the filter on the `DayListView`'s `onAttach()` needs to be done differently. Could possibly be causing issues too.

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -368,17 +371,13 @@ const ScheduleListView = CollectionView.extend({
...
+      if (!children.length || model.get('date') === 'null') return false;
 
-    // 'null' string is a key from groupBy
-    if (view.model.get('date') === 'null') {
-      return false;
-    }
-
-    return true;
+      return true;
+    });
```
nmajor25: Things appear to be working with the schedule now. Not sure if this is the best way to do it though 🤷

nmajor25:
There is a slight layout jump on the schedule when the app is restarted with a persisted find in list query ([screengrab](https://user-images.githubusercontent.com/35355575/234103986-8128f668-5be0-41f7-8ee2-27e3342fd377.mov)).

Which I don't like, but not sure how to avoid it.

src/js/views/shared/components/list-search/index.js
```diff
@@ -18,6 +17,9 @@ export default Component.extend({
     'watch:change': 'onViewWatchChange',
     'clear': 'onViewClear',
   },
+  initialize({ state }) {
+    if (state.query) this.setState('query', state.query);
```
paulfalgout: pretty sure this is the default functionality of how a component sets the initial state.  I don't think this `initialize` is needed

src/js/views/shared/components/list-search/list-search_views.js
```diff
@@ -9,8 +9,13 @@ import './list-search-component.scss';
...
   <span class="list-search__search-icon">{{far "magnifying-glass"}}</span>
-  <input class="list-search__input input-primary--small js-input w-100" type="text" {{#if isDisabled}}disabled{{/if}} placeholder="{{@intl.shared.components.listSearch.listSearchViews.placeholder}}"/>
-  <span class="list-search__clear-icon js-clear is-hidden">{{fas "circle-xmark"}}</span>
+  <input
+    class="list-search__input input-primary--small js-input w-100"
+    type="text"
+    placeholder="{{@intl.shared.components.listSearch.listSearchViews.placeholder}}"
+    value="{{query}}"
+  />
+  <span class="list-search__clear-icon js-clear {{#if shouldClearBeHidden}}is-hidden{{/if}}">{{fas "circle-xmark"}}</span>
```
paulfalgout: can just be `{{#unless query}}`

src/js/views/patients/schedule/schedule_views.js
paulfalgout: going to look at this more tomorrow to wrap my 🧠 around it.

paulfalgout (changes requested):

src/js/views/shared/components/list-search/index.js
nmajor25: It doesn't appear to be needed. Can't remember why I added that in the first place tbh...

But just removed it.

src/js/views/shared/components/list-search/list-search_views.js
nmajor25: Yup, just pushed that change.

src/js/views/patients/schedule/schedule_views.js
paulfalgout: Having a hard time with this one.  This was previously removing the collection that has no "day" because there's no where to put them..  _but_ it currently removes days that have no children with find in list.. so what about this PR requires this?

paulfalgout: I think the issue is that this filter first runs prior to the children rendering, so they're always empty.
This avoids this:
```js
  viewFilter(view) {
    if (this.isAttached() && this.state.get('searchQuery')) {
      return !view.isEmpty();
    }

    // 'null' string is a key from groupBy
    if (view.model.get('date') === 'null') {
      return false;
    }

    return true;
  },
```
but there might be a flash of content if there's lots of stuff.. still though.. it kinda has to render to be applied.

Thoughts?

nmajor25: Ah `this.isAttached()` is what I was looking for... nice. That works well. I just pushed that change

But ya that's the problem I was trying to solve with the `filterOutEmptyDays()` function. To have the view filter applied after the children were rendered. Otherwise, the view filter was clearing out all the days even when it shouldn't.

Ya I don't think there's a way to avoid the layout flash/shift.

paulfalgout: Ah this is essentially what you suggested in the earlier comments.. for whatever reason those are showing up on an "outdated" commit so I missed it till I came back to the main page 🤦

paulfalgout:
And I have no idea what that layout jump is... can't seem to catch why that'd happen

nmajor25:
@paulfalgout latest push adds the `searchQuery` to the local storage and should fix the broken tests.

I'm still working on adding tests for persisting the FIL.

src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_state.js
```diff
@@ -31,7 +31,7 @@ export default Backbone.Model.extend({
     return store.get(this.getStoreKey());
   },
   onChange() {
-    store.set(this.getStoreKey(), omit(this.attributes, 'searchQuery'));
+    store.set(this.getStoreKey(), omit(this.attributes));
```
paulfalgout: can just remove `omit` altogether here

src/js/views/patients/schedule/schedule_views.js
nmajor25: Haha all good!

paulfalgout (changes requested):
One small change

test/integration/patients/worklist/reduced-schedule.js
nmajor25, line 1: What do you think about creating a new `'Find In List'` test in this file.

It's currently being tested in `'display schedule'`, but it'd be easier to read/update in the future if it was split imo. Working on it just now was kind of confusing.

It could be in a future story. This PR is complicated enough 😩

src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_state.js
nmajor25: Oops yup, good call. Just removed that.

paulfalgout:
coverage drop is not part of this PR, it's the flake

PR #1034, opened 2023-04-26
Add a `Find in list...` specific test to the reduced schedule test file

nmajor25:
Shortcut Story ID: [sc-35924]

Currently the `Find in list...` functionality for the reduced schedule is tested in the `reduced schedule page => display schedule` test.

This PR creates a new `reduced schedule page => find in list` test in that same file.

Moving the `Find in list...` specific testing to it's own test will make it easier to understand and update those tests in the future.


nmajor25:
@paulfalgout still need to figure out why those test are flaky.

PR #1035, opened 2023-04-26
Remove collection clearing in the `onBeforeStart()` of the worklist, schedule, & clinician apps

nmajor25:
Shortcut Story ID: [sc-36004]

In PR #980 and #999, we added these to the worklist, schedule, and clinicians apps:

```js
onBeforeStop() {
   this.collection = null;
},
```

These were originally added to ensure the FIL input element was disabled properly when the list is loading.

After #1028, the FIL element is no longer being disabled and is now always in an enabled state. Therefore, we no longer need to clear the collection when these apps are stopped.

### DIFF
--- a/src/js/apps/clinicians/clinicians-all_app.js
+++ b/src/js/apps/clinicians/clinicians-all_app.js
@@ -18,9 +18,6 @@ export default SubRouterApp.extend({
   viewEvents: {
     'click:addClinician': 'onClickAddClinician',
   },
-  onBeforeStop() {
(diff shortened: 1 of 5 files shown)


paulfalgout (approved):

nmajor25:
@paulfalgout I couldn't figure out the flake with the `clinicians list > find in list` test.

So I reverted the `Fix some 'Find in list' related test flake` commit. And will create a separate shortcut story for it.

PR #1036, opened 2023-04-27
Give the `Find in list...` element a blue/white CSS color scheme when applied to lists

nmajor25:
Shortcut Story ID: [sc-34500]

To give the user a better visual indication that a FIL search query is being applied.

What the UI change looks like:

Clipboard 2023-26-04 at 4 48 03 PM (1)


paulfalgout (changes requested):
put `is-applied` conditionally on the `className` and modify it on `this.$el` I think.

paulfalgout (approved):

PR #1037, opened 2023-04-28
Move stored submission into onChange using the latest reduced submission

paulfalgout:
`prevSubmission` is always the result of the latest change reducers

Shortcut Story ID: [sc-###]

### DIFF
--- a/src/js/formapp.js
+++ b/src/js/formapp.js
@@ -58,6 +58,12 @@ function getContext(contextScripts) {
 
 let prevSubmission;
 
+function updateSubmission() {
+  router.request('update:storedSubmission', prevSubmission);
+}
+
+const updateSubmissionDebounce = debounce(updateSubmission, /* istanbul ignore next */ _TEST_ ? 100 : 2000);
+
 const onChange = function(form, changeReducers) {
   const data = getChangeReducers(form, changeReducers, structuredClone(form.submission.data), prevSubmission);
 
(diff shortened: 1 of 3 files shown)


zfixler (approved):

src/js/formapp.js
```diff
@@ -58,21 +58,24 @@ function getContext(contextScripts) {
...
+const updateSubmisionDebounce = debounce(updateSubmision, /* istanbul ignore next */ _TEST_ ? 100 : 2000);
+
 const onChange = function(form, changeReducers) {
   const data = getChangeReducers(form, changeReducers, structuredClone(form.submission.data), prevSubmission);
 
   form.data = data;
   form.setSubmission({ data }, { fromChangeReducers: true, fromSubmission: false });
 
   prevSubmission = structuredClone(form.submission.data);
+  updateSubmisionDebounce();
```
nmajor25: Missing an `s` in `submission` in a few places 👀 , `updateSubmision` vs. `updateSubmission`

nmajor25: Looks like that was there from before these changes...

paulfalgout: `s's are expenive

nmajor25: 😆

src/js/formapp.js
```diff
@@ -127,6 +128,8 @@ async function renderForm({ definition, isReadOnly, storedSubmission, formData,
     // Always run one last change event on submit
     onChangeDebounce.cancel();
     onChange(form, changeReducers);
+    updateSubmissionDebounce.cancel();
```
nmajor25, line 131: When would this be needed? Out of curiosity.

When a user submits the form before the `updateSubmissionDebounce()` finishes? To avoid a scenario where `router.request('update:storedSubmission', prevSubmission);` is called after a user hits the button to submit the form?

nmajor25 (approved):

src/js/formapp.js
paulfalgout: so `onChange` calls `updateSubmissionDebounce` now, but at this point it's never really wanted.  I suppose if you have an error _maybe_ it should run, but realistically we never want it to resolve after the form submission, because the form submission deletes the entry and if there's an error, the user is almost for sure going to make a change...  very low cost for any data loss.

nmajor25: Okay gotcha, thanks 👍

src/js/formapp.js
zfixler: > `s's are expenive

See, that's why I had no problem leaving those off. 😆

PR #1040, opened 2023-05-01
Don't persist `Find in list...` query when a user's session ends

nmajor25:
Shortcut Story ID: [sc-36060]

A user's `Find in list...` query should only persist during a specific browser session. If the user closes the app, the `Find in list...` query should be cleared.

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -41,6 +41,7 @@ export default App.extend({
     'change:flowsSortId': 'onChangeStateSort',
     'change:selectedFlows': 'onChangeSelected',
     'change:selectedActions': 'onChangeSelected',
+    'change:searchQuery': 'onChangeSearchQuery',
   },
   onChangeStateSort() {
     if (!this.isRunning()) return;
@@ -50,8 +51,13 @@ export default App.extend({
   onChangeSelected() {
     this.toggleBulkSelect();
   },
+  onChangeSearchQuery(state) {
+    this.currentSearchQuery = state.get('searchQuery');
+  },
   initListState() {
     const storedState = this.getState().getStore(this.worklistId);
+
+    this.setState({ searchQuery: this.currentSearchQuery });
 
     if (storedState) {
       this.setState(storedState);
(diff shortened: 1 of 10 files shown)


src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -50,9 +51,14 @@ export default App.extend({
   onChangeSelected() {
     this.toggleBulkSelect();
   },
+  onChangeSearchQuery(state) {
+    this.currentSearchQuery = state.get('searchQuery');
+  },
   initListState() {
     const storedState = this.getState().getStore(this.worklistId);
 
+    this.setState({ searchQuery: this.currentSearchQuery });
```
nmajor25, line 60: I originally had this at the top of the `initListState()` function like this:

```js
initListState() {
  this.setState({ searchQuery: this.currentSearchQuery });

  const storedState = this.getState().getStore(this.worklistId);

  . . .

}
```

But if we set the search query state before `const storedState = this.getState().getStore(this.worklistId);`, ` this.getState().getStore()` will return the state defaults instead of what was stored local storage.

Since `this.setState()` sets local storage:

```js
onChange() {
  store.set(this.getStoreKey(), omit(this.attributes, 'searchQuery'));
},
```

test/fixtures/test/roles.json
```diff
@@ -61,7 +61,8 @@
       "work:manage",
       "work:owned:manage",
       "app:schedule:reduced",
-      "work:own"
+      "work:own",
+      "dashboards:view"
```
nmajor25, line 65: In the `reduced-schedule => find in list` test, I need to verify that the FIL query persists on a route change. 

Adding the `"dashboards:view"` permission adds the `Dashboards` link to the app nav for reduced schedule users. Which allows me to easily go to the `/dashboards` page and then come back to the schedule to verify the FIL query persisted.

test/integration/patients/worklist/reduced-schedule.js
```diff
@@ -748,62 +742,30 @@ context('reduced schedule page', function() {
...
-      .find('[data-search-region] .list-search__container')
-      .should('have.class', 'is-applied');
+      }, 100)
+      .routeDashboards()
+      .visit('/schedule');
 
     cy
       .get('[data-count-region]')
-      .should('not.contain', '1 Action');
+      .should('not.contain', '20 Actions');
```
nmajor25, line 751: This was flakey locally. Curious to see if it's also flakey on CircleCI...

nmajor25:
When working on this, I realized we didn't add the persist functionality to the clinicians page.

I created a separate shortcut story for that: [36164](https://app.shortcut.com/roundingwell/story/36164/persist-find-in-list-query-on-the-clinicians-page).

paulfalgout (approved):

PR #1042, opened 2023-05-03
Make auth.js only dynamically exported and update appconfig.json

paulfalgout:
Shortcut Story ID: [sc-36309]

### DIFF
--- a/src/assets/appconfig.json
+++ b/src/assets/appconfig.json
@@ -1,17 +1,19 @@
 {
   "app": {
     "env": "develop",
-    "stack_id": "localhost"
+    "stackId": "localhost",
+    "name": "Salvation Clinic"
   },
   "auth0": {
     "domain": "auth.roundingwell.com",
     "clientId": "UVGIL27XqwGsJwsTfEyz50c1qg3Ll1p6",
-    "connection": "google-oauth2",
-    "name": "Salvation Clinic"
+    "authorizationParams": {
+      "connection": "google-oauth2"
+    }
   },
   "datadog": {
-    "client_token": "pubdf015273e710a5a7b64bd5c80f126b9a",
-    "app_id": "72496c17-540f-4aad-b5bb-848b8e836a3d"
+    "clientToken": "pubdf015273e710a5a7b64bd5c80f126b9a",
+    "applicationId": "72496c17-540f-4aad-b5bb-848b8e836a3d"
   },
   "versions": {
     "frontend": "develop",
--- a/src/js/auth.js
+++ b/src/js/auth.js
@@ -1,16 +1,24 @@
-import { extend, pick } from 'underscore';
-import { createAuth0Client } from '@auth0/auth0-spa-js';
+import { extend } from 'underscore';
 import Radio from 'backbone.radio';
+import { createAuth0Client } from '@auth0/auth0-spa-js';
 
-import { auth0Config as config } from './config';
+import { auth0Config as config, appConfig } from './config';
 
 import 'scss/app-root.scss';
 
 import { LoginPromptView } from 'js/views/globals/prelogin/prelogin_views';
 
-let auth0;
-const rwConnection = 'google-oauth2';
 const RWELL_KEY = 'rw';
+const RWELL_CONNECTION = 'google-oauth2';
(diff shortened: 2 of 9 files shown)


src/assets/appconfig.json
```diff
@@ -1,17 +1,19 @@
...
-    "stack_id": "localhost"
+    "stackId": "localhost",
+    "name": "Salvation Clinic"
   },
   "auth0": {
     "domain": "auth.roundingwell.com",
     "clientId": "UVGIL27XqwGsJwsTfEyz50c1qg3Ll1p6",
-    "connection": "google-oauth2",
-    "name": "Salvation Clinic"
+    "authorizationParams": {
```
paulfalgout, line 10: this nesting matches the auth0 expected config format

src/js/auth.js
```diff
@@ -29,24 +37,53 @@ function getToken() {
     });
 }
 
+/*
+ * Modifies the current history state
+ */
+function replaceState(state) {
+  window.history.replaceState({}, document.title, state);
```
paulfalgout, line 44: we were doing this same `history.replaceState` in a few places, but this makes it clearer what it's doing in context

src/js/auth.js
```diff
@@ -55,67 +92,56 @@ function authenticate(success) {
...
-
-    if (location.pathname === '/login') {
-      window.history.replaceState({}, document.title, '/');
-    }
-
-    success({ name: config.name });
-  });
+  createAuth0Client(getConfig())
+    .then(setAuth0)
+    .then(isAuthenticated => {
```
paulfalgout, line 108: nearly all of the changes in this function are from spacing changes from moving the `.then` to the next line

src/js/auth.js
```diff
@@ -55,67 +92,56 @@ function authenticate(success) {
...
+      if (location.pathname === '/logout') {
+        const federated = Radio.request('bootstrap', 'setting', 'federated_logout');
```
paulfalgout, line 110: Moved the `federated` setting to the users action of logout instead of doing it automatically in any logout..

src/js/auth.js
```diff
@@ -148,9 +174,12 @@ function forceLogin(appState = '/') {
   loginPromptView.render();
 }
 
-export {
+Radio.reply('auth', {
```
paulfalgout, line 177: moving the exports to `Radio` prevents this file from being included in more places as now it's only included by the `import()`

src/js/index.js
```diff
@@ -44,61 +43,49 @@ function startFormService() {
...
     return;
   }
 
+  const isForm = rootRoute === 'formapp';
+  const isOutreach = rootRoute === 'outreach';
+
   if ((_DEVELOP_ || _E2E_) && sessionStorage.getItem('cypress')) {
     versions.frontend = 'cypress';
+    appConfig.name = 'Cypress Clinic';
+    appConfig.cypress = sessionStorage.getItem('cypress');
```
paulfalgout, line 80: setting these things on the `appConfig` allowed less passing around of stuff and more standardization.

src/js/outreach/index.js
```diff
@@ -220,6 +219,17 @@ function startOutreachApp() {
   Backbone.history.start({ pushState: true });
 }
 
+let token;
+
+Radio.reply('auth', {
```
paulfalgout, line 224: Add `auth` Radio to `outreach` as now `js/auth` isn't in this js bundle at all, _but_ the `getToken` in the base fetch _will_ use this `Radio` for this particular bundle, but the `auth.js` `Radio` for the main bundle.

src/js/services/bootstrap.js
```diff
@@ -77,6 +77,8 @@ export default App.extend({
     return this.workspaces.clone();
   },
   getSetting(settingName) {
+    /* istanbul ignore if: difficult to test settings prior to bootstrap */
```
paulfalgout, line 80: we _could_ test this, but this would occur if we could force a logout prior to the initial bootstrap..  it's possible, but it's not easy and this particular scenario would break our coverage reports..  so we'd have to manufacture some other org setting to test it.. not impossible, but seems like overkill.

nmajor25 (approved):

PR #1043, opened 2023-05-03
Enable/disable uploading attachments per program action

nmajor25:
Shortcut Story ID: [sc-33784]

Adds a toggle on the program action sidebar between enabling ([screenshot](https://user-images.githubusercontent.com/35355575/236052819-54d12182-bc3b-40f8-aa22-6954a62b1dc7.png)) and disabling ([screenshot](https://user-images.githubusercontent.com/35355575/236052724-8915038b-6fc2-4ff2-a9f7-78699421d19e.png)) the ability to upload attachments on that specific action.

Enabling sets `action.attributes.uploads_allowed: ['pdf']` and disabling sets `action.attributes.uploads_allowed: []`.

If the `upload_attachments` org setting is `false` or `undefined`, the entire section will be hidden.

The patient action sidebar was also updated to not show the upload attachment button if uploads are not set to be allowed for the corresponding program action (i.e. equal to `uploads_allowed: []`).

### DIFF
--- a/src/js/apps/programs/sidebar/action-sidebar_app.js
+++ b/src/js/apps/programs/sidebar/action-sidebar_app.js
@@ -1,10 +1,11 @@
 import Radio from 'backbone.radio';
+import { size } from 'underscore';
 
 import { ACTION_OUTREACH } from 'js/static';
 
 import App from 'js/base/app';
 
-import { LayoutView, FormSharingButtonView, FormSharingView } from 'js/views/programs/sidebar/action/action-sidebar_views';
+import { LayoutView, FormSharingButtonView, FormSharingView, UploadsEnabledView } from 'js/views/programs/sidebar/action/action-sidebar_views';
 
 export default App.extend({
   beforeStart() {
@@ -19,8 +20,28 @@ export default App.extend({
     }));
 
     this.showFormSharing();
-
     this.listenTo(this.action, 'change:_form change:outreach', this.showFormSharing);
+
+    this.showUploadsEnabled();
+    this.listenTo(this.action, 'change:allowed_uploads', this.showUploadsEnabled);
+  },
+  showUploadsEnabled() {
+    if (!Radio.request('bootstrap', 'setting', 'upload_attachments')) return;
+
+    const uploadsEnabledView = new UploadsEnabledView({
+      isUploadsEnabled: !!size(this.action.get('allowed_uploads')),
+      isButtonDisabled: this.action.isNew(),
+    });
+
+    this.listenTo(uploadsEnabledView, 'click:enable', () => {
+      this.action.enableAttachmentUploads();
+    });
+
+    this.listenTo(uploadsEnabledView, 'click:disable', () => {
+      this.action.disableAttachmentUploads();
+    });
+
+    this.showChildView('allowUploads', uploadsEnabledView);
   },
   showFormSharing() {
     if (!Radio.request('bootstrap', 'setting', 'care_team_outreach')) return;
--- a/src/js/entities-service/entities/program-actions.js
+++ b/src/js/entities-service/entities/program-actions.js
@@ -57,6 +57,12 @@ const _Model = BaseModel.extend({
       _program_action: this.id,
     });
   },
+  enableAttachmentUploads() {
+    this.save({ allowed_uploads: ['pdf'] });
+  },
+  disableAttachmentUploads() {
+    this.save({ allowed_uploads: [] });
+  },
   getOwner() {
     const owner = this.get('_owner');
     if (!owner) return;
--- a/src/js/views/programs/sidebar/action/action-sidebar.hbs
+++ b/src/js/views/programs/sidebar/action/action-sidebar.hbs
@@ -17,7 +17,8 @@
     <div class="flex u-margin--t-8"><h4 class="sidebar__label u-margin--t-8">{{ @intl.programs.sidebar.action.actionSidebarTemplate.ownerLabel }}</h4><div class="flex-grow" data-owner-region></div></div>
     <div class="flex u-margin--t-8"><h4 class="sidebar__label u-margin--t-8">{{ @intl.programs.sidebar.action.actionSidebarTemplate.dueDayLabel }}</h4><div class="flex-grow" data-due-region></div></div>
     <div class="flex u-margin--t-8"><h4 class="sidebar__label u-margin--t-8">{{ @intl.programs.sidebar.action.actionSidebarTemplate.formLabel }}</h4><div class="flex-grow" data-form-region></div></div>
-    <div class="flex u-margin--t-24" data-form-sharing-region></div>
+    <div class="u-margin--t-24" data-form-sharing-region></div>
+    <div class="u-margin--t-24" data-allow-uploads-region></div>
     {{#if canTag}}
     <div class="sidebar__section">
       <div><h3 class="sidebar__section-label">{{ @intl.programs.sidebar.action.actionSidebarTemplate.tagsLabel }}</h3></div>
--- a/src/js/views/programs/sidebar/action/action-sidebar_views.js
+++ b/src/js/views/programs/sidebar/action/action-sidebar_views.js
@@ -113,7 +113,7 @@ const HeadingView = View.extend({
 
 const FormSharingButtonView = View.extend({
   tagName: 'button',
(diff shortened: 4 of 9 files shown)


src/scss/modules/sidebar.scss
```diff
@@ -53,7 +53,7 @@
 .sidebar__attachments {
   border-bottom: 1px solid $black-90;
   border-top: 1px solid $black-90;
-  padding: 16px 0 24px;
+  padding: 16px 0;
```
nmajor25, line 56: Not entirely sure why this had the `padding-bottom: 24px` before...

But both the program action sidebar and action sidebar looks good to me with `padding-bottom: 16px`:

Screenshot 2023-05-03 at 4 07 38 PM
Screenshot 2023-05-03 at 4 10 46 PM

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -1037,6 +1048,64 @@ context('action sidebar', function() {
...
+      .children()
+      .should('have.length', 1);
+
+    cy
+      .get('.sidebar')
+      .find('[data-attachments-region]')
+      .find('.js-add')
+      .should('not.exist');
+  });
+
```
nmajor25, line 1108: I need to test these two scenarios:

1. Upload attachment button is hidden because the `upload_attachments` org setting is set to `false` (or doesn't exist).
2. Upload attachment button is hidden because the `allowed_uploads: []` is empty or non-existent for the action.

These can't be tested together in the same test. Since if one or the either is `false`, the button will be hidden.

So I split them into two separate ones.

src/js/apps/programs/sidebar/action-sidebar_app.js
```diff
@@ -19,8 +20,28 @@ export default App.extend({
...
     this.listenTo(this.action, 'change:_form change:outreach', this.showFormSharing);
+
+    this.showUploadsEnabled();
+    this.listenTo(this.action, 'change:allowed_uploads', this.showUploadsEnabled);
+  },
+  showUploadsEnabled() {
+    if (!Radio.request('bootstrap', 'setting', 'upload_attachments')) return;
+
+    const uploadsEnabledView = new UploadsEnabledView({
+      uploadsEnabled: !!size(this.action.get('allowed_uploads')),
```
nmajor25: `size()` protects against `allowed_uploads` being `undefined`. Do we need to protect against that? Or can we assume an action will always have that attribute (`allowed_uploads: []`).

nmajor25:
We enforce what kind of file extensions are accepted when uploading an attachment [here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/views/patients/sidebar/action/action-sidebar-attachments-views.js#L98).

Do we need to update that to be whatever `action.attributes.allowed_uploads` is? Or can we just keep it hardcoded as `.pdf` for now?

src/js/apps/programs/sidebar/action-sidebar_app.js
nmajor25: Otherwise we could just do `this.action.get('allowed_uploads').length`.

src/js/entities-service/entities/program-actions.js
```diff
@@ -57,6 +57,12 @@ const _Model = BaseModel.extend({
       _program_action: this.id,
     });
   },
+  enableAttachmentUploads() {
+    this.save({ allowed_uploads: ['pdf'] });
```
nmajor25, line 61: @shadowhand for now, are we good just hardcoding this as only PDF?

src/js/apps/programs/sidebar/action-sidebar_app.js
paulfalgout: I think this should be `isUploadsEnabled`

src/js/views/programs/sidebar/action/action-sidebar_views.js
```diff
@@ -125,6 +125,34 @@ const FormSharingButtonView = View.extend({
...
+  template: hbs`
+    {{#if uploadsEnabled}}
+      <div class="flex sidebar__attachments">
+        <h3 class="flex-grow sidebar__heading">
+          {{far "paperclip"}}<span class="u-margin--l-8">{{ @intl.programs.sidebar.action.actionSidebarViews.uploadsEnabledView.attachmentsHeadingText }}</span>
+        </h3>
+        <button class="button--link js-disable">{{ @intl.programs.sidebar.action.actionSidebarViews.uploadsEnabledView.disableButtonText }}</button>
+      </div>
+    {{else}}
+      <button {{#if isButtonDisabled}}disabled{{/if}} class="button--blue w-100 u-text-align--center js-enable">
```
paulfalgout: isn't `u-text-align--center` the default?

src/js/views/programs/sidebar/action/action-sidebar_views.js
```diff
@@ -125,6 +125,34 @@ const FormSharingButtonView = View.extend({
   template: hbs`{{far "share-from-square"}} {{ @intl.programs.sidebar.action.actionSidebarViews.formSharingButtonView.buttonText }}`,
 });
 
+const UploadsEnabledView = View.extend({
+  className: 'w-100',
```
paulfalgout: this'll be a `div` by default so is the `w-100` necessary?

src/js/apps/programs/sidebar/action-sidebar_app.js
nmajor25: Ah yep, updated to that.

paulfalgout (changes requested):

src/js/views/programs/sidebar/action/action-sidebar_views.js
nmajor25: This is happens if it's removed: [screenshot](https://user-images.githubusercontent.com/35355575/236265615-57f5687f-7a37-4c12-b1b9-6b48878707a5.png).

The HTML structure is like this without it:

```html
<div class="flex u-margin--t-24" data-allow-uploads-region>
  <div>
    <button class="button--blue w-100 js-enable">
      <svg></svg>
    </button>
  </div>
</div>
```

`<div class="flex u-margin--t-24" data-allow-uploads-region>` has a `width: 100%`. But the empty div doesn't.

So maybe this could be structured differently?? But with the current structure I think we need it.

src/js/views/programs/sidebar/action/action-sidebar_views.js
nmajor25: Yup it is the default. I just removed that.

I copied it over from the `FormSharingButtonView` button. I removed `u-text-align--center` from that button as well.

src/js/views/programs/sidebar/action/action-sidebar_views.js
paulfalgout: is the `flex` needed on the region?

nmajor25: It doesn't look like it's needed... I just removed that and the `w-100`.

src/js/views/programs/sidebar/action/action-sidebar.hbs
```diff
@@ -17,7 +17,8 @@
     <div class="flex u-margin--t-8"><h4 class="sidebar__label u-margin--t-8">{{ @intl.programs.sidebar.action.actionSidebarTemplate.ownerLabel }}</h4><div class="flex-grow" data-owner-region></div></div>
     <div class="flex u-margin--t-8"><h4 class="sidebar__label u-margin--t-8">{{ @intl.programs.sidebar.action.actionSidebarTemplate.dueDayLabel }}</h4><div class="flex-grow" data-due-region></div></div>
     <div class="flex u-margin--t-8"><h4 class="sidebar__label u-margin--t-8">{{ @intl.programs.sidebar.action.actionSidebarTemplate.formLabel }}</h4><div class="flex-grow" data-form-region></div></div>
-    <div class="flex u-margin--t-24" data-form-sharing-region></div>
+    <div class="u-margin--t-24" data-form-sharing-region></div>
+    <div class="u-margin--t-24" data-allow-uploads-region></div>
```
nmajor25, line 21: I removed `flex` from the `data-form-sharing-region` region as well.

Everything looks the same without it.

paulfalgout (approved):

PR #1049, opened 2023-05-10
On the patient action sidebar, retrieve `allowed_uploads` from its related program action

nmajor25:
Shortcut Story ID: [sc-33784]

When this feature was built in #1043, FE was trying to get `action.attributes.allowed_uploads` on the patient action sidebar. 

But `allowed_uploads: []` is not stored on the patient action. It's stored on it's corresponding program action. So we need to get it from `action.included.program_action.allowed_uploads` instead.

### DIFF
--- a/test/integration/patients/sidebar/action-sidebar.js
+++ b/test/integration/patients/sidebar/action-sidebar.js
@@ -279,7 +279,6 @@ context('action sidebar', function() {
       attributes: {
         name: 'Name',
         details: 'Details',
-        allowed_uploads: ['pdf'],
         duration: 5,
         due_date: testDateSubtract(2),
         due_time: null,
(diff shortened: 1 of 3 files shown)


test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -699,13 +698,7 @@ context('action sidebar', function() {
     cy
       .get('.sidebar')
       .find('[data-attachments-region]')
-      .contains('No Attachments');
-
-    cy
-      .get('.sidebar')
-      .find('[data-attachments-region]')
-      .find('.js-add')
-      .should('exist');
```
nmajor25, line 708: The action used in this test is considered [adHoc](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/entities-service/entities/actions.js#L86-L88). Therefore, if I add a `program_action` relationship to the action used in this test, it breaks anything related to the name input field in this test.

So I figured it'd be easer if I just moved this check to the `display action from program action` test.

paulfalgout (approved):

PR #1050, opened 2023-05-11
Create new outreach opt-in page and views

nmajor25:
Shortcut Story ID: [sc-36527]

Initial version of the app views for outreach opt-in.

### DIFF
--- a/src/js/outreach/apps/opt-in_app.js
+++ b/src/js/outreach/apps/opt-in_app.js
@@ -1,0 +1,64 @@
+import Backbone from 'backbone';
+import App from 'js/base/app';
+
+import { optInPostRequest } from 'js/outreach/entities';
+
+import {
+  OptInView,
+  ResponseSuccessView,
+  ResponseErrorView,
+} from 'js/outreach/views/opt-in_views';
+
+import {
+  DialogView,
+} from 'js/outreach/views/dialog_views';
+
+const StateModel = Backbone.Model.extend({
+  defaults: {
+    firstName: '',
+    lastName: '',
+    dob: '',
+    phone: '',
+    email: '',
+  },
+});
+
+export default App.extend({
+  StateModel,
+  onStart() {
+    const dialogView = new DialogView();
+    this.showView(dialogView);
+
+    this.showOptInView();
+  },
+  showOptInView() {
+    const optInView = new OptInView({ model: this.getState() });
+
+    this.listenTo(optInView, 'click:submit', () => {
+      optInPostRequest({
+        inputData: this.getState(),
+      })
+        .then(() => {
+          this.showResponseSuccessView();
+        })
+        .catch(() => {
+          this.showResponseErrorView();
+        });
+    });
+
+    this.showChildView('content', optInView);
+  },
+  showResponseSuccessView() {
+    const responseSuccessView = new ResponseSuccessView();
+    this.showChildView('content', responseSuccessView);
+  },
+  showResponseErrorView() {
+    const responseErrorView = new ResponseErrorView();
+
+    this.listenTo(responseErrorView, 'click:tryAgain', () => {
+      this.showOptInView();
+    });
+
+    this.showChildView('content', responseErrorView);
+  },
+});
--- a/src/js/outreach/views/dialog_views.js
+++ b/src/js/outreach/views/dialog_views.js
@@ -1,7 +1,7 @@
 import hbs from 'handlebars-inline-precompile';
 import { View } from 'marionette';
 
-import { auth0Config } from 'js/config';
+import { appConfig } from 'js/config';
 
 import './dialog.scss';
 
@@ -15,7 +15,7 @@ const DialogView = View.extend({
   `,
   templateContext() {
     return {
-      name: auth0Config.name,
+      name: appConfig.name,
     };
   },
   regions: {
--- a/src/js/outreach/views/opt-in_views.js
+++ b/src/js/outreach/views/opt-in_views.js
@@ -1,0 +1,153 @@
+import hbs from 'handlebars-inline-precompile';
+import { View } from 'marionette';
+
+import 'scss/modules/buttons.scss';
+import 'scss/modules/forms.scss';
+
+import './opt-in.scss';
+
+const OptInView = View.extend({
+  template: hbs`
+    <div class="opt-in__icon opt-in__icon--warn">{{fat "hand-wave"}}</div>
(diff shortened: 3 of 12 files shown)


src/js/outreach/apps/opt-in_app.js
```diff
@@ -0,0 +1,68 @@
...
+      })
+        .then(() => {
+          this.showSuccessView();
+        })
+        .catch(() => {
+          this.showErrorView();
+        });
+    });
+
+    this.dialogViewRegion.show(optInView);
```
paulfalgout: since `dialogView` was used in `showView` this is the same as `this.showChildView('content', optInView)`

src/js/outreach/views/login_views.js
```diff
@@ -19,9 +19,22 @@ const LoginView = View.extend({
   template: hbs`
     <div class="dialog__icon">{{far "lock-keyhole"}}</div>
     <div>Enter your date of birth to access this form.</div>
-    <div><input type="date" class="js-date dialog__input{{#if hasError}} has-error{{/if}}" required pattern="\d{4}-\d{2}-\d{2}" placeholder="Your Date of Birth" value="{{ dob }}"></div>
-    {{#if hasError}}<div class="dialog__error">That date of birth does not match our records. Please try again.</div>{{/if}}
-    <div><button class="button--green dialog__button js-submit" {{#unless dob}}disabled{{/unless}}>Continue to Form {{fas "right-to-bracket"}}</button></div>
+    <div>
```
paulfalgout: don't need to change this right now

paulfalgout (changes requested):
Yeah I think this is essentially it

src/js/outreach/apps/opt-in_app.js
nmajor25: Good call, I updated it to that.

src/js/outreach/views/login_views.js
nmajor25: Oh forgot to change that back. I added spacing and line breaks so I could read the code easier when I was copying stuff from it.

src/js/outreach/views/dialog_views.js
```diff
@@ -15,7 +15,7 @@ const DialogView = View.extend({
   `,
   templateContext() {
     return {
-      name: auth0Config.name,
+      name: appConfig.name,
```
nmajor25, line 18: Use this instead of `auth0Config.name`? I don't know the reason for using that originally.

And I couldn't get `auth0Config.name` to work in cypress.

src/js/outreach/views/opt-in_views.js
```diff
@@ -0,0 +1,153 @@
...
+      return;
+    }
+
+    if (!phone && !email) {
+      this.disableSubmitButton();
+      return;
+    }
+
+    this.enableSubmitButton();
+  },
```
nmajor25, line 125: Handles whether the submit button should have the `disabled` property or not.

The submit button is enabled if the following are true:

* The user has entered their first name, last name, and date of birth.
* The user has entered at least one communication method. Either email address or phone number.

src/js/outreach/views/opt-in_views.js
```diff
@@ -0,0 +1,153 @@
...
+  onClickSubmit() {
+    this.disableSubmitButton();
+  },
```
nmajor25, line 128: Stops a user from double submitting the API request.

src/js/outreach/views/opt-in_views.js
```diff
@@ -0,0 +1,153 @@
...
+    phone: '.js-phone',
+    email: '.js-email',
+    submit: '.js-submit',
+  },
+  triggers: {
+    'keyup @ui.firstName': 'change:first:name',
+    'keyup @ui.lastName': 'change:last:name',
+    'keyup @ui.dob': 'change:dob',
+    'keyup @ui.phone': 'change:phone',
+    'keyup @ui.email': 'change:email',
```
nmajor25: I switched to using `keyup` here (instead of `onChange`).

With `keyup`, we can update the submit button state (disabled vs. enabled) as the user is typing. With `onChange`, we'd have to wait until the user un-focused the input field.

src/js/outreach/apps/opt-in_app.js
```diff
@@ -0,0 +1,64 @@
...
+    this.showChildView('content', optInView);
+  },
+  showResponseSuccessView() {
+    const responseSuccessView = new ResponseSuccessView();
+    this.showChildView('content', responseSuccessView);
+  },
+  showResponseErrorView() {
+    const responseErrorView = new ResponseErrorView();
+
+    this.listenTo(responseErrorView, 'click:try:again', () => {
```
paulfalgout: we should change this to `click:tryAgain` where events are `context:event:subject`

src/js/outreach/views/dialog_views.js
paulfalgout: oh yeah this changed recently

src/js/outreach/views/opt-in_views.js
paulfalgout: what about the `input` event?

paulfalgout (changes requested):
small stuff.  we could potentially just merge this and do any mark changes in a subsequent PR

src/js/outreach/apps/opt-in_app.js
nmajor25: Updated in a recent push.

src/js/outreach/views/opt-in_views.js
nmajor25: `input` is preferred, not sure why I originally used `keyup` instead.

src/js/outreach/views/opt-in_views.js
```diff
@@ -0,0 +1,153 @@
...
+        type="text"
+        class="input-primary opt-in__field-input js-last-name"
+        placeholder="Enter your last name"
+        value="{{ lastName }}"
+      />
+    </div>
+    <div class="opt-in__field">
+      <label class="opt-in__field-label">Your date of birth</label>
+      <input
+        type="text"
```
paulfalgout: I think we should make this the date input just like the dob on the outreach login.

paulfalgout (changes requested):

src/js/outreach/views/opt-in_views.js
nmajor25: I just updated it to use `type="date"`.

paulfalgout (approved):

PR #1053, opened 2023-05-15
On worklist/schedule, display total actions/flows count in the capped results message

nmajor25:
Shortcut Story ID: [sc-36700]

For example, display `Showing 500 of 600 actions.` instead of `Showing 500 of many actions.`, where `600` represents the total number of actions/flows stored in the database.

That total data is found via `actions.meta.total`/`flows.meta.total`. Added by backend here: https://github.com/RoundingWell/care-ops-backend/pull/5539

### DIFF
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -425,7 +425,7 @@ careOptsFrontend:
         countView:
           actionsCount: "{itemCount, plural, one {# Action} other {# Actions}}"
           flowsCount: "{itemCount, plural, one {# Flow} other {# Flows}}"
-          maximumListCount: Showing {maximumCount} of many {isFlowList, select, true {Flows} false {Actions}}.
+          maximumListCount: Showing {maximumCount} of {totalInDb, number} {isFlowList, select, true {Flows} false {Actions}}.
           maximumListCountNarrowed: Showing {itemCount} of {maximumCount} {isFlowList, select, true {Flows} false {Actions}}.
           narrowFilters: Try narrowing your filters.
     sidebar:
--- a/src/js/views/patients/shared/list_views.js
+++ b/src/js/views/patients/shared/list_views.js
@@ -1,8 +1,6 @@
 import hbs from 'handlebars-inline-precompile';
 import { View } from 'marionette';
 
-import { MAXIMUM_LIST_COUNT } from 'js/static';
-
 const ListCountTemplate = hbs`
   <strong>
     {{#if isFlowList}}
@@ -14,7 +12,7 @@ const ListCountTemplate = hbs`
 `;
 
 const MaximumCountTemplate = hbs`
-  <div>{{formatMessage (intlGet "patients.shared.listViews.countView.maximumListCount") maximumCount=maximumCount isFlowList=isFlowList}}</div>
+  <div>{{formatMessage (intlGet "patients.shared.listViews.countView.maximumListCount") maximumCount=maximumCount totalInDb=totalInDb isFlowList=isFlowList}}</div>
   <div>{{ @intl.patients.shared.listViews.countView.narrowFilters }}</div>
 `;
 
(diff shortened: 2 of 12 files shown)


src/js/views/patients/shared/list_views.js
```diff
@@ -51,6 +51,7 @@ const CountView = View.extend({
       maximumCount: MAXIMUM_LIST_COUNT,
       count: filteredCollection.length,
       isFlowList: !!this.getOption('isFlowList'),
+      totalInDb: this.getOption('totalInDb').toLocaleString(),
```
nmajor25: Added `.toLocaleString()` here to add thousand comma separators to this number. Since these have the potential to be over `1,000`? What do you think?

test/support/api/flows.js
```diff
@@ -101,7 +101,11 @@ Cypress.Commands.add('routeFlows', (mutator = _.identity) => {
   cy.route({
     url: '/api/flows?*',
     response() {
-      return mutator(generateData.call(this));
+      const apiData = mutator(generateData.call(this));
+
+      if (!apiData.meta) apiData.meta = { flows: { total: apiData.data.length } };
+
+      return apiData;
```
nmajor25, line 108: I wasn't sure the best way to add these `meta: {}` objects.

This allows a test to set a specific `meta.flows.total` value. Or a default one will be set based on`apiData.data.length`.

My initial solution was to add this to the `generateData()` function:

```js
function generateData(patients = _.sample(this.fxPatients, 5)) {
  const data = getResource(_.sample(this.fxFlows, 10), 'flows');

  . . .

  const meta = {
    flows: {
      total: 10
    }
  }

  return {
    data,
    included,
    meta,
  };
}
```

But then `meta.flows.total: 10` wouldn't be correct for tests where the number of flows is modified via the `mutator()`. Maybe that doesn't really matter that much. Since it wouldn't result in any errors or broken tests.

src/js/views/patients/shared/list_views.js
paulfalgout: actually I think in the i18n `maximumListCount` you just do `{# totalInDb}`

I dunno if we've ever done it, but https://github.com/formatjs/handlebars-intl#example

test/support/api/flows.js
paulfalgout: seems ok to me

paulfalgout (changes requested):
Cool.  Just the one handlebars-intl suggestion.. which I assume works.. and I think we're set.

paulfalgout (changes requested):
Cool.  Just the one handlebars-intl suggestion.. which I assume works.. and I think we're set.

src/js/i18n/en-US.yml
```diff
@@ -425,7 +425,7 @@ careOptsFrontend:
         countView:
           actionsCount: "{itemCount, plural, one {# Action} other {# Actions}}"
           flowsCount: "{itemCount, plural, one {# Flow} other {# Flows}}"
-          maximumListCount: Showing {maximumCount} of many {isFlowList, select, true {Flows} false {Actions}}.
+          maximumListCount: Showing {maximumCount} of {totalInDb} {isFlowList, select, true {Flows} false {Actions}}.
```
paulfalgout: Had another thought.. 
seems like we can remove `MAXIMUM_LIST_COUNT`. This message just shows up if the `count < total` and then the narrowed version only shows if `count < total && filteredCount < count` correct?

paulfalgout (changes requested):

src/js/views/patients/shared/list_views.js
```diff
@@ -48,9 +47,10 @@ const CountView = View.extend({
     const filteredCollection = this.getOption('filteredCollection');
 
     return {
-      maximumCount: MAXIMUM_LIST_COUNT,
+      maximumCount: this.collection.length,
```
nmajor25, line 50: Technically, this will be incorrect when the list hasn't reached the maximum. But it won't be used in those instances in any hbs template. Only when the list has reached the maximum.

paulfalgout (approved):

PR #1056, opened 2023-05-18
Add outreach UI for verifying users via a texted and/or emailed code

nmajor25:
Shortcut Story ID: [sc-36767]

Before a user can view the outreach form, they will need to request and enter a verification code that's sent via text or email.


src/js/outreach/views/verify_views.js
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        if (didInsertContent && !isLastInputEl) {
+          inputElements[index + 1].focus();
+          inputElements[index + 1].value = rest.join('');
+          inputElements[index + 1].dispatchEvent(new Event('input'));
+        }
+      });
+    });
+  },
```
nmajor25, line 104: When one input is entered, it will jump to the next one. And it also allows a code to be copy-pasted and you can backspace between the inputs.

Screen recording of the input functionality: 

https://github.com/RoundingWell/care-ops-frontend/assets/35355575/039cc118-d04b-48cf-a89c-f7930f3ad079

I couldn't get this to work via the Marionette.js `triggers: {}`, so need to investigate that more.

src/js/outreach/views/dialog.scss
```diff
@@ -20,6 +20,7 @@
   margin: 0 auto;
   max-width: 520px;
   padding: 48px 32px;
+  text-align: left;
```
nmajor25, line 23: I need to go through this file and remove any un-used code. And I could probably consolidate some styles that are identical between the `verify.scss` and `opt-in.scss` files.

nmajor25: Also need to update the error views ([example](https://github.com/RoundingWell/care-ops-frontend/assets/35355575/4a081a7e-7adb-4a8d-91d3-4fce3edf871a)) to use the new icons and spacing from the design docs.

src/js/outreach/index.js
```diff
@@ -6,7 +6,7 @@ import 'js/entities-service/forms';
 
 import RouterApp from 'js/base/routerapp';
 
-import LoginApp from 'js/outreach/apps/login_app';
+import VerifyApp from 'js/outreach/apps/verify_app';
```
nmajor25, line 9: Renamed everything from `login` => `verify`.

test/integration/outreach/outreach.js
```diff
@@ -140,6 +140,143 @@ context('Outreach', function() {
...
+      .get('.verify__code-fields')
+      .find('.js-input')
+      .first()
+      .type('1')
+      .next()
+      .type('2')
+      .next()
+      .type('3')
+      .next()
+      .type('4');
```
nmajor25: Couldn't get the inputs to automatically move to the next one in the tests. Also couldn't get the copy-paste to work either. 

So still need to investigate that more.

paulfalgout:
Cool good start.  I think we can get the marionette triggers working, but I can check.. there are a handful of events that don't work well when proxying off a parent element..  But if we use marionette since that utilizes jquery we can the `cypress.get('foo').trigger('jquery:event')` for tests.. it's about the only useful thing jquery gives us right now :-D. Anyhow.. I think we can call this done for now..  shouldn't be a merge conflict issue to leave it open a bit, but we can just merge it if it moves the epic along too.. we'll see..

src/js/outreach/views/verify_views.js
nmajor25: Just updated this to use marionette.js to handle the `input` and `keydown` input events.

src/js/outreach/views/verify_views.js
```diff
@@ -0,0 +1,124 @@
...
+      currentElement.focus();
+      currentElement.val(rest.join(''));
+      currentElement.trigger('input');
+    }
+  },
+  watchKeydown(event) {
+    const inputElements = this.$('.js-input');
+    const index = inputElements.index(event.target);
+
+    if (event.keyCode === 8 && event.target.value === '') {
```
paulfalgout: there's a util where we name the keycode you can pull in here

src/js/outreach/views/verify_views.js
```diff
@@ -0,0 +1,124 @@
...
+      <input class="input-primary verify__code-input js-input" />
+      <input class="input-primary verify__code-input js-input" />
+      <input class="input-primary verify__code-input js-input" />
+      <input class="input-primary verify__code-input js-input" />
+    </div>
+    <button class="verify__submit button--green w-100 js-submit" disabled>Send Verification Code</button>
+    <div class="verify__heading-text u-text-link js-resend">Send a new code</div>
+  `,
+  watchInput(event) {
+    const inputElements = this.$('.js-input');
```
paulfalgout: `this.ui.input`

src/js/outreach/views/verify_views.js
```diff
@@ -0,0 +1,124 @@
...
+    const index = inputElements.index(event.target);
+
+    const [first, ...rest] = event.target.value;
```
paulfalgout: this is to get the first character correct?   
I'm on the fence on the destructuring.. makes it look like an array.. maybe string functions would be clearer?
```js
const value = String(event.target.value);
const first = event.target.value = value.charAt(0);
const rest = value.substr(1);
```

src/js/outreach/views/verify_views.js
```diff
@@ -0,0 +1,124 @@
...
+
+    event.target.value = first ?? '';
+
+    const isLastInputEl = index === inputElements.length - 1;
+    const didInsertContent = first !== undefined;
+
+    const code = inputElements.map((i, el) => el.value).get().join('');
+
+    if (code.length === inputElements.length) {
+      this.enableSubmitButton();
```
paulfalgout: I think we should probably make one function like `this.disableSubmitButton(code.length !== inputElements.length);`

src/js/outreach/views/verify_views.js
```diff
@@ -0,0 +1,124 @@
...
+    if (didInsertContent && !isLastInputEl) {
+      const currentElement = inputElements.eq(index + 1);
+
+      currentElement.focus();
+      currentElement.val(rest.join(''));
+      currentElement.trigger('input');
+    }
+  },
+  watchKeydown(event) {
+    const inputElements = this.$('.js-input');
```
paulfalgout: `this.ui.input`

src/js/outreach/views/verify_views.js
```diff
@@ -0,0 +1,124 @@
...
+    }
+  },
+  enableSubmitButton() {
+    this.ui.submit.prop('disabled', false);
+  },
+  disableSubmitButton() {
+    this.ui.submit.prop('disabled', true);
+  },
+  onClickSubmit() {
+    const inputElements = this.$('.js-input');
```
paulfalgout: `this.ui` here too

paulfalgout (changes requested):

src/js/outreach/apps/verify_app.js
```diff
@@ -15,19 +15,26 @@ import {
...
 
 export default App.extend({
-  onStart({ actionId }) {
-    this.actionId = actionId;
+  beforeStart({ actionId }) {
+    return getPatientInfo({ actionId });
+  },
+  onStart(options, patient) {
+    this.patientPhoneEnd = patient.phone_end;
+    this.patientId = patient.relationships.patient.data.id;
```
nmajor25: I'm assuming there's a better way to handle this data... so need to look into that.

paulfalgout: should probably use `get`s.  `phone_end` will also be on `attributes` as this'll be a fully jsonapi.  I'm not sure if this is a now thing, but we're coming up to being close to needing to setup an outreach entities using the backbone models.

paulfalgout (changes requested):

src/js/outreach/apps/verify_app.js
nmajor25: I just updated to use `get`.

I'm down to come back after this is deployed and switch over to use backbone models.

src/js/outreach/views/verify_views.js
paulfalgout: @nmajor25 ping on these too

src/js/outreach/views/verify_views.js
nmajor25: Good call, updated this to do `event.keyCode === BACKSPACE_KEY`.

src/js/outreach/views/verify_views.js
nmajor25: Updated.

src/js/outreach/views/verify_views.js
nmajor25: Updated this to not do the destructuring.

src/js/outreach/views/verify_views.js
nmajor25: That works, updated to that.

src/js/outreach/views/verify_views.js
nmajor25: Updated.

src/js/outreach/views/verify_views.js
nmajor25: Updated.

src/js/outreach/views/opt-in_views.js
```diff
@@ -94,10 +83,6 @@ const OptInView = View.extend({
     this.model.set({ phone: this.ui.phone.val() });
```
paulfalgout: I think we need to run this though the phonenumber library for the US

paulfalgout (changes requested):

src/js/outreach/apps/verify_app.js
```diff
@@ -0,0 +1,90 @@
...
+        })
+        .catch(response => {
+          const status = response.status;
+
+          if (status === 409) {
+            this.showAlreadySubmittedView();
+            return;
+          }
+
+          if (status === 401 || status === 403 || status === 404) {
```
paulfalgout: I think all of these 400 codes were moved up to the `getPatientInfo` request such that before a code is requested/entered we're telling the user if it's not available.  So that can happen in an `onFail` which should get the same response object.  We can still catch errors here though for a bad code.

paulfalgout (changes requested):

src/js/outreach/apps/verify_app.js
```diff
@@ -20,6 +20,14 @@ export default App.extend({
   beforeStart({ actionId }) {
     return getPatientInfo({ actionId });
   },
+  onFail() {
```
paulfalgout: I think this needs:

```js
  if (status === 409) {
    this.showAlreadySubmittedView();
    return;
  }

  this.showNotAvailableView();
```
And the only state needed in the verification below is `this.showDialogErrorView();`

src/js/outreach/apps/verify_app.js
```diff
@@ -20,6 +20,14 @@ export default App.extend({
...
+    const dialogView = new DialogView();
+    this.showView(dialogView);
+
+    this.showNotAvailableView();
+
+    // this.stop();
```
nmajor25: Do we need the `this.stop()`? Other `onFail`s in the app have an `this.stop()`.

If I call `this.stop()` here, it starts the form app via this: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/outreach/index.js#L46-L49.

paulfalgout: No need to stop, the app is showing a view `onFail`

paulfalgout (changes requested):

src/js/outreach/views/opt-in_views.js
```diff
@@ -80,7 +81,12 @@ const OptInView = View.extend({
     this.setSubmitButtonState();
   },
   onChangePhone() {
-    this.model.set({ phone: this.ui.phone.val() });
+    const phone = parsePhoneNumber(this.ui.phone.val(), 'US');
+
+    this.model.set({
+      phone: phone ? phone.number : null,
```
nmajor25, line 87: The parsed `phone` will either equal `undefined` or be an object like this:

```js
"phone": {
  "country": "US",
  "countryCallingCode": "1",
  "nationalNumber": "8887771234",
  "number": "+18887771234"
}
```

paulfalgout: interesting.. so if the phone isn't formatted correctly the submit button will be disabled.. hmm..  we _might_ need to set the value on the model and validate it to show an error on submit or something..

src/js/outreach/apps/verify_app.js
nmajor25: Yup makes sense, just updated this.

src/js/outreach/views/opt-in_views.js
nmajor25: Agreed, seems better than just sending a `null` value to the backend.

`parsePhoneNumber` does seem to accept almost everything though in terms of formatting. The only time I can get it to be `undefined` is if I start entering non-numerical values.

nmajor25: Do we want to only allow numbers in that input?

paulfalgout: hmm people are going to type `-`s so it'd have to be numbers and dashes.. maybe `()`s

src/js/outreach/index.js
```diff
@@ -32,25 +32,25 @@ const OutreachApp = RouterApp.extend({
...
+    this.startVerify();
   },
-  startLogin() {
-    const loginApp = this.startCurrent('login', { actionId: this.actionId });
+  startVerify() {
+    const verifyApp = this.startCurrent('verify', { actionId: this.actionId });
 
-    this.listenTo(loginApp, 'stop', () => {
+    this.listenTo(verifyApp, 'stop', () => {
       // NOTE: Future non-authed stop if (!isAuthed) return;
```
paulfalgout, line 44: @nmajor25 I think this `if` likely fixes the issue.  user doesn't successfully auth so the app shouldn't load the form.

test/integration/outreach/outreach.js
```diff
@@ -308,6 +308,96 @@ context('Outreach', function() {
...
+
+    cy
+      .get('.verify__code-fields')
+      .find('.js-input.has-error')
+      .should('not.exist');
+
+    cy
+      .get('.verify__error-text')
+      .should('not.exist');
+  });
```
nmajor25, line 407: When the user enters an invalid OTP code:

1. The four input elements should have a red border.
2. The input fields should be cleared.
3. The `Incorrect verification code. Please try again.` error message should be shown.

~When the user types into one of the fields again, the red input border and error messaging should be removed.~

[Screenshot](https://github.com/RoundingWell/care-ops-frontend/assets/35355575/2937c913-536e-4947-b10b-4258a3cf41dc)

src/js/outreach/apps/verify_app.js
```diff
@@ -58,15 +58,23 @@ export default App.extend({
...
       validateVerificationCode({ patientId: this.patientId, code })
         .then(() => {
           this.stop({ isVerified: true });
         })
-        .catch(() => {
+        .catch(response => {
+          if (response.status === 403) {
+            this.showVerifyCodeView(true);
+            return;
+          }
```
nmajor25, line 76: Better way to do this?

If this request returns a `403` error, I reshow the `VerifyCodeView` view with `hasInvalidCodeError = true`.

src/js/outreach/views/verify_views.js
```diff
@@ -0,0 +1,146 @@
...
+    const inputElements = this.ui.input;
+    const index = inputElements.index(event.target);
+
+    if (event.keyCode === BACKSPACE_KEY && event.target.value === '') {
+      inputElements.eq(Math.max(0, index - 1)).focus();
+    }
+  },
+  removeErrorMessaging() {
+    this.ui.input.removeClass('has-error');
+    this.ui.errorMessage.remove();
```
paulfalgout: I don't love altering the DOM outside of the render.  This is also different behavior to other forms.  other forms keep the error state until the next submit.  Check the name input on the clinician sidebar.  Maybe we can bring this in line with those other form errors?

paulfalgout (changes requested):

src/js/outreach/views/verify_views.js
nmajor25: Okay, I've updated it so the red input border and error message stay visible.

paulfalgout (approved):

paulfalgout (approved):

PR #1057, opened 2023-05-18
Implement canEdit check for actions and flows

paulfalgout:
Shortcut Story ID: [sc-35582]

This includes a pattern for dealing with actions/flows that can and can't be edited regarding how they're displayed and selected via multi-select.

The read-only row item is only a POC

This is only for the worklist.. and these patterns would be applied to schedule and patient dashboard/flow

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -225,6 +225,22 @@ export default App.extend({
       },
     });
   },
+  showDeleteSuccess(itemCount) {
+    if (this.getState().isFlowType()) {
+      Radio.request('alert', 'show:success', renderTemplate(BulkDeleteFlowsSuccessTemplate, { itemCount }));
+      return;
+    }
+
+    Radio.request('alert', 'show:success', renderTemplate(BulkDeleteActionsSuccessTemplate, { itemCount }));
+  },
+  showUpdateSuccess(itemCount) {
+    if (this.getState().isFlowType()) {
+      Radio.request('alert', 'show:success', renderTemplate(BulkEditFlowsSuccessTemplate, { itemCount }));
+      return;
+    }
+
+    Radio.request('alert', 'show:success', renderTemplate(BulkEditActionsSuccessTemplate, { itemCount }));
+  },
   showDisabledSelectAll() {
     this.showChildView('selectAll', new SelectAllView({ isDisabled: true }));
   },
@@ -266,22 +282,6 @@ export default App.extend({
   getComparator() {
     const sortId = this.getState().getSort();
     return this.getSortOption(sortId).getComparator();
-  },
-  showDeleteSuccess(itemCount) {
-    if (this.getState().isFlowType()) {
-      Radio.request('alert', 'show:success', renderTemplate(BulkDeleteFlowsSuccessTemplate, { itemCount }));
-      return;
-    }
-
-    Radio.request('alert', 'show:success', renderTemplate(BulkDeleteActionsSuccessTemplate, { itemCount }));
-  },
-  showUpdateSuccess(itemCount) {
-    if (this.getState().isFlowType()) {
-      Radio.request('alert', 'show:success', renderTemplate(BulkEditFlowsSuccessTemplate, { itemCount }));
-      return;
-    }
-
-    Radio.request('alert', 'show:success', renderTemplate(BulkEditActionsSuccessTemplate, { itemCount }));
   },
   showCountView() {
     const countView = new CountView({
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -123,6 +123,7 @@ export default App.extend({
   showList() {
     const collectionView = new ListView({
       collection: this.collection,
+      editableCollection: this.editableCollection,
       state: this.getState(),
       viewComparator: this.getComparator(),
     });
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -1,4 +1,3 @@
-import { get } from 'underscore';
 import Radio from 'backbone.radio';
 
 import intl, { renderTemplate } from 'js/i18n';
@@ -13,15 +12,12 @@ import BulkEditFlowsApp from 'js/apps/patients/sidebar/bulk-edit-flows_app';
 
 import DateFilterComponent from 'js/views/patients/shared/components/date-filter';
 import SearchComponent from 'js/views/shared/components/list-search';
-import OwnerDroplist from 'js/views/patients/shared/components/owner_component';
 import { CountView } from 'js/views/patients/shared/list_views';
 
 import { getSortOptions } from './worklist_sort';
 
-import { ListView, SelectAllView, LayoutView, ListTitleView, TableHeaderView, SortDroplist, TypeToggleView, NoOwnerToggleView } from 'js/views/patients/worklist/worklist_views';
+import { ListView, SelectAllView, LayoutView, ListTitleView, TableHeaderView, SortDroplist, TypeToggleView } from 'js/views/patients/worklist/worklist_views';
 import { BulkEditButtonView, BulkEditFlowsSuccessTemplate, BulkEditActionsSuccessTemplate, BulkDeleteFlowsSuccessTemplate, BulkDeleteActionsSuccessTemplate } from 'js/views/patients/shared/bulk-edit/bulk-edit_views';
-
-const i18n = intl.patients.worklist.filtersApp;
 
 export default App.extend({
   StateModel,
@@ -38,10 +34,8 @@ export default App.extend({
   stateEvents: {
     'change:listType change:clinicianId change:teamId change:noOwner change:filters change:states': 'restart',
     'change:actionsDateFilters change:flowsDateFilters': 'restart',
-    'change:actionsSortId': 'onChangeStateSort',
-    'change:flowsSortId': 'onChangeStateSort',
-    'change:selectedFlows': 'onChangeSelected',
-    'change:selectedActions': 'onChangeSelected',
+    'change:actionsSortId change:flowsSortId': 'onChangeStateSort',
+    'change:actionsSelected change:flowsSelected': 'onChangeSelected',
     'change:searchQuery': 'onChangeSearchQuery',
   },
   onChangeStateSort() {
@@ -58,7 +52,7 @@ export default App.extend({
   initListState() {
     const storedState = this.getState().getStore(this.worklistId);
 
-    this.setState({ searchQuery: this.currentSearchQuery });
+    this.getState().setSearchQuery(this.currentSearchQuery);
 
     if (storedState) {
       this.setState(storedState);
@@ -73,19 +67,14 @@ export default App.extend({
     this.collection = null;
   },
   onBeforeStart({ worklistId }) {
-    const isFiltersSidebarOpen = this.getState('isFiltering');
-
     if (this.isRestarting()) {
+      const isFiltersSidebarOpen = this.getState('isFiltering');
       if (!isFiltersSidebarOpen) Radio.request('sidebar', 'close');
 
-      this.showListTitle();
-      this.showTypeToggleView();
-      this.showSortDroplist();
-      this.showTableHeaders();
-      this.showDateFilter();
-      this.getRegion('list').startPreloader();
-
       this.getRegion('count').empty();
+
+      this.showTypeViews();
+      this.getRegion('list').startPreloader();
 
       return;
     }
@@ -93,25 +82,16 @@ export default App.extend({
     this.worklistId = worklistId;
     this.initListState();
 
-    const currentClinician = Radio.request('bootstrap', 'currentUser');
-    this.shouldShowClinician = this.getState().id !== 'shared-by';
-    this.shouldShowTeam = this.getState().id !== 'owned-by';
-    this.canViewAssignedActions = currentClinician.can('app:worklist:clinician_filter');
+    this.setView(new LayoutView());
 
-    this.showView(new LayoutView({
-      worklistId: this.worklistId,
-      state: this.getState(),
-    }));
-
-    this.getRegion('list').startPreloader();
     this.showDisabledSelectAll();
-    this.showSortDroplist();
-    this.showTableHeaders();
-    this.showListTitle();
-    this.showTypeToggleView();
     this.showSearchView();
-    this.showDateFilter();
     this.startFiltersApp();
+
+    this.showTypeViews();
+    this.getRegion('list').startPreloader();
+
+    this.showView();
   },
   beforeStart() {
     return Radio.request('entities', `fetch:${ this.getState().getType() }:collection`, {
@@ -122,7 +102,22 @@ export default App.extend({
   onStart(options, collection) {
     this.collection = collection;
     this.filteredCollection = collection.clone();
+    this.editableCollection = collection.clone();
 
+    this.showCountView();
+    this.toggleBulkSelect();
+
+    this.showList();
+  },
+  // NOTE: Shows views dependent on getState().getType()
+  showTypeViews() {
+    this.showListTitle();
+    this.showTypeToggleView();
+    this.showDateFilter();
+    this.showSortDroplist();
+    this.showTableHeaders();
+  },
+  showList() {
     const collectionView = new ListView({
       collection: this.collection,
       state: this.getState(),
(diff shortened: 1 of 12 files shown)


src/js/views/patients/worklist/action_views.js
```diff
@@ -74,18 +74,24 @@ const ActionItemView = View.extend({
...
-    this.showState();
-    this.showOwner();
-    this.showDueDate();
-    this.showDueTime();
     this.showForm();
     this.showDetailsTooltip();
+
+    this.canEdit = this.model.canEdit();
+
+    if (this.canEdit) {
```
paulfalgout, line 87: So currently the idea would be if you can't edit it then we'll show the value of these components in the template..  I think otherwise we show a read-only version of the value which might be cheaper vs having to always render the values to be replaced..  _but_ I don't think we should make the components themselves handle this state.

src/js/views/patients/worklist/flow_views.js
```diff
@@ -98,7 +101,7 @@ const FlowItemView = View.extend({
     this.showChildView('check', checkComponent);
   },
   showState() {
-    if (!this.model.isDone()) {
+    if (!this.model.isDone() || !this.canEdit) {
```
paulfalgout, line 112: the flow already has a read only version of state.. so we can follow this pattern or this could be 
```js
if (!this.model.isDone() || !this.canEdit) return;
```
where the state is rendered inside the region in the template.

src/js/views/patients/worklist/action_views.js
```diff
@@ -74,18 +74,24 @@ const ActionItemView = View.extend({
...
   onRender() {
-    this.showCheck();
-    this.showState();
-    this.showOwner();
-    this.showDueDate();
-    this.showDueTime();
     this.showForm();
     this.showDetailsTooltip();
+
+    this.canEdit = this.model.canEdit();
```
paulfalgout, line 85: technically we only need to update this if the owner changes..  we might decide to do that in the future as a perf improvement for long lists...

src/js/views/patients/worklist/action-item.hbs
```diff
@@ -18,7 +18,7 @@
 </td>
 <td class="table-list__cell worklist-list__meta">
   <span class="worklist-list__display-icon u-margin--r-16" data-details-region></span>&#8203;{{~ remove_whitespace ~}}
-  <span class="table-list__meta" data-state-region></span>&#8203;<span class="worklist-list__search-helper">{{ state }}</span>&#8203;{{~ remove_whitespace ~}}
+  <span class="table-list__meta" data-state-region><span class="worklist-list__readonly action--{{ stateOptions.color }}">{{fa stateOptions.iconType stateOptions.icon}}</span></span>&#8203;<span class="worklist-list__search-helper">{{ state }}</span>&#8203;{{~ remove_whitespace ~}}
```
paulfalgout: so again as mentioned in the earlier commit.. this is rendered inside of, and will be replaced by the component if editable.  We can either do this, or move this to read-only views like the flow currently does..

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -73,9 +73,21 @@ export default Backbone.Model.extend({
   getStatesFilters() {
     return clone(this.get('states'));
   },
+  setSearchQuery(searchQuery = '') {
+    return this.set({
+      searchQuery: searchQuery.length > 2 ? searchQuery : '',
+      lastSelectedIndex: null,
```
paulfalgout, line 79: I moved all of the `lastSelectedIndex` management inside of the state model

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -38,8 +38,8 @@ export default Backbone.Model.extend({
       teamId: this.currentClinician.getTeam().id,
       noOwner: false,
       lastSelectedIndex: null,
-      selectedActions: {},
-      selectedFlows: {},
+      actionsSelected: {},
```
paulfalgout, line 41: renamed these to match ``` `${ type }DateFilters` ``` and ``` `${ type }SortId` ```

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -85,42 +97,6 @@ export default Backbone.Model.extend({
   setSort(sortId) {
     this.set(`${ this.getType() }SortId`, sortId);
   },
-  formatDateRange(date, rangeType, dateFormat) {
```
paulfalgout, line 88: simplified and moved this down nearer to the other date filter functions

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -138,30 +114,32 @@ export default Backbone.Model.extend({
...
+
+    if (selectedWeek) return this.formatDateRange(dateType, selectedWeek, 'week');
+
+    relativeDate = relativeDate || 'thismonth';
+    const { prev, unit } = relativeRanges.get(relativeDate).pick('prev', 'unit');
+    const relativeRange = dayjs().subtract(prev, unit);
+
+    return this.formatDateRange(dateType, relativeRange, unit);
+  },
+  getStaticDateFilter() {
```
paulfalgout, line 141: Changed this as we had a couple of checks for if the state id was `shared-by` or `owned-by` in relation to if a date is set, but really we want to know if it's already hard coded..  so this method is called within the app now and is used to look up the filter if it's not user-defined in `getEntityDateFilter`

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -138,30 +114,32 @@ export default Backbone.Model.extend({
...
-    return filter(selectedStates, id => contains(availableStateFilterIds, id)).join() || NIL_UUID;
-  },
   getFiltersState() {
     return {
       filters: this.getFilters(),
       states: this.getStatesFilters(),
       defaultStates: this.getDefaultSelectedStates(),
     };
   },
-  getEntityFilter() {
```
paulfalgout, line 154: This function was creating the filters for every worklist app every time instead of just the one that this state needs.  I also broke this up by filter-type

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -173,56 +151,92 @@ export default Backbone.Model.extend({
...
+    const filtersState = this.getFilters();
+    return reduce(filtersState, (filters, selected, slug) => {
+      if (selected !== null) filters[`@${ slug }`] = selected;
 
-      filters[this.id][`@${ slug }`] = selected;
-    });
+      return filters;
+    }, {});
+  },
+  getEntityFilter() {
```
paulfalgout, line 209: this was the super complicated filter builder

src/js/apps/patients/worklist/worklist_state.js
```diff
@@ -173,56 +151,92 @@ export default Backbone.Model.extend({
...
   },
   toggleSelected(model, isSelected, selectedIndex) {
-    const listName = this.isFlowType() ? 'selectedFlows' : 'selectedActions';
-    const currentList = clone(this.get(listName));
+    const selectedList = this.getSelectedList();
 
-    const newList = extend(currentList, {
-      [model.id]: isSelected,
-    });
+    selectedList[model.id] = isSelected;
```
paulfalgout, line 231: `getSelectedList` clones so no need to extend

src/js/views/patients/shared/list_views.js
```diff
@@ -24,13 +25,16 @@ const MaximumCountNarrowedTemplate = hbs`
 `;
 
 const CountView = View.extend({
+  initialize() {
+    const listType = this.getOption('isFlowList') ? 'flows' : 'actions';
```
paulfalgout, line 29: moved this into the view so we only have to pass in `isFlowList` and the `get(...` logic is only needed here.

src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -102,7 +102,6 @@
 
 .worklist-list__toggle {
   margin-left: auto;
-  margin-right: 16px;
```
paulfalgout, line 105: this seemed extra

src/js/views/patients/worklist/flow_views.js
```diff
@@ -12,7 +12,7 @@ import FlowItemTemplate from './flow-item.hbs';
 import 'scss/domain/action-state.scss';
 import './worklist-list.scss';
 
-const FlowTooltipTemplate = hbs`{{formatMessage (intlGet "patients.worklist.flowViews.flowListTooltips") title=worklistId team=team}}`;
+const FlowTooltipTemplate = hbs`{{formatMessage (intlGet "patients.worklist.flowViews.flowListTooltips") title=worklistId team=owner}}`;
```
paulfalgout, line 15: these are rendered in the list item and the `owner` will be the `team` when the `worklistId` needs the `team`

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -117,14 +118,12 @@ const TypeToggleView = View.extend({
 });
 
 const NoOwnerToggleView = View.extend({
+  className: 'u-margin--l-24 u-margin--r-16',
   template: hbs`
     <button class="button-filter-toggle {{#if noOwner}}button--blue{{/if}}">
       {{ @intl.patients.worklist.worklistViews.noOwnerToggleView.noOwner }}{{#if noOwner}}{{far "xmark"}}{{/if}}
     </button>
   `,
-  modelEvents: {
```
paulfalgout, line 125: this view is reinit'd after the `noOwner` change due to the app restart

src/js/views/patients/worklist/worklist_views.js
```diff
@@ -150,41 +171,81 @@ const ListTitleView = View.extend({
...
-      showOwnerDroplist: this.getOption('showOwnerDroplist'),
-      icons: worklistIcons[worklistId],
+      owner: this.owner.get('name'),
+      worklistId: underscored(this.model.id),
+      icons: worklistIcons[this.model.id],
     };
   },
+  initialize() {
+    const currentClinician = Radio.request('bootstrap', 'currentUser');
+    this.canViewAssignedActions = currentClinician.can('app:worklist:clinician_filter');
```
paulfalgout, line 191: I moved a lot of these `should` checks and the child view rendering from the app to this view and proxy the events back up to the app

test/integration/patients/worklist/worklist.js
```diff
@@ -3740,7 +3741,8 @@ context('worklist page', function() {
 
     cy
       .get('@listSearch')
-      .should('have.attr', 'value', 'Nurse');
+      .invoke('val')
```
paulfalgout, line 3755: the input didn't need to be re-rendered on restart, so the `value` attribute wasn't being re-rendered, so we have to read the input value instead.

test/integration/patients/worklist/worklist.js
```diff
@@ -2536,7 +2537,7 @@ context('worklist page', function() {
 
     cy
       .get('[data-owner-filter-region]')
-      .should('not.exist');
+      .should('be.empty');
```
paulfalgout, line 2551: the region always exists now, it's just empty when not used

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -122,7 +102,22 @@ export default App.extend({
   onStart(options, collection) {
     this.collection = collection;
     this.filteredCollection = collection.clone();
+    this.editableCollection = collection.clone();
```
nmajor25, line 105: Nice, this seems like the best way to do it. Just a separate collection like we already do with the `filteredCollection`.

nmajor25: I did notice a bug when selecting all via the bulk select button after changing actions to read only: [screengrab](https://github.com/RoundingWell/care-ops-frontend/assets/35355575/d8b659bf-320c-4d2b-9a66-cabf3a6e3038).

The total selected count is off and the newly changed read only action is edited along with the others when it shouldn't.

So I think we need to update the `editableCollection` when an action/flow in the worklist is changed from editable => read only. So any read only items are removed from the `editableCollection`. Currently those read only items are sticking around in there.

Everything else works well with it. Including the `shift`+`click` multi-select.

src/js/entities-service/entities/actions.js
```diff
@@ -89,6 +89,15 @@ const _Model = BaseModel.extend({
...
   },
+  canEdit() {
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+
+    if (currentUser.can('work:manage')) return true;
+
+    if (currentUser.can('work:owned:manage') && this.getOwner() === currentUser) return true;
+
+    return false;
+  },
```
nmajor25, line 100: That's a good way to do it, keeping this in the entity file.

src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -23,6 +23,13 @@
   color: color(blue);
 }
 
+.worklist-list__readonly {
+  border: solid 1px rgb(0 0 0 / 0%);
+  cursor: not-allowed;
+  display: inline-block;
+  vertical-align: middle;
+}
```
nmajor25: Probably want the `padding: 8px` on here too? So the list items maintain the same alignment if they're editable or not: [screenshot](https://github.com/RoundingWell/care-ops-frontend/assets/35355575/936b0b24-f617-4bd9-ac3e-33debc628109).

But it's just a POC, so doesn't really matter yet.

src/js/views/patients/worklist/action-item.hbs
nmajor25: I'm not pulled either way tbh... this is a clever way to do it though 😄 Hadn't thought of that 🤯

src/js/views/patients/worklist/worklist-list.scss
paulfalgout: yeah I wasn't sure as whatever this is it'd have to be applied to each element that is typically a component..  it _might_ be that this particular class ends up only being the `cursor: not-allowed`

src/js/views/patients/worklist/action-item.hbs
paulfalgout: So we _probably_ should go the route of how the flow currently works.. it's certainly more LoC and complexities, but it doesn't add extra rendering time to regular long lists.

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: good catch.

paulfalgout:
Alright so I implemented the `change:canEdit` event to update the `editableCollection`. It actually cleaned up a bit too.  So this'll drop the coverage until the feature is done, but we can probably move forward after :eyes:  I also pulled the POC commit of the action template and the style.  We probably need to go the route of the read only view of the flow.  However.. I noticed that there's already a case where actions are semi-read-only if an action is complete.  This is more similar to the direction you were taking it..  I think maybe we can do a bit of both.  In the cases where we _already_ have a read-only solution, we could go with it... and where we don't we add a read-only view.  I _don't_ want to modify any component to do this.  The one exception might be adding the no-cursor style to a disabled component.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -122,35 +102,56 @@ export default App.extend({
...
-    this.listenTo(collectionView, 'click:patientSidebarButton', ({ model }) => {
-      const patient = model.getPatient();
-      Radio.request('sidebar', 'start', 'patient', { patient });
+    this.listenTo(collectionView, {
+      'filtered'(filtered) {
+        this.filteredCollection.reset(filtered);
+        this.editableCollection.reset(this._getListEditable(collectionView));
+      },
+      'change:canEdit'() {
+        this.editableCollection.reset(this._getListEditable(collectionView));
```
nmajor25, line 137: There's the potential where an action/flow can be selected and then switched to `canEdit === false`. When that currently happens, the action/flow still has the `is-selected` classname when it shouldn't.

When `change:canEdit` is triggered, looks like that action/flow is still being stored in `state.actionsSelected`/`state.flowsSelected`.

Everything else with multi-select works well. Only thing I noticed.

Related screen recording:

https://github.com/RoundingWell/care-ops-frontend/assets/35355575/130e566c-8f7c-4f9b-8cd2-80ca472a980e

paulfalgout: so the `getSelected` on the state is already dealing with filtering out actions/flows that aren't in the editableCollection..  I _think_ this was so we didn't have to manage `actionSelected`/`flowsSelected` as users did the find in list.. _so_ I think we can just handle this visually on that view

paulfalgout: I haven't had time to test this fully, but I think the latest commit is the solution

paulfalgout:
Moving this work to [sc-37269] although it partially solves the card it's currently connected to, it didn't finish out those permissions and doesn't test for it.

nmajor25 (approved):

PR #1069, opened 2023-06-07
Work manage permission on worklists

paulfalgout:
Shortcut Story ID: [sc-35584]

Since the worklist swaps between action/flow childViews I kept most of the logic testing on the flow tests and did not duplicate them to actions, where I created tests for only the read-only row item.  Although now I'm not entirely sure if the action tests and some of the flow tests should move out of bulk-edit as it's more about the lists than bulk edit..


src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -122,13 +122,32 @@
...
-  padding: 0 8px;
+  font-size: 12px;
+  line-height: 14px;
+  overflow: hidden;
+  padding: 8px;
+  position: relative;
+  text-overflow: ellipsis;
   vertical-align: middle;
+  white-space: nowrap;
 }
```
nmajor25, line 137: We're still using the original design doc Mark commented on the shortcut story?

What you currently have built in this PR:

Screenshot 2023-06-07 at 10 39 33 AM

It looks like the read only items in that doc have the semi-bold font weight to match the editable version. And the clock icon has a lighter grey color.

nmajor25 (approved):

src/js/views/patients/worklist/worklist-list.scss
paulfalgout: confirmed we're ok on styles

paulfalgout:
[sc-37274]

PR #1076, opened 2023-06-21
Fix datadog console error when a response status is empty (i.e. cancelled API request)

nmajor25:
Shortcut Story ID: [sc-37555]

This was discovered on QA when an API request is cancelled when typing a query into the patient search modal.

### DIFF
--- a/src/js/datadog.js
+++ b/src/js/datadog.js
@@ -53,7 +53,7 @@ function initRum({ isForm }) {
     beforeSend(event, context) {
       // Add header/response context to api errors
       if (event.type === 'resource' && event.resource.type === 'fetch') {
-        if (context.response.status >= 400) {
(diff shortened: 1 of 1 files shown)


src/js/datadog.js
```diff
@@ -53,7 +53,7 @@ function initRum({ isForm }) {
     beforeSend(event, context) {
       // Add header/response context to api errors
       if (event.type === 'resource' && event.resource.type === 'fetch') {
-        if (context.response.status >= 400) {
+        if (context.response.status && context.response.status >= 400) {
```
nmajor25, line 56: Configured this way, the header/response stuff won't be included if `context.response.status` is empty.

We could do this instead:

```js
if (!context.response.status || context.response.status >= 400)
```

If we want to include it for empty status API requests.

paulfalgout (approved):

PR #1078, opened 2023-06-22
Replace `cy.route()` with `cy.intercept()` in cypress integration files

nmajor25:
Shortcut Story ID: [sc-30276]

This refactors all instances of `cy.route()` to `cy.intercept()` that are used in `test/integration/...` files.

The `test/support/api` files still use `cy.route()`. I figured we could tackle that in a separate PR since it gets a little more complicated.


paulfalgout:
feels nice

PR #1082, opened 2023-06-23
Update dependencies and Node

paulfalgout:
Shortcut Story ID: [sc-33205]

@zfixler we'll want to do this with the forms repo soon enough.   nothing should be incompatible, but we should keep them in sync


paulfalgout:
bah.. need to update cypress docker

.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ orbs:
 executors:
   with-chrome-and-firefox:
     docker:
-      - image: "cypress/browsers:node16.18.0-chrome107-ff106-edge"
+      - image: "cypress/browsers:node18.12.0-chrome103-ff107"
```
paulfalgout, line 9: the newer cypress orb should use less strict browser versioning, but that's a bigger update

paulfalgout:
https://github.com/cypress-io/cypress/pull/27127#event-9638691241 12.15.1 didn't happen so I guess this can get merged for now.

PR #1086, opened 2023-06-27
Add action specific context to the pdf form service

nmajor25:
Shortcut Story ID: [sc-37972]

Allows the form pdf service to load a form based on an action id.

### DIFF
--- a/src/js/entities-service/form-responses.js
+++ b/src/js/entities-service/form-responses.js
@@ -10,6 +10,7 @@ const Entity = BaseEntity.extend({
     'formResponses:collection': 'getCollection',
     'fetch:formResponses:submission': 'fetchSubmission',
     'fetch:formResponses:latestSubmission': 'fetchLatestSubmission',
+    'fetch:formResponses:submission:byAction': 'fetchLatestSubmissionByAction',
   },
   fetchSubmission(responseId) {
     if (!responseId) return [{}];
@@ -22,6 +23,14 @@ const Entity = BaseEntity.extend({
       return filters;
     }, { filter: {} });
     return fetcher(`/api/patients/${ patientId }/form-responses/latest`, { data }).then(handleJSON);
+  },
+  fetchLatestSubmissionByAction(actionId, filter) {
+    const data = reduce(filter, (filters, value, key) => {
+      if (!value) return filters;
+      filters.filter[key] = value;
+      return filters;
+    }, { filter: {} });
+    return fetcher(`/api/actions/${ actionId }/form-responses/latest`, { data }).then(handleJSON);
   },
 });
 
--- a/src/js/entities-service/form-responses.js
+++ b/src/js/entities-service/form-responses.js
@@ -10,7 +10,6 @@ const Entity = BaseEntity.extend({
     'formResponses:collection': 'getCollection',
     'fetch:formResponses:submission': 'fetchSubmission',
     'fetch:formResponses:latestSubmission': 'fetchLatestSubmission',
-    'fetch:formResponses:submission:byAction': 'fetchLatestSubmissionByAction',
   },
   fetchSubmission(responseId) {
     if (!responseId) return [{}];
@@ -23,14 +22,6 @@ const Entity = BaseEntity.extend({
       return filters;
     }, { filter: {} });
     return fetcher('/api/form-responses/latest', { data }).then(handleJSON);
-  },
-  fetchLatestSubmissionByAction(actionId, filter) {
-    const data = reduce(filter, (filters, value, key) => {
-      if (!value) return filters;
-      filters.filter[key] = value;
-      return filters;
-    }, { filter: {} });
-    return fetcher(`/api/actions/${ actionId }/form-responses/latest`, { data }).then(handleJSON);
   },
 });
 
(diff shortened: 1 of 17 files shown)


src/js/entities-service/form-responses.js
```diff
@@ -23,6 +24,14 @@ const Entity = BaseEntity.extend({
     }, { filter: {} });
     return fetcher(`/api/patients/${ patientId }/form-responses/latest`, { data }).then(handleJSON);
   },
+  fetchLatestSubmissionByAction(actionId, filter) {
+    const data = reduce(filter, (filters, value, key) => {
+      if (!value) return filters;
+      filters.filter[key] = value;
+      return filters;
+    }, { filter: {} });
+    return fetcher(`/api/actions/${ actionId }/form-responses/latest`, { data }).then(handleJSON);
```
nmajor25: This API endpoint doesn't exist on the backend yet. Shortcut story: https://app.shortcut.com/roundingwell/story/38123/.

I copied the `filter` stuff over from the `fetchSubmission`. I assumed we'd want that, but wasn't 100% sure.

nmajor25: Currently missing coverage for the `filter` portion.

paulfalgout: yeah so we'll want to pass the same the the regular form does for `action.tags` and `flow` for the filter.. but only those.. not the formId.

paulfalgout (changes requested):

src/js/formservice.js
```diff
@@ -7,7 +7,27 @@ import App from 'js/base/app';
...
 import 'js/entities-service';
 
-const Application = App.extend({
+const ActionFormApp = App.extend({
+  beforeStart({ actionId }) {
+    return [
+      Radio.request('entities', 'fetch:forms:byAction', actionId),
+      Radio.request('entities', 'fetch:forms:definition:byAction', actionId),
+      Radio.request('entities', 'fetch:forms:fields', actionId),
+      Radio.request('entities', 'fetch:formResponses:latestSubmission', { action: actionId }),
```
nmajor25: We talked about moving this to the `onStart()`. But does this work?

All this needs is the action id, correct?

paulfalgout:
This PR includes commits that are already in develop.  If we can make this a single commit that'd be better.

src/js/formservice.js
paulfalgout: this does not work.  we need to determine based on the tags on the action what needs to be loaded.  The logic is essentially the same as https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/services/forms.js#L139 but instead of `form` it'd be `action`

paulfalgout (changes requested):

test/integration/formservice/formservice.js
```diff
@@ -95,4 +95,114 @@ context('Formservice', function() {
...
+    cy
+      .get('iframe')
+      .should('have.attr', 'src', '/formservice/1');
+
+    cy
+      .window()
+      .then(win => {
+        win.postMessage({ message: 'form:pdf', args: {
+          definition: {
+            'components': [
```
paulfalgout: None of these keys need `'`s

paulfalgout (changes requested):
The tags test will look something like: https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/integration/forms/action.js#L492

test/integration/formservice/formservice.js
nmajor25: 👍 I just removed those

test/integration/formservice/formservice.js
```diff
@@ -95,4 +97,142 @@ context('Formservice', function() {
...
+      .routeLatestFormResponse()
+      .visit('/formapp/pdf/action/1', { noWait: true, isRoot: true });
+
+    cy
+      .wait('@routeLatestFormResponse')
+      .itsUrl()
+      .its('search')
+      .should('contain', 'filter[action.tags]=foo-tag')
+      .should('contain', 'filter[flow]=1')
+      .should('contain', 'filter[patient]=1');
```
nmajor25, line 191: Are we okay just checking the URL filter for this one?

In `display action form with a response`, we already test that the latest response displays in the form.

src/js/formservice.js
```diff
@@ -29,10 +73,16 @@ const Application = App.extend({
 
 const Router = Backbone.Router.extend({
   routes: {
+    'formservice/:actionId': 'startActionFormService',
```
paulfalgout: lets make this `formservice/action/:actionId` to future proof the pattern a bit more

test/integration/formservice/formservice.js
```diff
@@ -95,4 +97,142 @@ context('Formservice', function() {
...
+  })
```
paulfalgout: tests are failing due to a missing ';' here

paulfalgout:
you'll need to merge in `develop` to get the correct status checks, but either way, not a big deal.

paulfalgout (changes requested):

src/js/formservice.js
nmajor25: That's a good idea, just updated to that.

test/integration/formservice/formservice.js
nmajor25: Oops, fixed that.

paulfalgout (approved):

PR #1088, opened 2023-06-29
Switch to form versioning

paulfalgout:
Shortcut Story ID: [sc-37971]

@zfixler note that the version is passed along with the other contexts and will be made available at the "global" context as `version`. cool?

### DIFF
--- a/src/js/entities-service/entities/events.js
+++ b/src/js/entities-service/entities/events.js
@@ -33,8 +33,8 @@ const _Model = BaseModel.extend({
     return Radio.request('entities', 'programs:model', this.get('_program'));
   },
   getForm() {
-    if (!this.get('_form')) return;
-    return Radio.request('entities', 'forms:model', this.get('_form'));
+    if (!this.get('_form_revision')) return;
+    return Radio.request('entities', 'forms:model', this.get('_form_revision'));
   },
 });
 
--- a/src/js/entities-service/forms.js
+++ b/src/js/entities-service/forms.js
@@ -14,7 +14,11 @@ const Entity = BaseEntity.extend({
     'fetch:forms:byAction': 'fetchByAction',
     'fetch:forms:definition:byAction': 'fetchDefinitionByAction',
   },
-  fetchDefinition(formId) {
+  fetchDefinition(formId, actionId) {
+    if (actionId) {
+      return this.fetchDefinitionByAction(actionId);
+    }
+
     return fetcher(`/api/forms/${ formId }/definition`).then(handleJSON);
   },
   fetchFields(actionId, patientId, formId) {
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -113,7 +113,7 @@ export default App.extend({
   fetchForm() {
     const channel = this.getChannel();
 
-    return Promise.resolve(Radio.request('entities', 'fetch:forms:definition', this.form.id))
+    return Promise.resolve(Radio.request('entities', 'fetch:forms:definition', this.form.id), get(this.action, 'id'))
       .then(definition => {
         channel.request('send', 'fetch:form', {
           definition,
@@ -126,7 +126,7 @@ export default App.extend({
     const channel = this.getChannel();
 
     return Promise.all([
-      Radio.request('entities', 'fetch:forms:definition', this.form.id),
+      Radio.request('entities', 'fetch:forms:definition', this.form.id, get(this.action, 'id')),
     ]).then(([definition]) => {
       channel.request('send', 'fetch:form:data', {
         definition,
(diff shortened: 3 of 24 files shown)


src/js/entities-service/entities/events.js
```diff
@@ -33,8 +33,8 @@ const _Model = BaseModel.extend({
     return Radio.request('entities', 'programs:model', this.get('_program'));
   },
   getForm() {
-    if (!this.get('_form')) return;
-    return Radio.request('entities', 'forms:model', this.get('_form'));
+    if (!this.get('_form_revision')) return;
+    return Radio.request('entities', 'forms:model', this.get('_form_revision'));
```
shadowhand, line 37: I'm not sure that swapping `form` with `form_revision` is going to be right in _all_ cases.

src/js/entities-service/forms.js
```diff
@@ -14,7 +14,11 @@ const Entity = BaseEntity.extend({
     'fetch:forms:byAction': 'fetchByAction',
     'fetch:forms:definition:byAction': 'fetchDefinitionByAction',
   },
-  fetchDefinition(formId) {
+  fetchDefinition(formId, actionId) {
+    if (actionId) {
+      return this.fetchDefinitionByAction(actionId);
+    }
```
shadowhand, line 20: Would it make more sense to keep this method as-is, and switch to `fetchDefinitionByAction` with outside calls, rather than having to modify every call to `fetchDefinition`?

src/js/services/forms.js
```diff
@@ -113,10 +113,11 @@ export default App.extend({
   fetchForm() {
     const channel = this.getChannel();
 
-    return Promise.resolve(Radio.request('entities', 'fetch:forms:definition', this.form.id))
+    return Promise.resolve(Radio.request('entities', 'fetch:forms:definition', this.form.id), get(this.action, 'id'))
```
shadowhand, line 116:
```suggestion
    return Promise.resolve(Radio.request('entities', 'fetch:forms:definition', this.form.id, get(this.action, 'id')))
```

test/fixtures/test/action-events.json
```diff
@@ -294,7 +294,7 @@
           "type": "clinicians"
         }
       },
-      "form": {
+      "form_revision": {
         "data": {
           "id": "11111",
           "type": "forms"
```
shadowhand, line 300: Form revisions have `type: "form-revisions"`.

shadowhand (commented):
I think introducing form revision entity type and modeling the relationships accurately will be better in the long run.

src/js/services/forms.js
paulfalgout: 😬 ha.  apparently this is never used for action..at least in the tests

src/js/entities-service/entities/events.js
paulfalgout: it is here though isn't it?  (assuming it's the correct entity) because in the migration only program-action and workspace didn't get migrated.. plus this'd show whatever version of the form name went with that particular revision?

test/fixtures/test/action-events.json
paulfalgout: didn't realize there was a new type, but I guess that makes sense.

src/js/entities-service/entities/events.js
shadowhand: Yes, this is fine. I am worried that mixing form/form-revision is going to lead to confusion/bugs down the road.

src/js/entities-service/forms.js
paulfalgout: yeah, so I think the form service is going to be split soon into one that's action specific, but currently switching between requests is more painful than following the pattern on `fetchFields`.   But both this and `fetchFields` will/would be refactored if the services are split.

test/fixtures/test/form-kitchen-sink.json
```diff
@@ -791,6 +791,16 @@
         "disableOnInvalid": true,
         "input": true,
         "tableView": false
+    },
+    {
+        "label": "Version",
+        "spellcheck": true,
+        "tableView": true,
+        "calculateServer": false,
+        "key": "version",
```
zfixler, line 800: @paulfalgout , do we want to store the version as a string or an integer?

src/js/entities-service/forms.js
paulfalgout: nope.. this got a lot more complicated than I expected with the other entity.  we'll have to split the service now because for patient sidebar widgets are not really going to be using the form revisions and so the actions will always use the revision and the patient/preview will always use the specific form.id.  Might have to wait until the following release.

PR #1089, opened 2023-07-03
Switch FE to generic form-response/latest

paulfalgout:
Depends on https://github.com/RoundingWell/care-ops-backend/pull/5877

Shortcut Story ID: [sc-38243]

### DIFF
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -136,22 +136,21 @@ export default App.extend({
       });
     });
   },
-  _getPrefillFilters({ actionId, flowId, patientId }, form) {
+  _getPrefillFilters({ flowId, patientId }, form) {
     const prefillActionTag = form.getPrefillActionTag();
 
     if (prefillActionTag) {
       return {
-        'action': actionId,
         'action.tags': prefillActionTag,
         'flow': flowId,
-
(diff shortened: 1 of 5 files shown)


src/js/services/forms.js
```diff
@@ -136,31 +136,36 @@ export default App.extend({
...
     });
   },
-  _getPrefillFilters(flowId, form) {
+  _getPrefillFilters({ actionId, flowId, patientId }, form) {
     const prefillActionTag = form.getPrefillActionTag();
 
     if (prefillActionTag) {
       return {
+        'action': actionId,
         'action.tags': prefillActionTag,
```
shadowhand: The combination of `flow + action + action.tags` is illogical, because we cannot determine if `action` or `action.tags` should take precedence, given that `action` may not have `action.tags` at all.

Is the assumption that either `flow` or `action` will be filtered out because they have empty values?

paulfalgout: hmm yeah.. I'm actually not sure what use case we'd have for `action` right now.  Really I should have been doing this for patient, so I switched to that.. which I _think_ makes this essentially backwards compatible with how it's currently working

nmajor25 (approved):

PR #1097, opened 2023-07-14
Cypress component

paulfalgout:
This is a work in progress.  I'm going to finish converting the generic components.  However I think in the long run we could probably test more domain specific components such as the worklist date range selector as a component such that it reduces the testing load on the list itself and speeds up the tests. I'm not 100% sure what that looks like yet, but it seems ideal.


paulfalgout:
hmm linter extention fails on ci but not in vscode.. I'll probably need to extend from a common source for this to work without failure.

test/support/coverage.js
```diff
@@ -1,28 +1,36 @@
...
+function storeCoverage(win) {
+  const coverage = win.__coverage__;
+
+  if (!coverage) return;
+
+  windowCoverageObjects.push(coverage);
 }
 
 if (Cypress.env('COVERAGE')) {
+  beforeEach(function() {
```
paulfalgout, line 14: In attempting to switch to `@cypress/code-coverage` I found this library does not let us combine coverage reports across tests when just running cypress.  And it was actually slightly slower..  _but_ I copied the pattern of capturing the coverage object on the window load and then consuming in it the after.  This _should_ allow for us to use `cy.visit` without losing coverage.

nmajor25 (approved):
Sweet 👍

PR #1106, opened 2023-08-08
Outreach: Handle API Request 500 Errors

nmajor25:
Shortcut Story ID: [sc-39316]

When the `/api/outreach?filter[action]=11111` API request returns a `500` error, a blank page was displayed ([screenshot](https://github.com/RoundingWell/care-ops-frontend/assets/35355575/af616430-0d3c-4e9a-8c90-912f70addf2f)). 

Instead, this needs to show the `Uh-Oh...` error message we use in the main app ([screenshot](https://github.com/RoundingWell/care-ops-frontend/assets/35355575/02938724-851d-4f78-9c40-8b14841fc0db)).

### DIFF
--- a/src/js/outreach/apps/verify_app.js
+++ b/src/js/outreach/apps/verify_app.js
@@ -21,6 +21,8 @@ export default App.extend({
     return getPatientInfo({ actionId });
   },
   onFail(options, response) {
+    if (response.status >= 500) return;
+
     const dialogView = new DialogView();
     this.showView(dialogView);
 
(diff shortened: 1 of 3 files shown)


src/js/outreach/apps/verify_app.js
```diff
@@ -21,6 +21,8 @@ export default App.extend({
     return getPatientInfo({ actionId });
   },
   onFail(options, response) {
+    if (response.status >= 500) return;
```
paulfalgout, line 24: Does it even get here?  I would kind of assume the route takes over first, but I guess because there's no an error router app, it isn't stopping the previous app.. hmm that's interesting..

paulfalgout (approved):

PR #1108, opened 2023-08-09
Add Action's Flow state filter

paulfalgout:
Shortcut Story ID: [sc-38863]
Screenshot 2023-08-09 at 4 40 43 PM

This adds an additional state filter to filter the `flow.status` for action lists.

Because it made the filter sidebar `listType` specific this generated quite an organizational issue, and simply implementing it made duplicating state quite a bit..  so this includes a refactor to move the filter state exclusively into a shared filter state model.


src/js/apps/patients/shared/filters_state.js
```diff
@@ -0,0 +1,95 @@
...
+    const { done, notDone } = states.groupByDone();
+    this.states = { done, notDone, all: states };
+  },
+  defaults() {
+    const isDoneOnly = this.get('isDoneOnly');
+    const doneStates = this.states.done.map('id');
+    const notDoneStates = this.states.notDone.map('id');
+
+    return {
+      filters: {},
```
nmajor25: One thing I noticed and thought about that is probably not worth the effort to change.

But do we want to rename this to `customFilters` instead?

We have several different filters (date, states, flowStates, custom filters, etc.), but this `filters: {}` object only contains custom filters. The rest of our filters are stored outside of this object.

The local storage object as reference:

Screenshot 2023-08-10 at 11 04 33 AM

nmajor25 (approved):
Code looks good and everything works great when testing it locally 👍 

Just commented one thought I had when reviewing.

nmajor25:
Title of this PR should be `Flow status filter` instead of `Form status filter`, right?

src/js/apps/patients/shared/filters_state.js
paulfalgout: yeah I think you're correct on this.  I'll update it

nmajor25 (approved):

PR #1109, opened 2023-08-11
Add form drafts

paulfalgout:
Shortcut Story ID: [sc-38920]

This needs tests.  I also haven't tested this against the server as we need some modifications for user-specific drafts, but wanted to get 👀 on the code

### DIFF
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -6,6 +6,8 @@ import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
+import { FORM_RESPONSE_STATUS } from 'js/static';
+
 import { versions } from 'js/config';
 
 export default App.extend({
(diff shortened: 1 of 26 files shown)


paulfalgout:
Alright.. now I just need tests for the server drafts

src/js/services/forms.js
```diff
@@ -210,19 +240,32 @@ export default App.extend({
...
     ]).then(([definition, response]) => {
       channel.request('send', 'fetch:form:response', {
         definition,
-        formSubmission: get(response, 'data', {}),
+        formSubmission: response.getResponse(),
         contextScripts: this.form.getContextScripts(),
       });
     });
   },
+  /* istanbul ignore next: Testing the debounce is buggy with cy.clock */
```
paulfalgout, line 252: @nmajor25 I spent forever trying to figure out a reasonable way to test this with the 30 second debounce.  `cy.clock` doesn't work reliably because it doesn't always pass the `defer` that calls `Backbone.History.start` and everything fails.. I have some ideas to fix this, but it has to be in a future release.

paulfalgout: I just realized we're missing some other coverage of functions that only this calls.

PR #1111, opened 2023-08-17
Outreach: Handle 404 Errors

nmajor25:
Shortcut Story ID: [sc-39558]

When a malformed URL is loaded by a user (.i.e. `/outreach`, `/outreach/`, `/outreach/asdf/asdf`), show a `404` error:

Screenshot 2023-08-17 at 2 53 59 PM

### DIFF
--- a/src/js/outreach/apps/error_app.js
+++ b/src/js/outreach/apps/error_app.js
@@ -1,0 +1,35 @@
+import { bind } from 'underscore';
+
+import RouterApp from 'js/base/routerapp';
+
+import { DialogView } from 'js/outreach/views/dialog_views';
+
+import {
+  ErrorView,
+  Error404View,
+} from 'js/outreach/views/error_views';
+
+export default RouterApp.extend({
+  eventRoutes: {
+    'error': {
+      route: 'outreach/500',
+      action: 'show500',
+      root: true,
+    },
+  },
+  initialize() {
+    this.router.route('*unknown', 'show404', bind(this.show404, this));
+  },
+  show404() {
+    const dialogView = new DialogView();
+    this.showView(dialogView);
+
+    this.showChildView('content', new Error404View());
+  },
+  show500() {
+    const dialogView = new DialogView();
+    this.showView(dialogView);
+
+    this.showChildView('content', new ErrorView());
+  },
+});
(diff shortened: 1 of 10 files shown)


src/js/outreach/apps/error_app.js
```diff
@@ -0,0 +1,31 @@
...
+
+import RouterApp from 'js/base/routerapp';
+
+import { DialogView } from 'js/outreach/views/dialog_views';
+import { ErrorView } from 'js/outreach/views/error_views';
+
+export default RouterApp.extend({
+  eventRoutes: {
+    'error': {
+      route: 'outreach/500',
```
nmajor25, line 15: When a `500` error occurs, should we just keep the user on the same URL? Like we do for the `404`?

Currently, the URL is redirected to `roundingwell.com/500` and this is shown in the UI:

Screenshot 2023-08-17 at 3 05 05 PM

Reloading the page won't work since we redirected them to `roundingwell.com/outreach/500`. Or maybe the messaging in the UI needs to be changed.

paulfalgout: hmm yeah.. that message doesn't make any sense.  We could change that to a `Try again` link which is just a `history.back()`

src/js/outreach/views/error_views.js
```diff
@@ -0,0 +1,26 @@
...
+      <div class="dialog__error-header">Oops! The page you requested can’t be found.</div>
+      <div class="dialog__error-info">Return to the Outreach message and re-open the link.</div>
+    {{else}}
+      <div class="dialog__icon dialog__icon--error">{{fat "octagon-exclamation"}}</div>
+      <div class="dialog__error-header">Uh-oh, there was an error. Try reloading the page.</div>
+    {{/if}}
+  `,
+  templateContext() {
+    return {
+      is404: this.getOption('is404'),
```
paulfalgout: this flag can just be removed entirely and this can be `Error404View` or something

paulfalgout (changes requested):

src/js/outreach/views/error_views.js
nmajor25: Good call, just updated that.

src/js/outreach/apps/error_app.js
nmajor25: Okay, I'm going to do that in a different PR and shortcut story, so Sofia can track it better.

paulfalgout (approved):

src/js/outreach/apps/error_app.js
nmajor25: PR for reference: #1115

PR #1117, opened 2023-08-22
Don't include `fields[flows]=name,state` in `/flows` api request

nmajor25:
Shortcut Story ID: [sc-39819]

Bug introduced by this commit: https://github.com/RoundingWell/care-ops-frontend/pull/1116/commits/216b733b87d41cad56341a9a719feec0c9a7fdde.

If `fields[flows]=name,state` is included in the `/flows` api request, it throws this error on the flows version of the worklist:

Screenshot 2023-08-22 at 3 19 59 PM (1) (1)

Removing `fields[flows]=name,state` fixes the error. But it's still not clear why that causes the error in the first place.

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -113,13 +113,18 @@ export default App.extend({
     this.showView();
   },
   beforeStart() {
-    return Radio.request('entities', `fetch:${ this.getState().getType() }:collection`, {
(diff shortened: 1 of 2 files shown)


shadowhand (approved):
I have a feeling that the `fields` definition should move to something like:

```js
const data = {
  fields: this.getState().getFields(),
  // ...
}
```

But as a quick fix, this looks good.

PR #1119, opened 2023-08-23
Outreach: Update FE to coincide with backend api endpoint updates

nmajor25:
Shortcut Story ID: [sc-39842] [sc-39248]

Corresponding backend PR: [#6263](https://github.com/RoundingWell/care-ops-backend/pull/6263)

### DIFF
--- a/src/js/outreach/apps/opt-in_app.js
+++ b/src/js/outreach/apps/opt-in_app.js
@@ -1,7 +1,5 @@
-import Backbone from 'backbone';
+import Radio from 'backbone.radio';
 import App from 'js/base/app';
-
-import { optInPostRequest } from 'js/outreach/entities';
 
 import {
   OptInView,
@@ -11,30 +9,28 @@ import {
 
 import { DialogView } from 'js/outreach/views/dialog_views';
 
-const StateModel = Backbone.Model.extend({
-  defaults: {
-    firstName: '',
-    lastName: '',
-    dob: '',
-    phone: '',
-  },
-});
-
 export default App.extend({
-  StateModel,
-  onStart() {
+  beforeStart() {
+    return Radio.request('entities', 'outreach:model', {
+      first_name: '',
+      last_name: '',
+      birth_date: '',
+      phone: '',
+    });
+  },
+  onStart(options, outreach) {
+    this.outreach = outreach;
+
     const dialogView = new DialogView();
     this.showView(dialogView);
 
     this.showOptInView();
   },
   showOptInView() {
-    const optInView = new OptInView({ model: this.getState() });
+    const optInView = new OptInView({ model: this.outreach });
 
-    this.listenTo(optInView, 'click:submit', () => {
-      optInPostRequest({
-        inputData: this.getState(),
-      })
+    this.listenTo(optInView, 'click:submit', ({ model }) => {
+      this.outreach.registerPatient(model.attributes)
         .then(() => {
           this.showResponseSuccessView();
         })
--- a/src/js/outreach/entities-service/entities/outreach.js
+++ b/src/js/outreach/entities-service/entities/outreach.js
@@ -1,0 +1,62 @@
+import Store from 'backbone.store';
+import BaseCollection from 'js/base/collection';
+import BaseModel from 'js/base/model';
+import fetcher, { handleJSON } from 'js/base/fetch';
+
+const TYPE = 'outreach';
+
+const _Model = BaseModel.extend({
+  type: TYPE,
+  registerPatient(attributes) {
+    return fetcher('/api/outreach', {
+      method: 'POST',
+      data: JSON.stringify({
+        data: {
+          type: 'patients',
+          attributes,
+        },
+      }),
+    })
+      .then(handleJSON);
+  },
+  requestOtpCreation(actionId) {
+    return fetcher('/api/outreach/otp', {
+      method: 'POST',
+      data: JSON.stringify({
+        data: {
+          type: 'patient-actions',
+          id: actionId,
+        },
+      }),
+    })
+      .then(handleJSON);
+  },
+  requestOtpAuth(code) {
+    const url = '/api/outreach/auth';
+
+    return fetcher(url, {
+      method: 'POST',
+      data: JSON.stringify({
+        data: {
+          type: 'outreach',
+          id: this.id,
+          attributes: {
+            code,
+          },
+        },
+      }),
+    })
+      .then(handleJSON);
+  },
+});
+
+const Model = Store(_Model, TYPE);
+const Collection = BaseCollection.extend({
+  model: Model,
+});
+
+export {
+  _Model,
+  Model,
+  Collection,
+};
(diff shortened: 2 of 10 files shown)


src/js/outreach/entities-service/index.js
nmajor25, line 1: I'm not sure what we should do with the `getToken()` function and `Radio.reply('auth', ...)` code.  Where should it be placed?

nmajor25, line 1: Also need to look into the `postResponse()` function in this file as well.

src/js/outreach/apps/opt-in_app.js
```diff
@@ -11,30 +9,28 @@ import {
...
   showOptInView() {
-    const optInView = new OptInView({ model: this.getState() });
+    const optInView = new OptInView({ model: this.outreach });
 
-    this.listenTo(optInView, 'click:submit', () => {
-      optInPostRequest({
-        inputData: this.getState(),
-      })
+    this.listenTo(optInView, 'click:submit', ({ model }) => {
+      this.outreach.registerPatient(model.attributes)
```
paulfalgout, line 33: I think this'd be something like `Radio.request('entities', 'register:patient', model.attributes);` and the model can go back to being a state model.  This just isn't an outreach model.

src/js/outreach/entities-service/entities/outreach.js
```diff
@@ -0,0 +1,63 @@
...
+        data: {
+          type: 'patient-actions',
+          id: this.id,
+        },
+      }),
+    })
+      .then(handleJSON);
+  },
+  requestOtpAuth(code) {
+    const url = '/api/outreach/auth';
```
paulfalgout, line 35: this should be `return this.save({ code }, { url });`. This is an outreach model so we should use the model api

src/js/outreach/entities-service/entities/outreach.js
```diff
@@ -0,0 +1,63 @@
+import Store from 'backbone.store';
+import BaseCollection from 'js/base/collection';
+import BaseModel from 'js/base/model';
+import fetcher, { handleJSON } from 'js/base/fetch';
+
+const TYPE = 'outreach';
+
+const _Model = BaseModel.extend({
+  type: TYPE,
+  urlRoot: '/api/outreach',
```
paulfalgout: is the url even used by this model?

src/js/outreach/entities-service/entities/outreach.js
```diff
@@ -0,0 +1,63 @@
...
+      }),
+    })
+      .then(handleJSON);
+  },
+  requestOtpCreation() {
+    return fetcher('/api/outreach/otp', {
+      method: 'POST',
+      data: JSON.stringify({
+        data: {
+          type: 'patient-actions',
```
paulfalgout, line 27: is this correct?  I don't see `patient-actions` anywhere in the openapi documentation.. we're sure this shouldn't just be `return this.fetch({ url: '/api/outreach/otp' });`?  If this needs to be `patient-actions` then this isn't an outreach model and should just be it's own radio request outside of this model.

src/js/outreach/entities-service/entities/outreach.js
```diff
@@ -0,0 +1,63 @@
...
+import BaseCollection from 'js/base/collection';
+import BaseModel from 'js/base/model';
+import fetcher, { handleJSON } from 'js/base/fetch';
+
+const TYPE = 'outreach';
+
+const _Model = BaseModel.extend({
+  type: TYPE,
+  urlRoot: '/api/outreach',
+  registerPatient(attributes) {
```
paulfalgout, line 10: not an outreach model function

src/js/outreach/entities-service/index.js
paulfalgout: in the mainapp that's in an `auth.js` probably makes sense to move the setToken/getToken there.  well the function getToken with `patient-tokens` isn't in use at all so that just needs to be removed.

src/js/outreach/entities-service/index.js
```diff
@@ -41,62 +41,6 @@ function getToken({ dob, actionId }) {
...
-  };
-
-  return fetcher('/api/outreach', {
-    method: 'POST',
-    data: JSON.stringify({ data }),
-  })
-    .then(handleJSON);
-}
-
 function postResponse({ formId, actionId, response }) {
```
paulfalgout, line 44: this post response can probably just stay here with the handler for `registerPatient`, but instead of exporting the functions we can use the `entiities` channel for these.

However this function could be simplified since we have the models..
```js
const Model from 'js/base/model';

const FormResponse = Model.extend({
  url() {
    return `/api/actions/${ this.get('_action') }/relationships/form-responses`;
  },
  saveAll() {
    const attrs = this.attributes;

    const relationships = {
      'form': this.toRelation(attrs._form, 'forms'),
      'action': this.toRelation(attrs._action, 'patient-actions'),
    };

    return this.save(attrs, { relationships }, { wait: true });
  },
});

function postResponse({ formId, actionId, response }) {
  const response = new FormResponse({ _action: actionId, _form: formId, response });
  return response.saveAll();
}
```

We can probably utilize the base model for the patient request as well.

paulfalgout (changes requested):

src/js/outreach/entities-service/entities/outreach.js
```diff
@@ -0,0 +1,63 @@
...
+    })
+      .then(handleJSON);
+  },
+  requestOtpCreation() {
+    return fetcher('/api/outreach/otp', {
+      method: 'POST',
+      data: JSON.stringify({
+        data: {
+          type: 'patient-actions',
+          id: this.id,
```
paulfalgout: ok so this'll need to be the `actionId` not the outreach id and should move to the base entities with the postResponse etc.

src/js/outreach/entities-service/entities/outreach.js
nmajor25: It's here in the docs:

Screenshot 2023-08-24 at 10 03 25 AM

@shadowhand is this correct that the `/api/outreach/otp` endpoint has a `type: patient-actions`?

paulfalgout: yeah I talked to him after.  This needs to move out of the outreach model to just be in the general requests

src/js/outreach/entities-service/entities/outreach.js
nmajor25: I switched this to use `actionId`. But haven't switched it to the base entities yet. Still figuring out how to do that.

src/js/outreach/entities-service/entities/outreach.js
nmajor25: I don't think so. That's been removed.

src/js/outreach/entities-service/entities/outreach.js
nmajor25: Couldn't get this to work.

Switched it to this:

```js
  requestOtpAuth(code) {
    const url = '/api/outreach/auth';

    return this.save({ code }, { url });
  },
```

But it sends the request to `POST /api/outreach` instead of `POST api/outreach/auth`. The `{ url }` portion doesn't seem to be doing anything.

nmajor25:
Closed and replaced by: https://github.com/RoundingWell/care-ops-frontend/pull/1122.

PR #1122, opened 2023-08-24
Outreach: API Updates & Misc Fixes

nmajor25:
Shortcut Story ID: [sc-39842]

Corresponding backend PR: [#6263](https://github.com/RoundingWell/care-ops-backend/pull/6263)

### DIFF
--- a/src/js/outreach/views/opt-in_views.js
+++ b/src/js/outreach/views/opt-in_views.js
@@ -17,7 +17,7 @@ const OptInView = View.extend({
         type="text"
         class="input-primary opt-in__field-input js-first-name"
         placeholder="Enter your first name"
-        value="{{ firstName }}"
+        value="{{ first_name }}"
       />
     </div>
     <div class="opt-in__field">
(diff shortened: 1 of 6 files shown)


src/js/outreach/views/opt-in_views.js
```diff
@@ -54,30 +54,30 @@ const OptInView = View.extend({
...
+    'input @ui.birthDate': 'change:birthDate',
     'input @ui.phone': 'change:phone',
     'click @ui.submit': 'click:submit',
   },
   onRender() {
     this.setSubmitButtonState();
   },
   onChangeFirstName() {
-    this.model.set({ firstName: this.ui.firstName.val() });
+    this.model.set({ first_name: this.ui.firstName.val() });
```
paulfalgout, line 72: when we do cleanup this ought to be:
```
  modelEvents: {
    'change': 'setSubmitButtonState'
  },
  setSubmitButtonState() {
     this.ui.submit.prop('disabled', !this.model.canSubmit());
  },
```

test/integration/outreach/outreach.js
```diff
@@ -692,9 +600,8 @@ context('Outreach', function() {
       .visit('/outreach/11111', { noWait: true, isRoot: true });
 
     cy
-      .intercept('POST', '/api/outreach/1', {
-        delay: 100,
-        body: { data: {} },
+      .intercept('POST', '/api/outreach/otp', {
```
paulfalgout, line 603: when we come back for clean up I think any `POST` should have

```js
    .wait('@routeCreateVerifyCodeRequest')
     .its('request.body')
      .should(({ data }) => {
        expect(postData).to.be({ what: 'we think we sent' });
```

src/js/outreach/views/opt-in_views.js
```diff
@@ -98,10 +98,10 @@ const OptInView = View.extend({
   setSubmitButtonState() {
     const firstName = this.ui.firstName.val();
     const lastName = this.ui.lastName.val();
-    const dob = this.ui.dob.val();
+    const birthDate = this.ui.birthDate.val();
     const phone = this.ui.phone.val();
```
paulfalgout, line 102: This actually might be a little broken.. if `parsePhoneNumber` fails this will be true, but `phone` in the state model would be `null`.  But that's probably and edge case for now and will be fixed by doing the event driven state changes.

paulfalgout (approved):
We need to do a little clean up and make sure coverage is 💯  but :shipit: for now.

paulfalgout:
Since this went out in the release I released it, but we can come back around and do some refactoring

PR #1123, opened 2023-08-29
Sort state statuses by `sequence` attribute

nmajor25:
Shortcut Story ID: [sc-39838]

### DIFF
--- a/test/fixtures/test/states.json
+++ b/test/fixtures/test/states.json
@@ -2,7 +2,7 @@
   {
     "id": "22222",
     "name": "To Do",
-    "sequence": 1,
+    "sequence": 100,
     "status": "queued",
     "options": {
       "iconType": "fas",
@@ -13,7 +13,7 @@
   {
     "id": "33333",
     "name": "In Progress",
-    "sequence": 2,
+    "sequence": 200,
     "status": "started",
     "options": {
       "iconType": "fas",
@@ -24,7 +24,7 @@
   {
     "id": "55555",
     "name": "Done",
-    "sequence": 3,
+    "sequence": 300,
     "status": "done",
     "options": {
       "iconType": "fas",
@@ -35,7 +35,7 @@
   {
     "id": "66666",
     "name": "Unable to Complete",
-    "sequence": 4,
+    "sequence": 400,
     "status": "done",
     "options": {
       "iconType": "far",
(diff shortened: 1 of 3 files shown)


test/fixtures/test/states.json
```diff
@@ -2,7 +2,7 @@
   {
     "id": "22222",
     "name": "To Do",
-    "sequence": 1,
+    "sequence": 100,
```
nmajor25, line 5: Changed these to match what’s returned by the api in production.

Those sequence values are `100`, `200`, `300`, etc. instead of `1`, `2`, `3`, etc.

test/fixtures/test/states.json
```diff
@@ -46,18 +35,29 @@
...
+    "id": "55555",
+    "name": "Done",
+    "sequence": 300,
+    "status": "done",
+    "options": {
+      "iconType": "fas",
+      "icon": "circle-check",
+      "color": "green"
+    }
+  },
```
nmajor25: This confirms that the `comparator: 'sequence'` works. With the order of these changed, removing the `comparator: 'sequence'` causes tests to fail.

But I’m worried that this file might be changed in the future and we’ll unknowingly no longer be testing the comparator.

So maybe we want to add a test to the filters sidebar where we do a `routeStates(fx => {})` and we test this specifically without modifying the `fixtures/test/states.json` file?

nmajor25: Well, maybe that won't work 🤔 . Since any changes in the `routeStates(fx => {})` would be applied after the data is already received from the api and therefore after the `comparator` is applied?

paulfalgout: `routeStates` would modify the order if place before the `.visit`. It's probably worth testing in case `66666` and `77777` are removed

paulfalgout (changes requested):

test/integration/patients/sidebar/filter-sidebar.js
```diff
@@ -1343,5 +1343,62 @@ context('filter sidebar', function() {
...
+      })
+      .routeWorkspaces(fx => {
+        fx.data[0].relationships.states.data = [
+          { id: '22222', type: 'states' },
+          { id: '33333', type: 'states' },
+          { id: '55555', type: 'states' },
+        ];
+
+        return fx;
+      })
```
nmajor25, line 1374: Without this `routeWorkspaces()`, this happens:

Screenshot 2023-08-29 at 2 04 28 PM

I think because the states set on the workspace are overriding things in the test.

test/fixtures/test/states.json
nmajor25: I agree, I just added a test for that.

paulfalgout (approved):

PR #1124, opened 2023-08-30
Implement `work:team:manage` permission

nmajor25:
Shortcut Story ID: [sc-39262]

When a user has the `work:team:manage` permission, they can only edit actions that belong to team members (or the team itself).

And when that user is editing an action/flow, the owner droplist should be restricted so they can only assign work to users from their own team (or the team itself).

Still need to do the following:

- [x] Restrict owner component to only include the user’s team and team members (in all contexts)
- [x] Patient dashboard tests
- [x] Patient archive tests
- [x] Worklist bulk edit
- [x] Patient flow tests
- [x] Patient action sidebar tests
- [x] Patient flow sidebar tests

### DIFF
--- a/src/js/entities-service/entities/actions.js
+++ b/src/js/entities-service/entities/actions.js
@@ -109,6 +109,14 @@ const _Model = BaseModel.extend({
 
     if (currentUser.can('work:owned:manage') && this.getOwner() === currentUser) return true;
 
+    if (currentUser.can('work:team:manage')) {
+      const owner = this.getOwner();
+      const currentUsersTeam = currentUser.getTeam();
+      const ownersTeam = owner.type === 'teams' ? owner : owner.getTeam();
+
+      if (currentUsersTeam === ownersTeam) return true;
+    }
+
     return false;
   },
   saveDueDate(date) {
--- a/src/js/entities-service/entities/clinicians.js
+++ b/src/js/entities-service/entities/clinicians.js
@@ -120,6 +120,14 @@ const Collection = BaseCollection.extend({
   url: '/api/clinicians',
   model: Model,
   comparator: 'name',
+  filterAssignable() {
+    const assignable = this.filter(clinician => {
+      return clinician.isActive() && clinician.get('enabled') && clinician.can('work:own');
+    });
+
+    this.reset(assignable);
+    return this;
+  },
 });
 
 export {
--- a/src/js/entities-service/entities/clinicians.js
+++ b/src/js/entities-service/entities/clinicians.js
@@ -121,12 +121,15 @@ const Collection = BaseCollection.extend({
   model: Model,
   comparator: 'name',
   filterAssignable() {
+    const clone = this.clone();
+
     const assignable = this.filter(clinician => {
       return clinician.isActive() && clinician.get('enabled') && clinician.can('work:own');
     });
 
-    this.reset(assignable);
-    return this;
+    clone.reset(assignable);
+
+    return clone;
   },
 });
 
--- a/src/js/entities-service/entities/teams.js
+++ b/src/js/entities-service/entities/teams.js
@@ -11,9 +11,7 @@ const _Model = BaseModel.extend({
   getAssignableClinicians() {
     const clinicians = Radio.request('entities', 'clinicians:collection', this.get('_clinicians'));
 
-    clinicians.filterAssignable();
-
-    return clinicians;
+    return clinicians.filterAssignable();
   },
 });
 
--- a/src/js/entities-service/entities/teams.js
+++ b/src/js/entities-service/entities/teams.js
@@ -11,11 +11,7 @@ const _Model = BaseModel.extend({
   getAssignableClinicians() {
     const clinicians = Radio.request('entities', 'clinicians:collection', this.get('_clinicians'));
 
-    const assignableClinicians = clinicians.filter(clinician => {
-      return clinician.isActive() && clinician.get('enabled') && clinician.can('work:own');
-    });
-
-    clinicians.reset(assignableClinicians);
+    clinicians.filterAssignable();
 
     return clinicians;
   },
--- a/src/js/entities-service/entities/teams.js
+++ b/src/js/entities-service/entities/teams.js
@@ -1,3 +1,4 @@
+import Radio from 'backbone.radio';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
@@ -7,6 +8,17 @@ const TYPE = 'teams';
 const _Model = BaseModel.extend({
   type: TYPE,
   urlRoot: '/api/teams',
+  getAssignableClinicians() {
+    const clinicians = Radio.request('entities', 'clinicians:collection', this.get('_clinicians'));
+
+    const assignableClinicians = clinicians.filter(clinician => {
+      return clinician.isActive() && clinician.get('enabled') && clinician.can('work:own');
+    });
+
+    clinicians.reset(assignableClinicians);
+
+    return clinicians;
+  },
 });
 
 const Model = Store(_Model, TYPE);
--- a/test/integration/patients/patient/archive.js
+++ b/test/integration/patients/patient/archive.js
@@ -317,7 +317,7 @@ context('patient archive page', function() {
     cy.clock().invoke('restore');
   });
 
-  specify('work without work:manage permission', function() {
+  specify('work with only work:owned:manage permission', function() {
     cy
       .routeCurrentClinician(fx => {
         fx.data.relationships.role = { data: { id: '66666' } };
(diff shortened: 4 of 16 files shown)


src/js/entities-service/entities/actions.js
```diff
@@ -104,10 +104,15 @@ const _Model = BaseModel.extend({
   },
   canEdit() {
     const currentUser = Radio.request('bootstrap', 'currentUser');
+    const owner = this.getOwner();
+    const currentUsersTeam = currentUser.getTeam();
+    const ownersTeam = owner.type === 'teams' ? owner : this.getOwner().getTeam();
```
nmajor25: If an action/flow doesn’t have an owner, it’ll cause issues here.

Do we need to account for that?

test/integration/patients/worklist/worklist.js
```diff
@@ -3866,7 +3866,7 @@ context('worklist page', function() {
       .should('not.exist');
   });
 
-  specify('flows without work:manage permission', function() {
+  specify('flows with only work:owned:manage permission', function() {
```
nmajor25: Before a user would either have the `work:manage` or `work:owned:manage` permission. But now they can have the `work:team:manage` permission as well.

test/integration/patients/worklist/worklist.js
```diff
@@ -3968,4 +3968,185 @@ context('worklist page', function() {
...
+
+        fx.data[2].attributes.name = 'Owned by team member';
+        fx.data[2].attributes.created_at = testTsSubtract(3);
+        fx.data[2].relationships.state = { data: { id: '33333' } };
+        fx.data[2].relationships.owner = { data: { id: '22222', type: 'clinicians' } };
+
+        fx.data[3].attributes.name = 'Owned by non team member';
+        fx.data[3].attributes.created_at = testTsSubtract(4);
+        fx.data[3].relationships.state = { data: { id: '33333' } };
+        fx.data[3].relationships.owner = { data: { id: '33333', type: 'clinicians' } };
```
nmajor25: These tests create a worklist for a user with the `work:team:manage` permission that includes 4 actions/flows in the list:

1. Owned by current clinician’s team (allowed to edit)
2. Owned by another team (not allowed to edit)
3. Owned by a fellow team member (allowed to edit)
4. Owned by a non-team member (not allowed to edit)

Then I test whether or not they are correctly editable vs. non-editable.

src/js/entities-service/entities/actions.js
paulfalgout: owner is always expected.  I think these `const`s should move to above the last `return` they're needed on because this is a very hot function on a list.

paulfalgout: actually I think the additional logic and look up should be inside:
```js
if (currentUser.can('work:team:manage') {
   const currentUsersTeam = currentUser.getTeam();
   const ownersTeam = owner.type === 'teams' ? owner : this.getOwner().getTeam();
}
```

we don't need to even attempt doing extra work until we know if we have to.

nmajor25: I like it 👍

nmajor25: That's been updated.

src/js/entities-service/entities/actions.js
```diff
@@ -104,10 +104,18 @@ const _Model = BaseModel.extend({
   },
   canEdit() {
     const currentUser = Radio.request('bootstrap', 'currentUser');
+    const owner = this.getOwner();
```
paulfalgout: I think even this.. we should call it twice.. once in `if ('work:owned:manage')` and once for team.. there should never be an instance in which both of those flags are true, so we're never saving anything by calling it once, but we do risk not needing to call it.

test/integration/patients/worklist/worklist.js
paulfalgout: I think we can remove `only` here right now. Some of these permissions require that they're separate from each other and some can be combined and I don't know that we're really accounting for those combinations right now.  Might be worth looking into ideal ways to test rbac, but it seems impossible to test all variations over time.

src/js/entities-service/entities/actions.js
nmajor25: Yup good call, that's been updated.

nmajor25:
Don't understand this coverage branch miss: https://coveralls.io/builds/62482254/source?filename=src%2Fjs%2Fentities-service%2Fentities%2Factions.js#L172.

But looking into it... 👀

test/integration/patients/sidebar/flow-sidebar.js
```diff
@@ -617,4 +617,273 @@ context('flow sidebar', function() {
...
+        fx.data[2].attributes.name = 'Owned by team member';
+        fx.data[2].attributes.updated_at = testTsSubtract(3);
+        fx.data[2].relationships.state = { data: { id: '33333' } };
+        fx.data[2].relationships.owner = { data: { id: '22222', type: 'clinicians' } };
+
+        fx.data[3].id = '4';
+        fx.data[3].attributes.name = 'Owned by non team member';
+        fx.data[3].attributes.updated_at = testTsSubtract(4);
+        fx.data[3].relationships.state = { data: { id: '33333' } };
+        fx.data[3].relationships.owner = { data: { id: '33333', type: 'clinicians' } };
```
nmajor25: Looking at these, we could probably shorten up the `action-sidebar.js`, `flow-sidebar.js`, and `patient-flow.js` tests by only testing the restrictive effects of the permission.

So in this test we'd only have two flows: `'Owned by another team'` and `'Owned by non team member'`. And make sure those flows can't be edited.

`'Owned by current clinician’s team'` and `'Owned by team member'` test that the user can edit the flow.

Idk not sure... but some of these tests got kind of big, so it'd be nice to shrink them somehow.

test/integration/patients/sidebar/flow-sidebar.js
nmajor25, line 1: In this test, I set up a patient dashboard with four flows:

1. `'Owned by current clinician’s team'`
2. `'Owned by another team'`
3. `'Owned by team member'`
4. `'Owned by non team member'`

Then I went through each flow in the patient dashboard's list by clicking through to the `/flow:id` page and testing that the user can/can't make edits in the flow sidebar.

This way I could test all four scenarios in the same test. Instead of doing separate tests for each.

To shorten this, we could possibly remove the `'Owned by current clinician’s team'` and `'Owned by team member'` flows. And only test the flows that should be restricted/non-editable.

test/integration/patients/sidebar/action-sidebar.js
nmajor25, line 1: In these tests, I created a list of 4 actions:

1. `'Owned by current clinician’s team'`
2. `'Owned by another team'`
3. `'Owned by team member'`
4. `'Owned by non team member'`

Then opened each of their action sidebars and checked their editable/non-editable state.

Tests were added for both the `/patient/:id/action/:id` and `/flow/:id/action/:id` page's action sidebars.

test/integration/patients/patient/patient-flow.js
```diff
@@ -847,6 +847,172 @@ context('patient flow page', function() {
       .should('not.exist');
   });
 
+  specify('flow with only work:team:manage permission', function() {
```
nmajor25: In this test, I set up a patient dashboard with four flows:

1. `'Owned by current clinician’s team'`
2. `'Owned by another team'`
3. `'Owned by team member'`
4. `'Owned by non team member'`

Then I went through each flow in the patient dashboard's list, click through to the `/flow:id` page, and test that the user can/can't make edits to the flow (state and/or owner in header section of the page).

It's similar to the patient flow sidebar test, but tests the header section instead of the sidebar.

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -1768,7 +1768,254 @@ context('action sidebar', function() {
...
+
+    cy
+      .get('[data-action-region]')
+      .find('button')
+      .should('have.length', 0);
+
+    cy
+      .get('[data-action-region]')
+      .should('contain', 'Permissions')
+      .and('contain', 'You are not able to change settings on actions.');
```
nmajor25: Do we want to get more specific with this permission message?

In this `work:team:manage` scenario, we'll be showing a `'You are not able to change settings on actions.'` message to the user. 

Which is misleading because the user can edit other actions, just not this one specifically. This message implies they can't edit any actions.

Not a change for this PR/story, but something I noticed along the way.

Screenshot 2023-09-06 at 3 24 22 PM

paulfalgout: Interestingly that was true when we added the message the first time as it was with when we added `work:owned:manage` permission. Should like be "on this action" instead of "on actions"

src/js/entities-service/entities/teams.js
```diff
@@ -7,6 +8,17 @@ const TYPE = 'teams';
 const _Model = BaseModel.extend({
   type: TYPE,
   urlRoot: '/api/teams',
+  getAssignableClinicians() {
```
paulfalgout, line 11: I think this function should be:
```js
getAssignableClinicians() {
  const clinicians = Radio.request('entities', 'clinicians:collection', this.get('_clinicians'));

  clinicians.filterAssignable();

  return clinicians;
},
```
where filterAssignable is something like this on the clinician:
```js
const assignable = this.filter(clinician => {
  return clinician.isActive() && clinician.get('enabled') && clinician.can('work:own');
});

this.reset(assignable);
return this;
```
so that how we determine what an assignable clinician is, is in one location... so on workspace we'd use `filterAssignable` as well

test/integration/patients/patient/archive.js
```diff
@@ -419,4 +419,101 @@ context('patient archive page', function() {
       .find('button')
       .should('not.exist');
   });
+
+  specify('work with only work:team:manage permission', function() {
```
paulfalgout: I don't think we need the "only"s in the description.  There's already contextual knowledge required as this user does have other permissions, so this means "only" one `work:__:manage` permission, and currently it is a requirement that all users only have one of that type of permission and it's doubtful we'd ever support a combination.  Although perhaps ironically, they would functionally work to have any combination.. But the more permissible will always win.

test/integration/patients/sidebar/flow-sidebar.js
paulfalgout: Yeah it's funny..  I'm not sure on where the consistency is.  Semantically, the default stuff we shouldn't need to test for is the least permissive state, and a "user _with_ a permission" would be tested to see if they can do things, but as is the nature of things, most of our tests are built from full permissions and we're adding tests for restrictions.  

I don't know the best way to do this long term, but it does seem like in an ideal world, the "generic" tests would be with as little permissions as possible, and all other functionality is tested in the context of permissions, but that seems like an overhaul.  Possibly we should be testing the least permissive scenario and the most permissive scenario for functionality, and then beyond that, the most limited tests possible for combinations in between.  

For instance these tests are all checking to make sure that both the state and the owner are available or not in the sidebar.. but we have already tested that those things would be available or not based on the results of the `flow.canEdit` function.. For other permissions that change the result of `canEdit` we should have the most minimal tests possible to determine that result.  We don't need each test of a `canEdit` result to re-test everything that happens if that is `true` or `false`

paulfalgout (changes requested):
Summary of change requests:
- Move the logic for determining an assignable clinician into the clinician rather than in workspace and team.
- Test the interface results for `[action/flow].canEdit` once in it's most permissible and least permissible permissions.  For every other permission that affects the result of `canEdit` use the most limited test possible to determine the result of `canEdit`

test/integration/patients/patient/archive.js
shadowhand: > Although perhaps ironically, they would functionally work to have any combination.. **But the more permissible will always win.**

(Emphasis added by me) this last sentence is the key point, the most permissive permission (heh) always wins. We always structure our permission checks this way on backend, looking for the broadest permission first.

test/integration/patients/sidebar/flow-sidebar.js
shadowhand: Not sure this is helpful, but the way we test permissions in BE is to create a fixture with the smallest set of permissions possible for a scenario. We don't bother testing the whole matrix, because (as above) the most permissive always takes precedence.

test/integration/patients/sidebar/action-sidebar.js
shadowhand: Agree that the messaging should be specific to _this_ action, instead of a general statement about actions.

src/js/entities-service/entities/clinicians.js
```diff
@@ -120,6 +120,14 @@ const Collection = BaseCollection.extend({
   url: '/api/clinicians',
   model: Model,
   comparator: 'name',
+  filterAssignable() {
```
paulfalgout, line 123: @nmajor25 I'm running into a place where we have a similar pattern of wanting to filter a collection, but this pattern is mutating.  Now I think we should make it not mutating like:
```js
const assignable = ...;
return this.clone().reset(assignable);
```

Then the uses would need to change to:
```js
return clinicians.filterAssignable();
```

I don't think we need to update the release, but we should update this branch.

test/integration/patients/sidebar/flow-sidebar.js
nmajor25: For a start, I just pushed a commit to refactor the worklist/schedule tests.

In those tests, I stripped down the test to only include the two restrictive scenarios:

1. `Owned by another team`
2. `Owned by non team member`

And then made the test simply verify that those two actions/flows in the list are in a `canEdit = false` state. Everything else in the test was removed.

Is that all that needs to be done to refactor these tests?

*One caveat is that I left the `Owned by current clinician’s team` scenario in the one of the worklist tests so I could test the contents of the owner droplist (where it should only include the user's team and team members). That can be removed later when we create a component test for the owner droplist.*

test/integration/patients/sidebar/action-sidebar.js
nmajor25: This was updated.

src/js/entities-service/entities/clinicians.js
nmajor25: I can't get this to work... 

When I update the `filterAssignable()` to do the `return this.clone().reset(assignable)`:

```js
  filterAssignable() {
    const assignable = this.filter(clinician => {
      return clinician.isActive() && clinician.get('enabled') && clinician.can('work:own');
    });

    return this.clone().reset(assignable);
  }
```

This error is thrown:

Screenshot 2023-09-08 at 10 00 17 AM

So for some reason when we clone, we lose all the collection methods? At least that's my guess as to why this error is happening.

paulfalgout: Nope, sorry the `reset` returns an array of the new models.

So this has to be:
```js
const clone = this.clone();
clone.reset(assignable);
return clone;
```

test/integration/patients/worklist/worklist.js
```diff
@@ -4076,12 +4076,17 @@ context('worklist page', function() {
...
       .routeFlows(fx => {
-        fx.data = _.sample(fx.data, 2);
+        fx.data = _.sample(fx.data, 3);
 
-        fx.data[0].attributes.name = 'Owned by another team';
+        fx.data[0].attributes.name = 'Owned by current clinician’s team';
         fx.data[0].attributes.created_at = testTsSubtract(1);
         fx.data[0].relationships.state = { data: { id: '33333' } };
-        fx.data[0].relationships.owner = { data: { id: '22222', type: 'teams' } };
+        fx.data[0].relationships.owner = { data: { id: '11111', type: 'teams' } };
```
nmajor25, line 4084: This should fix the coverage drop in the flow's `canEdit()` function.

I *think* this is the only coverage drop related to this PR. I believe the rest are due to forms, outreach, `flow.states`, and the initial `canEdit()` implementation.

Well, there's also one in the `/behaviors/picklist-transport.js` that I don't understand. That's potentially related to this one? Not sure...

paulfalgout (approved):
I think https://coveralls.io/builds/62577152/source?filename=src%2Fjs%2Fentities-service%2Fentities%2Factions.js#L167 is a legitimate coverage loss _somehow_ but we have enough coverage issues to check, let's get this in and we can clean that one branch loss with the others.

PR #1125, opened 2023-08-30
Set popWidth option of pop region component before calculating height

paulfalgout:
Shortcut Story ID: [sc-39090]

This was bugging because it was measuring the height of the picker prior to constraining it's width... which meant it wasn't nearly as tall as it needed to be.

- [ ] Figure out a way to test this in the component testing


nmajor25 (approved):
Sweet, nice to have this fixed 😌

PR #1126, opened 2023-08-30
Use `/flow` route as a fallback when using the form `Go Back` link or `Submit + Go Back` button

nmajor25:
Shortcut Story ID: [sc-38686]

When a user loads a form URL directly (no browser route history) and then clicks the `Go Back` link or `Submit + Go Back` button, we take them to the patient's dashboard as the fallback.

Instead we want to take them to the `/flow` page (when possible). On an action form, if the action doesn't belong to a flow, we route them to the patient dashboard.

Since patient forms don't have an action or flow, they'll continue to route to the patient dashboard in this situation.

### DIFF
--- a/test/integration/forms/action.js
+++ b/test/integration/forms/action.js
@@ -1247,6 +1247,9 @@ context('Patient Action Form', function() {
       .routeFormDefinition()
       .routeFormResponse()
       .routeFormActionFields()
+      .routePatientByFlow()
+      .routeFlow()
+      .routeFlowActions()
       .visit('/patient-action/1/form/11111')
       .wait('@routeAction')
       .wait('@routePatientByAction');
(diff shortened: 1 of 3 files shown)


nmajor25:
If you load a preview form URL directly (i.e. `/form/:id/preview`) and then click the `Go Back` link, weird stuff happens where  the app is routed to the worklist URL where the left nav is showing, but the form itself is still showing instead of the worklist.

I'm not sure what context the form preview is used in so I don't know how big of a deal that is or how high priority it is to investigate.


https://github.com/RoundingWell/care-ops-frontend/assets/35355575/458460fe-0065-442c-b998-ce3d27f5e058

test/integration/forms/action.js
nmajor25, line 1: I created duplicate stripped down versions of the `"routing to form"` and `"submit and go back button"`, which are used to test that the `Go Back` or `Submit + Go Back` buttons go to either the `/flow/:id` or `/patient/dashboard/:id` routes.

1. `"routing to form"`: action belongs to a flow, so the `/flow/:id` route is used.
2. `"routing to form - action without a flow"`: action doesn't belong to a flow, so the `/patient/dashboard/:id` route is used. Otherwise a duplicate of the `"routing to form"` test with anything unnecessary stripped out.
3. `"submit and go back button"`: action belongs to a flow, so the `/flow/:id` route is used.
4. `"submit and go back button - action without a flow"`: action doesn't belong to a flow, so the `/patient/dashboard/:id` route is used. Otherwise a duplicate of the `"submit and go back button"` test with anything unnecessary stripped out.

paulfalgout:
Going to wait on getting the drafts in as this is going to cause merge conflicts and need test updates to account for the new endpoints

paulfalgout (approved):

shadowhand:
😻

PR #1131, opened 2023-09-11
Fix issues with draft submission and history display

paulfalgout:
Shortcut Story ID: [sc-40419] [sc-40447]

Includes a bit of clean up, but largely adds an existing id to draft or submissions if there's a latestDraft.

### DIFF
--- a/src/js/entities-service/entities/form-responses.js
+++ b/src/js/entities-service/entities/form-responses.js
@@ -67,6 +67,14 @@ const Collection = BaseCollection.extend({
   getFirstSubmission() {
     return this.find({ status: FORM_RESPONSE_STATUS.SUBMITTED });
   },
+  filterSubmissions() {
+    const clone = this.clone();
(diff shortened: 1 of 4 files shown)


src/js/entities-service/entities/form-responses.js
```diff
@@ -58,15 +53,16 @@ const Collection = BaseCollection.extend({
...
-  getSubmission() {
+  getFirstSubmission() {
     return this.find({ status: FORM_RESPONSE_STATUS.SUBMITTED });
   },
+  filterSubmissions() {
+    const submissions = this.filter({ status: FORM_RESPONSE_STATUS.SUBMITTED });
+
+    const clone = this.clone();
+    clone.reset(submissions);
+    return clone;
```
shadowhand:
```suggestion
    const clone = this.clone();
    const submissions = this.filter({ status: FORM_RESPONSE_STATUS.SUBMITTED });

    clone.reset(submissions);

    return clone;
```

nmajor25 (approved):

PR #1133, opened 2023-09-12
Remove form draft from loading after clearing submission

paulfalgout:
Shortcut Story ID: [sc-40620]

### DIFF
--- a/test/integration/forms/action.js
+++ b/test/integration/forms/action.js
@@ -296,6 +296,9 @@ context('Patient Action Form', function() {
         fx.data.id = '1';
         fx.data.relationships.form.data = { id: '11111' };
         fx.data.relationships['program-action'] = { data: { id: '11111' } };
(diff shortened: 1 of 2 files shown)


test/integration/forms/action.js
```diff
@@ -305,7 +308,22 @@ context('Patient Action Form', function() {
...
       })
       .routeFormByAction(_.identity, '11111')
-      .routeLatestFormResponse()
+      .routeLatestFormResponse(() => {
+        return {
+          data: {
+            id: '1',
+            attributes: {
+              status: 'draft',
+              created_at: testTsSubtract(1),
```
paulfalgout, line 317: This adds a "cloud draft" older that localstorage.  if the `this.latestResponse = null;` is removed, this test fails and shows `bazinga` in the foo field

nmajor25 (approved):

PR #1134, opened 2023-09-12
Support refresh tokens and utilize a localstorage cacheLocation

paulfalgout:
Shortcut Story ID: [sc-40637]

We'll need to set these two appconfig.json settings for all tenants along with flipping that setting everywhere.

### DIFF
--- a/src/assets/appconfig.json
+++ b/src/assets/appconfig.json
@@ -9,7 +9,9 @@
     "clientId": "UVGIL27XqwGsJwsTfEyz50c1qg3Ll1p6",
     "authorizationParams": {
       "connection": "google-oauth2"
-    }
+    },
+    "cacheLocation": "localstorage",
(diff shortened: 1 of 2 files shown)


src/assets/appconfig.json
```diff
@@ -9,7 +9,9 @@
     "clientId": "UVGIL27XqwGsJwsTfEyz50c1qg3Ll1p6",
     "authorizationParams": {
       "connection": "google-oauth2"
-    }
+    },
+    "cacheLocation": "localstorage",
```
shadowhand, line 13: Does BE need to define this for deploys?

shadowhand (approved):

src/assets/appconfig.json
paulfalgout: yep

PR #1135, opened 2023-09-12
Fix `canEdit()` related coverage drop

nmajor25:
Shortcut Story ID: [sc-40641]

Missing branch coverage on this [line of code](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/entities-service/entities/actions.js#L167).

For the `Add an Attachment` button to show in the action sidebar, these things need to exist:

1. `upload_attachments: true` org setting.
2. `included.program_action.allowed_uploads: ['pdf']` data on the action.
3. Action must be in a `canEdit = true` state.

We need a test where the `canEdit` status of an action is the only thing stopping the `Add an Attachment` button from showing.

### DIFF
--- a/test/integration/patients/sidebar/action-sidebar.js
+++ b/test/integration/patients/sidebar/action-sidebar.js
@@ -1654,7 +1654,7 @@ context('action sidebar', function() {
         return fx;
       })
       .routeSettings(fx => {
-        fx.data.push({ id: 'upload_attachments', attributes: { value: false } });
+        fx.data.push({ id: 'upload_attachments', attributes: { value: true } });
 
         return fx;
       })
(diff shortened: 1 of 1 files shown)


test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -1712,8 +1721,7 @@ context('action sidebar', function() {
 
     cy
       .get('[data-action-region]')
-      .find('.js-input')
-      .should('have.length', 2);
```
nmajor25, line 1716: Since this action now has `_program_action` data, this action is no longer ad hoc. 

So it won’t have the action name textfield anymore when its in the `canEdit = true` state. Only the details text field.

Screenshot 2023-09-12 at 1 32 49 PM

PR #1138, opened 2023-09-14
Outreach: change form submit success icon

nmajor25:
Shortcut Story ID: [sc-40841]

Changed from this:

Screenshot 2023-09-14 at 10 12 46 AM

To this:

Screenshot 2023-09-14 at 10 52 26 AM

PR #1140, opened 2023-09-14
Outreach: `.scss` refactoring

nmajor25:
Shortcut Story ID: [sc-39249]

### DIFF
--- a/src/js/outreach/views/dialog.scss
+++ b/src/js/outreach/views/dialog.scss
@@ -23,21 +23,26 @@
 
 .dialog__icon {
   color: $black-20;
-  margin-bottom: 24px;
-  text-align: left;
+  margin: 0 auto 24px;
+
+  &.dialog__icon--success {
+    color: $success-color;
+  }
+
+  &.dialog__icon--blue {
+    color: color(blue, dark);
+  }
+
+  &.dialog__icon--warn {
+    color: $warning-color;
+  }
+
+  &.dialog__icon--error {
+    color: $error-color;
+  }
 
   .icon {
     font-size: 48px;
     width: 1em;
   }
-}
-
-.dialog__icon--success {
-  color: $success-color;
-  margin: 0 auto 24px;
-}
-
-.dialog__icon--error {
-  color: $error-color;
-  margin: 0 auto 24px;
 }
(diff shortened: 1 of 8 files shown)


src/js/outreach/views/opt-in.scss
```diff
@@ -44,12 +44,6 @@
   display: block;
   padding: 12px 16px;
   width: 100%;
-
-  &.has-error {
-    border-color: $error-color;
-    color: $error-color;
-    margin-bottom: 6px;
-  }
```
nmajor25, line 52: We're only doing error styling for inputs on the verify code view. The opt in view doesn't show any errors for input fields. So this styling never gets used.

src/js/outreach/views/dialog.scss
```diff
@@ -10,10 +10,6 @@
   text-transform: uppercase;
 }
 
-.icon {
-  font-size: 32px;
-}
```
nmajor25, line 15: Only thing using this was the form submit success icon. But it's no longer being used after that was changed in #1138.

All icons are now using `.dialog__icon > .icon`.

src/js/outreach/views/dialog.scss
```diff
@@ -10,104 +10,38 @@
...
-  padding: 0 16px;
-  position: relative;
-  text-align: center;
-  width: 100%;
 
-  &.has-error {
-    border-color: $error-color;
-    color: $error-color;
-    margin-bottom: 6px;
+  &.dialog__icon--success {
```
paulfalgout: I don't think the `&`s are necessary here.  these classes look to be BEM'd enough that they do not need to be nested.  They just need to be either after the `dialog_icon` in the cascade or since it's all in the same file these could extend the base such that only this class is necessary in the DOM.  Ideally `&` is saved for adding states generally prefixed by something like `&.has-`, `&.is-` etc

paulfalgout (changes requested):

src/js/outreach/views/dialog.scss
nmajor25: I went the `@extend` route. Works well. I like that we now only need the `.dialog__icon--` classname in the dom. Instead of needing `.dialog__icon .dialog__icon--`. Good call 👍

paulfalgout (approved):

PR #1141, opened 2023-09-15
Update dependencies

nmajor25:
Shortcut Story ID: [sc-40888]


package.json
```diff
@@ -178,7 +178,7 @@
     "cross-env": "^7.0.3",
     "css-loader": "^6.2.0",
     "cssnano": "^6.0.0",
-    "cypress": "^12.14.0",
+    "cypress": "^13.2.0",
```
nmajor25: I think Cypress v13 is causing tests to fail.

But only on CircleCI. Everything looks good when I run them locally. So I'm not sure what's going on.

On the test replays it looks like the CSS isn't being loaded properly:

Screenshot 2023-09-15 at 2 29 04 PM

paulfalgout: So what's cool about v13 is that we should be able to inspect the source maybe on cloud.cypress.io.. I'm not entirely sure though how in depth it gets or if it is still just essentially screenshots..

nmajor25:
@paulfalgout there's two more outdated packages we could update:

1. `@cypress/webpack-preprocessor`: `v5.17.1` => `v6.0.0`
    * Nothing new, they accidentally created a release: https://github.com/cypress-io/cypress/blob/develop/npm/webpack-preprocessor/CHANGELOG.md

2. `stylelint-config-standard-scss`: `10.0.0` => `11.0.0`
    * Nothing too significant: https://github.com/stylelint-scss/stylelint-config-standard-scss/blob/main/CHANGELOG.md#1100

paulfalgout:
Yeah we should do all the updates.. if we have to wait on cypress we can pull that one back.  I did fill out a report with them

nmajor25:
@paulfalgout okay I've got Cypress switched back to use the newest iteration of v12. And `@cypress/webpack-preprocessor` and `stylelint-config-standard-scss` have been updated.

The `npm outdated` command will now show only `formio` and `cypress` as needing to be updated.

paulfalgout:
🥁 🧻

test/integration/services/alert.js
```diff
@@ -62,57 +62,63 @@ context('Alert Service', function() {
...
       });
 
-    cy
-      .clock();
-
     cy
       .get('.alert-box')
       .find('.js-dismiss')
       .click()
-      .click()
```
nmajor25, line 89: There were two `.click()`s here... I removed it, but maybe that was needed?

test/integration/services/alert.js
```diff
@@ -62,57 +62,63 @@ context('Alert Service', function() {
...
       .then(() => {
-        expect(options.onUndo).to.be.calledOnce;
-      });
+        expect(onUndo).to.be.calledOnce;
+      })
+      .tick(1000);
+
+    cy
+      .get('.alert-box')
+      .should('not.exist');
```
nmajor25, line 122: I added this. Just checks that the alert is no longer there after the undo button is clicked. Before we were only checking that the radio was sent/called.

test/integration/services/alert.js
nmajor25: Well I guess this is why there were 2 clicks there: https://coveralls.io/builds/62792353/source?filename=src%2Fjs%2Fviews%2Fglobals%2Falert-box%2Falert-box_views.js#L57.

Is for coverage purposes.

paulfalgout (changes requested):
Alright.. I think we should keep everything but the `is-active` removal and upgrading to Cypress 13 for now.  the other stuff is either good, or won't hurt for now.

nmajor25:
@paulfalgout I reverted the`.is-active` change and downgraded back to Cypress v12.

I also did a quick update of the `sass` and `autoprefixer` packages. Since those are new since the last time I did a `npm update` a few days ago.

PR #1146, opened 2023-09-18
Add temporary patch for draft history bug

paulfalgout:
Shortcut Story ID: [sc-41025]

@shadowhand I also snuck in changing the cloud draft timeout to 15 seconds as was discussed over the weekend.

### DIFF
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -21,7 +21,7 @@ export default App.extend({
     return `form${ this.getOption('form').id }`;
   },
   initialize(options) {
-    this.updateDraft = debounce(this.updateDraft, 30000);
+    this.updateDraft = debounce(this.updateDraft, 15000);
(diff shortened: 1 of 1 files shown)


src/js/services/forms.js
```diff
@@ -151,6 +156,9 @@ export default App.extend({
       });
   },
   fetchFormStoreSubmission({ submission }) {
+    // NOTE: Remove after 2023/10/19
+    patchEvernorthHistory(submission);
```
zfixler, line 160: Does `fetchFormStoreSubmission` cover both a cloud draft and loading from local storage?

paulfalgout: yep, this is the only way a submission is preloaded into a form, right before it's sent to the form app.

zfixler (approved):

nmajor25 (approved):

PR #1147, opened 2023-09-18
Bugfix: add menu showing empty action/flow items

nmajor25:
Shortcut Story ID: [sc-40923]


test/integration/patients/patient/patient-flow.js
nmajor25, line 1: Added two actions for the add action menu where one is `published: false` and the other has `behavior: automated`. Then verified in the test that they don't show up.

The patient dashboard tests are already testing those scenarios. So those don't need to be updated.

paulfalgout (approved):

PR #1148, opened 2023-09-21
Outreach: remove unused code that caused coverage drops

nmajor25:
Shortcut Story ID: [sc-39330]

Remove unused code that's causing drops in code coverage.

### DIFF
--- a/src/js/outreach/apps/verify_app.js
+++ b/src/js/outreach/apps/verify_app.js
@@ -20,8 +20,6 @@ export default App.extend({
     return getPatientInfo({ actionId: this.actionId });
   },
   onFail(options, response) {
-    if (response.status >= 500) return;
-
     const dialogView = new DialogView();
     this.showView(dialogView);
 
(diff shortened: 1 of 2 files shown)


src/js/outreach/apps/verify_app.js
```diff
@@ -20,8 +20,6 @@ export default App.extend({
     return getPatientInfo({ actionId: this.actionId });
   },
   onFail(options, response) {
-    if (response.status >= 500) return;
```
nmajor25, line 23: This handles errors on the `GET /api/outreach?filter[action]=` request.

If that request `500`s, this `onFail()` doesn’t seem to get called. It’s handled directly by the error app. So I believe this will always be `false` and therefore not needed.

src/js/outreach/views/verify_views.js
```diff
@@ -77,7 +77,7 @@ const VerifyCodeView = View.extend({
     const first = value.charAt(0);
     const rest = value.substring(1);
 
-    event.target.value = first ?? '';
+    event.target.value = first;
```
nmajor25, line 80: I can’t think of a situation where `first` would be `null` or `undefined` here? 

`event.target` would have to be `null`/`undefined` somehow? Is that possible here? This is an event watcher function, so I would imagine it's not possible.

I can’t remember why I put that in there originally.

PR #1149, opened 2023-09-21
Outreach: refactor error handling in verify app

nmajor25:
Shortcut Story ID: [sc-39330]

### DIFF
--- a/src/js/outreach/apps/verify_app.js
+++ b/src/js/outreach/apps/verify_app.js
@@ -52,7 +52,9 @@ export default App.extend({
         .then(() => {
           this.showVerifyCodeView();
         })
-        .catch(() => {
+        .catch(response => {
+          if (response.status >= 500) return;
+
           this.showDialogErrorView();
         });
     });
@@ -71,12 +73,9 @@ export default App.extend({
           this.stop({ isVerified: true });
         })
         .catch(response => {
-          if (response.status === 403) {
-            this.showVerifyCodeView(true);
-            return;
-          }
+          if (response.status >= 500) return;
 
-          this.showDialogErrorView();
+          this.showVerifyCodeView(true);
         });
     });
 
--- a/src/js/outreach/apps/verify_app.js
+++ b/src/js/outreach/apps/verify_app.js
@@ -7,10 +7,10 @@ import {
   VerifyCodeView,
   AlreadySubmittedView,
   NotAvailableView,
+  GeneralErrorView,
 } from 'js/outreach/views/verify_views';
 
 import { DialogView } from 'js/outreach/views/dialog_views';
-import { ErrorView } from 'js/outreach/views/error_views';
 
 export default App.extend({
   onBeforeStart({ actionId }) {
@@ -55,7 +55,7 @@ export default App.extend({
         .catch(response => {
           if (response.status >= 500) return;
 
-          this.showDialogErrorView();
+          this.showGeneralErrorView();
         });
     });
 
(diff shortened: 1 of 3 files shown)


src/js/outreach/apps/verify_app.js
nmajor25, line 1: If a `500` error occurs on either the `POST /api/outreach/otp` (create new OTP code) or `POST /api/outreach/auth` (verify OTP code) requests, the [ErrorView](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/outreach/views/error_views.js#L6-L18) was being instantiated twice.

In these two spots:

1. The base fetcher [here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/base/fetch.js#L98-L101).
2. And then the `.catch()` in this verify app file.

So the goal is for the error app to completely handle `500` errors. And any other errors be handled by this verify app.

src/js/outreach/apps/verify_app.js
```diff
@@ -52,7 +52,9 @@ export default App.extend({
         .then(() => {
           this.showVerifyCodeView();
         })
-        .catch(() => {
+        .catch(response => {
+          if (response.status >= 500) return;
+
           this.showDialogErrorView();
         });
```
nmajor25, line 57: This request either returns `204` (success), `400`, or a `500` server error.

With this code update, we’re bailing out of the `.catch()` if it’s a `500` error and letting the error app handle it.

But what should we do if it’s a `400` error? I’m not sure what scenario would cause a `400` error to be returned when a user is requesting a new code be created. So I don't know what we want to show to the user here.

This current code will show the same `Uh oh…` view as the `500` error without routing to the `/outreach/500` URL. Which means the `Try Again` link won’t work, since `history.back()` will have nothing to go to.

So I guess we could either route to the `/outreach/500` via a `Radio.trigger('event-router', 'error’)`. Or create a new view in the `/outreach/views/verify_views.js` file that has a different error message.

src/js/outreach/apps/verify_app.js
```diff
@@ -71,12 +73,9 @@ export default App.extend({
...
         .catch(response => {
-          if (response.status === 403) {
-            this.showVerifyCodeView(true);
-            return;
-          }
+          if (response.status >= 500) return;
 
-          this.showDialogErrorView();
+          this.showVerifyCodeView(true);
         });
```
nmajor25, line 77: This is request either returns `201` (success), `403`, or a `500` server error.

If `500`, we’ll bail from this `.catch()` and have the error app handle things. Otherwise, we can assume it was the `403` incorrect code error.

test/integration/outreach/outreach.js
nmajor25, line 1: I moved these `500` error scenarios into the `”500 error”` test and removed the `verify_app` specific one:

1. When initially loading the verify app (`/api/outreach?filter[action]=`).
2. When the user requests a new code be created (`/api/outreach/otp`).
3. When the user attempts to verify their code (`/api/outreach/auth`).

src/js/outreach/apps/verify_app.js
paulfalgout: The previous error page said something about reloading the page, but that didn't work for 500s because it would reload the 500 route.  Really we need one view for 500s and the other view for 400s that can be the original message instead of using one solution for both.

paulfalgout (changes requested):
We should split the errors, but otherwise a good improvement.

src/js/outreach/apps/verify_app.js
nmajor25: I added a general error for the verify app. Which is now shown for that `400` error. Or any other errors besides `500` on the `/api/outreach/otp` request.

Outreach coverage is now at 100% 😄

paulfalgout (approved):

PR #1154, opened 2023-09-25
Outreach: in tests, verify POST requests send correct data

nmajor25:
Shortcut Story ID: [sc-41200]

### DIFF
--- a/test/integration/outreach/outreach.js
+++ b/test/integration/outreach/outreach.js
@@ -45,7 +45,7 @@ context('Outreach', function() {
     cy
       .get('.js-phone')
       .clear()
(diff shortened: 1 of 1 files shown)


test/integration/outreach/outreach.js
```diff
@@ -45,7 +45,7 @@ context('Outreach', function() {
     cy
       .get('.js-phone')
       .clear()
-      .type('+18887771234');
+      .type('8887771234');
```
nmajor25, line 48: By removing the `+1` here, we can test that the `+1` formatting is sent in the api request whether the user types it or not.

paulfalgout (approved):

PR #1156, opened 2023-09-27
Outreach: move boilerplate `intercept`s to `test/support/api` directory

nmajor25:
Shortcut Story ID: [sc-41201]

### DIFF
--- a/test/support/api/outreach.js
+++ b/test/support/api/outreach.js
@@ -1,0 +1,20 @@
+import _ from 'underscore';
+
+Cypress.Commands.add('routeOutreachStatus', (mutator = _.identity) => {
+  cy.route({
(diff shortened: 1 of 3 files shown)


test/support/api/outreach.js
```diff
@@ -0,0 +1,20 @@
...
+    url: '/api/outreach?*',
+    response() {
+      return mutator({
+        data: {
+          type: 'Outreach',
+          id: '11111',
+          attributes: {
+            phone_end: '1234',
+          },
+        },
```
nmajor25, line 14: Worth it to move this to it's own fixture file?

Seemed overkill since it would only be one fixture, but not sure...

paulfalgout (approved):

PR #1157, opened 2023-09-28
Refactor cypress `.route()` usage to `.intercept()` in `test/support/api` files

nmajor25:
Shortcut Story ID: [sc-38071]

`.route()` was deprecated by Cypress, this switches the `test/support/api` files to `.intercept()`.

First phase was complete in https://github.com/RoundingWell/care-ops-frontend/pull/1078.

### DIFF
--- a/test/helpers/json-api.js
+++ b/test/helpers/json-api.js
@@ -1,6 +1,8 @@
 import _ from 'underscore';
 
 function getResource(data, type) {
+  data = JSON.parse(JSON.stringify(data));
+
   if (_.isArray(data)) {
     return _.map(data, _.partial(getResource, _, type));
   }
--- a/test/integration/patients/worklist/bulk-edit.js
+++ b/test/integration/patients/worklist/bulk-edit.js
@@ -4,8 +4,6 @@ import formatDate from 'helpers/format-date';
 import { testTs, testTsSubtract } from 'helpers/test-timestamp';
 import { testDateAdd } from 'helpers/test-date';
 
-import fxFlows from 'fixtures/collections/flows';
-
 const tomorrow = testDateAdd(1);
 
 context('Worklist bulk editing', function() {
@@ -42,11 +40,15 @@ context('Worklist bulk editing', function() {
         fx.data[3].attributes.due_time = '07:00:00';
         fx.data[3].relationships.state = { data: { id: '22222' } };
 
-
         fx.included.push({
           id: '1',
           type: 'flows',
-          attributes: _.extend({}, _.sample(fxFlows), { name: 'Test Flow' }),
+          attributes: {
+            name: 'Test Flow',
+            details: null,
+            created_at: testTs(),
+            updated_at: testTs(),
+          },
         });
 
         return fx;
@@ -665,7 +667,12 @@ context('Worklist bulk editing', function() {
         fx.included.push({
           id: '1',
           type: 'flows',
-          attributes: _.extend({}, _.sample(fxFlows), { name: 'Test Flow' }),
+          attributes: {
+            name: 'Test Flow',
+            details: null,
+            created_at: testTs(),
+            updated_at: testTs(),
+          },
         });
 
         return fx;
@@ -1143,8 +1150,12 @@ context('Worklist bulk editing', function() {
         fx.included.push({
           id: '1',
           type: 'flows',
-          attributes: _.extend({}, _.sample(fxFlows), { name: 'Test Flow' })
-          ,
+          attributes: {
+            name: 'Test Flow',
+            details: null,
+            created_at: testTs(),
+            updated_at: testTs(),
+          },
         });
 
         return fx;
--- a/test/integration/patients/worklist/bulk-edit.js
+++ b/test/integration/patients/worklist/bulk-edit.js
@@ -12,15 +12,6 @@ context('Worklist bulk editing', function() {
   specify('date and time components', function() {
     cy
       .routeActions(fx => {
-        const flowInclude = {
-          id: '1',
-          type: 'flows',
-          attributes: _.extend(_.sample(fxFlows), {
-            name: 'Test Flow',
-            id: '1',
-          }),
-        };
-
         fx.data = _.sample(fx.data, 4);
 
         fx.data[0].id = '1';
@@ -52,7 +43,11 @@ context('Worklist bulk editing', function() {
         fx.data[3].relationships.state = { data: { id: '22222' } };
 
 
-        fx.included.push(flowInclude);
+        fx.included.push({
+          id: '1',
+          type: 'flows',
+          attributes: _.extend({}, _.sample(fxFlows), { name: 'Test Flow' }),
+        });
 
         return fx;
       })
@@ -606,15 +601,6 @@ context('Worklist bulk editing', function() {
   specify('bulk actions editing', function() {
     cy
       .routeActions(fx => {
-        const flowInclude = {
-          id: '1',
-          type: 'flows',
-          attributes: _.extend(_.sample(fxFlows), {
-            name: 'Test Flow',
-            id: '1',
-          }),
-        };
-
         fx.data = _.sample(fx.data, 4);
         fx.data[0] = {
           id: '1',
@@ -676,7 +662,11 @@ context('Worklist bulk editing', function() {
         fx.data[3].attributes.created_at = testTsSubtract(2);
         fx.data[3].relationships.state = { data: { id: '33333' } };
 
-        fx.included.push(flowInclude);
+        fx.included.push({
+          id: '1',
+          type: 'flows',
+          attributes: _.extend({}, _.sample(fxFlows), { name: 'Test Flow' }),
+        });
 
         return fx;
       })
@@ -1143,15 +1133,6 @@ context('Worklist bulk editing', function() {
     cy
       .routeFlows()
       .routeActions(fx => {
-        const flowInclude = {
-          id: '1',
-          type: 'flows',
-          attributes: _.extend(_.sample(fxFlows), {
-            name: 'Test Flow',
-            id: '1',
-          }),
-        };
-
         fx.data = _.sample(fx.data, 2);
         fx.data[0].id = '1';
         fx.data[0].relationships.state = { data: { id: '22222' } };
@@ -1159,7 +1140,12 @@ context('Worklist bulk editing', function() {
         fx.data[1].id = '3';
         fx.data[1].relationships.state = { data: { id: '55555' } };
 
-        fx.included.push(flowInclude);
+        fx.included.push({
+          id: '1',
+          type: 'flows',
+          attributes: _.extend({}, _.sample(fxFlows), { name: 'Test Flow' })
+          ,
+        });
 
         return fx;
       })
--- a/test/integration/patients/worklist/bulk-edit.js
+++ b/test/integration/patients/worklist/bulk-edit.js
@@ -4,6 +4,8 @@ import formatDate from 'helpers/format-date';
 import { testTs, testTsSubtract } from 'helpers/test-timestamp';
 import { testDateAdd } from 'helpers/test-date';
 
+import fxFlows from 'fixtures/collections/flows';
+
 const tomorrow = testDateAdd(1);
 
 context('Worklist bulk editing', function() {
@@ -13,7 +15,7 @@ context('Worklist bulk editing', function() {
         const flowInclude = {
           id: '1',
           type: 'flows',
-          attributes: _.extend(_.sample(this.fxFlows), {
+          attributes: _.extend(_.sample(fxFlows), {
             name: 'Test Flow',
             id: '1',
           }),
@@ -607,7 +609,7 @@ context('Worklist bulk editing', function() {
         const flowInclude = {
           id: '1',
           type: 'flows',
-          attributes: _.extend(_.sample(this.fxFlows), {
+          attributes: _.extend(_.sample(fxFlows), {
             name: 'Test Flow',
             id: '1',
           }),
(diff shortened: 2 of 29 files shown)


nmajor25:
@paulfalgout [worklist page > action sorting - patient field](https://cloud.cypress.io/projects/ep9zr6/runs/5589/overview/d5413fcb-9311-45a4-9498-0e1cf12dad4d?reviewViewBy=FAILED&utm_source=github&utm_medium=failed&utm_campaign=view%20test)

It's the only failing test holding up this PR at this point. So still looking into it.

It's failing locally as well, but only when it's ran with all the other worklist tests. If it runs with `.only`, it never fails. So I'm perplexed.

test/integration/patients/worklist/worklist.js
```diff
@@ -2819,6 +2819,14 @@ context('worklist page', function() {
...
     cy
       .routesForPatientAction()
+      .routeSettings(fx => {
+        const sortingSettings = _.find(fx.data, setting => setting.id === 'sorting');
+        _.each(sortingSettings.attributes.value, function(sortMethod) {
+          sortMethod.sort_type = 'alphanumeric';
+        });
+
+        return fx;
+      })
```
nmajor25: I don't know why this happening, but there's a [previous test](https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/integration/patients/worklist/worklist.js#L2375) in this file where `sortMethod.sort_type` is set to `numeric`. 

Which then carries over to this test, where the `sort_type` remains as `numeric` and causes the test to fail intermittently.

This started happening because of the different way we are now importing the fixtures in the `test/support/api/settings.js` file. A change in this PR.

It was originally like this when using `.route()`:

```js
Cypress.Commands.add('routeSettings', (mutator = _.identity) => {
  cy
    .fixture('test/settings').as('fxTestSettings');

  cy.route({
  
    . . .

  })
});
```

Which is changed in this PR to this when using `.intercept()`:

```js
import fxTestSettings from 'fixtures/test/settings';

Cypress.Commands.add('routeSettings', (mutator = _.identity) => {
  cy.intercept({

    . . .

  })
});
```

paulfalgout: We should ensure we're not introducing mutating elsewhere, rather than add it here I think.  That is an interesting concern though.. we may need to be more aware of mutation where we've been lax up till now in tests.

test/integration/patients/worklist/worklist.js
```diff
@@ -3761,7 +3768,7 @@ context('worklist page', function() {
         const flowInclude = {
           id: '1',
           type: 'flows',
-          attributes: _.extend(_.sample(this.fxFlows), {
+          attributes: _.extend(_.sample(fxFlows), {
```
paulfalgout: even this is mutating..  I suppose we need to figure out how imported fixtures work in cypress an go from there.

paulfalgout (changes requested):
We should prove whether or not the fixture mutation issue is a problem.  I can see how it might be, but I don't really know the scope of cypress' imports, particularly regarding the support files.  I can certainly understand how mutating an import would be an issue within the same file, but how does it affect the imports in `support` for instance.  We should be able to prove this by setting up a simple test and support file to determine where mutation matters.

test/integration/patients/worklist/worklist.js
nmajor25: I can't prove that we're mutating elsewhere.

It seems to be only the `.routeSettings()` command in the `/patients/worklist/worklist.js` file. Not even other `routeSettings()` in other test files mutate across tests like that one. Which doesn't make much sense to me. What is different about that one?

So I'm not entirely sure how to proceed.

If we update the `.routeSettings()` command to this it fixes the mutation issue:

```js
Cypress.Commands.add('routeSettings', (mutator = _.identity) => {
  cy.fixture('test/settings').then(fixture => {
    cy.intercept('GET', '/api/settings', {
      body: mutator({
        data: getResource(fixture, 'settings'),
        included: [],
      }),
    });
  })
    .as('routeSettings');
})
```

Maybe we do that for just the `.routeSettings()` command and then keep an eye out in the future for other mutation issues.

Or maybe all of our commands should be switched to the `cy.fixture().then()` pattern. That seems to be the suggested way in the Cypress documentation.

nmajor25: For posterity, instead of doing the above we decided to add a deep clone to the top of `getResource()`:

```js
function getResource(data, type) {
  data = JSON.parse(JSON.stringify(data));

  . . .

}
```

test/helpers/json-api.js
```diff
@@ -10,8 +10,8 @@ function getResource(data, type) {
   return {
     id: data.id,
     type,
-    attributes: _.omit(data, 'id'),
-    relationships: {},
+    attributes: _.extend(_.omit(data, 'id'), attributes),
+    relationships: relationships || {},
```
nmajor25: There's a few test files that were adding `fx.included.relationship` data along with `fx.included.attributes`. So I added `relationships` as another function argument.

test/integration/patients/worklist/schedule.js
```diff
@@ -81,32 +81,8 @@ context('schedule page', function() {
...
-          {
-            id: '2',
-            type: 'patients',
-            attributes: _.extend(_.sample(fxPatients), {
-              first_name: 'LongTest',
-              last_name: 'PatientName',
-              id: '1',
-            }),
-          },
+          getResource(_.extend(_.sample(fxPatients), { id: '1' }), 'patients', { id: '1', first_name: 'Test', last_name: 'Patient' }),
```
nmajor25: `data.id: '1'` needs to be set on the fixture data before it's sent to the `getResource()`.

Which is the reason for the `_.extend(_.sample(fxPatients), { id: '1' })`

test/helpers/json-api.js
nmajor25: This causes this test to fail: https://cloud.cypress.io/projects/ep9zr6/runs/5614/test-results/1cd5e357-fc6f-4a44-96f2-c17d0ffb2345

test/integration/services/patient-quick-search.js
```diff
@@ -1,10 +1,12 @@
...
-import { getRelationship, getIncluded } from 'helpers/json-api';
+
+import { getResource, getRelationship, getIncluded } from 'helpers/json-api';
+
+import fxPatients from 'fixtures/collections/patients';
 
 context('Patient Quick Search', function() {
   beforeEach(function() {
-    const patients = _.map(_.sample(patientFixture, 10), (patient, index) => {
+    const patients = _.map(getResource(_.sample(fxPatients, 10), 'patients'), (patient, index) => {
```
paulfalgout: I don't think this works.  we just need to clone it here.. it's being jsonapi'd in the `getRelationship`, the `getInclude` and manually in the map below.

test/integration/patients/worklist/schedule.js
paulfalgout: Yeah I think that's a bad design on my part in the `getResource`.

We should revert the changes to `getResource` and just do:
```js
getResource(_.extend({}, _.sample(fxPatients), { id: '1', first_name: 'Test', last_name: 'Patient' }, 'patients');
```
then if there's any relationships it can be:
```js
const thingy = getResource(...., 'thingys');
thingy.relationships.foo = getRelationship(...);
```

paulfalgout (changes requested):

test/helpers/json-api.js
nmajor25: I fixed that failing test. But I sense this is going to cause problems.

test/integration/patients/worklist/bulk-edit.js
```diff
@@ -52,7 +43,11 @@ context('Worklist bulk editing', function() {
         fx.data[3].relationships.state = { data: { id: '22222' } };
 
 
-        fx.included.push(flowInclude);
+        fx.included.push({
+          id: '1',
+          type: 'flows',
+          attributes: _.extend({}, _.sample(fxFlows), { id: '1', name: 'Test Flow' }),
+        });
```
nmajor25, line 52: Do we need to run this through `getResource()`? Or can we just do this instead?

The goal is to avoid mutation issues. Doing just `_.extend({}, fixtureData, { changes })` should achieve that, right? Since it creates a new object and does modifications without changing the source fixture data.

I don't get the benefit of running it through `getResource()` as well...

paulfalgout: well there'd be no `id` in the `attributes` `getResource` just helps us setup the jsonapi correctly.  we don't _need_ it, but it helps us store flat fixtures.  But in simple situations, maybe it's ok..
So what's better:
```js
const includeFlow =  _.extend({}, _.sample(fxFlows), { id: '1', name: 'Test Flow' });
fx.included.push(getResource(includeFlow, 'flows');
```

Although I guess in this case it'd be:

```js
const includeFlow =  _.extend({}, _.sample(fxFlows), { id: '1', name: 'Test Flow' });
fx.included = getIncluded(fx.included, includeFlow, 'flows');
```

But that helper is even longer..  maybe `getIncluded` isn't useful..

we're probably ok here.. _but_ we should make it consistent

test/integration/services/patient-quick-search.js
```diff
@@ -1,10 +1,13 @@
...
-import patientFixture from 'fixtures/collections/patients.json';
+
 import { getRelationship, getIncluded } from 'helpers/json-api';
 
+import fxPatients from 'fixtures/collections/patients';
+
 context('Patient Quick Search', function() {
   beforeEach(function() {
-    const patients = _.map(_.sample(patientFixture, 10), (patient, index) => {
+    const clonedfxPatients = _.extend({}, _.sample(fxPatients, 10));
```
paulfalgout: this doesn't work `sample(foo, 10)` returns an array. not an object.

could be
```js
_.map(_.sample(patientFixture, 10), (patient, index) => {
  return _.extend({}, patient, {
    id: `${ index }`,
    first_name: 'Test',
....
  });
});
```

test/integration/patients/worklist/worklist.js
```diff
@@ -3831,7 +3812,11 @@ context('worklist page', function() {
         fx.data[3].relationships.owner = { data: { id: '11111', type: 'teams' } };
         fx.data[3].relationships.state = { data: { id: '33333' } };
 
-        fx.included.push(flowInclude);
+        fx.included.push({
+          id: '1',
+          type: 'flows',
+          attributes: _.extend(_.sample(fxFlows), { name: 'Test Flow', id: '1' }),
```
paulfalgout: needs `_.extend({}, _.sample..`

paulfalgout (changes requested):
need to remove `id` from the attributes

test/integration/services/patient-quick-search.js
nmajor25: Good call, that's been updated.

test/integration/patients/worklist/worklist.js
nmajor25: That's been added.

test/integration/patients/worklist/bulk-edit.js
```diff
@@ -674,7 +662,11 @@ context('Worklist bulk editing', function() {
         fx.data[3].attributes.created_at = testTsSubtract(2);
         fx.data[3].relationships.state = { data: { id: '33333' } };
 
-        fx.included.push(flowInclude);
+        fx.included.push({
+          id: '1',
+          type: 'flows',
+          attributes: _.extend({}, _.sample(fxFlows), { id: '1', name: 'Test Flow' }),
```
paulfalgout: we need to pull all of the `id`s from the `attributes`

nmajor25: Ah ya, those have been removed now.

test/integration/patients/worklist/reduced-schedule.js
```diff
@@ -49,16 +47,25 @@ context('reduced schedule page', function() {
...
-          {
-            id: '1',
-            type: 'flows',
-            attributes: _.extend({}, _.sample(fxFlows), { name: 'Complex Care Management' }),
+            attributes: {
+              first_name: 'Test',
+              last_name: 'Patient',
+            },
           },
         );
```
nmajor25, line 70: What if we just don't import the fixture data. And just manually create the patient/flow fixture instead? Then we don't have to worry about any mutating issues.

Doesn't really seem necessary to import the entire fixture json data in these instances.

paulfalgout: That's probably ok. The benefit of the fixtures data is that we can update the base fixtures and anywhere they're used in the app will also update.. so if in the future we add something to some model, and if all instances of that model across the code base need that new data not to break, then fixture data where we're modifying a couple of items will work, but the incomplete fixtures we're storing in the test will not.

test/integration/patients/worklist/worklist.js
```diff
@@ -906,12 +902,13 @@ context('worklist page', function() {
         fx.included.push({
           id: '1',
           type: 'flows',
-          attributes: _.extend(_.sample(this.fxFlows), {
+          attributes: {
             name: 'Test Flow',
-          }),
-          relationships: {
-            state: { data: { id: '55555' } },
+            details: null,
```
paulfalgout, line 907: were the details, created_at, updated_at needed in all the tests?

test/integration/patients/worklist/reduced-schedule.js
nmajor25: Hmm ya I didn't think of that...

test/integration/patients/worklist/worklist.js
nmajor25: I would guess no. I added them to match what data was being returned before.

Kind of related, but there's a few of these tests were if I remove that specific `fx.include` data, the test still passes. I hadn't looked into the details as to why that `fx.include` was originally put there.

PR #1163, opened 2023-10-05
Refactor Cypress relationships

paulfalgout:
Shortcut Story ID: [sc-41873]


test/integration/patients/patient/patient-flow.js
nmajor25, line 1: Add these to the worklist and schedule shift+click tests as well? The same debounce exists on the schedule/worklist.

nmajor25 (approved):
I'm digging it, some good changes 👍 Just one question/comment.

PR #1164, opened 2023-10-08
Revert "Add styles for consistent table borders in PDF"

paulfalgout:
- [x] Reverts RoundingWell/care-ops-frontend#843
- [x] Shortcut Story: [sc-41905]

### DIFF
--- a/src/scss/formapp/print.scss
+++ b/src/scss/formapp/print.scss
@@ -1,10 +1,5 @@
 // NOTE: This is used by the form PDF generator.
 @media print {
-  html {
-    filter: opacity(1);
-    print-color-adjust: exact;
-  }
-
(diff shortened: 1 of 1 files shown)


shadowhand (approved):

zfixler (approved):
The revert is fine, but I think we need to address the inconsistent borders in some way with this. It was originally a request from the customer that we should try to resolve.

paulfalgout:
To some degree it can be addressed, particularly for rendering at 100%.  But at some point, this is how PDFs work.  But in order to fix it we need a form where at 100% this is an issue.  I can't find one.

shadowhand:
@zfixler customer request or not, visual discrepancy is not more important than having textual data in PDFs.

PR #1165, opened 2023-10-09
Merge json api

paulfalgout:
This includes #1163 

The two additional commits introduce the `mergeJsonApi` helper and refactor the worklist tests.

The purpose of this helper is to deal with the complexities of merging a fixture that has already been made jsonapi.

We generate generic fixtures for the purpose of composing fixtures for stubs.  This reduces the overall lift in accounting for things outside of the functionality we're trying to test for, as well as reducing maintenance costs.  So in most situations it becomes one of two patterns:

```js
.routeFoo(fx => {
  fx.data = mergeApiJson(fx.data, { ...stuff to test...});
  return fx;
});
```
and
```js
const fxTestFoo = _.defaults({ ...stuff to test...}, fxFoos[0]);

const testFoo = getResource(fxTestFoo, 'foos', { relationships });

// Then maybe if testFoo is a base fixture
const otherTest = mergeJsonApi(testFoo, { id: 'test2', attributes: { ..other stuff.. }, relationships: { ...merge these too... });
```

Generally I think we may want to adopt a naming rule where the `fx` prefix indicates that the data is a flat fixture and not-jsonapi.

I'm not sure this solves our issues overall, but I am already seeing some inconsistencies and errors this is cleaning up while refactoring.  I do think it will cut down on flake.


.circleci/config.yml
```diff
@@ -83,6 +83,7 @@ workflows:
                 parallel: true
       - run-tests-in-parallel:
           name: Run Component Tests in Parallel
+          parallelism: 8
```
paulfalgout, line 86: @nmajor25 🤦 I actually did set this up so they could be set differently

nmajor25 (approved):

PR #1167, opened 2023-10-11
Fix coverage reports

paulfalgout:


package.json
```diff
@@ -10,13 +10,13 @@
   "scripts": {
     "analyze": "cross-env NODE_ENV=production webpack --profile --json > webpack.stats.json && webpack-bundle-analyzer webpack.stats.json dist/",
     "build": "cross-env NODE_ENV=production webpack",
-    "coverage": "npm run coverage:clear & npm start & cypress run --env COVERAGE=true --config video=false && npm run coverage:report",
```
paulfalgout, line 13: `video=false` is now the default

.circleci/config.yml
```diff
@@ -85,7 +85,7 @@ workflows:
           name: Run Component Tests in Parallel
           requires:
             - Install & Persist To Workspace
-          cypress-command: npx cypress run --component --env COVERAGE=true --parallel --record --group care-ops-component
+          cypress-command: npx cypress run --component --env COVERAGE=true --parallel --record --group care-ops-component --config watchForFileChanges=false
```
paulfalgout, line 88: no need to setup file watches on CI.  _should_ help some stability

test/plugins/coverage.js
```diff
@@ -1,30 +1,34 @@
...
-        coverageMap.merge(coverage);
-        return coverageMap;
-      },
-      write() {
-        fs.outputJson(coverageFile, coverageMap);
-        return coverageMap;
-      },
-    });
-  }
+  if (!config.env.COVERAGE) return;
```
paulfalgout, line 7: making this a guard made line changes for the whole file, but the major changes here are.. we're only re-reading coverage maps if the cypress UI is open.. this is because merging files happens on CI on coveralls.

test/support/coverage.js
```diff
@@ -21,16 +21,22 @@ if (Cypress.env('COVERAGE')) {
...
+
   afterEach(function() {
     each(uniq(windowCoverageObjects), coverage => {
       cy.task('coverage', coverage);
     });
 
-    cy.task('write');
+    cy.task('coverage', global.__coverage__);
+
+    cy.window().then(win => {
```
paulfalgout, line 32: there are some circumstances where adding the coverage at this moment fixes the report issues.  I suspect this is somehow related to adding the report async slightly too early., but I can't quite put my finger on the issue.

paulfalgout:
The coverage missing is actually correct.  I need to add a test for the owner relationship parsing.

nmajor25 (approved):

PR #1168, opened 2023-10-12
Support archived status on programs

nmajor25:
Shortcut Story ID: [sc-40867]

Add a `Published` and `Archived` on/off toggle for programs, program flows, and program actions. And create a simplified picklist for only updating program-flow/program-action's behavior.

Example of the program flow sidebar:

Screenshot 2023-10-12 at 5 07 38 PM

This PR also switches from the deprecated `program.published`, `program-flow.published`, and `program-action.published` boolean => `.published_at` `timestamp | null`.

Still to do:

- [x] Investigate why setting `program-action.archived_at` doesn't work.
- [x] Do additional manual testing in local dev environment.
- [x] Refactor toggle view component code.
- [x] Lots of test updates.


src/js/views/programs/sidebar/action/action-sidebar_views.js
```diff
@@ -286,20 +296,51 @@ const LayoutView = View.extend({
...
+    const isDisabled = this.action.isNew();
+
+    const toggleView = new ToggleComponent({
+      status: archived,
+      isDisabled,
+    });
+
+    this.listenTo(toggleView, 'click', () => {
+      const newArchivedAt = archived ? null : dayjs.utc().format();
+      this.action.save({ archived_at: newArchivedAt });
```
nmajor25, line 337: This currently doesn't work. `program-action.archived_at` isn't being set in the db when I test this locally. It's always `null`, regardless of whether I send `archived_at: null` or `archived_at: utc_date` in the `PATCH` api request.

It's the same method I'm using for the program and program-flow (which work), so I'm not sure what's up. Either a bug by me or something on the backend side of things.

src/js/views/programs/shared/components/toggle_component.js
```diff
@@ -0,0 +1,30 @@
+import hbs from 'handlebars-inline-precompile';
+import { View } from 'marionette';
+
+export default View.extend({
```
paulfalgout: I think it might be better to make this view something that can re-render on it's own... which means ideally the toggle is in the template and not the className  This also really isn't a component.  could make it one... where the state of the component is stored with it, but this could just be a shared view.    Although I'm not sure it even needs to be, https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/views/programs/sidebar/program/programs-sidebar_views.js#L92 is pretty much the required functionality if the view is passed the model, so some duplication might be ok
But if you go with a shared view it could be configurable via an `attr` defined like: `const FlowPublishedToggle = ToggleComponent.extend({ attr: 'published_at' });`
```js
export default View.extend({
  template: ProgramStateTemplate,
  triggers: {
    'click': 'click',
  },
  modelEvents: () {
     return {
         [`change:${ this.getOption('attr') }`]: 'render',
     };
  },
  templateContext() {
    status: this.model.get(this.getOption('attr'));
  },
});
```

if this change is made, there wouldn't be need for the parent view to rerender on the change, with exception of if it is disabled.

paulfalgout (changes requested):

src/js/views/programs/shared/components/toggle_component.js
nmajor25: Ah ya, good call. I had three instances of that `ToggleView` created that were almost identical. So I felt compelled to do something to stop the duplication.

I'll play around with the options you mentioned. And probably go with whichever one is simpler in the end. I'm guessing that'll be just the view that re-renders on it's own.

src/js/views/programs/sidebar/action/action-sidebar_views.js
nmajor25: I'm *pretty sure* I'm doing everything correct for this.

I'm sending this data with the `/api/program-actions/:id` `PATCH` request:

```js
{
    "data": {
        "id": "46f30482-62a8-11ec-8efa-b558ead68b59",
        "type": "program-actions",
        "attributes": {
            "archived_at": "2023-10-13T16:39:44Z"
        }
    }
}
```

But always get `archived_at: null` back from the backend for that action.

I think it's a backend issue.

shadowhand: @nmajor25 do you have an up to date `develop` checked out locally? The code is definitely in place:

image

nmajor25: Yup, I do. And I also did a `./bin/docker-rebuild` just to be sure.

shadowhand: @nmajor25 hmmm, okay. I know you're off this afternoon, but make a bug card for me when you get a chance, please.

nmajor25: @shadowhand  I just created that story here: [sc-42170](https://app.shortcut.com/roundingwell/story/42170/unable-to-set-archived-at-on-program-actions).

paulfalgout (commented):
Looks good.  Just need to get the coverage up it looks like

src/js/views/programs/sidebar/program/programs-sidebar_views.js
```diff
@@ -142,10 +142,12 @@ const LayoutView = View.extend({
...
-    this.listenTo(this.program, {
-      'change:published_at': this.showPublished,
-      'change:archived_at': this.showArchived,
-    });
+    if (!this.model.isNew()) {
+      this.listenTo(this.program, {
+        'change:published_at': this.showPublished,
+        'change:archived_at': this.showArchived,
+      });
+    }
```
nmajor25, line 150: This fixes a bug where a console error is thrown after successfully creating a new program. Which is caused by the listener I added in a previous commit.

What was happening:

1. Open new program sidebar. At this time, the `[data-published-region]` and `[data-archived-region]` regions are hidden because `isNew() === true`.
2. Hit `Save` button to create a new program. When the `POST` request comes back, the `change:published_at` and `change:archived_at` listeners are triggered. Which means `showPublished()` and `showArchived()` try to do a render on those `[data-published-region]` and `[data-archived-region]` regions. `isNew()` now equals `false`, but those regions don't exist yet. And this all happens before the program sidebar closes.

So to avoid this, I only add the listeners if the program model is not new.

test/fixtures/config/program-flows.js
```diff
@@ -7,8 +7,15 @@ module.exports = {
...
-      published: faker.datatype.boolean(),
-      behavior: faker.random.arrayElement(['standard', 'conditional']),
+      published_at: faker.random.arrayElement([faker.date.between(
+        dayjs().subtract(2, 'week').format(),
+        dayjs().subtract(1, 'week').format(),
+      ), null]),
+      archived_at: faker.random.arrayElement([faker.date.between(
+        dayjs().subtract(2, 'week').format(),
+        dayjs().subtract(1, 'week').format(),
+      ), null]),
```
nmajor25, line 17: When generating fixture data, this randomly sets either a timestamp or `null` for `published_at` and `archived_at`. The same pattern is used across the programs, program flows and program actions fixtures.

test/fixtures/config/program-flows.js
```diff
@@ -7,8 +7,15 @@ module.exports = {
...
+      behavior: faker.random.arrayElement(['standard', 'automated']),
```
nmajor25, line 18: My understanding is that program flows can't have the `Conditional` behavior. So I replaced that with the `Automated` behavior in the program flow fixture.

My understanding of which behaviors they can have:

1. Program flows: `Standard` or `Automated`
2. Program actions: `Standard`, `Automated`, or `Conditional` (but only if the action belongs to a flow)

test/integration/programs/sidebar/action-sidebar.js
```diff
@@ -441,61 +457,92 @@ context('program action sidebar', function() {
...
+      .contains('Off')
       .click();
 
     cy
-      .get('.picklist')
-      .contains('Conditional')
+      .wait('@routePatchAction')
+      .its('request.body')
+      .should(({ data }) => {
+        expect(data.attributes.archived_at).to.not.be.null;
```
nmajor25, line 494: Not sure how we want to test this.

This is when the `Published`/`Archived` toggle is switched to `On`. Which means we generate a timestamp (`dayjs.utc().format()`) and send that in the `PATCH` request.

I'm currently just testing if this value is `null` or not.

Do we need to test the exact timestamp? Is that possible to do without creating flake?

test/support/api/program-actions.js
```diff
@@ -34,6 +34,7 @@ Cypress.Commands.add('routeProgramActions', (mutator = _.identity, programId) =>
     action.relationships = getProgramActionRelationships({
       program,
       owner: _.random(1) ? _.sample(fxTestTeams) : null,
+      programFlow: _.sample(fxProgramFlows),
```
nmajor25, line 37: This wasn't causing failures in the tests, but it was causing this:

Screenshot 2023-10-17 at 5 01 46 PM

Where program actions have `attributes.behavior: 'conditional'`, but also `relationships._program_flow: null`. Which means the behavior is set to something that's not in the picklist.

test/integration/programs/program/program.js
```diff
@@ -125,4 +128,86 @@ context('program page', function() {
...
+      .find('[data-behavior-region]')
+      .click();
+
+    cy
+      .get('.picklist')
+      .find('.js-picklist-item')
+      .should('have.length', 2)
+      .contains('Conditional')
+      .should('not.exist');
+  });
```
nmajor25, line 212: This tests that a program action with no flow (`relationships.program_flow = null`) doesn't have the `Conditional` behavior picklist option.

Tests that in both the program workflows list and action sidebar.

src/js/entities-service/entities/program-actions.js
```diff
@@ -133,7 +133,7 @@ const Collection = BaseCollection.extend({
     const clone = this.clone();
 
     const addable = this.filter(action => {
-      return action.get('published') && action.get('behavior') !== PROGRAM_BEHAVIORS.AUTOMATED;
+      return !!action.get('published_at') && action.get('behavior') !== PROGRAM_BEHAVIORS.AUTOMATED;
```
paulfalgout: just noticing this.. actions that are archived shouldn't be addable either.

test/integration/programs/sidebar/action-sidebar.js
paulfalgout: not null is ok to me.  we could test this with `visitOnClock` by setting the now and expecting that now to be this timestamp

paulfalgout (changes requested):

src/js/entities-service/entities/program-actions.js
nmajor25: That's a good catch, I just pushed an update for that.

src/js/views/programs/sidebar/action/action-sidebar_views.js
nmajor25, line 1: This fixes a bug I created where a click outside of the button would trigger the click event. Since the click event was being placed on the wrapping empty `<div>` instead of the actual toggle `<button>` element.

paulfalgout (approved):

PR #1174, opened 2023-10-19
Improve form test flake

paulfalgout:
The previous `startOf(‘minute’)` fix was still flaky.  Really we needed store the time and test based on what we expected not relative to the test runners time which changes as the test runs.

Shortcut Story ID: [sc-###]

### DIFF
--- a/test/integration/forms/action.js
+++ b/test/integration/forms/action.js
@@ -91,8 +91,6 @@ context('Patient Action Form', function() {
   });
 
   specify('storing stored submission', function() {
-    const currentTs = dayjs().startOf('minute');
-
     cy
       .routeAction(fx => {
         fx.data.id = '1';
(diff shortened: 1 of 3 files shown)


paulfalgout:
Merging this cause it can't get worse

test/integration/forms/action.js
```diff
@@ -325,10 +321,8 @@ context('Patient Action Form', function() {
   });
 
   specify('discarding stored submission', function() {
```
nmajor25, line 323: Same pattern could be used in the `Patient Form` => `discarding stored submission` test as well? Looks like that one is still using the `currentTs` method from before.

Code link: https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/integration/forms/patient.js#L309

nmajor25: Well nvm, I guess that one is just doing `const currentTs = dayjs();`... so maybe that one won't cause issues...

paulfalgout: yeah.. should probably be updated for consistency

PR #1177, opened 2023-10-23
Always redirect and handle the logout route

paulfalgout:
This is cleaner in every sense.

Shortcut Story ID: [sc-42618]


shadowhand (commented):
👏🏼

nmajor25 (approved):

PR #1178, opened 2023-10-24
Keep `.is-selected` on the program-flow header when it's edited via the sidebar

nmajor25:
Shortcut Story ID: [sc-42651]

`.is-selected` gives the program flow header a grey background to indicate that it's currently being edited via the sidebar:

Screenshot 2023-10-24 at 4 13 45 PM

When a user makes a change via the program-flow sidebar, the `.is-selected` class was being dropped when that header view re-rendered. But as long as the sidebar is open and the user is editing the program-flow, we want the `.is-selected` class to remain.

### DIFF
--- a/src/js/views/programs/program/flow/flow_views.js
+++ b/src/js/views/programs/program/flow/flow_views.js
@@ -51,12 +51,13 @@ const ContextTrailView = View.extend({
 });
 
 const HeaderView = View.extend({
+  className: 'program-flow__header',
   modelEvents: {
     'editing': 'onEditing',
     'change': 'render',
   },
   onEditing(isEditing) {
-    this.ui.flow.toggleClass('is-selected', isEditing);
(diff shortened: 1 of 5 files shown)


src/js/views/programs/program/flow/flow_views.js
```diff
@@ -97,6 +94,17 @@ const HeaderView = View.extend({
   },
 });
 
+const AddActionView = View.extend({
+  template: hbs`
```
paulfalgout, line 99: `program-flow__actions` should just be the `className` here and the template just the button

paulfalgout (changes requested):
Pretty good refactor

src/js/views/programs/program/flow/flow_views.js
nmajor25: Good call, that's been updated 👍

paulfalgout (approved):

PR #1179, opened 2023-10-24
Add `this.startRoute(currentRoute)` to program flow app

nmajor25:
Shortcut Story ID: [sc-42671]

To fix a bug where the action sidebar wasn't opening when directly loading the `/program-flow/:id/action/:id` url.

Still need to:

- [x] Find somewhere in the tests to verify this happens correctly.


paulfalgout (approved):
Probably need a new test in program that just `visit`s to a flow action

PR #1181, opened 2023-10-27
Support `work:submit` permissions to control action form submit

nmajor25:
Shortcut Story ID: [sc-42759]

Restricts a user's ability to submit action forms based on the ownership of the form's corresponding action.

Permissions to be supported:

* `work:submit`: allow the user to submit a form on any action.
* `work:owned:submit`: allow the user to submit a form on any action they own.
* `work:team:submit`: allow the user to submit a form on any action owned by their team.

When a user doesn't have the correct permissions to submit a form, a locked message is shown and the form renders in read only mode.

Screenshot 2023-10-27 at 4 11 57 PM

Changes do not apply to patient forms. Only patient-action forms.

- [x] Incorporate `isLocked()` with the permissions added in this PR.
- [x] For `work:team:submit`, verify the submit is locked for an action owned by a non team member. Currently only testing that for the team itself.
- [x] Refactor the locked submit view scss styles that were added.

### DIFF
--- a/src/js/apps/forms/form/form_app.js
+++ b/src/js/apps/forms/form/form_app.js
@@ -19,6 +19,7 @@ import {
   FormStateActionsView,
   StatusView,
   ReadOnlyView,
+  LockedSubmitView,
   SaveView,
   UpdateView,
   HistoryView,
@@ -80,7 +81,8 @@ export default App.extend({
     this.action = action;
     this.responses = action.getFormResponses();
     this.latestResponse = latestResponse;
-    this.isReadOnly = action.isLocked() || form.isReadOnly();
+    this.isReadOnly = form.isReadOnly();
+    this.isLocked = action.isLocked() || !action.canSubmit();
     this.isSubmitHidden = form.isSubmitHidden();
 
     this.listenTo(action, 'destroy', function() {
@@ -112,7 +114,7 @@ export default App.extend({
       latestResponse: this.latestResponse,
     });
 
-    if (!this.isReadOnly) this.bindEvents(formService, this.serviceEvents);
+    if (!this.isReadOnly && !this.isLocked) this.bindEvents(formService, this.serviceEvents);
   },
   serviceEvents: {
     'submit': 'onFormServiceSubmit',
@@ -207,7 +209,7 @@ export default App.extend({
   showContent() {
     const responseId = this.getState('responseId');
 
-    if (this.isReadOnly || responseId) {
+    if (this.isReadOnly || this.isLocked || responseId) {
       this.showForm();
       return;
     }
@@ -281,6 +283,11 @@ export default App.extend({
       return;
     }
 
+    if (this.isLocked) {
+      this.showLockedSubmit();
+      return;
+    }
+
     if (this.getState('responseId')) {
       this.showFormUpdate();
       return;
@@ -290,6 +297,9 @@ export default App.extend({
   },
   showReadOnly() {
     this.showChildView('formAction', new ReadOnlyView());
+  },
+  showLockedSubmit() {
+    this.showChildView('formAction', new LockedSubmitView());
   },
   showFormStatus() {
     if (!this.responses.getFirstSubmission()) return;
--- a/src/js/entities-service/entities/actions.js
+++ b/src/js/entities-service/entities/actions.js
@@ -54,6 +54,15 @@ const _Model = BaseModel.extend({
     const owner = this.get('_owner');
     return Radio.request('entities', `${ owner.type }:model`, owner.id);
   },
+  isSameTeamAsUser() {
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    const currentUsersTeam = currentUser.getTeam();
+
+    const owner = this.getOwner();
+    const ownersTeam = owner.type === 'teams' ? owner : owner.getTeam();
+
+    return currentUsersTeam === ownersTeam;
+  },
   getAuthor() {
     return Radio.request('entities', 'clinicians:model', this.get('_author'));
   },
@@ -107,13 +116,18 @@ const _Model = BaseModel.extend({
 
     if (currentUser.can('work:owned:manage') && this.getOwner() === currentUser) return true;
 
-    if (currentUser.can('work:team:manage')) {
-      const owner = this.getOwner();
-      const currentUsersTeam = currentUser.getTeam();
-      const ownersTeam = owner.type === 'teams' ? owner : owner.getTeam();
+    if (currentUser.can('work:team:manage') && this.isSameTeamAsUser()) return true;
 
-      if (currentUsersTeam === ownersTeam) return true;
-    }
+    return false;
+  },
+  canSubmit() {
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+
+    if (currentUser.can('work:submit')) return true;
+
+    if (currentUser.can('work:owned:submit') && this.getOwner() === currentUser) return true;
+
+    if (currentUser.can('work:team:submit') && this.isSameTeamAsUser()) return true;
 
     return false;
   },
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -47,7 +47,10 @@ export default App.extend({
     if (feVersion !== versions.frontend) window.location.reload();
   },
   isReadOnly() {
-    return (this.action && this.action.isLocked()) || this.form.isReadOnly();
+    const isLocked = this.action && this.action.isLocked();
+    const isSubmitRestricted = this.action && !this.action.canSubmit();
+
+    return this.form.isReadOnly() || isLocked || isSubmitRestricted;
   },
   getStoreId() {
     const actionId = get(this.action, 'id');
--- a/src/js/views/forms/form/form.scss
+++ b/src/js/views/forms/form/form.scss
@@ -189,6 +189,30 @@
   font-size: 12px;
 }
 
+.form__submit-locked {
+  align-items: center;
+  border-left: 1px solid $black-80;
+  display: flex;
+  margin-left: 20px;
+  padding-left: 20px;
+}
+
+.form__submit-locked-icon {
+  font-size: 20px;
+  margin-right: 12px;
+
+  .icon {
+    color: $black-60;
+    font-size: 20px;
+    vertical-align: -0.125em;
+  }
+}
+
+.form__submit-locked-text {
+  font-size: 12px;
+  width: 160px;
+}
+
 .form__form-action {
   border-left: 1px solid $black-80;
   display: flex;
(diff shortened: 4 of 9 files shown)


src/js/entities-service/entities/actions.js
```diff
@@ -117,6 +117,23 @@ const _Model = BaseModel.extend({
...
+    if (currentUser.can('work:team:submit')) {
+      const owner = this.getOwner();
+      const currentUsersTeam = currentUser.getTeam();
+      const ownersTeam = owner.type === 'teams' ? owner : owner.getTeam();
+
+      if (currentUsersTeam === ownersTeam) return true;
+    }
+
+    return false;
+  },
```
nmajor25, line 133: My assumption is that we want this to be independent of the `work:manage` (`canEdit()`) permissions.

Because of that assumption, I did not include a `if (!this.canEdit()) return false;` check at the top of this function.

src/js/views/forms/form/form.scss
```diff
@@ -189,6 +189,30 @@
...
+  .icon {
+    color: $black-60;
+    font-size: 20px;
+    vertical-align: -0.125em;
+  }
+}
+
+.form__locked-text {
+  font-size: 12px;
+  width: 160px;
```
nmajor25: With `width: 160px;`:

Screenshot 2023-10-27 at 4 11 57 PM

Without:
Screenshot 2023-10-27 at 4 41 21 PM

src/js/views/forms/form/form_views.js
```diff
@@ -270,6 +270,18 @@ const ReadOnlyView = View.extend({
...
+  className: 'form__locked',
+  template: hbs`
+    <div class="form__locked-icon">
+      {{far "lock-keyhole"}}
+    </div>
+    <div class="form__locked-text">
+      {{ @intl.forms.form.formViews.restrictedSubmitView.permissionMessage }}
+    </div>
+  `,
+});
```
nmajor25, line 283: Any thoughts on the naming I'm using? `RestrictedSubmitView`, `isSubmitRestricted`... I wasn't 100% what we want to name this stuff.

I didn't want to use any `lock` wording to avoid confusion with `action.isLocked()` (`locked_at` attribute). Which is a separate thing from this permission message we're showing here.

nmajor25: If we stick with the current naming, I should change these class names to `form__restricted-submit` or something like that.

test/integration/forms/action.js
```diff
@@ -1988,4 +1988,375 @@ context('Patient Action Form', function() {
       .contains('Submit')
       .should('not.exist');
   });
+
+  specify('user has only work:owned:submit permission', function() {
```
nmajor25: Quick test overview:

1. Set the current clinician to have the `work:owned:submit` permission.
2. Load a patient dashboard with two actions. One that is owned by the current clinician and another that is not.
3. Then I click through to both of those action's forms and verify that the submit button is properly restricted. And that the form was rendered as read only.

nmajor25: May potentially name these tests something different too. Not sure yet though...

test/integration/forms/action.js
```diff
@@ -1988,4 +1988,375 @@ context('Patient Action Form', function() {
...
+      .should('not.exist');
+
+    cy
+      .iframe()
+      .find('[name="data[patient.fields.foo]"]')
+      .should('have.value', 'bar')
+      .should('be.disabled');
+  });
+
+  specify('user has only work:team:submit permission', function() {
```
nmajor25: Quick test overview:

1. Set the current clinician to have the `work:team:submit` permission.
2. Load a patient dashboard with two actions. One that is owned by the current clinician's team and another that is not.
3. Then I click through to both of those action's forms and verify that the submit button is properly restricted. And that the form was rendered as read only.

nmajor25: In this test, I need to also verify the submit button is enabled for an action owned by a fellow team member. That's the reason for the coverage drop in the `action.canSubmit()` function.

src/js/apps/forms/form/form_app.js
```diff
@@ -81,6 +82,7 @@ export default App.extend({
     this.responses = action.getFormResponses();
     this.latestResponse = latestResponse;
     this.isReadOnly = action.isLocked() || form.isReadOnly();
```
nmajor25: My assumption is that `action.isLocked()` (based on `!!action.locked_at`) is separate from this permission. In that the reason for locking an action is not related to what permissions the user has.

So if that's not the case, I'll need to rework some things.

paulfalgout: Currently the behavior of `action.isLocked()` is the same as `form.isReadOnly()` what needs to change is that the behavior fo `action.isLocked()` should be the same as `isSubmitRestricted`

src/js/entities-service/entities/actions.js
```diff
@@ -117,6 +117,23 @@ const _Model = BaseModel.extend({
...
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+
+    if (currentUser.can('work:submit')) return true;
+
+    if (currentUser.can('work:owned:submit') && this.getOwner() === currentUser) return true;
+
+    if (currentUser.can('work:team:submit')) {
+      const owner = this.getOwner();
+      const currentUsersTeam = currentUser.getTeam();
+      const ownersTeam = owner.type === 'teams' ? owner : owner.getTeam();
```
paulfalgout: I think we're already doing this logic in `action` we should probably make a `getOwnerTeam()`

src/js/services/forms.js
```diff
@@ -47,7 +47,10 @@ export default App.extend({
     if (feVersion !== versions.frontend) window.location.reload();
   },
   isReadOnly() {
-    return (this.action && this.action.isLocked()) || this.form.isReadOnly();
+    const isLocked = this.action && this.action.isLocked();
```
paulfalgout, line 50: I don't really think `isLocked` is different than `isSubmitRestricted`. If action is locked, then no permission has the permission to submit it.  Maybe in the future we add a permission for users that can submit locked actions

src/js/views/forms/form/form.scss
paulfalgout: how about `max-width`

src/js/apps/forms/form/form_app.js
paulfalgout: essentially: https://app.shortcut.com/roundingwell/story/42758/split-locked-and-read-only-form-display

src/js/views/forms/form/form.scss
```diff
@@ -189,6 +189,30 @@
   font-size: 12px;
 }
 
+.form__locked {
```
paulfalgout: I think this can be generalized with the css above.  Something like `form__status` as it seems like the design here will always be `| [icon] some text` so rather than having styles that describe what it does for business logic, it should describe what is does in the design.

src/js/views/forms/form/form.scss
paulfalgout: so the other messages just use two separate text lines.  that'd be an option as well.

paulfalgout (changes requested):

src/js/apps/forms/form/form_app.js
nmajor25: I've updated this in the last push, let me know if I missed anything 🤔

src/js/entities-service/entities/actions.js
nmajor25: Good call, created that in my last push.

src/js/entities-service/entities/actions.js
```diff
@@ -108,9 +112,24 @@ const _Model = BaseModel.extend({
     if (currentUser.can('work:owned:manage') && this.getOwner() === currentUser) return true;
 
     if (currentUser.can('work:team:manage')) {
-      const owner = this.getOwner();
+      const ownersTeam = this.getOwnerTeam();
```
paulfalgout: I think this was a good change, but seeing where the other check was used, it might make more sense for this to be
```js
if (currentUser.can(...) &&  this.isSameTeamAsUser()) return true;
```
or some name more concise?

src/js/apps/forms/form/form_app.js
```diff
@@ -112,7 +114,7 @@ export default App.extend({
       latestResponse: this.latestResponse,
     });
 
-    if (!this.isReadOnly) this.bindEvents(formService, this.serviceEvents);
+    if (!this.isReadOnly && !this.isLocked) this.bindEvents(formService, this.serviceEvents);
```
paulfalgout, line 117: I don't think the `& !this.isLocked` is necessary due to the nature of `form.isReadOnly`

src/js/apps/forms/form/form_app.js
```diff
@@ -207,7 +209,7 @@ export default App.extend({
   showContent() {
     const responseId = this.getState('responseId');
 
-    if (this.isReadOnly || responseId) {
+    if (this.isReadOnly || this.isLocked || responseId) {
```
paulfalgout, line 212: same as above

test/integration/forms/action.js
paulfalgout: I think for now we can assume the "only"

paulfalgout (changes requested):

src/js/apps/forms/form/form_app.js
nmajor25: `form.isReadOnly()` (via `/js/entities-service/entities/forms.js`) only checks for `form.options.read_only`:

```js
isReadOnly() {
  return get(this.get('options'), 'read_only');
},
```

So I think we still need these `isLocked` checks...

paulfalgout: oh yep.. burned by `forms.js` being in `/services`.  I was looking at the wrong `isReadOnly`

nmajor25: I did the same thing when first looking at this too.

src/js/entities-service/entities/actions.js
nmajor25: I like that idea, added the `isSameTeamAsUser()` function just now.

But just realized `canEdit()` and `canSubmit()` are now both doing `const currentUser = Radio.request('bootstrap', 'currentUser');` twice. Maybe that's not a big deal 🤔

test/integration/forms/action.js
nmajor25: This has been updated to not include `"only"`.

src/js/views/forms/form/form.scss
```diff
@@ -166,15 +166,15 @@
   }
 }
 
-.form__last-updated {
+.form__submit-status {
```
nmajor25, line 169: Wasn't 100% sure what to name this...

paulfalgout (approved):

PR #1182, opened 2023-11-01
Show error alert on `formResponse.saveAll()` 403 error

nmajor25:
Shortcut Story ID: [sc-42760]

### DIFF
--- a/src/js/apps/forms/form/form_app.js
+++ b/src/js/apps/forms/form/form_app.js
@@ -152,8 +152,12 @@ export default App.extend({
     this.responses.unshift(response);
     this.setState({ responseId: response.id });
   },
-  onFormServiceError() {
+  onFormServiceError(errors) {
     if (this.loadingModal) this.loadingModal.destroy();
+
+    if (errors[0].status === 403) {
+      Radio.request('alert', 'show:error', intl.forms.form.formViews.lockedSubmitView.permissionMessage);
+    }
 
     this.showFormSave();
   },
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -310,8 +310,10 @@ export default App.extend({
       }).catch(({ responseData }) => {
         /* istanbul ignore next: Don't handle non-API errors */
         if (!responseData) return;
+
+        this.trigger('error', responseData.errors);
+
         const errors = map(responseData.errors, 'detail');
-        this.trigger('error', errors);
         channel.request('send', 'form:errors', errors);
       });
   },
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -282,7 +282,13 @@ export default App.extend({
 
     this.latestResponse = formResponse;
 
-    return formResponse.saveAll();
+    return formResponse.saveAll()
+      .catch(({ responseData }) => {
+        /* istanbul ignore next: Don't handle non-API errors */
+        if (!responseData) return;
+
+        this.trigger('error', responseData.errors);
+      });
   },
   submitForm({ response }) {
     // Cancel any pending draft updates
--- a/test/integration/forms/action.js
+++ b/test/integration/forms/action.js
@@ -1872,7 +1872,7 @@ context('Patient Action Form', function() {
       .routeLatestFormResponse()
       .routeFormDefinition()
       .routeFormActionFields()
-      .visit('/patient-action/1/form/11111')
+      .visitOnClock('/patient-action/1/form/11111', { now: testTs() })
       .wait('@routeAction')
       .wait('@routeFormByAction')
       .wait('@routeFormDefinition');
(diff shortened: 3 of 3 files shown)


src/js/apps/forms/form/form_app.js
nmajor25, line 1: Current functionality with this code configuration:

https://github.com/RoundingWell/care-ops-frontend/assets/35355575/8062b36b-447c-4059-8981-f7b61c71785b

nmajor25: We'll want to not show that error in the form, correct? The `Please fix the following errors before submitting.` message.

src/js/services/forms.js
nmajor25, line 1: When this is a `403` error, this is the structure of the response data:

```js
{
    "errors": [
        {
            "id": "1",
            "status": 403,
            "title": "Forbidden",
            "detail": "Insufficient permissions"
        }
    ]
}
```

nmajor25: This `catch()` passes the errors to the form app and down to the actual iframe.

If it's a `403`, we would want to not pass the error down to the form? And only show that error message in the alert pop-up?

This error handling is also used for patient forms, so we have to make sure we don't break that since the `work:submit` permissions don't apply to it.

src/js/apps/forms/form/form_app.js
```diff
@@ -152,9 +152,21 @@ export default App.extend({
...
-  onFormServiceError() {
+  onFormServiceError(errors) {
     if (this.loadingModal) this.loadingModal.destroy();
 
+    if (errors[0].status === 403) {
+      this.isLocked = true;
+
+      this.showLockedSubmit();
+
+      this.getRegion('formUpdated').empty();
```
nmajor25: Added this to avoid:

Screenshot 2023-11-01 at 12 21 53 PM

My assumption is that we don't want to show those messages simultaneously.

nmajor25: Well... but if the user keeps working in the form after the 403 error, that last updated info will show again. So I guess we'll need a `if (this.isLocked) return` check in the `showLastUpdated()` function.

src/js/apps/forms/form/form_app.js
```diff
@@ -152,9 +152,21 @@ export default App.extend({
...
     this.setState({ responseId: response.id });
   },
-  onFormServiceError() {
+  onFormServiceError(errors) {
     if (this.loadingModal) this.loadingModal.destroy();
 
+    if (errors[0].status === 403) {
+      this.isLocked = true;
+
+      this.showLockedSubmit();
```
paulfalgout: I'm on the fence about this:

really I think the only thing we need to do here is show the error and let it go back to save.  If we lock it, the user can't do anything to resolve the issue.  if their assignment gets flipped, even by them, they could potentially flip it back and submit.

Plus the other issue is that plenty of our forms, the major ones, have submit buttons within the form themselves..

src/js/services/forms.js
```diff
@@ -310,8 +310,10 @@ export default App.extend({
       }).catch(({ responseData }) => {
         /* istanbul ignore next: Don't handle non-API errors */
         if (!responseData) return;
+
+        this.trigger('error', responseData.errors);
+
         const errors = map(responseData.errors, 'detail');
-        this.trigger('error', errors);
```
paulfalgout, line 314: we're not currently using the map anywhere I assume

src/js/services/forms.js
```diff
@@ -310,8 +310,10 @@ export default App.extend({
       }).catch(({ responseData }) => {
         /* istanbul ignore next: Don't handle non-API errors */
         if (!responseData) return;
+
+        this.trigger('error', responseData.errors);
```
paulfalgout, line 320: we'll also need to do this up on L285

src/js/apps/forms/form/form_app.js
paulfalgout: I think it's ok to show it.  the alert will go away, but not all forms handle errors the same.

paulfalgout (changes requested):

src/js/apps/forms/form/form_app.js
nmajor25: That's a good point. Didn't think of that. I agree the submit should stay, without doing the locking.

src/js/services/forms.js
nmajor25: Not that I could I find... but I'll double check that.

I think it's only used for the `channel.request('send', 'form:errors', errors);` in the `/js/formapp.js` file [here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/formapp.js#L105-L115).

paulfalgout (changes requested):
I think all that's needed now is to add the same catch for the `saveAll` for the draft

paulfalgout (changes requested):
https://coveralls.io/builds/63701257/source?filename=src%2Fjs%2Fapps%2Fforms%2Fform%2Fform_app.js#L158

Just 1 line

nmajor25:
For the coverage, I think we just need to test a non 403 error. Just pushed that update.

nmajor25:
Looks like the missing `form_app` line is now covered. I don't know what's going on with this though: https://coveralls.io/builds/63713734/source?filename=src%2Fjs%2Fviews%2Fpatients%2Fpatient%2Fflow%2Fflow_views.js#L255

PR #1184, opened 2023-11-06
Update Datadog SDKs from `v4` => `v5`

nmajor25:
Shortcut Story ID: [sc-42902]

### DIFF
--- a/src/js/datadog.js
+++ b/src/js/datadog.js
@@ -49,7 +49,6 @@ function initRum({ isForm }) {
     useCrossSiteSessionCookie: true,
     allowedTracingUrls: [window.origin],
     trackLongTasks: true,
-    trackFrustrations: true,
     trackUserInteractions: true,
     defaultPrivacyLevel: 'allow',
     enableExperimentalFeatures: ['clickmap'],
--- a/src/js/datadog.js
+++ b/src/js/datadog.js
@@ -47,7 +47,7 @@ function initRum({ isForm }) {
     version: versions.frontend,
     useSecureSessionCookie: true,
     useCrossSiteSessionCookie: true,
-    allowedTracingOrigins: [window.origin],
+    allowedTracingUrls: [window.origin],
     trackLongTasks: true,
     trackFrustrations: true,
     trackUserInteractions: true,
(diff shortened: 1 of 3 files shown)


src/js/datadog.js
nmajor25, line 1: Pertaining to this: https://docs.datadoghq.com/real_user_monitoring/guide/browser-sdk-upgrade/#automatic-recording-of-sessions-sampled-for-session-replay

Does `datadogRum.startSessionReplayRecording();` need to be called in the `setUser()` function? Or can we remove that line of code and have it automatically record at the beginning of the session?

But maybe we need to keep that there until after the `setUser()` is called? I’m not sure.

nmajor25: If we remove `startSessionReplayRecording()`, we’ll need to set `sessionReplaySampleRate: 100`. The default for that changed from `100` => `0`. ([reference](https://docs.datadoghq.com/real_user_monitoring/guide/browser-sdk-upgrade/#default-session-replay-sampling-rate))  

Otherwise, it’d be set to `0` by default and no replays would be recorded.

nmajor25: If we need to keep using `startSessionReplayRecording()`, we’ll need to set the `startSessionReplayRecordingManually: true` init parameter.

nmajor25, line 1: There were also some small changes to the `beforeSend()` method. But I don't think they effect anything for us.

paulfalgout: I think setting the manual would be ok.  There's very very little the user can do prior to bootstrap that we could glean from watching a session replay.  That way we're also only watching the session for a logged in user.

paulfalgout (changes requested):

src/js/datadog.js
nmajor25: Ya that makes sense. I'll work on getting that updated. 

And I'll go through the change log again and make sure we're all good in terms of handling any other breaking changes.

nmajor25: Okay I just added the `startSessionReplayRecordingManually: true` init option.

Since we use `startSessionReplayRecording()`, my understanding is that we don't need to set `sessionReplaySampleRate: 100`. That's only used if the sessions recordings are automatically started when the session begins (without `startSessionReplayRecording()`)

paulfalgout (approved):

PR #1186, opened 2023-11-07
Bugfix: error alert is not shown on `formResponse.saveAll()` 403 error

nmajor25:
Shortcut Story ID: [sc-42760]

We're checking for `status === 403` but in that context the status code is a string (`status === '403'`).

### DIFF
--- a/src/js/apps/forms/form/form_app.js
+++ b/src/js/apps/forms/form/form_app.js
@@ -155,7 +155,9 @@ export default App.extend({
   onFormServiceError(errors) {
     if (this.loadingModal) this.loadingModal.destroy();
 
-    if (errors[0].status === 403) {
+    const status = parseInt(errors[0].status, 10);
+
+    if (status === 403) {
       Radio.request('alert', 'show:error', intl.forms.form.formViews.lockedSubmitView.permissionMessage);
     }
 
--- a/test/integration/forms/action.js
+++ b/test/integration/forms/action.js
@@ -1809,7 +1809,7 @@ context('Patient Action Form', function() {
           errors: [
             {
               id: '1',
-              status: 403,
+              status: '403',
               title: 'Forbidden',
               detail: 'Insufficient permissions',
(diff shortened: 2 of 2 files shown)


src/js/apps/forms/form/form_app.js
```diff
@@ -155,7 +155,7 @@ export default App.extend({
   onFormServiceError(errors) {
     if (this.loadingModal) this.loadingModal.destroy();
 
-    if (errors[0].status === 403) {
+    if (errors[0].status === '403') {
```
shadowhand: Seems like it might be better to allow both, casting `status` to an integer:

```suggestion
    const status = parseInt(errors[0].status);

    if (status === 403) {
```

test/integration/forms/action.js
```diff
@@ -1885,7 +1885,7 @@ context('Patient Action Form', function() {
           errors: [
             {
               id: '1',
-              status: 403,
+              status: '403',
```
shadowhand: Maybe leave one of these as an int, just so that both situations are tested?

src/js/apps/forms/form/form_app.js
nmajor25: That's been updated to use `parseInt()`.

src/js/apps/forms/form/form_app.js
```diff
@@ -155,7 +155,9 @@ export default App.extend({
   onFormServiceError(errors) {
     if (this.loadingModal) this.loadingModal.destroy();
 
-    if (errors[0].status === 403) {
+    const status = parseInt(errors[0].status, 10);
```
nmajor25, line 158: `eslint` complains if I don't include the radix (`10` is the default, so I just went with that).

test/integration/forms/action.js
nmajor25: Ah yup good call, that's updated.

PR #1187, opened 2023-11-15
Add form submit reducers

paulfalgout:
Shortcut Story ID: [sc-43837]

This also does a bit of renaming.

What was `form:fields` is now `form:data` and `reducers` were renamed to `loaderReducers` where possible without needing to update form definitions.

We should be able to release this without affecting any current forms.

- [x] Add Tests
- [x] Add a default reducer

### DIFF
--- a/src/js/entities-service/entities/forms.js
+++ b/src/js/entities-service/entities/forms.js
@@ -32,6 +32,7 @@ const _Model = BaseModel.extend({
       loaderReducers: this.getLoaderReducers(),
       changeReducers: this.getChangeReducers(),
       beforeSubmit: this.getBeforeSubmit(),
+      submitReducers: this.getSubmitReducers(),
     };
   },
   getContextScripts() {
@@ -45,6 +46,9 @@ const _Model = BaseModel.extend({
   },
   getBeforeSubmit() {
     return get(this.get('options'), 'beforeSubmit', defaultBeforeSubmit);
+  },
+  getSubmitReducers() {
+    return get(this.get('options'), 'submitReducers', []);
   },
   getWidgets() {
     const formWidgets = get(this.get('options'), 'widgets');
--- a/src/js/entities-service/entities/forms.js
+++ b/src/js/entities-service/entities/forms.js
@@ -26,22 +26,22 @@ const _Model = BaseModel.extend({
   isSubmitHidden() {
     return get(this.get('options'), 'submit_hidden');
   },
-  getReducers() {
-    return get(this.get('options'), 'reducers', [defaultReducer]);
-  },
   getContext() {
     return {
       contextScripts: this.getContextScripts(),
-      reducers: this.getReducers(),
+      loaderReducers: this.getLoaderReducers(),
       changeReducers: this.getChangeReducers(),
       beforeSubmit: this.getBeforeSubmit(),
     };
   },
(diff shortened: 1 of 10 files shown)


paulfalgout:
last minute change I made to the reducer in the submit is causing issues, but I see it

src/js/entities-service/entities/forms.js
```diff
@@ -15,6 +15,12 @@ const defaultReducer = `
   return subm;
 `;
 
+const defaultSubmitReducer = `
+  formData.fields = formSubmission.fields || _.get(formSubmission, 'patient.fields');
+
+  return formData;
```
zfixler, line 21: Per our convo, we probably need to make sure that `formData.fields` is never submitted as `{}` here if the form does not have any fields.

src/js/formapp.js
```diff
@@ -145,7 +146,19 @@ async function renderForm({ definition, isReadOnly, storedSubmission, formData,
       return;
     }
 
-    router.request('submit:form', { response: extend({}, response, { data }) });
+    try {
+      const submitResponse = extend(getResponse(form, submitReducers, data), response, { data });
+
+      // Remove empty data to prevent { __empty__: true }
+      each(['fields', 'flow', 'action'], key => {
+        if (isEmpty(submitResponse[key])) delete submitResponse[key];
```
shadowhand, line 154: @zfixler re https://github.com/RoundingWell/care-ops-frontend/pull/1187/files#r1394594132, it looks like the empties are removed here.

src/js/formapp/utils.js
```diff
@@ -70,8 +70,16 @@ function getChangeReducers(form, changeReducers, curSubmission, prevSubmission)
   }, curSubmission);
 }
 
+function getResponse(form, submitReducers, formSubmission) {
+  return reduce(submitReducers, (memo, reducer) => {
+    const context = form.evalContext({ formSubmission, formData: memo });
+    return evaluate(reducer, context) || memo;
+  }, { fields: {}, action: {}, flow: {} });
```
shadowhand:
```suggestion
  const submission = { fields: {}, action: {}, flow: {} };

  return reduce(submitReducers, (memo, reducer) => {
    const context = form.evalContext({ formSubmission, formData: memo });
    return evaluate(reducer, context) || memo;
  }, submission);
```

src/js/entities-service/entities/forms.js
```diff
@@ -48,7 +54,7 @@ const _Model = BaseModel.extend({
     return get(this.get('options'), 'beforeSubmit', defaultBeforeSubmit);
   },
   getSubmitReducers() {
-    return get(this.get('options'), 'submitReducers', []);
+    return get(this.get('options'), 'submitReducers', [defaultSubmitReducer]);
```
zfixler: See my comment [here](https://github.com/RoundingWell/care-ops-forms/pull/1275/commits/133b240e1581dc49c84cd30cb7ead8d8329eda86#r1394773881). I'm not sure where it is best to handle this edge case, but `defaultSubmitReducer` won't run on forms that are deployed from the form builder branch, even if they do not have any `submitReducers` added.

zfixler (approved):

nmajor25 (approved):

PR #1196, opened 2023-12-11
Allow for a custom form for manual patient creation

paulfalgout:
Shortcut Story ID: [sc-43818]

```json
{
  "id": "patient_creation_form",
  "attributes": {
    "value": {
      "form_id": "11111",
      "submit_text": "Continue to Form 11111"
    }
  }
}
```

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -257,6 +257,9 @@ export default RouterApp.extend({
   onStart() {
     const currentUser = Radio.request('bootstrap', 'currentUser');
 
+    const hasManualPatientCreate = Radio.request('bootstrap', 'setting', 'manual_patient_creation');
+    this.canPatientCreate = hasManualPatientCreate && currentUser.can('patients:manage');
+
     if (!currentUser.can('clinicians:manage')) {
       adminNavMenu.remove('CliniciansApp');
     }
@@ -291,6 +294,7 @@ export default RouterApp.extend({
 
     const bottomNavView = new BottomNavView({
       model: this.getState(),
+      canPatientCreate: this.canPatientCreate,
     });
 
     this.showChildView('bottomNavContent', bottomNavView);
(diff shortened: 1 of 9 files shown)


test/integration/services/patient-quick-search.js
```diff
@@ -226,4 +227,40 @@ context('Patient Quick Search', function() {
...
+    cy
+      .get('.modal')
+      .as('searchModal')
+      .find('.patient-search__input')
+      .type('None')
+      .wait('@routeEmptyPatientSearch');
+
+    cy
+      .get('@searchModal')
+      .should('contain', 'No results match your query')
```
nmajor25, line 258: Doesn't really matter that much, but could be `No results match your query.` (period included at the end). To match the `i18n` file.

nmajor25: Doesn't really matter that much, but could be `No results match your query.` (period included at the end). To match the `i18n` file.

src/js/apps/globals/search_app.js
nmajor25, line 1: Not caused by this PR, but something I've noticed when testing.

The `No results match your query.` is often shown before the api request is finished. Would we prefer to show the `Searching...` loading text whenever there's a pending api request?

https://github.com/RoundingWell/care-ops-frontend/assets/35355575/de82a1b4-e0ae-4961-be14-a06d9f37c5c3

src/js/apps/globals/nav_app.js
```diff
@@ -398,8 +410,14 @@ export default RouterApp.extend({
         patientModal.disableSubmit();
         patient.saveAll(patientClone.attributes)
           .then(({ data }) => {
-            Radio.trigger('event-router', 'patient:dashboard', data.id);
             patientModal.destroy();
+
+            if (patientFormId) {
+              Radio.trigger('event-router', 'form:patient', data.id, patientFormId);
```
nmajor25, line 416: When we redirect to the patient form, this is what's shown for the breadcrumb link. A link to the patient dashboard. I think that makes sense...?

Screenshot 2023-12-11 at 4 09 59 PM

nmajor25 (approved):
Looks good 👍 . Everything worked well when testing it locally.

I just added a few comments/thinking out loud things.

src/js/apps/globals/nav_app.js
paulfalgout: this is the desired behavior

paulfalgout:
@nmajor25 I pulled out the fix for the no results issue.  I'm going to have to follow up with that big

nmajor25:
@paulfalgout sounds good 👍 that technically was an already existing bug/issue.

PR #1197, opened 2023-12-13
Support activity source attribute for action and flow events

nmajor25:
Shortcut Story ID: [sc-44817]


src/js/entities-service/entities/events.js
```diff
@@ -36,6 +36,9 @@ const _Model = BaseModel.extend({
     if (!this.get('_form')) return;
     return Radio.request('entities', 'forms:model', this.get('_form'));
   },
+  getSource() {
+    return this.get('source') || 'system';
```
shadowhand: Not sure if this is just defensive coding, but `source` will [always exist](https://github.com/RoundingWell/care-ops-backend/blob/b9aba8d847d8553f48f3aca113f590c99d8f5f5e/src/Event/Api/EventMapping.php#L31).

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -20,93 +20,93 @@ import 'scss/modules/sidebar.scss';
 import './action-sidebar.scss';
 
 const CreatedTemplate = hbs`
-  {{formatHTMLMessage (intlGet "patients.sidebar.action.activityViews.created") name = name team = team}}
+  {{formatHTMLMessage (intlGet (getI18nSource "created")) name = name team = team}}
```
shadowhand, line 23: This `name = name team = team` syntax doesn't bother anyone? 🙀 Makes me feel like my eyes are not working right... 😅

src/js/i18n/en-US.yml
```diff
@@ -487,32 +487,49 @@ careOptsFrontend:
...
+            dueTimeCleared: '<strong>Due Time</strong> cleared'
+            dueTimeUpdated: '<strong>Due Time</strong> changed to { time }'
+            durationCleared: '<strong>Duration</strong> cleared'
+            durationUpdated: '<strong>Duration</strong> updated to { duration }'
+            formUpdated: '<strong>Form</strong> { form } added'
+            nameUpdated: '<strong>Action name</strong> updated from { from_name } to { to_name }'
+            sharingCanceled: '<strong>Form sharing</strong> ({ team }) cancelled'
+            sharingSent: 'Form shared with <strong>{ recipient }</strong>. Waiting for response.'
+            stateUpdated: '<strong>State</strong> changed to { to_state }'
+            teamAssigned: '<strong>Owner</strong> changed to { to_team }'
```
nmajor25, line 529: For the `system` version of the action events, I removed `formRecipientResponded`, `formRemoved`, and `formResponded`. In Mark's google doc with the language updates, those are listed as not needed (`n/a`).

Is it safe to do that? The app will error if any of those three event types have `source: 'system'`.

In the local dev environment, I just noticed there are a few of these that cause app errors. But don't know if that translates to production.

src/js/entities-service/entities/events.js
nmajor25, line 1: In the `getEditor()` function of this file, do we still need the `if (!this.get('_editor')) {}` portion?

src/js/entities-service/entities/events.js
nmajor25: 👍 I just removed that.

src/js/i18n/en-US.yml
shadowhand: It is possible for `FormRemoved` to be a system event. The other two could only be created a form submission and therefore are implicitly user events.

src/js/entities-service/entities/events.js
nmajor25: Disregard. Yes, we do need this.

src/js/i18n/en-US.yml
nmajor25: Gotcha. I've added support for all three to the `i18n` file in that last commit.

paulfalgout (approved):

PR #1200, opened 2023-12-18
Refactor form tests for jsonapi

paulfalgout:
Shortcut Story ID: [sc-45142]

There's a reasonable chance something is failing somewhere, but I'm pushing for 👀 rather than sitting on it.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -99,19 +99,14 @@ In Cypress, API routes are setup in `test/support/api/`. Each added file must be
 API's are organized by model and collection requests. Multiple versions of the same route might be added to support different scenarios. For instance both a check-in and a clinician response share the same API, but return different data depending on what is requested from the `id`. Each route follows the same format so that in practice the data can be mutated for a particular test scenario.
 
 ```js
-Cypress.addParentCommand({
-  routeCheckIn(mutator = _.identity) {
-    cy.fixture('checkIn').as('fxCheckIn');
-    cy.route({
-      url: /ajax\/response\/\d+?/,
-      response() {
-        return mutator({
-          data: this.fxCheckIn.checkin
-        });
-      }
+import fxCheckIn from 'fixtures/test/check-in';
+
+Cypress.Commands.add('routeCheckIn', (mutator = _.identity) => {
+  cy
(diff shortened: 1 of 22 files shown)


test/README.md
nmajor25, line 1: Not a big deal, but we could update the `cy.route` from this example code as well: https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/README.md?plain=1#L62

test/integration/forms/action.js
```diff
@@ -4,34 +4,46 @@ import dayjs from 'dayjs';
...
 
 context('Patient Action Form', function() {
   beforeEach(function() {
     cy.routesForDefault();
   });
 
   specify('deleted action', function() {
+    const errors = getErrors({
+      status: '410',
+      title: 'Not Founddddd',
```
nmajor25, line 31: `Not Founddddd`? Was that intentional?

nmajor25 (approved):

test/integration/forms/action.js
paulfalgout: ha.. kind of.   I initially made it the same as what we were getting back, but when I would flip between 404 and 410 for testing I wasn't seeing any change to what the endpoint was returning.  So I made it obviously different, but still wasn't seeing the change..  So I wasn't sure if I was somehow hitting my docker or what was going on which led to https://github.com/RoundingWell/care-ops-frontend/pull/1200/commits/e27b65d1e18d18b6dcb51c16c4eabb710be51282  I guess I could remove that, but honestly that text doesn't matter.  If we ever need to support multiple languages that value would be returned in the language of the user so we shouldn't use it as a value to be checked against.

test/README.md
paulfalgout: Yeah might as well, though really the whole thing needs new eyes https://app.shortcut.com/roundingwell/story/30274/test-docs-are-outdated

test/integration/forms/action.js
nmajor25: Gotcha, makes sense 👍

shadowhand: This should really be `Gone` because that's the description of HTTP 410.

PR #1209, opened 2024-01-04
Refactor fetch data handling

paulfalgout:
Shortcut Story ID: [sc-44818]

There's a lot here, but this should standardize the fetch data and errors as well as fix the search abort bug we were seeing.

I'll add comments in the PR for things to look at specifically.


src/js/app.js
```diff
@@ -132,14 +133,17 @@ const Application = App.extend({
     ];
   },
 
-  onFail(options, response) {
-    addError(response);
-    this.getRegion('preloader').show(new PreloaderView({ notSetup: true }));
+  onFail(options, error = {}) {
+    const { response, responseData } = error;
+
+    addError(responseData || error);
```
paulfalgout: This `onFail` might handle an api error or a code error.  But most errors by this point will be handled generically.

src/js/apps/clinicians/clinicians-all_app.js
```diff
@@ -93,14 +93,14 @@ export default SubRouterApp.extend({
       onSubmit: () => {
         clinicianModal.disableSubmit();
         clinician.saveAll(clinicianClone.attributes)
-          .then(({ data }) => {
+          .then(() => {
```
paulfalgout, line 96: all backbone requests now return the model instead of the data, which we already have above.

src/js/base/app.js
```diff
@@ -1,6 +1,8 @@
-import { bind, isArray, noop, uniqueId, isError } from 'underscore';
+import { bind, isArray, noop, uniqueId } from 'underscore';
 import { App } from 'marionette.toolkit';
 
+import handleErrors from 'js/utils/handle-errors';
```
paulfalgout, line 4: Currently only used in one place, but I moved it to a util, because we could potentially use it elsewhere and I can unit test it there.

src/js/base/backbone-fetch.js
```diff
@@ -1,56 +1,30 @@
...
-        error.response = response;
-        error.responseData = responseData;
-        if (options.error) options.error(error);
-        else throw error;
-        return Promise.reject(error);
-      });
-    })
-    .then(responseData => {
-      fetcher.readyState = 'DONE';
+    .then(async response => {
```
paulfalgout, line 11: Logic in the backbone section of the fetch was greatly simplified.  This should look similar to the fetchBy and is generally handling whether to pass back the response or reject the response and data.

src/js/base/collection.js
```diff
@@ -1,24 +1,17 @@
...
-      fetcher = Backbone.Collection.prototype.fetch.call(this, options);
-
-      registerFetcher(baseUrl, fetcher);
-    }
+    const fetcher = Backbone.Collection.prototype.fetch.call(this, options);
 
     // Resolve with entity if successful
     return fetcher.then(response => {
-      return this;
+      if (!response || response.ok) return this;
```
paulfalgout, line 12: no response is an abort in which case it'll return the current entity

src/js/base/fetch.js
```diff
@@ -1,9 +1,79 @@
 //  Similar to https://github.com/akre54/Backbone.Fetch
-import { isObject, isArray, defaults, extend, map, flatten, reduce, first, rest } from 'underscore';
+import { isObject, isArray, defaults, extend, map, flatten, reduce, first, rest, get } from 'underscore';
 import Radio from 'backbone.radio';
 
 import { logResponse } from 'js/datadog';
 
+const fetchers = [];
+
+function registerFetcher(baseUrl, fetcher, controller) {
```
paulfalgout, line 9: The "control" file for registering fetchers was moved into the base fetcher to simplify how the abort controller is understood since it can't reliably be attached to the fetcher itself.

src/js/base/fetch.js
```diff
@@ -48,59 +118,52 @@ function getUrl(url, data) {
...
-        if (response.status >= 400) {
-          logResponse(url, options, response.clone());
-        }
+        const responseClone = response.clone();
 
         if (response.status === 401) {
-          Radio.request('auth', 'logout');
-          return Promise.reject(response);
+          const responseData = await responseClone.json();
+          if (get(responseData, ['errors', 0, 'code']) !== '5000') Radio.request('auth', 'logout');
```
paulfalgout, line 154: with the exception of 1 401 we currently can receive, we want to logout at this point.

src/js/entities-service/entities/patient-search-results.js
```diff
@@ -47,7 +47,9 @@ const Collection = BaseCollection.extend({
     const filter = { search };
 
     delete this._hasIdentifiers;
-    this.fetcher = this.fetch({ data: { filter } });
+    this.controller.abort();
+    this.controller = new AbortController();
+    this.fetcher = this.fetch({ data: { filter }, signal: this.controller.signal });
```
paulfalgout, line 52: we need a custom abort controller for this request as we're aborting under certain circumstances that are different than the default aborts.

src/js/services/bootstrap.js
```diff
@@ -43,6 +43,8 @@ export default App.extend({
   _getWorkspace(slug) {
     const workspaces = this.currentUser.getWorkspaces();
 
+    if (!workspaces.length) throw 'No workspaces found';
```
paulfalgout, line 46: Previously this function would return `undefined` if there were no workspaces and we'd show the hold up message when `workspace.id` failed... this is a more explict error that we handle explicitly now.

nmajor25 (approved):

PR #1210, opened 2024-01-05
Refactor clinicians tests for jsonapi

nmajor25:


test/integration/clinicians/clinician-sidebar.js
```diff
@@ -449,9 +459,9 @@ context('clinician sidebar', function() {
   specify('view clinician', function() {
```
nmajor25: Don't think this test does anything unique? It just checks the clinician name and disabled email in the clinician sidebar.

The same things are done in the `edit clinician` test. Unless I'm missing something...

test/support/api/workspaces.js
```diff
@@ -23,6 +23,12 @@ export function getWorkspaces(data, { depth = 0 } = {}) {
   return mergeJsonApi(resource, data, { VALID: { relationships: _.keys(defaultRelationships) } });
 }
 
+const workspaces = getResource(fxTestWorkspaces, TYPE);
+
+// Exporting only workspaces needed for testing variance
+export const workspaceOne = _.find(workspaces, { id: '11111' });
+export const workspaceTwo = _.find(workspaces, { id: '22222' });
```
nmajor25, line 30: This is for the `find in list` test above, where we need each clinician needs to only be part of one workspace.

Attempting to follow the same pattern in roles and teams support files.

test/support/api/workspaces.js
```diff
@@ -23,6 +23,12 @@ export function getWorkspaces(data, { depth = 0 } = {}) {
   return mergeJsonApi(resource, data, { VALID: { relationships: _.keys(defaultRelationships) } });
 }
 
+const workspaces = getResource(fxTestWorkspaces, TYPE);
```
nmajor25, line 26: I tried to do `const workspaces = getWorkspaces()` here but the `clinicians: getRelationship(getClinicians({}, { depth })),` was throwing an error in the `getWorkspaces()` function.

Not really sure why.

test/integration/clinicians/clinician-sidebar.js
paulfalgout: yeah if we can remove this without coverage drop, we should.  edit was probably added later, duplicating this when it should have just been modified

test/integration/clinicians/clinician-sidebar.js
```diff
@@ -1,10 +1,12 @@
...
 import stateColors from 'helpers/state-colors';
-import fxWorkspaces from 'fixtures/test/workspaces.json';
+
+import { teamCoordinator, teamNurse } from 'support/api/teams';
+import { roleAdmin, roleEmployee, roleManager } from 'support/api/roles';
+import { getWorkspaces } from 'support/api/workspaces';
+import { getClinician, getCurrentClinician } from 'support/api/clinicians';
 
 const testClinician = {
   id: '1',
```
paulfalgout, line 12: I think rather than storing a pojo here we could get `getClinician` at this point?  Then we don't need to set the `id` and we can test with `testClinician.id` in the expects

test/integration/clinicians/clinicians-all.js
```diff
@@ -1,24 +1,38 @@
...
-        fx.data[0].attributes.last_active_at = testTs();
-        fx.data[0].relationships.role.data.id = '33333';
-        fx.data[0].relationships.team.data.id = '11111';
-
-        fx.data[1].attributes.name = 'Baron Baronson';
-        fx.data[1].attributes.enabled = true;
-        fx.data[1].attributes.last_active_at = null;
+        fx.data = [
+          getClinician({
+            id: '1',
```
paulfalgout, line 16: I don't think these ids are needed

test/integration/clinicians/clinicians-all.js
```diff
@@ -186,18 +200,30 @@ context('clinicians list', function() {
...
+            },
+          }),
+          getClinician({
+            id: '2',
+            attributes: {
+              name: 'Baron Baronson',
+              enabled: true,
+            },
+            relationships: {
+              workspaces: [getRelationship([workspaceTwo])],
```
paulfalgout: there's an extra set of arrays here..

test/support/api/workspaces.js
paulfalgout: This is a dependency issue..  I don't know how to fix it because we `import` all of the files in one file..  I think we'd have to always export functions..

paulfalgout (changes requested):

test/integration/clinicians/clinicians-all.js
nmajor25: Oops I was playing around with something previously and forgot to remove those. Should be removed now.

test/integration/clinicians/clinicians-all.js
nmajor25: If I don't set the `id` here, an error is thrown:

Screenshot 2024-01-09 at 10 37 10 AM

test/integration/clinicians/clinician-sidebar.js
nmajor25: I just updated this to be `const testClinician = getClinician({})` and then used that throughout the tests.

test/support/api/workspaces.js
nmajor25: So leave this how I have it for now?

test/integration/clinicians/clinician-sidebar.js
nmajor25: This was removed and test coverage remained at 100% 👍

test/integration/clinicians/clinicians-all.js
paulfalgout: oh ha.  well that's why I put that error in there because this one is unique

paulfalgout (approved):

PR #1211, opened 2024-01-09
Prevent older fetchers from triggering search

paulfalgout:
Shortcut Story ID: [sc-46054]

Fairly simple check to make sure the latest fetch is the only one that resolves.

Not certain how to test this successfully if we're not getting coverage.

### DIFF
--- a/test/integration/services/patient-quick-search.js
+++ b/test/integration/services/patient-quick-search.js
@@ -39,13 +39,16 @@ context('Patient Quick Search', function() {
         url: 'api/patients?filter*',
       }, req => {
         if (req.url.includes('None')) {
-          req.reply({ data: [] });
+          req.reply({ body: { data: [] } });
           req.alias = 'routeEmptyPatientSearch';
           return;
         }
         req.reply({
-          data,
-          included: [...getResource(patients, 'patients')],
+          body: {
+            data,
+            included: [...getResource(patients, 'patients')],
+          },
+          delay: 300,
         });
         req.alias = 'routePatientSearch';
       });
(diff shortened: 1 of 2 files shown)


nmajor25 (approved):

nmajor25:
Just noticed the coverage miss for the `if (this.fetcher !== fetcher) return;` line.

paulfalgout:
I figured out a coverage test

test/integration/services/patient-quick-search.js
nmajor25, line 1: This is purely for coverage, right? Like if we remove the `entities-service/entities/patient-search-results.js` file changes, this test will still pass.

nmajor25: Well I guess the `if (this.fetcher !== fetcher) return;` line still isn't covered 🤔

paulfalgout: yep it does.. although.. I didn't look at the `req.reply` API, but it's odd to be that the data wasn't nested in `body`....

paulfalgout: it's covered isn't it?

paulfalgout: hmm lemme run it again
it works locally.

nmajor25: Ya it's probably just something weird with the coveralls report...

paulfalgout: ok.. I think I got it working on CI by slowing down the test a hair.. ironically that flow.owner flake on coverage happened.. so running it.. _again_

PR #1212, opened 2024-01-10
Make sure fetch response is always cloned

paulfalgout:
Shortcut Story ID: [sc-46105]

### DIFF
--- a/src/js/base/fetch.js
+++ b/src/js/base/fetch.js
@@ -139,6 +139,14 @@ export function handleError(error) {
   if (error.name !== 'AbortError') throw error;
 }
 
+async function shouldLogout(response) {
+  if (response.status !== 401) return false;
+
+  const responseData = await response.clone().json();
+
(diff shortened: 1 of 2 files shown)


src/js/base/fetch.js
```diff
@@ -139,6 +139,14 @@
   if (error.name !== 'AbortError') throw error;
 }
 
+async function shouldLogout(response) {
+  if (!response.status === 401) return false;
```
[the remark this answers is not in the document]
paulfalgout: 🤦

src/js/datadog.js
nmajor25, line 1: Code looks good 👍 

So the datadog error was caused by `response.json()` being called twice on the same response object (when a `401` occurred)?

To fix that we just need to make sure `.json()` is called on a new `response.clone()` each time?

Or was the datadog error caused by something else?

paulfalgout: right `.json()` was called twice without a `.clone`

nmajor25 (approved):

PR #1221, opened 2024-01-22
Cache bust index.html

paulfalgout:
This is the only FE file that requires cache-busting after a deploy.  We could remove that need with this change.

It only saves < 100k to cache this file.

Shortcut Story ID: [sc-###]

https://medium.com/swlh/cache-busting-for-an-angular-app-deployed-with-aws-s3-and-cloudfront-86cd52359578


shadowhand (changes requested):
This is no longer required, because:

https://github.com/RoundingWell/care-ops-backend/pull/7191/files#diff-8cf16232f3b7094ea431234555d38ad71e6e62764c3e29a5841e25172aea5049R873-R874

PR #1222, opened 2024-01-23
Edit/View patient modal bugs

nmajor25:
Shortcut Story ID: [sc-46539]

We only want the `patient_creation_form.value.submit_text` org setting value to be the submit button text when creating a new patient.

We're currently showing that submit button text for both the edit and view patient modals. For example:

Clipboard 2024-18-01 at 9 16 02 AM (1)

Instead the edit patient modal should have `Save` for it's submit button. And the view patient modal should have `Done`. 

For reference, the `patient_creation_form` org setting was initially implemented here: https://github.com/RoundingWell/care-ops-frontend/pull/1196.

### DIFF
--- a/src/js/views/globals/patient-modal/patient-modal_views.js
+++ b/src/js/views/globals/patient-modal/patient-modal_views.js
@@ -234,12 +234,6 @@ const PatientModal = View.extend({
 mixinState(PatientModal);
 
 function getPatientModal(opts) {
-  const { submit_text: submitText } = Radio.request('bootstrap', 'setting', 'patient_creation_form') || {};
-
-  if (submitText) {
-    extend(opts, { submitText });
-  }
-
   const patient = opts.patient;
   const bodyView = new PatientModal({
     model: patient,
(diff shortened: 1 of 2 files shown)


src/js/views/globals/patient-modal/patient-modal_views.js
```diff
@@ -249,6 +243,11 @@ function getPatientModal(opts) {
 
   if (canEdit) {
     const type = patient.isNew() ? 'add' : 'edit';
+    const { submit_text: submitText } = Radio.request('bootstrap', 'setting', 'patient_creation_form') || {};
+
+    if (submitText && patient.isNew()) {
+      extend(opts, { submitText });
+    }
```
nmajor25, line 250: My understanding is that these bugs are just with the submit button text. 

On QA currently, the submit buttons behave correctly in the edit and view patient modals. They're just showing the `patient_creation_form` setting value when they shouldn't be. Everything else in those different versions of the patient modal looks correct to me.

nmajor25:
No idea why the `program workflows page => add action` test would be failing. But looking into that...

paulfalgout (approved):
I can't reproduce what I think I was seeing.. could just have been this button that was throwing me off.. but this fixes something that needs to be fixed either way.

paulfalgout:
The coveralls issue is just missing some component test coverage.. it's an incorrect report

PR #1223, opened 2024-01-23
Remove patient workspace modifications

nmajor25:
Shortcut Story ID: [sc-46927]

This PR makes it so users cannot add or remove the workspaces a patient belong to in the UI. So when a patient is created, the FE won't send any workspace data to the backend. And the patient will be added to the user's workspaces automatically.

Changes to the different versions of the patient modal:

* `Add New Patient`: entire workspaces section will not be shown. And the workspaces data will not be sent with the api request. ([screenshot](https://github.com/RoundingWell/care-ops-frontend/assets/35355575/ddca1b03-62aa-4726-9a9c-d1278d877b4b))
* `Edit Patient Details`: workspaces section will be shown as before, but the droplist component is disabled and only displays a list of that patient's workspaces. So workspaces will be in a read only state. ([screenshot](https://github.com/RoundingWell/care-ops-frontend/assets/35355575/fc173968-ff50-46b8-978b-495290f71496))
* `View Patient Details`: remains the same as before where the workspaces are in a read only state. The droplist component is disabled and only displays a list of that patient's workspaces. ([screenshot](https://github.com/RoundingWell/care-ops-frontend/assets/35355575/defb5564-974b-43a5-b89b-34a8125aeb96))

To do:

- [x] Remove ability to select workspaces in add patient modal.
- [x] Make workspace list view only in edit patient modal.
- [x] Confirm the view patient model remains unchanged.
- [x] Figure out why `schedule => display schedule` test is failing. Potentially addressed in a separate PR. I don't think it's related to this PR. Fixed in a separate PR: https://github.com/RoundingWell/care-ops-frontend/pull/1224.

### DIFF
--- a/src/js/entities-service/entities/patients.js
+++ b/src/js/entities-service/entities/patients.js
@@ -12,12 +12,11 @@ const _Model = BaseModel.extend({
   type: TYPE,
   urlRoot: '/api/patients',
   /* eslint-disable complexity */
-  validate({ first_name, last_name, birth_date, sex, _workspaces }) {
+  validate({ first_name, last_name, birth_date, sex }) {
     const errors = {};
 
     if (!first_name || !last_name) errors.name = 'required';
     if (!sex) errors.sex = 'required';
-    if (!_workspaces || !_workspaces.length) errors.workspaces = 'required';
 
     if (!birth_date) errors.birth_date = 'required';
     else if (dayjs(birth_date).isAfter()) errors.birth_date = 'invalidDate';
@@ -34,28 +33,14 @@ const _Model = BaseModel.extend({
     const fields = this.getFields();
     return fields.find({ name });
   },
-  addWorkspace(workspace) {
-    const workspaces = this.getWorkspaces();
-    workspaces.add(workspace);
-    this.set('_workspaces', workspaces.map(model => model.pick('id')));
-  },
-  removeWorkspace(workspace) {
-    const workspaces = this.getWorkspaces();
-    workspaces.remove(workspace);
-    this.set('_workspaces', workspaces.map(model => model.pick('id')));
-  },
   saveAll(attrs) {
     attrs = extend({}, this.attributes, attrs);
-
-    const relationships = {
-      'workspaces': this.toRelation(attrs._workspaces, 'workspaces'),
-    };
 
     const opts = { wait: true };
 
     if (this.isNew()) opts.type = 'PUT';
 
-    return this.save(attrs, { relationships }, opts);
+    return this.save(attrs, {}, opts);
   },
   canEdit() {
     return this.isNew() || this.get('source') === 'manual';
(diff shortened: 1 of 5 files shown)


src/js/views/globals/patient-modal/patient-modal.hbs
```diff
@@ -2,4 +2,3 @@
 <div class="u-margin--b-8 flex"><div class="modal__form-label">{{ @intl.globals.patientModal.patientModalTemplate.lastName }}</div><div class="flex-grow" data-last-name-region></div></div>
 <div class="u-margin--b-8 flex"><div class="modal__form-label">{{ @intl.globals.patientModal.patientModalTemplate.dateOfBirth }}</div><div data-dob-region></div></div>
 <div class="u-margin--b-8 flex"><div class="modal__form-label">{{ @intl.globals.patientModal.patientModalTemplate.sex }}</div><div data-sex-region></div></div>
-<div class="u-margin--b-8 flex"><div class="modal__form-label">{{ @intl.globals.patientModal.patientModalTemplate.workspaces }}</div><div class="modal__form-component" data-workspaces-region></div></div>
```
nmajor25, line 5: Realized just now that we'll need this region for the edit patient and view patient modals. And just have it not show if it's the `Add Patient` version of the modal.

So, need to redo some work on this.

paulfalgout: probably could be `{{#unless isNew}}`.. if we need an empty region for add for the view not to fail could be a blank div in an `{{else}}`

paulfalgout (changes requested):

nmajor25:
Something weird is happening with the `schedule => display schedule` test. It's failing for me off & on. When a patient sidebar is opened in that test, this error is shown:

Screenshot 2024-01-24 at 10 50 28 AM

I'm investigating it still. Think it's something to do with the status widget? Not sure yet. Can't figure out how or if the changes in this PR would cause this.

nmajor25:
I've confirmed this is only an error in this test. It all works as expected in my local dev environment.

For some reason in the `schedule => display schedule` test, the status widget loads a workspace patient with no attribute data. So that status widget's handlebars template fails when it tries to display an `undefined` value from `patient.model.getStatus()`.

It's failing consistently locally for me. But it seems to flaky on CircleCI.

I think it's unrelated to this PR.

src/js/views/globals/patient-modal/patient-modal.hbs
nmajor25: I went with just an `{{@unless isNew}}` in the template here. And then did a `if (!this.model.isNew()) this.showWorkspacesComponent();` when showing the region.

I don't feel strongly about either way. But it seemed cleaner than doing an empty div 🤷. I'd be fine with either.

src/js/entities-service/entities/patients.js
```diff
@@ -12,12 +12,11 @@ const _Model = BaseModel.extend({
   type: TYPE,
   urlRoot: '/api/patients',
   /* eslint-disable complexity */
-  validate({ first_name, last_name, birth_date, sex, _workspaces }) {
```
nmajor25, line 15: My understanding is that this is only used when determining enabled/disabled state of the `Save` button when adding a new patient. I couldn't find any other place where this is used.

So I felt safe removing this.

src/js/views/globals/patient-modal/patient-modal_views.js
```diff
@@ -165,7 +165,7 @@ const PatientModal = View.extend({
     this.showLastNameView();
     this.showSexDroplist();
     this.showBirthDatePicker();
-    this.showWorkspacesComponent();
+    if (!this.model.isNew()) this.showWorkspacesComponent();
```
paulfalgout: I think this'd be better inside the function as a guard

paulfalgout (changes requested):
My guess is the schedule needs a `routeWorkspacePatient` for the test that opens a sidebar.  Perhaps the test doesn't have time to fail in some cases.. however because the workspace-patient is a uuid v5 the workspace and patient id ha to align with the workspace-patient id.

src/js/views/globals/patient-modal/patient-modal_views.js
nmajor25: I agree, good call. I just updated it to that.

paulfalgout (approved):

PR #1224, opened 2024-01-25
Fix schedule sidebar test

paulfalgout:
Test didn’t explicitly set `routePatient` so it was attempting to load a completely different status in the sidebar on render.

### DIFF
--- a/test/integration/patients/worklist/schedule.js
+++ b/test/integration/patients/worklist/schedule.js
@@ -1,5 +1,6 @@
 import _ from 'underscore';
 import dayjs from 'dayjs';
+import { v5 as uuid } from 'uuid';
 
 import formatDate from 'helpers/format-date';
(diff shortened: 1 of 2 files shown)


test/integration/patients/worklist/schedule.js
```diff
@@ -139,6 +141,18 @@ context('schedule page', function() {
...
         return fx;
       })
+      .routePatient(fx => {
+        fx.data = testPatient1;
+
+        return fx;
+      })
+      .routeWorkspacePatient(fx => {
+        fx.data = getWorkspacePatient({
+          id: uuid(testPatient1.id, workspaceOne.id),
```
nmajor25, line 151: Ah okay that's how you get those to match 👍

nmajor25 (approved):

PR #1225, opened 2024-01-25
Expose form response data to loader reducers

paulfalgout:
Shortcut Story ID: [sc-47117]

Previous `response` data that isn't the form.io response is now exposed to the script context available to loader reducers as `responseData` which should equate to the `formData` of the prefilled form submission with the `state` and `metadata` that is attached from form.io

### DIFF
--- a/src/js/entities-service/entities/form-responses.js
+++ b/src/js/entities-service/entities/form-responses.js
@@ -1,4 +1,4 @@
-import { get } from 'underscore';
+import { get, omit } from 'underscore';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
@@ -41,6 +41,9 @@ const _Model = BaseModel.extend({
   },
   getResponse() {
     return get(this.get('response'), 'data', {});
+  },
+  getFormData() {
+    return omit(this.get('response'), 'data');
   },
   parseRelationship: _parseRelationship,
 });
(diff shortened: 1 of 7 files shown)


paulfalgout:
Fixing the coverage issue here: https://github.com/RoundingWell/care-ops-frontend/pull/1224/commits/52354c9b70df237018efaa8432a3ca1434f3b58a

Although it looks like some of the component testing is missing.. but maybe related to the auth0.js file?

src/js/entities-service/entities/form-responses.js
```diff
@@ -42,6 +42,9 @@ const _Model = BaseModel.extend({
   getResponse() {
     return get(this.get('response'), 'data', {});
   },
+  getFormData() {
```
zfixler, line 45: Should this be `getResponseData`? Although that's confusing with `getResponse`...

nmajor25 (approved):

zfixler (approved):

src/js/entities-service/entities/form-responses.js
paulfalgout: it's all a bit confusing in terms, but the object that get's preloaded is `formData` so this is kind of the response's `formData`. but `response.data` is also a thing.. so.. it's not entirely obvious without renaming a bunch of stuff.

zfixler: Yeah, that's true. There isn't a more obvious name, so it works.

PR #1226, opened 2024-01-26
Switch from created_at to updated_at for form-responses

paulfalgout:
Shortcut Story ID: [sc-47063]

Previously we always created new responses.  When we switched to updating responses we needed to sort by and display the `updated_ts` date


nmajor25 (commented):
Update this to `updated_at` as well in the tests?

https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/support/api/form-responses.js#L23

paulfalgout:
Well `updated_at` should be there with `create_at` in the stubs.  Added

nmajor25 (approved):

PR #1227, opened 2024-01-26
Add patient sidebar menu options for patient status

nmajor25:
Shortcut Story ID: [sc-46925]

Adds two new options to the patient sidebar menu to allow a user to update a patient's status:

1. Toggle between `active` & `inactive` status.
2. Set a patient as `archived`.

This PR also implements the `patients:manage` permission. These menu options won't be shown to users that don't have that permission.

Still yet to do:

- [x] Refactoring
- [x] Add confirmation modal before setting archived status
- [x] Add/update tests
- [x] Update language in menu & archive confirmation modal

### DIFF
--- a/src/js/apps/patients/patient/patient_app.js
+++ b/src/js/apps/patients/patient/patient_app.js
@@ -42,9 +42,11 @@ export default SubRouterApp.extend({
     return Radio.request('entities', 'fetch:patients:model', patientId);
   },
 
-  onFail() {
-    Radio.trigger('event-router', 'notFound');
-    this.stop();
+  onFail(options, { response }) {
+    if (response.status === 410) {
+      Radio.trigger('event-router', 'notFound');
+      this.stop();
+    }
   },
 
   onStart({ currentRoute }, patient) {
--- a/src/js/apps/patients/patient/patient_app.js
+++ b/src/js/apps/patients/patient/patient_app.js
@@ -42,6 +42,11 @@ export default SubRouterApp.extend({
     return Radio.request('entities', 'fetch:patients:model', patientId);
   },
 
+  onFail() {
+    Radio.trigger('event-router', 'notFound');
+    this.stop();
+  },
+
   onStart({ currentRoute }, patient) {
     this.patient = patient;
 
--- a/src/js/apps/patients/patient/sidebar/sidebar_app.js
+++ b/src/js/apps/patients/patient/sidebar/sidebar_app.js
@@ -30,9 +30,6 @@ export default App.extend({
     const currentUser = Radio.request('bootstrap', 'currentUser');
     this.canManagePatients = currentUser.can('patients:manage');
 
-    this.showSidebar();
-  },
-  showSidebar() {
     const widgets = Radio.request('bootstrap', 'sidebarWidgets');
 
     this.showView(new SidebarView({
@@ -46,10 +43,8 @@ export default App.extend({
   },
   toggleActiveStatus() {
     this.patient.toggleActiveStatus();
-    this.showSidebar();
   },
   archivePatient() {
     this.patient.setArchivedStatus();
-    this.showSidebar();
   },
 });
--- a/src/js/apps/patients/patient/sidebar/sidebar_app.js
+++ b/src/js/apps/patients/patient/sidebar/sidebar_app.js
@@ -27,15 +27,12 @@ export default App.extend({
   },
   onStart({ patient }) {
     this.patient = patient;
-    const currentUser = Radio.request('bootstrap', 'currentUser');
-    this.canManagePatients = currentUser.can('patients:manage');
 
     const widgets = Radio.request('bootstrap', 'sidebarWidgets');
 
     this.showView(new SidebarView({
       model: this.patient,
       collection: widgets,
-      canManagePatients: this.canManagePatients,
     }));
   },
   showPatientModal() {
--- a/src/js/apps/patients/patient/sidebar/sidebar_app.js
+++ b/src/js/apps/patients/patient/sidebar/sidebar_app.js
@@ -9,6 +9,8 @@ export default App.extend({
   viewEvents: {
     'click:patientEdit': 'showPatientModal',
     'click:patientView': 'showPatientModal',
+    'click:activeStatus': 'toggleActiveStatus',
+    'click:archivedStatus': 'archivePatient',
   },
   onBeforeStart({ patient }) {
     this.showView(new SidebarView({ model: patient }));
@@ -25,15 +27,29 @@ export default App.extend({
   },
   onStart({ patient }) {
     this.patient = patient;
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    this.canManagePatients = currentUser.can('patients:manage');
 
+    this.showSidebar();
+  },
+  showSidebar() {
     const widgets = Radio.request('bootstrap', 'sidebarWidgets');
 
     this.showView(new SidebarView({
       model: this.patient,
       collection: widgets,
+      canManagePatients: this.canManagePatients,
     }));
   },
   showPatientModal() {
     Radio.request('nav', 'patient', this.patient);
+  },
+  toggleActiveStatus() {
+    this.patient.toggleActiveStatus();
+    this.showSidebar();
+  },
+  archivePatient() {
+    this.patient.setArchivedStatus();
+    this.showSidebar();
   },
 });
--- a/src/js/entities-service/entities/patients.js
+++ b/src/js/entities-service/entities/patients.js
@@ -52,6 +52,18 @@ const _Model = BaseModel.extend({
     const workspacePatient = Radio.request('entities', 'get:workspacePatients:model', this.id);
     return workspacePatient.get('status');
   },
+  toggleActiveStatus() {
+    const workspacePatient = Radio.request('entities', 'get:workspacePatients:model', this.id);
+    const currentStatus = workspacePatient.get('status');
+    const newStatus = currentStatus !== 'active' ? 'active' : 'inactive';
+
+    workspacePatient.setNewStatus(newStatus, this);
+  },
+  setArchivedStatus() {
+    const workspacePatient = Radio.request('entities', 'get:workspacePatients:model', this.id);
+
+    workspacePatient.setNewStatus('archived', this);
+  },
 });
 
 const Model = Store(_Model, TYPE);
--- a/src/js/entities-service/entities/patients.js
+++ b/src/js/entities-service/entities/patients.js
@@ -6,6 +6,8 @@ import dayjs from 'dayjs';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
 
+import { PATIENT_STATUS } from 'js/static';
+
 const TYPE = 'patients';
 
 const _Model = BaseModel.extend({
@@ -55,14 +57,14 @@ const _Model = BaseModel.extend({
   toggleActiveStatus() {
     const workspacePatient = Radio.request('entities', 'get:workspacePatients:model', this.id);
     const currentStatus = workspacePatient.get('status');
-    const newStatus = currentStatus !== 'active' ? 'active' : 'inactive';
+    const newStatus = currentStatus !== PATIENT_STATUS.ACTIVE ? PATIENT_STATUS.ACTIVE : PATIENT_STATUS.INACTIVE;
 
     workspacePatient.saveAll({ status: newStatus });
   },
   setArchivedStatus() {
     const workspacePatient = Radio.request('entities', 'get:workspacePatients:model', this.id);
 
-    workspacePatient.saveAll({ status: 'archived' });
+    workspacePatient.saveAll({ status: PATIENT_STATUS.ARCHIVED });
   },
 });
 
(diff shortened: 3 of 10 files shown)


src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -25,15 +28,29 @@ export default App.extend({
   },
   onStart({ patient }) {
     this.patient = patient;
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    this.canManagePatients = currentUser.can('patients:manage');
```
nmajor25: Based on what's in the backend, `patients:manage` is needed in order to update the patient's status. And the api request fails if the user doesn't have it.

There is also a `patients:owned:manage` permission in the backend. But it doesn't look like it's used for anything.

src/js/entities-service/entities/workspace-patients.js
```diff
@@ -6,6 +7,19 @@ const TYPE = 'workspace-patients';
...
+    const attrs = { status: newStatus };
+    const opts = { type: 'PUT' };
+
+    const relationships = {
+      'workspace': this.toRelation(currentWorkspace, 'workspaces'),
+      'patient': this.toRelation(patient, 'patients'),
+    };
+
+    this.save(attrs, { relationships }, opts);
+  },
```
nmajor25, line 19: A lot of this probably needs to be refactored. This is in the "get it to work" state.

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -61,21 +61,35 @@ const SidebarView = View.extend({
...
+      menuOptions.push({
+        text: patientStatus !== 'active' ? i18n.menuOptions.activate : i18n.menuOptions.inactivate,
+        onSelect: bind(this.triggerMethod, this, 'click:activeStatus'),
+      });
+
+      if (patientStatus !== 'archived') {
+        menuOptions.push({
+          text: i18n.menuOptions.archive,
+          onSelect: bind(this.triggerMethod, this, 'click:archivedStatus'),
+        });
```
nmajor25, line 89: Archiving a patient seems like a big deal. And something that can't be easily reversed in most cases.

Should we add a `Are you sure?` confirmation modal? Could also help users avoid setting `status = archived` on accident.

src/js/apps/patients/patient/sidebar/sidebar_app.js
nmajor25: When a patient is archived, that patient should be removed from patient search results, correct? (in that specific workspace)

In my local sandbox, the patients are still being included. Not sure if that translates to production, but wanted to make a note of that.

src/js/apps/patients/patient/sidebar/sidebar_app.js
```diff
@@ -25,15 +28,29 @@ export default App.extend({
...
       collection: widgets,
+      canManagePatients: this.canManagePatients,
     }));
   },
   showPatientModal() {
     Radio.request('nav', 'patient', this.patient);
   },
+  toggleActiveStatus() {
+    this.patient.toggleActiveStatus();
+    this.showSidebar();
```
nmajor25: When the status of the patient is changed, the text option in the menu and the status widget will need to re-render to show the updated status value.

Not sure how to re-render those individually... so I'm just re-rendering the entire sidebar to get those values to update. Seems okay to me, but open to a better way to do it.

src/js/apps/patients/patient/patient_app.js
```diff
@@ -42,6 +42,11 @@ export default SubRouterApp.extend({
     return Radio.request('entities', 'fetch:patients:model', patientId);
   },
 
+  onFail() {
+    Radio.trigger('event-router', 'notFound');
+    this.stop();
+  },
+
```
nmajor25, line 51: After a patient is archived, retrieving that patient returns a `410` error from the api. (for that specific workspace)

In that scenario, this shows the `404` page error. Not sure what we want to show.

Screenshot 2024-01-26 at 2 36 34 PM

src/js/apps/patients/patient/sidebar/sidebar_app.js
paulfalgout: I don't know on this one.. I would assume not, but not sure

src/js/apps/patients/patient/patient_app.js
paulfalgout: Yep. so with our api 404 means that api endpoint doesn't exist.  410 means it's not found on an existing api endpoint.  So this is what we want to show

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -61,21 +61,35 @@ const SidebarView = View.extend({
...
       },
     ]);
 
+    if (canEdit && canManagePatients) {
+      menuOptions.push({
+        text: patientStatus !== 'active' ? i18n.menuOptions.activate : i18n.menuOptions.inactivate,
+        onSelect: bind(this.triggerMethod, this, 'click:activeStatus'),
+      });
+
+      if (patientStatus !== 'archived') {
```
paulfalgout: can we move these strings into the static.js?

src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -61,21 +61,35 @@ const SidebarView = View.extend({
...
+    const canManagePatients = this.getOption('canManagePatients');
 
     const menuOptions = new Backbone.Collection([
       {
+        text: canEdit ? i18n.menuOptions.edit : i18n.menuOptions.view,
         onSelect: bind(this.triggerMethod, this, canEdit ? 'click:patientEdit' : 'click:patientView'),
       },
     ]);
 
+    if (canEdit && canManagePatients) {
```
paulfalgout: is this really `canEdit`?  Can only manually added patients status be adjusted?

I wonder if really this check should be `this.model.getWorkspacePatient().canEdit()`

src/js/views/patients/patient/sidebar/sidebar_views.js
paulfalgout: Definitely could use a modal here

src/js/apps/patients/patient/patient_app.js
paulfalgout: could do: `onFail(options, { response }) { if (response.status === 410`

src/js/entities-service/entities/workspace-patients.js
paulfalgout: Let's make this function a `saveAll` and instead of `newStatus` just pass in `{ status: 'foo' }`

paulfalgout: Rather than needing to pass the patient here, let's modify the initial entity service request to do:
```js
    return new Model({ id: uuid(patientId, workspaceId), _patient: patientId, _workspace: workspaceId });
```
so that we do not need to re-get the workspace or pass the patient.. the data should always match.

src/js/entities-service/entities/patients.js
```diff
@@ -52,6 +52,18 @@ const _Model = BaseModel.extend({
     const workspacePatient = Radio.request('entities', 'get:workspacePatients:model', this.id);
     return workspacePatient.get('status');
   },
+  toggleActiveStatus() {
+    const workspacePatient = Radio.request('entities', 'get:workspacePatients:model', this.id);
```
paulfalgout: Let's remove the `getStatus` function and just do `getWorkspacePatient` for a single way to look it up on the patient

src/js/apps/patients/patient/sidebar/sidebar_app.js
paulfalgout: So on the status widget we can add an `initialize` where we get the `workspacePatient` and then just listenTo the `change:status` on that and re-render.

paulfalgout (changes requested):

src/js/apps/patients/patient/sidebar/sidebar_app.js
nmajor25: In that scenario, if a user clicks on an archived patient in the search results, the user will hit that 410/404 error page.

src/js/apps/patients/patient/sidebar/sidebar_app.js
nmajor25: That worked nicely 👍

src/js/entities-service/entities/workspace-patients.js
nmajor25: Good call, updated to that 👍

src/js/views/patients/patient/sidebar/sidebar_views.js
nmajor25: Done 👍

src/js/apps/patients/patient/patient_app.js
nmajor25: Updated this to do the `if (response.status === 410)`.

src/js/views/patients/widgets/widgets.js
```diff
@@ -64,14 +64,21 @@ const widgets = {
     },
     template: hbs`{{formatMessage (intlGet "patients.widgets.widgets.sex") sex=sex}}`,
   },
-  status: {
+  status: View.extend({
```
nmajor25, line 67: Had to do `View.extend` here to get the `initialize()` function to work.

src/js/entities-service/entities/patients.js
nmajor25: Done 👍

src/js/entities-service/entities/workspace-patients.js
```diff
@@ -16,6 +17,12 @@ const _Model = BaseModel.extend({
 
     this.save(attrs, { relationships }, opts);
   },
+  canEdit() {
+    const patient = Radio.request('entities', 'patients:model', this.get('_patient'));
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+
+    return patient.isNew() || currentUser.can('patients:manage');
```
nmajor25: Do we need the `isNew()` check here? Not 100% sure...

src/js/views/patients/patient/sidebar/sidebar_views.js
nmajor25: Hmm right, probably don't need to check if it's a manually added patient.

In https://github.com/RoundingWell/care-ops-frontend/pull/1227/commits/11b4960b624c3b100536e0d2f743af89e3773186, I created a `workspacePatient.canEdit()` function and use that here instead. Where I check `isNew()` and the `currentUser.can('patient:manage')` permission.

src/js/views/patients/widgets/widgets.js
```diff
@@ -64,14 +64,21 @@ const widgets = {
...
     template: hbs`<span class="widgets__status-{{ status }}">{{formatMessage (intlGet "patients.widgets.widgets.status") status=status}}</span>`,
+    initialize() {
+      const workspacePatient = Radio.request('entities', 'get:workspacePatients:model', this.model.get('id'));
```
paulfalgout: never have to do `.get('id')` `.id` is always better.  But instead can this call `this.model.getWorkspacePatient()`?  We can also probably just `this.listenTo(workspacePatient, 'change:status', this.render)`

src/js/views/patients/patient/sidebar/sidebar_views.js
paulfalgout: I guess I'm not sure if we need the isNew check.  I don't think this sidebar can be shown for a new patient, so we'll never be able to hit that coverage

src/js/entities-service/entities/patients.js
paulfalgout: I mean can we replace the `getStatus()` function from this entity up on 53 and make a `getWorkspacePatient`?

paulfalgout (changes requested):

src/js/views/patients/widgets/widgets.js
nmajor25: I updated the listener to this in the status widget:

```js
initialize() {
  this.workspacePatient = this.model.getWorkspacePatient();

  this.listenTo(this.workspacePatient, 'change:status', this.render);
},
templateContext() {
  return {
    status: this.workspacePatient.get('status'),
  };
},
```

src/js/entities-service/entities/patients.js
nmajor25: I believe that's the update I made... this `/entities/patients.js` file no longer has the `getStatus()` function. And I created the `getWorkspacePatient()` and use that instead of `patient.getStatus()`.

src/js/views/patients/patient/sidebar/sidebar_views.js
nmajor25: Okay, I removed the `isNew()` check from `workspacePatient.canEdit()`. Which now only checks that the current user has the `patient:manage` permission.

src/js/views/patients/patient/sidebar/sidebar_views.js
nmajor25: Confirmation modal has been added:


https://github.com/RoundingWell/care-ops-frontend/assets/35355575/2a2b04b7-896e-41c7-b246-09fbb42c2c81

paulfalgout (approved):

PR #1233, opened 2024-02-08
Add is_report option to forms

paulfalgout:
Shortcut Story ID: [sc-47120]

This adds it to both the form app and the pdf form service.

We should probably really only be using this option with `action.tags`

### DIFF
--- a/src/js/formservice.js
+++ b/src/js/formservice.js
@@ -34,25 +34,27 @@ const ActionFormApp = App.extend({
       });
   },
   _getPrefillFilters(form, action) {
+    const opts = {
+      status: FORM_RESPONSE_STATUS.SUBMITTED,
+      flow: action.get('_flow'),
+      patient: action.get('_patient'),
+    };
+
+    const isReport = form.isReport();
+
+    if (isReport) opts.created = `<=${ action.get('created_at') }`;
+
     const prefillActionTag = form.getPrefillActionTag();
-    const flowId = action.get('_flow');
-    const patientId = action.get('_patient');
 
     if (prefillActionTag) {
-      return {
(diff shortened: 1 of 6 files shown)


test/integration/formservice/formservice.js
```diff
@@ -184,7 +188,8 @@ context('Formservice', function() {
       .its('search')
       .should('contain', 'filter[action.tags]=foo-tag')
       .should('contain', 'filter[flow]=1')
-      .should('contain', 'filter[patient]=1');
+      .should('contain', 'filter[patient]=1')
+      .should('contain', `filter[created]=<=${ createdAt }`);
```
shadowhand, line 192: This would be URL encoded in a live scenario, right?

shadowhand (approved):

zfixler (approved):

test/integration/formservice/formservice.js
paulfalgout: Screenshot 2024-02-08 at 10 50 09 PM

yeah the URL encode happens on the entities and then that `itsUrl` translates it back for the tests as well

src/js/formservice.js
```diff
@@ -34,25 +34,27 @@ const ActionFormApp = App.extend({
...
   _getPrefillFilters(form, action) {
+    const opts = {
+      status: FORM_RESPONSE_STATUS.SUBMITTED,
+      flow: action.get('_flow'),
+      patient: action.get('_patient'),
+    };
+
+    const isReport = form.isReport();
+
+    if (isReport) opts.created = `<=${ action.get('created_at') }`;
```
shadowhand, line 45: @paulfalgout I missed that this was using a complete timestamp...

PR #1234, opened 2024-02-13
Fix coverage drop related to handling `410` error in the patient app

nmajor25:
Shortcut Story ID: [sc-47198]

Fixes coverage drop here: https://coveralls.io/builds/65582818/source?filename=src%2Fjs%2Fapps%2Fpatients%2Fpatient%2Fpatient_app.js#L46 that was introduced in https://github.com/RoundingWell/care-ops-frontend/pull/1227.

Which is an `else` branch not being covered when we show a `410` not found error on patient app related pages.

To fix it, we added an `istanbul ignore else` since other error scenarios besides `410` are handled elsewhere in the codebase. So there's nothing to do in the tests for this specific `else` scenario.

### DIFF
--- a/src/js/apps/patients/patient/patient_app.js
+++ b/src/js/apps/patients/patient/patient_app.js
@@ -43,6 +43,7 @@ export default SubRouterApp.extend({
   },
 
   onFail(options, { response }) {
+    /* istanbul ignore else: other error scenarios handled elsewhere */
     if (response.status === 410) {
       Radio.trigger('event-router', 'notFound');
       this.stop();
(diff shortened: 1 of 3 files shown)


test/integration/patients/patient/dashboard.js
nmajor25, line 1: Having this test in the patient dashboard test file makes more sense to me.

src/js/apps/patients/patient/patient_app.js
```diff
@@ -43,6 +43,7 @@ export default SubRouterApp.extend({
   },
 
   onFail(options, { response }) {
+    /* istanbul ignore else: other error scenarios hanlded elsewhere */
```
paulfalgout: handled

paulfalgout (changes requested):

src/js/apps/patients/patient/patient_app.js
nmajor25: Oops, just fixed that 👍

paulfalgout (approved):

PR #1235, opened 2024-02-13
In add clinician modal, workspaces should be added back to the droplist after removal

nmajor25:
Shortcut Story ID: [sc-48049]

Below is an example of the bug this fixes. In that scenario, the `All Patients` workspace should be added back to the droplist after being removed as a selection.

https://github.com/RoundingWell/care-ops-frontend/assets/35355575/eee1dac6-35d9-4ab2-9f51-61fb1fba9814

Also fixes this coverage drop: https://coveralls.io/jobs/135883378/source_files/20988475378#L112. Which was originally caused by https://github.com/RoundingWell/care-ops-frontend/pull/1223.


paulfalgout (approved):
No coveralls comment 😞 

Anyhow.. https://coveralls.io/builds/65680023 only missing branch is the 410 branch that's in develop.

PR #1236, opened 2024-02-15
Add handlebars template widget

paulfalgout:
Shortcut Story ID: [sc-47977]

- [x] Add tests

So this was spicier than expected.  Mainly because our two handlebars solutions hardcode `handlebars/runtime`.

I could have required templates be precompiled so we could use the runtime, but then we'd be `eval()`ing the precompiled templates.. that seemed less of an issue than adding the handlebars compiler.

So the easiest fix was to add the helpers to both instances.  It's almost no duplicate code in the build, but there will be multiple instances of the handlebars runtime.. but it's relatively small.

some widget example:
```json
{
  "type": "widgets",
  "id": "template",
  "attributes": {
    "definition": {
      "widget_type": "widget",
      "template": "<hr><div>{{far 'calendar-days'}}Sex: <b>{{ sex }}</b></div><hr>",
      "display_name": "Template"
    },
    "values": {
      "sex": "@patient.sex"
    }
  }
}
```

### DIFF
--- a/src/js/apps/patients/patient/sidebar/sidebar_app.js
+++ b/src/js/apps/patients/patient/sidebar/sidebar_app.js
@@ -15,6 +15,8 @@ export default App.extend({
   onBeforeStart({ patient }) {
     this.showView(new SidebarView({ model: patient }));
 
+    this.widgets = Radio.request('bootstrap', 'sidebarWidgets');
+
     this.getRegion('widgets').startPreloader();
   },
   beforeStart({ patient }) {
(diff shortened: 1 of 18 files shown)


src/js/views/patients/widgets/widgets.js
```diff
@@ -45,6 +47,19 @@ function getWidgetValue({ fields, name, key, childValue }) {
 // NOTE: These widgets are documented in ./README.md
```
nmajor25, line 47: Update the docs at some point?

src/js/apps/patients/patient/sidebar/sidebar_app.js
nmajor25, line 1: Do we need/want to make these same modifications to the `widgets_header_app`?

nmajor25 (approved):

PR #1237, opened 2024-02-26
Re-render all nav after change:workspace

paulfalgout:
`change:workspace` was only calling `this.showMainNavDroplist` but due to workspace related settings, all the nav needs to be re-rendered.

Shortcut Story ID: [sc-48192]


nmajor25 (commented):
👍

nmajor25 (approved):

PR #1238, opened 2024-03-12
Create GraphQL entities service for `icd($term)`

nmajor25:
Shortcut Story ID: [sc-48884]

Implemented on the backend here: https://github.com/RoundingWell/care-ops-backend/pull/7540

To do:

- [x] Add cypress test for retrieving icd code
- [x] Handle `400` api request errors properly
- [x] Add cypress test when `getIcd()` request returns a `400` error

### DIFF
--- a/src/js/entities-service/graphql.js
+++ b/src/js/entities-service/graphql.js
@@ -1,5 +1,5 @@
 import BaseEntity from 'js/base/entity-service';
-import fetcher from 'js/base/fetch';
+import fetcher, { handleJSON } from 'js/base/fetch';
 
 const Entity = BaseEntity.extend({
   radioRequests: {
@@ -32,7 +32,7 @@ const Entity = BaseEntity.extend({
       method: 'POST',
       body: JSON.stringify({ query, variables }),
     })
-      .then(response => response.json());
+      .then(handleJSON);
   },
 });
 
--- a/src/js/entities-service/graphql.js
+++ b/src/js/entities-service/graphql.js
@@ -1,0 +1,40 @@
+import BaseEntity from 'js/base/entity-service';
+import fetcher from 'js/base/fetch';
+
+const Entity = BaseEntity.extend({
+  radioRequests: {
+    'fetch:icd:byTerm': 'fetchIcdByTerm',
+  },
+  fetchIcdByTerm(term) {
+    const variables = { term };
+    const query = `query ($term: String!) {
+      icdCodes(term: $term) {
+        code
+        description
+        hcc_v24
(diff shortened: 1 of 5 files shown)


src/js/entities-service/graphql.js
```diff
@@ -0,0 +1,40 @@
...
+        hcc_v24
+        hcc_v28
+        parent {
+          code
+          description
+        }
+        children {
+          code
+          description
+        }
```
shadowhand, line 23: @zfixler do you need the parent/child in the form? Taking it out will improve performance, if that is possible.

zfixler: I think we will need it for [this story.](https://app.shortcut.com/roundingwell/story/48868/ability-to-flag-invalid-codes)

It's very likely we will want to show the children as options in the form if a parent is added.

src/js/services/forms.js
```diff
@@ -160,6 +161,17 @@ export default App.extend({
...
+  fetchIcd({ term, requestId }) {
+    const channel = this.getChannel();
+
+    return Promise.resolve(Radio.request('entities', 'fetch:icd:byTerm', term))
+      .then(icd => {
+        channel.request('send', 'fetch:icd', { value: get(icd, ['data', 'icdCodes']), requestId });
+      })
+      .catch(({ responseData }) => {
+        channel.request('send', 'fetch:icd', { error: responseData, requestId });
+      });
```
nmajor25, line 173: This catch isn't doing anything when the api request returns a `400`. 

Was thinking about doing something along these lines:

```js
return Promise.resolve(Radio.request('entities', 'fetch:icd:byTerm', term))
  .then(responseData => {
    if (responseData.errors) {
      channel.request('send', 'fetch:icd', { error: responseData.errors, requestId });
      return;
    }

    channel.request('send', 'fetch:icd', { value: get(responseData, ['data', 'icdCodes']), requestId });
  });
```

But that doesn't seem 100% correct to me. Maybe something needs to be changed in the `/entities-service/graphql.js` file? To reject the promise and trigger the `catch()`.

nmajor25: I added a `.then(handleJSON)` in the `src/js/entities-service/graphql.js` file here: https://github.com/RoundingWell/care-ops-frontend/pull/1238/commits/6cf646db80a4ee0356cce8e7f093ebb27449ea9c.

Instead of doing the code snippet I commented above.

The promise is now being rejected on a `400`, so the `catch()` is triggered now.

paulfalgout (approved):

PR #1243, opened 2024-03-20
Refactor `integration/patients/patient/archive.js` tests for jsonapi

nmajor25:
Shortcut Story ID: [sc-48960]

### DIFF
--- a/test/integration/patients/patient/archive.js
+++ b/test/integration/patients/patient/archive.js
@@ -482,7 +482,7 @@ context('patient archive page', function() {
         return fx;
       })
       .routeWorkspaceClinicians(fx => {
-        const nonTeamMemberClinician = {
+        const nonTeamMemberClinician = getClinician({
           id: '22222',
           attributes: {
             name: 'Non Team Member',
@@ -490,7 +490,7 @@ context('patient archive page', function() {
           relationships: {
             team: getRelationship(teamNurse),
           },
-        };
+        });
 
         fx.data = [currentClinican, nonTeamMemberClinician];
 
--- a/test/integration/patients/patient/archive.js
+++ b/test/integration/patients/patient/archive.js
@@ -1,8 +1,18 @@
-import _ from 'underscore';
 import dayjs from 'dayjs';
 
 import { testTs, testTsSubtract } from 'helpers/test-timestamp';
 import { testDate, testDateSubtract } from 'helpers/test-date';
+import { getRelationship } from 'helpers/json-api';
+
+import { getAction } from 'support/api/actions';
+import { getFlow } from 'support/api/flows';
+import { getPatient } from 'support/api/patients';
+import { workspaceOne } from 'support/api/workspaces';
+import { testForm } from 'support/api/forms';
+import { stateDone, stateInProgress, stateTodo } from 'support/api/states';
+import { getClinician, getCurrentClinician } from 'support/api/clinicians';
+import { roleNoFilterEmployee, roleTeamEmployee } from 'support/api/roles';
+import { teamCoordinator, teamNurse } from 'support/api/teams';
 
 context('patient archive page', function() {
   specify('action, flow and events list', function() {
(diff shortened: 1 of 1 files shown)


test/integration/patients/patient/archive.js
```diff
@@ -11,76 +21,101 @@ context('patient archive page', function() {
...
-        fx.data[2].relationships.state = { data: { id: '55555' } };
-        fx.data[2].attributes.updated_at = testTsSubtract(2);
-        fx.data[2].attributes.due_time = '09:00:00';
-        fx.data[2].attributes.due_date = testDateSubtract(2);
-
-        fx.data[1].attributes.name = 'Not In List';
-        fx.data[1].relationships.state = { data: { id: '33333' } };
-        fx.data[1].attributes.updated_at = testTsSubtract(6);
+            relationships: {
+              patient: getRelationship('11111', 'patients'),
```
paulfalgout: is this patient relationship doing anything?  I suspect this relationship doesn't need to be set for the test here

test/integration/patients/patient/archive.js
```diff
@@ -317,69 +352,86 @@ context('patient archive page', function() {
...
-        };
-
-        fx.data[1].attributes.name = 'Third In List';
-        fx.data[1].relationships.state = { data: { id: '55555' } };
-        fx.data[1].relationships.owner = { data: { id: '11111', type: 'teams' } };
-        fx.data[1].attributes.updated_at = testTsSubtract(2);
-        fx.data[1].attributes.due_time = '09:00:00';
-        fx.data[1].attributes.due_date = testDateSubtract(2);
+            relationships: {
+              patient: getRelationship('11111', 'patients'),
```
paulfalgout: maybe not needed?  at least it doesn't match the `getPatient` id so is it doing anything?

test/integration/patients/patient/archive.js
```diff
@@ -317,69 +352,86 @@ context('patient archive page', function() {
...
       })
       .routesForPatientAction()
       .routePatient(fx => {
-        fx.data.id = '1';
-        fx.data.relationships.workspaces.data = [
-          {
-            id: '11111',
-            type: 'workspaces',
+        fx.data = getPatient({
+          id: '1',
```
paulfalgout: I wonder if we need to set the ids on these `getPatient`?

test/integration/patients/patient/archive.js
```diff
@@ -418,46 +470,69 @@ context('patient archive page', function() {
...
       .routeCurrentClinician(fx => {
-        fx.data.relationships.role = { data: { id: '77777' } };
-        fx.data.relationships.team = { data: { id: '11111', type: 'teams' } };
+        fx.data = currentClinican;
 
         return fx;
       })
       .routeWorkspaceClinicians(fx => {
-        fx.data = _.first(fx.data, 2);
+        const nonTeamMemberClinician = {
```
paulfalgout: seems like this should be in a `getClinician`?  Also we could not set the id and then use `nonTeamMemberClinician.id` in the below tests

test/integration/patients/patient/archive.js
```diff
@@ -317,69 +352,86 @@ context('patient archive page', function() {
...
-        fx.data[1].attributes.updated_at = testTsSubtract(6);
+        fx.data = [
+          getFlow({
+            attributes: {
+              name: 'Second In List',
+              updated_at: testTsSubtract(1),
+            },
+            relationships: {
+              state: getRelationship(stateDone),
+              owner: getRelationship('11111', 'clinicians'),
```
paulfalgout: I believe all of the `getRelationship('11111', 'clinicians')` could be

```js
const currentClinician = getCurrentClinician();

... getRelationship(currentClinician),
```

and the `11111` tests could look at `currentClinician.id`

test/integration/patients/patient/archive.js
```diff
@@ -11,76 +21,101 @@ context('patient archive page', function() {
...
-          relationships: {
-            patient: { data: { id: '11111' } },
-            owner: {
-              data: {
-                id: '11111',
-                type: 'clinicians',
-              },
+        fx.data = [
+          getAction({
+            id: '1',
```
paulfalgout, line 42: one thing I tried in the recent tests that might be better.. I saved the resources that are used to test urls such that rather than setting an id to `1` and testing `visit('/foo/1')` I switched to ````visit(`/foo/${ resource.id }`)```

I think we could probably accomplish that here which will probably safer in the future.

paulfalgout (changes requested):

test/integration/patients/patient/archive.js
nmajor25: Doesn't appear to be, that's been removed.

test/integration/patients/patient/archive.js
nmajor25: Removed without incident.

test/integration/patients/patient/archive.js
nmajor25: I don't think we need them, I deleted them without any issues.

test/integration/patients/patient/archive.js
nmajor25: I updated this to `getClinician()`. Good call, I missed that.

Looks like we're required to include an `id` with `getClinician()`: https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/support/api/clinicians.js#L31. I don't fully understand why though.

test/integration/patients/patient/archive.js
nmajor25: I took a crack at this in this commit: https://github.com/RoundingWell/care-ops-frontend/pull/1243/commits/dc624e02c295ae0810340622428d9b0c577864f8

test/integration/patients/patient/archive.js
nmajor25: Why is that method the safer option?

paulfalgout: well the trend to move towards uuid v5 requires uuid ids

paulfalgout (approved):

PR #1245, opened 2024-03-22
Refactor `test/integration/patients/patient/dashboard.js` tests for jsonapi

nmajor25:
Shortcut Story ID: [sc-49091]

### DIFF
--- a/test/integration/patients/patient/dashboard.js
+++ b/test/integration/patients/patient/dashboard.js
@@ -405,7 +405,7 @@ context('patient dashboard page', function() {
       .should('contain', `patient-action/${ testAction.id }/form/${ testForm.id }`);
   });
 
-  specify('add action and flow', function() {
+  specify.only('add action and flow', function() {
     const currentClinican = getCurrentClinician({
       relationships: {
         role: getRelationship(roleEmployee),
@@ -583,7 +583,7 @@ context('patient dashboard page', function() {
               archived_at: null,
             },
             relationships: {
-              porgram: getRelationship(testPrograms[0].id, 'programs'),
+              program: getRelationship(testPrograms[0]),
               state: getRelationship(stateTodo),
               owner: getRelationship(teamOther),
             },
@@ -597,7 +597,7 @@ context('patient dashboard page', function() {
               archived_at: null,
             },
             relationships: {
-              program: getRelationship(testPrograms[1].id, 'programs'),
+              program: getRelationship(testPrograms[1]),
             },
           }),
           getFlow({
@@ -609,7 +609,7 @@ context('patient dashboard page', function() {
               archived_at: null,
             },
             relationships: {
-              program: getRelationship(testPrograms[1].id, 'programs'),
+              program: getRelationship(testPrograms[1]),
             },
           }),
           getFlow({
@@ -621,7 +621,7 @@ context('patient dashboard page', function() {
               archived_at: null,
             },
             relationships: {
-              program: getRelationship(testPrograms[1].id, 'programs'),
+              program: getRelationship(testPrograms[1]),
             },
           }),
           getFlow({
@@ -633,7 +633,7 @@ context('patient dashboard page', function() {
               archived_at: testTs(),
             },
             relationships: {
-              program: getRelationship(testPrograms[1].id, 'programs'),
+              program: getRelationship(testPrograms[1]),
             },
           }),
           getFlow({
@@ -645,7 +645,7 @@ context('patient dashboard page', function() {
               archived_at: null,
             },
             relationships: {
-              program: getRelationship(testPrograms[1].id, 'programs'),
+              program: getRelationship(testPrograms[1]),
             },
           }),
         ];
--- a/test/integration/patients/patient/dashboard.js
+++ b/test/integration/patients/patient/dashboard.js
@@ -2,7 +2,7 @@ import dayjs from 'dayjs';
 
 import { testTs, testTsSubtract } from 'helpers/test-timestamp';
 import { testDate, testDateSubtract } from 'helpers/test-date';
-import { getRelationship } from 'helpers/json-api';
+import { getRelationship, mergeJsonApi } from 'helpers/json-api';
 
 import { getAction } from 'support/api/actions';
 import { getFlow } from 'support/api/flows';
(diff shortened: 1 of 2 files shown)


test/integration/patients/patient/dashboard.js
```diff
@@ -31,7 +40,7 @@ function createActionPostRoute(id) {
 context('patient dashboard page', function() {
   specify('action and flow list', function() {
     const testTime = dayjs(testDate()).hour(12).valueOf();
-    const actionData = {
+    const testActionData = getAction({
       id: '1',
```
paulfalgout: it would be nice to not set this and just use `testActionData.id` where `1` is currently needed.

test/integration/patients/patient/dashboard.js
```diff
@@ -42,72 +51,86 @@ context('patient dashboard page', function() {
...
+            attributes: {
+              name: 'Second In List',
+              updated_at: testTsSubtract(1),
+            },
+            relationships: {
+              state: getRelationship(stateInProgress),
+            },
+          }),
+          getFlow({
+            id: '2',
```
paulfalgout: similarly would be great to `getFlow` above this into some sort of `testFlow` for `testFlow.id`

test/integration/patients/patient/dashboard.js
```diff
@@ -380,189 +403,254 @@ context('patient dashboard page', function() {
...
-        fx.data[3].relationships['program-actions'] = { data: [] };
-        fx.data[3].relationships['program-flows'] = { data: [] };
-
-        fx.data[4].id = 5;
-        fx.data[4].attributes.name = 'Should not show - archived';
-        fx.data[4].attributes.published_at = testTs();
-        fx.data[4].attributes.archived_at = testTs();
-        fx.data[4].relationships['program-actions'] = { data: [] };
-        fx.data[4].relationships['program-flows'] = { data: [] };
+        fx.data = [
```
paulfalgout: this test setup would be best to create the elements at the top using the uuids that come back.. _however_ there is a circular relationship so some of the `getFoo`s will need to be passed `uuid`s since they need to be used in a relationship prior to their instantiation.

paulfalgout (changes requested):

test/integration/patients/patient/dashboard.js
nmajor25: Updated 👍

test/integration/patients/patient/dashboard.js
nmajor25: Updated 👍

test/integration/patients/patient/dashboard.js
nmajor25: I made some updates to this test. But probably more that can be done here.

test/integration/patients/patient/dashboard.js
```diff
@@ -376,198 +402,258 @@ context('patient dashboard page', function() {
...
+          getFlow({
+            id: '4',
+            attributes: {
+              name: '1 Flow',
+              behavior: 'standard',
+              published_at: testTs(),
+              archived_at: null,
+            },
+            relationships: {
+              porgram: getRelationship(testPrograms[0].id, 'programs'),
```
paulfalgout: porgram
also `program: getRelationship(testPrograms[0].id, 'programs'),` and `program: getRelationship(testPrograms[0])` should be the same result

test/integration/patients/patient/dashboard.js
```diff
@@ -376,198 +402,258 @@ context('patient dashboard page', function() {
...
+      getProgram({
+        attributes: {
+          name: 'Two Actions, One Published, One Flow',
+          published_at: testTs(),
+          archived_at: null,
+        },
+        relationships: {
+          'program-flows': getRelationship([{ id: '4' }], 'flows'),
+          'program-actions': getRelationship(
+            [{ id: '1' }, { id: '4' }, { id: '5' }, { id: '6' }],
```
paulfalgout: all of the ids and the ones below should be able to be done by using only `getProgram`, `getProgramAction` and `getProgramFlow` except that you'll need `uuid`s to setup the circular ids.. should be like on patient-flow or program-flow tests as well.

paulfalgout (changes requested):

test/integration/patients/patient/dashboard.js
```diff
@@ -405,7 +405,7 @@ context('patient dashboard page', function() {
       .should('contain', `patient-action/${ testAction.id }/form/${ testForm.id }`);
   });
 
-  specify('add action and flow', function() {
+  specify.only('add action and flow', function() {
```
nmajor25: Forgot the `.only` here 🤦, but it's removed in the next commit.

test/integration/patients/patient/dashboard.js
nmajor25: This is updated.

test/integration/patients/patient/dashboard.js
nmajor25: I've got a solution for this. But it feels overly complicated. Open to any better ideas.

test/integration/patients/patient/dashboard.js
```diff
@@ -405,13 +406,16 @@ context('patient dashboard page', function() {
...
-  specify.only('add action and flow', function() {
+  specify('add action and flow', function() {
     const currentClinican = getCurrentClinician({
       relationships: {
         role: getRelationship(roleEmployee),
       },
     });
 
+    const testProgramActionIds = [uuid(), uuid(), uuid(), uuid(), uuid(), uuid()];
+    const testProgramFlowIds = [uuid(), uuid(), uuid(), uuid(), uuid(), uuid()];
```
nmajor25: I needed to set these all at the top. And then use them throughout.

Otherwise using e.g. `testProgramActions[0].id` or `testProgramFlows[0].id` in `testPrograms = []` would cause errors. Since we'd be trying to use data/variables that weren't instantiated yet.

paulfalgout: I think if you flipped it and made programIds with uuid then you could define the flows/actions first and use the ids generated from there for relations on the program.
You could also use `_.times` with `uuid`

paulfalgout (changes requested):
nearly there

test/integration/patients/patient/dashboard.js
```diff
@@ -405,46 +407,203 @@ context('patient dashboard page', function() {
...
+      }),
+      getFlow({
+        attributes: {
+          name: '2 Flow',
+          behavior: 'standard',
+          published_at: testTs(),
+          archived_at: null,
+        },
+        relationships: {
+          program: getRelationship(testProgramIds[1], 'programs'),
```
nmajor25, line 508: The test still passes if these are removed... 🤔

test/integration/patients/patient/dashboard.js
nmajor25: Think that works well 👍 , and `_.times` is also much better

paulfalgout (approved):

PR #1249, opened 2024-03-29
Refactor `/patients/patient/patient-flow.js` tests for jsonapi

nmajor25:
Shortcut Story ID: [sc-49400]

### DIFF
--- a/test/integration/patients/patient/patient-flow.js
+++ b/test/integration/patients/patient/patient-flow.js
@@ -2,64 +2,66 @@ import _ from 'underscore';
 
 import { testTs, testTsSubtract } from 'helpers/test-timestamp';
 import { testDateAdd, testDateSubtract } from 'helpers/test-date';
-import { getErrors } from 'helpers/json-api';
+import { getErrors, getRelationship, mergeJsonApi } from 'helpers/json-api';
+
+import { getFlow } from 'support/api/flows';
+import { getPatient } from 'support/api/patients';
+import { getAction } from 'support/api/actions';
+import { getProgramAction } from 'support/api/program-actions';
+import { getProgramFlow } from 'support/api/program-flows';
+import { getClinician, getCurrentClinician } from 'support/api/clinicians';
+import { getForm, testForm } from 'support/api/forms';
+import { stateInProgress, stateDone, stateTodo } from 'support/api/states';
+import { teamCoordinator, teamNurse, teamPharmacist, teamPhysician } from 'support/api/teams';
+import { roleNoFilterEmployee, roleTeamEmployee } from 'support/api/roles';
 
 const tomorrow = testDateAdd(1);
 
 context('patient flow page', function() {
+  const testFlow = getFlow({
+    attributes: {
+      name: 'Test Flow',
+    },
+  });
+
+  const testAction = getAction({
+    attributes: {
+      name: 'Test Action',
+    },
+    relationships: {
+      flow: getRelationship(testFlow),
+    },
+  });
+
   beforeEach(function() {
     cy
       .routesForPatientDashboard();
   });
 
   specify('context trail', function() {
+    const testPatient = getPatient({
+      attributes: {
+        first_name: 'Test',
+        last_name: 'Patient',
+      },
+    });
+
     cy
       .routesForDefault()
-      .routeFlows(fx => {
-        fx.data = _.sample(fx.data, 1);
-
-        fx.data[0].id = '1';
-        fx.data[0].attributes.name = 'Test Flow';
-        fx.data[0].relationships.state.data.id = '33333';
-        fx.data[0].relationships.patient.data.id = '1';
-
-        fx.included = [{
-          id: '1',
-          type: 'patients',
-          attributes: {
-            first_name: 'Test',
-            last_name: 'Patient',
-          },
-        }];
-
-        return fx;
-      })
+      .routeFlows()
       .routeFlow(fx => {
-        fx.data.id = '1';
-
-        fx.data.attributes.name = 'Test Flow';
-        fx.data.attributes.updated_at = testTs();
-        fx.data.relationships.patient = { data: { id: '1' } };
-
-        fx.included.push({
-          id: '1',
-          type: 'patients',
-          attributes: {
-            first_name: 'Test',
-            last_name: 'Patient',
+        fx.data = mergeJsonApi(testFlow, {
+          relationships: {
+            patient: getRelationship(testPatient),
           },
         });
+
+        fx.included = [testPatient];
 
         return fx;
       })
       .routePatientByFlow()
-      .routeFlowActions(fx => {
-        const flow = _.find(fx.included, { type: 'flows' });
-
-        flow.attributes.name = 'Test Flow';
-
-        return fx;
-      })
+      .routeFlowActions()
       .visit('/worklist/owned-by')
       .wait('@routeActions');
 
@@ -85,7 +87,7 @@ context('patient flow page', function() {
 
     cy
       .url()
-      .should('contain', '/patient/dashboard/1');
+      .should('contain', `/patient/dashboard/${ testPatient.id }`);
 
     cy
       .go('back');
@@ -103,19 +105,20 @@ context('patient flow page', function() {
   specify('patient flow action sidebar', function() {
     cy
       .routesForPatientAction()
-      .routeFlow()
+      .routeFlow(fx => {
+        fx.data = testFlow;
+
+        return fx;
+      })
       .routeFlowActions()
       .routeAction(fx => {
-        fx.data.id = '1';
-        fx.data.attributes.name = 'Test Action';
-
-        fx.data.relationships.flow = { data: { id: '1' } };
+        fx.data = testAction;
 
         return fx;
       })
       .routePatientByFlow()
       .routeActionActivity()
-      .visit('/flow/1/action/1')
+      .visit(`/flow/${ testFlow.id }/action/${ testAction.id }`)
       .wait('@routeFlow')
       .wait('@routePatientByFlow')
       .wait('@routeFlowActions');
@@ -130,24 +133,19 @@ context('patient flow page', function() {
     cy
       .routesForPatientAction()
       .routeFlow(fx => {
-        fx.data.id = '1';
-        fx.data.relationships.state = {
-          data: { id: '55555' },
-        };
+        fx.data = testFlow;
+
         return fx;
       })
       .routeFlowActions()
       .routeAction(fx => {
-        fx.data.id = '1';
-        fx.data.attributes.name = 'Test Action';
-
-        fx.data.relationships.flow = { data: { id: '1' } };
+        fx.data = testAction;
 
         return fx;
       })
       .routePatientByFlow()
       .routeActionActivity()
-      .visit('/flow/1/action/1')
+      .visit(`/flow/${ testFlow.id }/action/${ testAction.id }`)
       .wait('@routeFlow')
       .wait('@routePatientByFlow')
       .wait('@routeFlowActions');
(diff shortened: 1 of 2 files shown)


paulfalgout (changes requested):
Similar to the other PRs I'd like to see us use the ids from`getPatient`/`getAction` etc as uuids.  And if there's a circular relationship only then utilize `uuid()` to create the ids passed to the `getAction` or whichever

paulfalgout:
https://github.com/RoundingWell/care-ops-frontend/blob/develop/test/integration/programs/program/program-flow.js#L14 an example of the need for `uuid`

nmajor25:
@paulfalgout took a crack at it, let me know what you think 👍

test/integration/patients/patient/patient-flow.js
```diff
@@ -1,65 +1,71 @@
...
+import { teamCoordinator, teamNurse, teamPharmacist, teamPhysician } from 'support/api/teams';
+import { roleNoFilterEmployee, roleTeamEmployee } from 'support/api/roles';
 
 const tomorrow = testDateAdd(1);
 
 context('patient flow page', function() {
+  const testFlowId = uuid();
+
+  const testFlow = getFlow({
+    id: testFlowId,
```
paulfalgout: you don't need the `testFlowId` here you can use `testFlow.id` later and just use `getRelationship(testFlow)` below.. you would need a `uuid` if the flow also related to the action

test/integration/patients/patient/patient-flow.js
```diff
@@ -1,65 +1,71 @@
...
-
-        fx.included.push({
-          id: '1',
-          type: 'patients',
-          attributes: {
-            first_name: 'Test',
-            last_name: 'Patient',
+        fx.data = mergeJsonApi(testFlow, {
+          relationships: {
+            patient: getRelationship(testPatient.id, 'patients'),
```
paulfalgout: `getRelationship(testPatient)`

test/integration/patients/patient/patient-flow.js
```diff
@@ -162,69 +164,67 @@ context('patient flow page', function() {
...
-          id: '44444',
-          type: 'teams',
-        };
-
-        fx.included = _.reject(fx.included, { type: 'flows' });
-
-        fx.included.push({ id: '1', type: 'forms', attributes: { name: 'Test Form' } });
+        fx.data = [
+          getAction({
+            id: '1',
```
paulfalgout: do we need to set the ids in this array?

test/integration/patients/patient/patient-flow.js
```diff
@@ -162,69 +164,67 @@ context('patient flow page', function() {
...
+            attributes: {
+              name: 'First In List',
+              due_date: testDateSubtract(1),
+              created_at: testTsSubtract(1),
+              sequence: 1,
+              outreach: 'patient',
+              sharing: 'sent',
+            },
+            relationships: {
+              flow: getRelationship(testFlowId, 'flows'),
```
paulfalgout: just `getRelationship(testFlow)` here and below

test/integration/patients/patient/patient-flow.js
```diff
@@ -290,9 +290,14 @@ context('patient flow page', function() {
 
     cy
       .routeAction(fx => {
-        fx.data.id = '2';
-        fx.data.relationships.state = { data: { id: '22222' } };
-        fx.data.relationships.form = { data: { id: '1' } };
+        fx.data = getAction({
+          id: '2',
```
paulfalgout: can we not set the id here too

test/integration/patients/patient/patient-flow.js
```diff
@@ -432,139 +434,115 @@ context('patient flow page', function() {
...
-            behavior: 'conditional',
-            details: '',
-            days_until_due: 0,
-            sequence: 0,
-          },
+        fx.data = mergeJsonApi(testFlow, {
           relationships: {
-            owner: { data: null },
-            form: { data: { id: '11111' } },
+            'program-flow': getRelationship('1', 'program-flows'),
```
paulfalgout: I think we can use `getProgramFlow` here instead of an arbitrary id

test/integration/patients/patient/patient-flow.js
```diff
@@ -432,139 +434,115 @@ context('patient flow page', function() {
...
-            sequence: 3,
-          },
-          relationships: {
-            owner: { data: { id: '11111', type: 'teams' } },
-            form: { data: { id: '11111' } },
-          },
-        });
+        const testProgramActions = [
+          getProgramAction({
+            id: '1',
```
paulfalgout: how about these ids?

test/integration/patients/patient/patient-flow.js
```diff
@@ -586,8 +564,13 @@ context('patient flow page', function() {
 
     cy
       .routeAction(fx => {
-        fx.data.id = 'test-1';
-        fx.data.attributes.name = 'Conditional';
+        fx.data = getAction({
+          id: 'test-1',
```
paulfalgout: I think we can remove this too

test/integration/patients/patient/patient-flow.js
```diff
@@ -686,79 +670,80 @@ context('patient flow page', function() {
...
-          type: 'teams',
-        };
-
-        fx.data[3].relationships.owner.data = {
-          id: '22222',
-          type: 'teams',
-        };
+        fx.data = [
+          getAction({
+            id: '1',
```
paulfalgout: here too

paulfalgout (commented):
I think we should attempt to not use `id:` if we can

test/integration/patients/patient/patient-flow.js
nmajor25: Updated 👍

test/integration/patients/patient/patient-flow.js
nmajor25: Updated 👍 And switched to `getRelationship(testFlow)` throughout.

test/integration/patients/patient/patient-flow.js
nmajor25: The second one needs some sort of id set on it (`id: '2'`). Since we do this `PATCH` later on in the test:

```js
.intercept('PATCH', '/api/actions/2', {
  statusCode: 204,
  body: {},
})
```

The test errors with a `401` error without that `id: '2'`.

test/integration/patients/patient/patient-flow.js
nmajor25: Yup, updated 👍

test/integration/patients/patient/patient-flow.js
nmajor25: Same comment as above. Think we need to set some sort of it here because of the `PATCH` request this test is making.

test/integration/patients/patient/patient-flow.js
nmajor25: Updated 👍 had to do a little extra since that same program flow needs to be in the `fx.included`.

test/integration/patients/patient/patient-flow.js
nmajor25: The test fails without the `id: '1'` on the first one. 

Since we do a check against it later in the test:

```js
expect(data.relationships['program-action'].data.id).to.equal('1');
```

But the other ones can be removed.

test/integration/patients/patient/patient-flow.js
nmajor25: Removed 👍

test/integration/patients/patient/patient-flow.js
nmajor25: Removed 👍

test/integration/patients/patient/patient-flow.js
paulfalgout: well why can't we store the id that comes back from the `getProgramAction` and then test against that instead?

test/integration/patients/patient/patient-flow.js
```diff
@@ -162,69 +160,65 @@ context('patient flow page', function() {
...
+            relationships: {
+              flow: getRelationship(testFlow),
+              state: getRelationship(stateTodo),
+              owner: getRelationship(teamNurse),
+              form: getRelationship(testForm),
+              files: getRelationship([{ id: '1' }], 'files'),
+            },
+          }),
+          getAction({
+            id: '2',
```
paulfalgout: id

test/integration/patients/patient/patient-flow.js
paulfalgout: the patch should use the id returned from the `getAction`

test/integration/patients/patient/patient-flow.js
```diff
@@ -627,7 +600,7 @@ context('patient flow page', function() {
     cy
       .wait('@routeAction')
       .url()
-      .should('contain', 'flow/1/action/test-1');
+      .should('contain', `flow/${ testFlow.id }/action/test-1`);
```
paulfalgout: I'm surprised this works... it should use the id from the `getAction`.. how is `test-1` not failing here...

test/integration/patients/patient/patient-flow.js
```diff
@@ -1211,70 +1218,68 @@ context('patient flow page', function() {
...
-          id: '44444',
-          type: 'teams',
-        };
-
-        fx.included = _.reject(fx.included, { type: 'flows' });
-
-        fx.included.push({ id: '11111', type: 'forms', attributes: { name: 'Test Form' } });
+        fx.data = [
+          getAction({
+            id: '1',
```
paulfalgout: ids here and down

test/integration/patients/patient/patient-flow.js
```diff
@@ -1323,8 +1328,12 @@ context('patient flow page', function() {
 
     cy
       .routeAction(fx => {
-        fx.data.id = '1';
-        fx.data.relationships.state.data.id = '22222';
+        fx.data = getAction({
+          id: '1',
```
paulfalgout: id

paulfalgout (changes requested):

test/integration/patients/patient/patient-flow.js
nmajor25: Updated in https://github.com/RoundingWell/care-ops-frontend/pull/1249/commits/e422b40b6a59a5afa856450c31590c79b3781071.

test/integration/patients/patient/patient-flow.js
nmajor25: Updated to that in https://github.com/RoundingWell/care-ops-frontend/pull/1249/commits/213eb815c98f99c91386d8d0beedf05721ea99e5

test/integration/patients/patient/patient-flow.js
nmajor25: Updated in https://github.com/RoundingWell/care-ops-frontend/pull/1249/commits/bd245f8de17726f7f73611135d8f4a0377d26f45

test/integration/patients/patient/patient-flow.js
```diff
@@ -371,17 +369,14 @@ context('patient flow page', function() {
       .should('contain', '11:15 AM');
 
     cy
-      .intercept('DELETE', '/api/actions/2', {
+      .intercept('DELETE', '/api/actions/*', {
```
paulfalgout: I think `${ testListAction.id }` would probably be better.

test/integration/patients/patient/patient-flow.js
```diff
@@ -429,142 +424,116 @@ context('patient flow page', function() {
...
+        name: 'Conditional',
+        published_at: testTs(),
+        archived_at: null,
+        behavior: 'conditional',
+        details: '',
+        days_until_due: 0,
+        sequence: 0,
+      },
+      relationships: {
+        owner: getRelationship(null),
```
paulfalgout: I think this is just `getRelationship()`

test/integration/patients/patient/patient-flow.js
```diff
@@ -429,142 +424,116 @@ context('patient flow page', function() {
...
-                },
-                {
-                  id: '5',
-                  type: 'program-actions',
-                },
-              ],
-            },
-          },
-        });
+        fx.included.push(testProgramFlow);
```
paulfalgout, line 529: I just realized this is pushing program data onto the include of getFlow.. which isn't where this data is coming from.. we should be able to construct the test data above the `.routeFoos` and then use them for relationships in this route but still add them to where they should be coming from..

test/integration/patients/patient/patient-flow.js
```diff
@@ -686,79 +660,76 @@ context('patient flow page', function() {
...
+              state: getRelationship(stateDone),
+              owner: getRelationship(teamNurse),
+              flow: getRelationship(testFlow),
+            },
+          }),
+        ];
 
-        fx.included = _.reject(fx.included, { type: 'flows' });
-        fx.included.push({
+        fx.included.push(getClinician({
```
paulfalgout: similarly it's odd that this is being pushed to the included.  we should be able to set this on the bootstrapped `routeWorkspaceClinicians` like we would elsewhere

test/integration/patients/patient/patient-flow.js
```diff
@@ -1035,39 +1035,42 @@ context('patient flow page', function() {
...
-        fx.data[0].relationships.state.data.id = '22222';
-        fx.data[1].relationships.state.data.id = '22222';
-        fx.data[2].relationships.state.data.id = '22222';
-
-        fx.data[0].relationships.flow.data = { id: '1' };
-        fx.data[1].relationships.flow.data = { id: '1' };
-        fx.data[2].relationships.flow.data = { id: '1' };
-
-        fx.included = _.reject(fx.included, { type: 'flows' });
+        fx.data = [
```
paulfalgout: I actually think this might be a good case for `getActions` since all the relationships are the same.

test/integration/patients/patient/patient-flow.js
```diff
@@ -1323,8 +1327,12 @@ context('patient flow page', function() {
 
     cy
       .routeAction(fx => {
-        fx.data.id = '1';
-        fx.data.relationships.state.data.id = '22222';
+        fx.data = getAction({
```
paulfalgout: this is a little odd.  seems like it should be `fx.data = testFlowActions[0];` Seems like the relation is already set there.

test/integration/patients/patient/patient-flow.js
```diff
@@ -1757,38 +1765,43 @@ context('patient flow page', function() {
...
       .routeFlowActions(fx => {
-        fx.data = _.first(fx.data, 3);
-
-        _.each(fx.data, (action, index) => {
-          action.id = `${ index + 1 }`;
-          action.relationships.flow = { data: { id: '1' } };
-          action.relationships.state.data.id = '22222';
-        });
+        fx.data = [
+          getAction({
```
paulfalgout: maybe `getActions`

test/integration/patients/patient/patient-flow.js
```diff
@@ -1852,69 +1865,84 @@ context('patient flow page', function() {
...
+            attributes: {
+              name: 'First In List',
+              due_date: testDateAdd(5),
+              sequence: 0,
+            },
+            relationships: {
+              flow: getRelationship(testFlow),
+              state: getRelationship(stateTodo),
+              owner: getRelationship('11111', 'clinicians'),
```
paulfalgout: `getCurrentClinician`

test/integration/patients/patient/patient-flow.js
```diff
@@ -1852,69 +1865,84 @@ context('patient flow page', function() {
...
+          getAction({
+            attributes: {
+              name: 'Last In List',
+              due_date: testDateAdd(5),
+              sequence: 3,
+            },
+            relationships: {
+              flow: getRelationship(testFlow),
+              state: getRelationship(stateTodo),
+              owner: getRelationship('11111', 'clinicians'),
```
paulfalgout: `getCurrentClinician`

test/integration/patients/patient/patient-flow.js
```diff
@@ -1852,69 +1865,84 @@ context('patient flow page', function() {
...
-        fx.data.id = '1';
-
-        const flowActions = _.sample(fx.data.relationships.actions.data, 4);
-
-        _.each(flowActions, (action, index) => {
-          action.id = `${ index + 1 }`;
+        fx.data = mergeJsonApi(testFlow, {
+          relationships: {
+            state: getRelationship(stateInProgress),
+            owner: getRelationship('11111', 'clinicians'),
```
paulfalgout: `getCurrentClinician`

paulfalgout (changes requested):
wow.. huge refactor, but the consistency is showing.. caught a couple other things this pass.

test/integration/patients/patient/patient-flow.js
nmajor25: Updated in https://github.com/RoundingWell/care-ops-frontend/pull/1249/commits/2668ece9915bf3a5089523336e0ddb58f572073e.

test/integration/patients/patient/patient-flow.js
nmajor25: Updated in https://github.com/RoundingWell/care-ops-frontend/pull/1249/commits/93742de5055d28fe1db71aa6540c4f6fdc142151.

test/integration/patients/patient/patient-flow.js
nmajor25: Updated in https://github.com/RoundingWell/care-ops-frontend/pull/1249/commits/547129d020b12f89b46152ff55d9bde5fcc8e573.

test/integration/patients/patient/patient-flow.js
nmajor25: Updated in https://github.com/RoundingWell/care-ops-frontend/pull/1249/commits/38ecf9ddfd0517a6583e7ff1cf53b77c0bc39fff.

test/integration/patients/patient/patient-flow.js
nmajor25: Updated in https://github.com/RoundingWell/care-ops-frontend/pull/1249/commits/dff45dd23f28609ead160543dac53a42dcff8024.

test/integration/patients/patient/patient-flow.js
nmajor25: Updated in https://github.com/RoundingWell/care-ops-frontend/pull/1249/commits/fc0b51aeca4b760a1e42007ce49934791773ca60.

test/integration/patients/patient/patient-flow.js
nmajor25: Updated three of these in https://github.com/RoundingWell/care-ops-frontend/pull/1249/commits/95fdae5462810ae830c55e4bdd8ff2ff3ed711ad.

test/integration/patients/patient/patient-flow.js
nmajor25: Still working on this.

I'm not tracking where the data is supposed to be coming from?

paulfalgout: I believe it's all or one of these:
```
    .routePrograms()
    .routeAllProgramActions()
    .routeAllProgramFlows();
```

nmajor25: Still can't get this to work.

I've tried several things, but just spinning my wheels. Latest attempt is here: https://github.com/RoundingWell/care-ops-frontend/pull/1249/commits/6f94a791fe0ecfa2e23d3983486ad0d2c6b65612.

In the `flow_app.js`, I can't get `this.flow.getProgramFlow()` to return the full program flow model. It just returns `{id: 'uuid'}` without any of the associated actions data.

paulfalgout: man..  I put you on a goose chase..  so.. I must've been looking at the add on patient dashboard.. this one is using that `programFlow.getAddableActions()` which is using the `includes`. 🤦   So the include was correct..........

nmajor25: Okay I was wondering about that, but couldn't tell for sure. I reverted that last commit. I'll also need to fix some merge conflicts before this is ready as well.

paulfalgout: It would be more ideal to reset the head prior to the commit rather than reverting it for a new PR to reduce the blame depth.

nmajor25: 👍 , just reset it back.

paulfalgout (approved):

PR #1250, opened 2024-03-29
Refactor `/patients/patient/patient.js` tests for jsonapi

nmajor25:
Shortcut Story ID: [sc-49487]

### DIFF
--- a/test/integration/patients/patient/patient.js
+++ b/test/integration/patients/patient/patient.js
@@ -1,16 +1,27 @@
+import { getRelationship } from 'helpers/json-api';
+
+import { getPatient } from 'support/api/patients';
+import { getAction } from 'support/api/actions';
+import { stateDone } from 'support/api/states';
+
 context('patient page', function() {
+  const testPatient = getPatient({
+    attributes: {
+      first_name: 'First',
+      last_name: 'Last',
+    },
+  });
+
   specify('context trail', function() {
     cy
       .routesForPatientDashboard()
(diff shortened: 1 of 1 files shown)


test/integration/patients/patient/patient.js
```diff
@@ -1,16 +1,30 @@
...
   specify('context trail', function() {
     cy
       .routesForPatientDashboard()
       .routeActions()
       .routePatient(fx => {
-        fx.data.id = '1';
-        fx.data.attributes.first_name = 'First';
-        fx.data.attributes.last_name = 'Last';
+        fx.data = getPatient({
+          id: testPatientId,
```
paulfalgout: I don't know that an id needs to be set?  `getPatient` will return an id that the visit can then use.

paulfalgout (changes requested):
I think we can remove the `uuid` and just use the ids returned from` getFoo` for the `visits`

test/integration/patients/patient/patient.js
nmajor25: I think that makes sense as well. I just pushed an updated commit...

nmajor25: I'll work on making these same types of changes to the other PRs as well.

nmajor25:
@paulfalgout I think this PR is good to go? Unless I'm missing something else that needs to be changed...

paulfalgout (approved):

PR #1251, opened 2024-04-08
Update widgets for new API attribute naming

paulfalgout:
Shortcut Story ID: [sc-49711]

### DIFF
--- a/src/js/entities-service/entities/forms.js
+++ b/src/js/entities-service/entities/forms.js
@@ -1,9 +1,8 @@
-import { get, size } from 'underscore';
+import { get, size, invoke, map, compact } from 'underscore';
 import Radio from 'backbone.radio';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
-import collectionOf from 'js/utils/formatting/collection-of';
 
 const TYPE = 'forms';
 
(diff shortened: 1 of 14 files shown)


src/js/entities-service/entities/forms.js
```diff
@@ -62,14 +61,20 @@ const _Model = BaseModel.extend({
...
-    const formWidgets = get(this.get('options'), 'widgets');
+    const formWidgets = get(this.get('options'), ['widgets', 'widgets']);
 
-    return Radio.request('entities', 'widgets:collection', collectionOf(get(formWidgets, 'widgets'), 'id'));
+    if (!formWidgets) return Radio.request('entities', 'widgets:collection');
+
+    const allWidgets = Radio.request('bootstrap', 'widgets');
+
+    const widgets = map(formWidgets, slug => {
+      return allWidgets.find({ slug }).omit('id');
```
paulfalgout: all defined widgets should always be available.  `omit`ing `id` here creates a json object so that the widget collection can calculate the `modelId` for purposes of display.  It's possible in the future we may want the request to be `widgets:collection:forDisplay` or something where the collection is different than the standard collection.

src/js/entities-service/entities/forms.js
```diff
@@ -62,14 +61,20 @@ const _Model = BaseModel.extend({
...
+    });
+
+    return Radio.request('entities', 'widgets:collection', widgets);
   },
   getWidgetFields() {
-    const formWidgets = get(this.get('options'), 'widgets');
-
-    return get(formWidgets, 'fields');
+    return get(this.get('options'), ['widgets', 'fields']);
```
paulfalgout, line 75: no need for 2 `get`s when one will do.

src/js/views/patients/widgets/README.md
```diff
@@ -47,11 +47,11 @@ Deprecated:
...
-The following widget with a `widget_type: "workspaces"`:
+The following widget with a `category: "workspaces"`:
 
 ```json
   {
-    "widget_type": "workspaces",
+    "category": "workspaces",
     "id": "workspaces",
     "definition": {
       "display_name": "Display Name"
```
shadowhand, line 57: FWIW, widgets do have a `name` that is independent of `slug` now, so we probably don't need to have display name any more. That being said, the default `name` is the value of `slug`, so there would need to be coordinated clean up and/or migration.

test/fixtures/test/widgets.json
```diff
@@ -1,40 +1,40 @@
 [
   {
-    "widget_type": "dob",
-    "id": "dob",
+    "category": "dob",
+    "slug": "dob",
     "definition": {
       "display_name": "Date of Birth"
```
shadowhand, line 6: Exactly like this, we can move `definition.display_name > name`.

shadowhand (commented):
Some thoughts on future changes, otherwise looks good from an API standpoint.

src/js/views/patients/widgets/README.md
paulfalgout: I thought about that, but display name is an optional label that gets added in the sidebar.. And really you could utilize the same widget differently with different display names

nmajor25 (approved):
Looks good 👍

PR #1252, opened 2024-04-08
Move filterable directories request from `bootstrap` => `app-frame_app`

nmajor25:
Shortcut Story ID: [sc-49745]

Backend will now filter directories by workspace. On the FE, this ensures the filterable directories request is made when a user switches workspaces (`GET /api/directories?filter[filterable]=true`).

Related backend PR: https://github.com/RoundingWell/care-ops-backend/pull/7696.

### DIFF
--- a/test/integration/globals/app-nav.js
+++ b/test/integration/globals/app-nav.js
@@ -233,8 +233,23 @@ context('App Nav', function() {
 
         return fx;
       })
+      .routeDirectories(fx => {
+        fx.data = [
+          {
+            attributes: {
+              name: 'Insurance Plans',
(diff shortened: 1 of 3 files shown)


test/integration/globals/app-nav.js
nmajor25, line 1: Loads `workspaceOne` with the `Insurance Plans` filterable directory.

And when the user switches to `workspaceTwo`, simulate the `/api/directories?filter[filterable]=true` request returning a different filterable directory (`ACO` instead).

paulfalgout (approved):
We might want to wait on the BE merge just to keep it simple

PR #1253, opened 2024-04-09
Ownership menu: keep the current user at the top of the picklist

nmajor25:
Shortcut Story ID: [sc-49433]

- [x] Fix any failing tests

### DIFF
--- a/src/js/views/patients/shared/components/owner_component.js
+++ b/src/js/views/patients/shared/components/owner_component.js
@@ -1,4 +1,3 @@
-import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 import hbs from 'handlebars-inline-precompile';
 
@@ -70,6 +69,8 @@ export default Droplist.extend({
       infoText: this.getOption('infoText'),
       headingText: this.getOption('headingText'),
       placeholderText: this.getOption('placeholderText'),
+      canClear: this.getOption('hasCurrentClinician') && this.clinicians.get(this.currentUser),
+      clearText: this.currentUser.get('name'),
     };
   },
   viewOptions() {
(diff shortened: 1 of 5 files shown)


src/js/views/patients/shared/components/owner_component.js
```diff
@@ -116,31 +117,30 @@ export default Droplist.extend({
...
       });
     }
 
     this.setState({ selected: owner });
   },
+  onPicklistSelect({ model }) {
+    this.setState('selected', model || this.currentUser);
+
+    this.popRegion.empty();
+  },
```
nmajor25, line 143: I couldn't get this to work without overriding this `Droplist` function.

If I don't override this, it's called before the `onChangeSelected()` function below. Which makes things go 💥 since this owner component then is attempting to access a `selected` object that is `null`.

nmajor25: So when I attempted this: 

```js
onChangeSelected(selected) {
  if (!selected) {
    this.setState({ selected: currentUser });
      return;
  }

  this.triggerMethod('change:owner', selected);
},
```

The app would 💥 before this function even got the chance to run.

src/js/views/patients/shared/components/owner_component.js
```diff
@@ -70,6 +69,8 @@ export default Droplist.extend({
       infoText: this.getOption('infoText'),
       headingText: this.getOption('headingText'),
       placeholderText: this.getOption('placeholderText'),
+      canClear: this.getOption('hasCurrentClinician') && this.clinicians.get(this.currentUser),
+      clearText: this.currentUser.get('name'),
```
nmajor25, line 73: I feel like this would look look better if we added the team abbreviation to the current user name. But I'm not sure how to do that and keep the same styling as the others below it in the picklist.

Can we pass a `hbs` template as the `clearText`?

Screenshot 2024-04-09 at 4 41 33 PM

src/js/views/patients/shared/components/owner_component.js
nmajor25, line 1: It's not related to these code changes, but I'm noticing a bug with the owner component. That it sometimes doesn't trigger the change every time you click to switch owners. And to make it work you have to close the droplist and re-open it.

So I'm planning on investigating that more as well.

src/js/views/patients/shared/components/owner_component.js
paulfalgout: I don't think we should do this right now.  That'll be a fairly substantial change to the component.  If Mark doesn't like that it doesn't have the team we should change this to "Assign to Me"

src/js/views/patients/shared/components/owner_component.js
paulfalgout: I think that's probably because of `const isTeam = selected.type === 'teams';` but this seems ok to me.

paulfalgout (approved):

src/js/views/patients/shared/components/owner_component.js
nmajor25: Cool, works for me 👍

src/js/views/patients/shared/components/owner_component.js
nmajor25: I'm going to create a separate story to investigate this more.

paulfalgout (approved):

paulfalgout:
I keep getting a "Merge already in progress" message so I'm going to close and reopen to see if it fixes anything

PR #1256, opened 2024-04-23
Refresh stale forms after 30 min of no change

nmajor25:
Shortcut Story ID: [sc-50208]

- [x] Add tests

### DIFF
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -27,7 +27,10 @@ export default App.extend({
   },
   initialize(options) {
     this.updateDraft = debounce(this.updateDraft, 15000);
+    this.refreshForm = debounce(this.refreshForm, 1800000);
+
     this.mergeOptions(options, ['action', 'form', 'patient', 'responses', 'latestResponse']);
+
     this.currentUser = Radio.request('bootstrap', 'currentUser');
   },
   radioRequests: {
@@ -48,6 +51,8 @@ export default App.extend({
   },
   readyForm() {
     this.trigger('ready');
+
+    this.refreshForm();
   },
   checkVersion(feVersion) {
     /* istanbul ignore if: can't test reload */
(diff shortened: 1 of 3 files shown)


src/js/services/forms.js
nmajor25, line 1: When the form initially loads or reloads (on `ready:form`), set the `refreshForm()` function to run in 30 min.

When the user makes a change to the form (represented by the `updateStoredSubmission()` function being called), set a new `refreshForm()` to be called in 30 min.

When the user submits the form, cancel the scheduled `refreshForm()` function.

src/js/services/forms.js
```diff
@@ -48,6 +51,8 @@ export default App.extend({
   },
   readyForm() {
     this.trigger('ready');
+
+    this.refreshForm();
```
nmajor25, line 55: I initially put this in the `initialize()` function. But if a user is prompted with the keep/discard work prompt, `this.refreshForm()` would be called before the actual form has been shown to the user.

So I put it here instead. To wait until the form content is loaded.

paulfalgout (approved):
Looks correct to me

nmajor25:
@paulfalgout cool, I'll get some tests created for this.

test/integration/forms/action.js
nmajor25, line 1: The test is designed to simulate a user having a form open in their browser that becomes "stale". And after 30 minutes of inactivity, we reload the form so the current state of the form is displayed to the user.

Test sequence:

1. Load an empty form.
2. Restart the form after 30 minutes of no changes (via `cy.tick()`). The form is reloaded with a new draft latest form response.
    - Simulating a form draft having been created in a different tab/window/session.
4. Verify that the form loaded the latest form response data into the form. 
5. Restart the form after 30 more minutes of no changes (via `cy.tick()`). The form is reloaded with a new submitted form response.
    - Simulating a form draft having been submitted in a different tab/window/session.
6. Verify that the form is now submitted. Content shown should be from the submitted form response and the submit button should be in the `Update` state.

PR #1259, opened 2024-05-07
If worklist is restarting, don't set new sort comparator

nmajor25:
Shortcut Story ID: [sc-50540]

### DIFF
--- a/src/js/base/app.js
+++ b/src/js/base/app.js
@@ -23,14 +23,16 @@ export default App.extend({
   },
   beforeStart: noop,
   onSyncData(fetchId, options, args = []) {
+    if (!this._isRunning || this._fetchId !== fetchId) return;
+
     this._isLoading = false;
-    if (!this.isRunning() || this._fetchId !== fetchId) return;
 
     this.finallyStart.call(this, options, ...args);
   },
   triggerSyncFail(fetchId, options, ...args) {
+    if (!this._isRunning || this._fetchId !== fetchId) return;
+
     this._isLoading = false;
-    if (!this.isRunning() || this._fetchId !== fetchId) return;
 
     this.triggerMethod('fail', options, ...args);
   },
--- a/test/integration/patients/worklist/worklist.js
+++ b/test/integration/patients/worklist/worklist.js
@@ -4132,4 +4132,38 @@ context('worklist page', function() {
       .find('button')
       .should('not.exist');
   });
+
+  // NOTE: needs to be moved to base class component tests
+  specify('change sort before list is done loading', function() {
+    cy
+      .routesForPatientAction()
+      .routeFlows()
+      .routeActions()
+      .routeFlow()
+      .routeFlowActions()
+      .routePatientByFlow()
+      .routePatientField();
+
+    cy
+      .intercept('GET', '/api/actions*', { delay: 1000, body: { data: [] } })
+      .visit('/worklist/owned-by');
+
+    cy
+      .get('[data-date-filter-region]')
+      .should('contain', 'Added:')
+      .should('contain', 'This Month')
+      .click();
+
+    cy
+      .get('.app-frame__pop-region')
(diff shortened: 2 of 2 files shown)


src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -55,7 +55,7 @@ export default App.extend({
     this.setState(filterState.getFiltersState());
   },
   onChangeStateSort() {
-    if (!this.isRunning()) return;
+    if (!this.isRunning() || this.isRestarting()) return;
```
nmajor25: I can't reproduce this bug in my dev environment, so can't tell if this actually fixes the issue.

Also still contemplating how to test this.

paulfalgout: Here's the test, but `isRestarting` doesn't fix the issue:
```js
    cy
      .intercept('GET', '/api/flows*', { delay: 1000, body: { data: [] } })
      .visit('/worklist/owned-by');

    cy
      .get('[data-date-filter-region]')
      .should('contain', 'Added:')
      .should('contain', 'This Month')
      .click();

    cy
      .get('.app-frame__pop-region')
      .contains('Last Week')
      .click();

    cy
      .get('.worklist-list__filter-sort')
      .click()
      .get('.picklist')
      .contains('Patient Last: A')
      .click();
```

paulfalgout (changes requested):

src/js/apps/patients/worklist/worklist_app.js
nmajor25: What about checking that the `list` view has a collection before setting the comparator?

Like:

```js
  onChangeStateSort() {
    if (!this.isRunning()) return;

    const listView = this.getChildView('list');

    if (listView.collection) {
      listView.setComparator(this.getComparator());
    }
  },
```

That seems to work...

paulfalgout:
OK so.. this fixed seemed oddly unnecessary..  how was it that the app was running.. wasn't restarting and still had the preloader running.

It turns out the issue was deeper and it might subsequently also solve the issue we're seeing in the program action bug.  This'll actually be an odd place to test this.. Really at some point we need to add component tests for the base classes.  But I think we can keep this test and just add a comment that links to this PR and a message about moving it to a component test.

So in `base/app.js` it was setting loading to false, but that `fetchId` check prevents older fetches from resolving..  so.. if you make two requests one after another where the 2nd should take precedent.. the first one clears the loading flag which modifies running..

Currently:
```js
  onSyncData(fetchId, options, args = []) {
    this._isLoading = false;
    if (!this.isRunning() || this._fetchId !== fetchId) return;

    this.finallyStart.call(this, options, ...args);
  },
  triggerSyncFail(fetchId, options, ...args) {
    this._isLoading = false;
    if (!this.isRunning() || this._fetchId !== fetchId) return;

    this.triggerMethod('fail', options, ...args);
  },
```

Change to:  (notice the switch from `isRunning()` to `_isRunning`)
```js
onSyncData(fetchId, options, args = []) {
    if (!this._isRunning || this._fetchId !== fetchId) return;

    this._isLoading = false;

    this.finallyStart.call(this, options, ...args);
  },
  triggerSyncFail(fetchId, options, ...args) {
    if (!this._isRunning || this._fetchId !== fetchId) return;

    this._isLoading = false;

    this.triggerMethod('fail', options, ...args);
  },
```

shadowhand:
@paulfalgout that explanation makes a ton of sense. My spidey senses were tingling when I saw this story because it seemed like something we would have noticed before, as the particular trigger is relatively common.

nmajor25:
@paulfalgout interesting... I updated the PR to your solution and it seems to fix this worklist bug (via the test passing).

But it doesn't fix the program action sidebar bug in #1260.

paulfalgout (approved):

PR #1260, opened 2024-05-08
Set `this.action` in the `onBeforeStart()` in program action sidebar

nmajor25:
Shortcut Story ID: [sc-50520]

### DIFF
--- a/src/js/apps/programs/sidebar/action-sidebar_app.js
+++ b/src/js/apps/programs/sidebar/action-sidebar_app.js
@@ -11,9 +11,10 @@ export default App.extend({
   beforeStart() {
     return Radio.request('entities', 'fetch:tags:collection');
   },
-  onStart({ action }, tags) {
+  onBeforeStart({ action }) {
     this.action = action;
-
+  },
+  onStart(options, tags) {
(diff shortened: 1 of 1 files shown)


src/js/apps/programs/sidebar/action-sidebar_app.js
```diff
@@ -11,9 +11,10 @@ export default App.extend({
   beforeStart() {
     return Radio.request('entities', 'fetch:tags:collection');
   },
-  onStart({ action }, tags) {
+  onBeforeStart({ action }) {
     this.action = action;
-
+  },
+  onStart(options, tags) {
```
nmajor25, line 17: This appears to fix the issue.

Instead of doing the `if (this.action)` in the `onStop()`. Which stopped the app from borking but caused weirdness in the UI since the `this.action.trigger('editing', false)` doesn't get called correctly.

paulfalgout:
This fix seems perfectly fine, but I'm interested to see if https://github.com/RoundingWell/care-ops-frontend/pull/1259#issuecomment-2101970524 fixes it

paulfalgout (approved):

PR #1263, opened 2024-05-23
Remove adding adhoc actions

paulfalgout:
Shortcut Story ID: [sc-51161]

This did affect quite a bit of things.

There's likely some further clean up around `action.trigger('editing', true);` which I don't love.  the lists can probably state-wise understand if an action is being edited at render rather than waiting for it to run to then trigger.

- [x] Fix coverage drop

### DIFF
--- a/test/integration/patients/sidebar/action-sidebar.js
+++ b/test/integration/patients/sidebar/action-sidebar.js
@@ -11,272 +11,6 @@ import stateColors from 'helpers/state-colors';
 import { workspaceOne } from 'support/api/workspaces';
 
 context('action sidebar', function() {
-  specify('display new action sidebar', function() {
-    cy
-      .routesForPatientAction()
-      .routePatient(fx => {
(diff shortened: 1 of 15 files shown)


test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -11,272 +11,6 @@ import stateColors from 'helpers/state-colors';
...
-    cy
-      .get('.sidebar')
-      .find('[data-name-region] .js-input')
-      .type('Test Name');
-
-    cy
-      .get('.sidebar')
-      .find('[data-save-region] .js-cancel')
-      // Need force because Cypress does not recognize the element is typeable
-      .type('{enter}', { force: true });
```
paulfalgout, line 90: need to move this test

nmajor25 (approved):
Code looks good. Didn't find anything when testing the app locally 👍

PR #1264, opened 2024-05-28
Add `visible_to` filter to the add UI for program actions/flows

nmajor25:
Shortcut Story ID: [sc-51292]

### DIFF
--- a/src/js/entities-service/entities/program-actions.js
+++ b/src/js/entities-service/entities/program-actions.js
@@ -1,4 +1,4 @@
-import { extend, first } from 'underscore';
+import { extend, first, size } from 'underscore';
 import Radio from 'backbone.radio';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
@@ -82,6 +82,15 @@ const _Model = BaseModel.extend({
   hasOutreach() {
     return this.get('outreach') !== ACTION_OUTREACH.DISABLED;
   },
+  isVisibleToCurrentUser() {
+    const visibleToList = this.get('_visible_to');
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    const currentUserTeam = currentUser.getTeam();
+
+    if (!size(visibleToList)) return true;
+
+    return !!visibleToList.find(team => team.id === currentUserTeam.id);
+  },
   saveForm(form) {
     form = this.toRelation(form);
     const saveData = { _form: form.data };
@@ -136,8 +145,9 @@ const Collection = BaseCollection.extend({
       const isPublished = !!action.get('published_at');
       const isArchived = !!action.get('archived_at');
       const isAutomated = action.get('behavior') === PROGRAM_BEHAVIORS.AUTOMATED;
+      const isVisible = action.isVisibleToCurrentUser();
 
-      return isPublished && !isArchived && !isAutomated;
+      return isPublished && !isArchived && !isAutomated && isVisible;
     });
 
     clone.reset(addable);
--- a/test/integration/patients/patient/dashboard.js
+++ b/test/integration/patients/patient/dashboard.js
@@ -429,6 +429,7 @@ context('patient dashboard page', function() {
         relationships: {
           owner: getRelationship(teamCoordinator),
           form: getRelationship(testForm),
+          visible_to: getRelationship([teamNurse]),
         },
       }),
       getAction({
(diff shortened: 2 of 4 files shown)


test/integration/patients/patient/dashboard.js
nmajor25, line 1: Goal is to cover these scenarios in the tests for both addable flows and actions:

1. `relationship.visible_to` does not exist (treated the same as being an empty array). Action/flow should be addable.
2. `relationship.visible_to` exists and contains the current user's team. Action/flow should be addable.
3. `relationship.visible_to` exists but does not contain the current user's team. Action/flow should not be addable.

src/js/entities-service/entities/program-actions.js
```diff
@@ -129,15 +129,25 @@ const Collection = BaseCollection.extend({
       data: JSON.stringify({ data }),
     });
   },
+  isVisibleToCurrentUser(action) {
+    const visibleToList = action.get('_visible_to');
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    const currentUserTeamId = currentUser.get('_team');
+
+    if (!size(visibleToList)) return true;
```
nmajor25: If `relationships.visible_to` is `null`/`undefined` or an empty `[]`, assume it's addable by every team/user.

test/integration/patients/patient/dashboard.js
```diff
@@ -552,6 +568,18 @@ context('patient dashboard page', function() {
...
+      getFlow({
+        attributes: {
+          name: 'Should not show - not visible to current user team',
+          behavior: 'standard',
+          published_at: testTs(),
+          archived_at: null,
+        },
+        relationships: {
+          program: getRelationship(testProgramIds[1], 'programs'),
+          visible_to: { data: [{ type: 'teams', id: teamNurse.id }] },
```
nmajor25: Better way to do this?

src/js/entities-service/entities/program-actions.js
```diff
@@ -129,15 +129,25 @@ const Collection = BaseCollection.extend({
       data: JSON.stringify({ data }),
     });
   },
+  isVisibleToCurrentUser(action) {
```
nmajor25: Open to any better naming conventions for these functions/variables.

test/integration/patients/patient/dashboard.js
shadowhand: If `teamNurse` has `id, type` defined, then we could do this ... ?

```suggestion
          visible_to: { data: [ teamNurse ] },
```

nmajor25: Yep that works 👍 . Via `visible_to: getRelationship([teamNurse])`. I forgot we had examples of doing that elsewhere in our tests.

src/js/entities-service/entities/program-actions.js
```diff
@@ -129,15 +129,25 @@ const Collection = BaseCollection.extend({
...
+    const visibleToList = action.get('_visible_to');
+    const currentUser = Radio.request('bootstrap', 'currentUser');
+    const currentUserTeamId = currentUser.get('_team');
```
paulfalgout: I think we should us `getTeam()` here.  we can use `.id`

src/js/entities-service/entities/program-actions.js
paulfalgout: I think we can move this function up into the action/flow such that it ends up `const isVisible = action.isVisibleToCurrentUser();`

test/integration/patients/patient/patient-flow.js
```diff
@@ -424,6 +424,12 @@ context('patient flow page', function() {
   });
 
   specify('add action', function() {
+    const currentClinician = getCurrentClinician({
```
paulfalgout: Can we flip these tests here and in patient-action to use the default `teamNurse` that the test `currentClinician` is by default so we don't need to set it?

paulfalgout (changes requested):

test/integration/patients/patient/patient-flow.js
nmajor25: Good call, updated 👍

src/js/entities-service/entities/program-actions.js
nmajor25: Updated 👍

src/js/entities-service/entities/program-actions.js
nmajor25: Updated 👍

paulfalgout (approved):

PR #1265, opened 2024-05-29
Fix small coverage drops

paulfalgout:


paulfalgout:
Not sure where the coveralls comment is, but https://coveralls.io/builds/67773933

nmajor25 (approved):

PR #1266, opened 2024-05-30
Update eslint from `v8` => `v9`

nmajor25:
Shortcut Story ID: [sc-50941]

Also updated to `eslint-webpack-plugin`. Which needed to release a new version to be compatible with eslint v9.

### DIFF
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -103,7 +103,7 @@ export default App.extend({
     try {
       store.set(this.getStoreId(), { submission, updated });
       this.trigger('update:submission', updated);
-    } catch (e) /* istanbul ignore next: Tested locally, test runner borks on CI */ {
+    } catch /* istanbul ignore next: Tested locally, test runner borks on CI */ {
       store.each((value, key) => {
         if (String(key).startsWith('form-subm-')) store.remove(key);
       });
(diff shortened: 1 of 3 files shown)


nmajor25:
Still looking into why the tests are borking. Not happening locally.

paulfalgout (changes requested):
The borking tests are just the linter failing on this line: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/services/forms.js#L106

It's saying the `(e)` should be `()`

nmajor25:
Yep, good call. Think we should be good now 🤞

src/js/services/forms.js
```diff
@@ -103,7 +103,7 @@ export default App.extend({
     try {
       store.set(this.getStoreId(), { submission, updated });
       this.trigger('update:submission', updated);
-    } catch (e) /* istanbul ignore next: Tested locally, test runner borks on CI */ {
```
paulfalgout, line 106: the coverage drop is from the comment being removed

paulfalgout (changes requested):

src/js/services/forms.js
nmajor25: Oops, added that back

paulfalgout (approved):

PR #1274, opened 2024-06-26
Split bootstrap service into smaller services

paulfalgout:
Shortcut Story ID: [sc-52305]

We've never really tested the services directly.  I added tests for both the widgets and settings services, but haven't yet for workspaces / bootstrap.

This is for two reasons.. I'm about to change some of this functionality in the next portion of this work. And second these two are quite a bit harder to implement and so I'll come back around to it after the feature is complete.  We should still see line coverage from the e2e tests.

### DIFF
--- a/src/js/services/bootstrap.js
+++ b/src/js/services/bootstrap.js
@@ -2,10 +2,13 @@ import { includes, reject } from 'underscore';
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 
+import getWorkspaceRoute from 'js/utils/root-route';
+
 import App from 'js/base/app';
 
 import SettingsService from './settings';
 import WidgetsService from './widgets';
+import WorkspaceService from 'js/services/workspace';
 
 // NOTE: Roles are set only at login so they can be cached
 let activeRolesCache;
@@ -55,6 +58,9 @@ export default App.extend({
   },
   initialize({ name }) {
     this.organization = new Backbone.Model({ name });
+
+    // NOTE: handle pre-init'd workspace requests
+    Radio.reply('workspace', 'current');
   },
   beforeStart() {
     return [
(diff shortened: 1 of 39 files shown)


src/js/services/settings.js
```diff
@@ -0,0 +1,27 @@
...
+
+export default App.extend({
+  channelName: 'settings',
+  radioRequests: {
+    'get': 'getSetting',
+  },
+  initialize({ settings }) {
+    this.settings = settings;
+  },
+  getSetting(settingName) {
```
nmajor25, line 14: No longer need the `if (!this.isRunning()) return;` check?

src/js/services/bootstrap.js
```diff
@@ -1,12 +1,15 @@
...
 import Radio from 'backbone.radio';
-import store from 'store';
 
 import getWorkspaceRoute from 'js/utils/root-route';
 
 import App from 'js/base/app';
 
+import SettingsService from './settings';
+import WidgetsService from './widgets';
+import WorkspaceService from 'js/services/workspace';
```
nmajor25: Want to import these all the same?

```js
import SettingsService from 'js/services/settings';
import WidgetsService from 'js/services/widgets';
import WorkspaceService from 'js/services/workspace';
```

src/js/services/settings.js
paulfalgout: correct because it's now started after the bootstrap is started it's init'd only when needed and it no long is `start`ed.  It _might_ be possible to get a warning if this setting gets requested prior to it starting, but I _think_ that was only oddly possible in testing, and generally the warning is more useful than the `return;` I would expect.

src/js/services/bootstrap.js
paulfalgout: good catch.. late copy/paste

nmajor25 (approved):

PR #1275, opened 2024-06-30
Bootstrap clinicians per program shared workspaces

paulfalgout:
Shortcut Story ID: [sc-52417]

### DIFF
--- a/src/js/entities-service/clinicians.js
+++ b/src/js/entities-service/clinicians.js
@@ -1,4 +1,5 @@
 import { setUser } from 'js/datadog';
+import Radio from 'backbone.radio';
 import BaseEntity from 'js/base/entity-service';
 import { _Model, Model, Collection } from './entities/clinicians';
 
@@ -21,8 +22,12 @@ const Entity = BaseEntity.extend({
   },
   fetchByWorkspace(workspaceId) {
     const url = `/api/workspaces/${ workspaceId }/relationships/clinicians`;
+    const workspace = Radio.request('entities', 'workspaces:model', workspaceId);
 
-    return this.fetchCollection({ url });
+    return this.fetchCollection({ url })
+      .then(clinicians => {
+        workspace.updateClinicians(clinicians);
+      });
   },
 });
 
--- a/src/js/services/workspace.js
+++ b/src/js/services/workspace.js
@@ -42,23 +42,45 @@ export default App.extend({
   getDirectories() {
     return this.directories.clone();
   },
+  _getSharedWorkspaces(programs) {
+    const sharedWorkspaces = Radio.request('entities', 'workspaces:collection');
+
+    programs.each(program => {
+      const workspaces = program.getUserWorkspaces();
+
+      sharedWorkspaces.add(workspaces.models);
+    });
+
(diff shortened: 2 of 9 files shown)


src/js/entities-service/clinicians.js
```diff
@@ -21,8 +22,12 @@ const Entity = BaseEntity.extend({
   },
   fetchByWorkspace(workspaceId) {
     const url = `/api/workspaces/${ workspaceId }/relationships/clinicians`;
+    const workspace = Radio.request('entities', 'workspaces:model', workspaceId);
 
-    return this.fetchCollection({ url });
+    return this.fetchCollection({ url })
+      .then(clinicians => {
+        workspace.updateClinicians(clinicians);
```
paulfalgout, line 29: this makes more sense here but is essentially only necessary for added/disabled clinicians.  nearly unnecessary.

src/js/services/workspace.js
```diff
@@ -42,23 +42,55 @@ export default App.extend({
   getDirectories() {
     return this.directories.clone();
   },
+  _getSharedWorkspaceIds(programs) {
```
paulfalgout: This is the crux of the logic of this PR.  However to really tested it aside from adding component tests we'll need the owner droplist UI in the following work

src/js/services/workspace.js
```diff
@@ -42,23 +42,55 @@ export default App.extend({
...
 
-    this.currentWorkspace.updateClinicians(clinicians);
+    const workspaceIds = this._getSharedWorkspaceIds(programs);
+
+    const clinicianRequests = map(workspaceIds, id => {
+      return Radio.request('entities', 'fetch:clinicians:byWorkspace', id);
+    });
 
-    this.resolvePromise(this.currentWorkspace);
+    Promise.all([...clinicianRequests])
```
paulfalgout, line 76: this work does add a serialized request which is not ideal, but it works

test/support/commands.js
```diff
@@ -98,7 +100,7 @@ Cypress.Commands.add('visitOnClock', (url, options = {}) => {
 
   // NOTE: this is a hack to fix the clock around defer around Backbone.history.loadUrl
   cy
-    .wait(200)
+    .wait(50)
```
paulfalgout, line 103: Adding the serial request seems to lessen the need for this wait

test/support/commands.js
```diff
@@ -66,6 +66,8 @@ Cypress.Commands.overwrite('visit', (originalFn, url = '/', options = {}) => {
 
   let waits = [
     '@routeWorkspaceClinicians',
+    '@routeDirectories',
```
paulfalgout, line 69: routeDirectories probably always should have been here.. these are the requests for when a workspace changes.. _we could_ make this just `routeWorkspaceClinicians` as well since it's currently last.. adding directories meant I needed to change one test from `wait` to `get`

src/js/services/workspace.js
paulfalgout: considering the next step of this epic, there's a better way to do this

zfixler (approved):

nmajor25 (approved):

PR #1276, opened 2024-07-03
Owner component workspaces by programs

paulfalgout:
Includes https://github.com/RoundingWell/care-ops-frontend/pull/1275

Shortcut Story ID: [sc-52510]


paulfalgout:
This needs some more tests, but it might be good to go

zfixler (approved):

src/js/i18n/en-US.yml
```diff
@@ -511,7 +511,6 @@ careOptsFrontend:
           system:
             actionCopiedFromProgram: '<strong>Action</strong> added from the { program } program'
             clinicianAssigned: '<strong>Owner</strong> changed to { to_name }'
-            created: '<strong>Action</strong> ({ team }) added'
```
shadowhand, line 514: Why was this deleted?

paulfalgout: At one point long ago "system" creates were filtered out by assuming that actions that came from program-actions were added by the system here: https://github.com/RoundingWell/care-ops-frontend/pull/1276/files#diff-fb5d041965ec94c9fc646e053667635bd8c984c188d4e687937fc816b8eacadbL249

Whenever source was added, the filter wasn't updated, and all events were duplicated are reformatted.  I don't think this one was ever intended to be displayed in the first place, and it is being filtered out.

related: https://github.com/RoundingWell/care-ops-frontend/pull/129

nmajor25 (approved):

PR #1278, opened 2024-07-09
Add link to clinician profile to access owned-by list

nmajor25:
Shortcut Story ID: [sc-52624] [sc-52625]

Screenshot 2024-07-11 at 4 59 56 PM

To do:

- [x] `Owned By` nav sidebar item doesn't have active css state on the route change.
- [x] Get full code coverage for changes made. (update: looks like we have full coverage after all)

### DIFF
--- a/src/js/apps/globals/nav_app.js
+++ b/src/js/apps/globals/nav_app.js
@@ -242,7 +242,7 @@ export default RouterApp.extend({
   },
   _navMatch(navCollection, event, eventArgs) {
     return navCollection.find(model => {
-      return model.get('event') === event && isEqual(model.get('eventArgs'), eventArgs);
+      return model.get('event') === event && isEqual(model.get('eventArgs')[0], eventArgs[0]);
     });
   },
   onBeforeStart() {
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -72,11 +72,14 @@ export default App.extend({
 
     if (storedState) {
       this.setState(storedState);
+      this.getState().setClinicianId(this.clinicianId);
       this.startFiltersApp();
       return;
     }
 
     this.setState({ id: this.worklistId });
+
+    this.getState().setClinicianId(this.clinicianId);
 
     this.startFiltersApp({ setDefaults: true });
   },
@@ -84,7 +87,7 @@ export default App.extend({
     this.collection = null;
     if (!this.isRestarting()) this.stopChildApp('filters');
   },
-  onBeforeStart({ worklistId }) {
+  onBeforeStart({ worklistId, clinicianId }) {
     if (this.isRestarting()) {
       const filtersApp = this.getChildApp('filters');
 
@@ -99,6 +102,7 @@ export default App.extend({
     }
 
     this.worklistId = worklistId;
+    this.clinicianId = clinicianId;
     this.initListState();
 
     this.setView(new LayoutView());
--- a/src/js/views/clinicians/sidebar/clinician-sidebar.scss
+++ b/src/js/views/clinicians/sidebar/clinician-sidebar.scss
@@ -15,3 +15,14 @@
   right: 16px;
   top: 8px;
 }
+
+.clinician-sidebar__worklist {
+  border-top: 1px solid $black-90;
+  padding-top: 16px;
+}
+
+.clinician-sidebar__worklist-info {
+  color: $black-60;
+  font-size: 14px;
+  margin-top: 12px;
+}
--- a/src/js/views/clinicians/sidebar/clinician-sidebar_views.js
+++ b/src/js/views/clinicians/sidebar/clinician-sidebar_views.js
@@ -113,6 +113,7 @@ const WorklistView = View.extend({
     'click .js-button': 'click:button',
   },
   onClickButton() {
+    Radio.request('sidebar', 'close');
     Radio.trigger('event-router', 'worklist', 'owned-by', this.clinician.id);
   },
 });
--- a/src/js/views/clinicians/sidebar/clinician-sidebar_views.js
+++ b/src/js/views/clinicians/sidebar/clinician-sidebar_views.js
@@ -86,6 +86,37 @@ const InfoView = View.extend({
   `,
 });
 
+const WorklistView = View.extend({
+  className: 'clinician-sidebar__worklist',
+  template: hbs`
+    <h3 class="sidebar__heading">
+      {{formatHTMLMessage (intlGet "clinicians.sidebar.clinicianSidebarViews.worklistView.workspaceName") name=workspaceName}}
+    </h3>
+    <button class="button-primary w-100 u-margin--t-8 js-button">
+      {{far "list"}}<span>{{formatHTMLMessage (intlGet "clinicians.sidebar.clinicianSidebarViews.worklistView.worklistBtn") name=clinicianName}}</span>
+    </button>
+    <div class="clinician-sidebar__worklist-info">
+      {{ @intl.clinicians.sidebar.clinicianSidebarViews.worklistView.workspacesInfo }}
+    </div>
+  `,
+  initialize({ clinician, workspace }) {
+    this.clinician = clinician;
+    this.workspace = workspace;
+  },
+  templateContext() {
+    return {
+      clinicianName: this.clinician.get('name'),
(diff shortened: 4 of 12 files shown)


src/js/apps/patients/worklist/worklist_app.js
nmajor25, line 1: Tested this locally and it seems to work. No cypress tests yet. I'm not sure how to test it effectively until we do [sc-52625](https://app.shortcut.com/roundingwell/story/52625) next.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -72,19 +72,22 @@ export default App.extend({
...
     if (storedState) {
       this.setState(storedState);
+      this.getState().setClinicianId(this.clinicianId);
       this.startFiltersApp();
       return;
     }
 
     this.setState({ id: this.worklistId });
 
+    this.getState().setClinicianId(this.clinicianId);
```
nmajor25, line 82: This seems simpler to me than going the `initState()` route. But maybe there's a reason we don't want to use a state method here 🤔

paulfalgout, line 82: Maybe this is ok, but having two sets in a row means we are hitting the localstorage twice.. This would be better in a single set if possible

paulfalgout (changes requested):

src/js/apps/patients/worklist/worklist_app.js
nmajor25: What's the downside to hitting local storage twice?

I'm not quite sure how to implement it in a way that sets it once (yet). It seems like we have to do it in a second call because we have to check for `if (this.clinicianId)` before changing the state. But I can come back to this after I get the link/s added in the UI.

paulfalgout: what's wrong with:
```js
initState(state) {
  if (this.clinicianId) { 
    this.setState(extend({}, state, { clinicianId: this.clinicianId });
    return;
  }
  this.setState(state);
} 
```
or if/else 🤮 
```js
  if (this.clinicianId) {  this.setState({ id: this.worklistId, clinicianId: this.clinicianId }); } else { this.setState({ id: this.worklistId }); }
```

it's likely ok to hit localstorage a bunch.. and because it's before the start we _can_ probably hit the state twice..  it's not _ideal_ to do two subsequent `change` events but it's almost more ideological at this point.. there's no particular issue right now if we want to hit it twice...  but there'd certainly be other similar cases where we _couldn't_ do two subsequent state changes without it rendering twice or something like that..

src/js/views/patients/patient/dashboard/dashboard_views.js
```diff
@@ -130,10 +130,8 @@ const ActionItemView = View.extend({
       return;
     }
 
-    const program = this.model.getProgram();
```
paulfalgout: these all shouldn't be removed

src/js/views/clinicians/sidebar/clinician-sidebar_views.js
```diff
@@ -86,6 +86,37 @@ const InfoView = View.extend({
...
+    </div>
+  `,
+  initialize({ clinician, workspace }) {
+    this.clinician = clinician;
+    this.workspace = workspace;
+  },
+  templateContext() {
+    return {
+      clinicianName: this.clinician.get('name'),
+      workspaceName: this.workspace.get('name'),
```
paulfalgout, line 109: I want to determine that if we aren't listing other workspaces that we still want to show the workspace name

src/js/views/clinicians/sidebar/clinician-sidebar_views.js
```diff
@@ -86,6 +86,37 @@ const InfoView = View.extend({
...
+    };
+  },
+  triggers: {
+    'click .js-button': 'click:button',
+  },
+  onClickButton() {
+    Radio.trigger('event-router', 'worklist', 'owned-by', this.clinician.get('id'));
```
paulfalgout: `this.clinician.id`

src/js/views/clinicians/sidebar/clinician-sidebar.scss
```diff
@@ -15,3 +15,14 @@
   right: 16px;
   top: 8px;
 }
+
+.clinician-sidebar__worklist {
+  border-top: 1px solid $black-90;
+  padding-top: 16px;
+
+  .clinician-sidebar__worklist-info {
```
paulfalgout: should not need to nest this

src/js/views/clinicians/sidebar/clinician-sidebar.scss
```diff
@@ -15,3 +15,14 @@
   right: 16px;
   top: 8px;
 }
+
+.clinician-sidebar__worklist {
```
paulfalgout, line 19: the fact that either of these domain styles need to exist make me feel like our generic sidebar styles need refactoring.  it's surprising that we would need to add styles for this generally.

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: I'm ok with it.  outside of the `onBeforeStart` it would probably cause real issues of multiple restart/renders

paulfalgout (changes requested):
So for the highlight the `_navMatch` in the nav app will need to be modified to do some sort of include instead of equals.

src/js/views/clinicians/sidebar/clinician-sidebar.scss
nmajor25: Updated 👍

src/js/views/clinicians/sidebar/clinician-sidebar_views.js
nmajor25: Updated 👍

src/js/views/patients/patient/dashboard/dashboard_views.js
nmajor25: Not sure where these came from. I think I messed up a rebase/merge. But I removed those changes.

src/js/views/clinicians/sidebar/clinician-sidebar.scss
nmajor25: `.sidebar__section` is almost a perfect match. But it has too much top margin than is shown in Mark's designs.

For the informational text we have `.sidebar__info`. But it has a yellow background we don't want here.

src/js/apps/clinicians/sidebar/clinician-sidebar_app.js
```diff
@@ -26,6 +26,5 @@ export default App.extend({
     this.clinician.trigger('editing', false);
 
     Radio.request('sidebar', 'close');
-    Radio.trigger('event-router', 'clinicians:all');
```
nmajor25: We'll need to keep this line.

And somehow have it only run when the sidebar is being closed via the `X` icon button. Not when routing to the worklist (or elsewhere).

src/js/apps/globals/nav_app.js
```diff
@@ -242,7 +242,11 @@ export default RouterApp.extend({
   },
   _navMatch(navCollection, event, eventArgs) {
     return navCollection.find(model => {
-      return model.get('event') === event && isEqual(model.get('eventArgs'), eventArgs);
+      const eventArgsMatch
+        = contains(model.get('eventArgs'), eventArgs[0])
+        || isEqual(model.get('eventArgs'), eventArgs);
```
nmajor25: My original change broke the `.is-selected` styling functionality for the schedule nav item. It's because the schedule nav item has an empty `[]` for its event arguments.

paulfalgout: maybe just `isEqual(model.get('eventArgs')['0'], eventArgs[0])` would work?  Like I don't think `eventArgs[0]` being any part of the array is correct.. really it's that `[0] === [0]` `if([1]) [1] === [1]` etc..  But _probably_ just the `[0]`s check is ok for now?  we don't really deal with deeply nested routes right now.

paulfalgout (changes requested):

src/js/apps/globals/nav_app.js
nmajor25: Yep, that definitely works 👍 I updated it to that just now

paulfalgout (approved):

PR #1282, opened 2024-07-23
Add prefixes graphql support for icd10 query

paulfalgout:
Shortcut Story ID: [sc-52760]

FE here is basically just passing term and/or prefixes through to the query.  It does support the current api for older forms using `getIcd('By this string')`


paulfalgout:
I have a bug I'll fix shortly.

shadowhand (approved):

nmajor25 (approved):

PR #1285, opened 2024-07-24
Add cache buster to patient endpoint

paulfalgout:
Shortcut Story ID: [sc-52969]

This uses `filter[workspace]={id}`
For our usecase because workspace is in the header it doesn’t actually matter if this API is functional on it’s own for filtering.

It does affect PUT/PATCH, but I don't think that should be an issue either.

But this will prevent cross workspace caching of this particular endpoint.  This ought to be enough even including other `/api/patient/{id}/relationships/foo` endpoints as at least in our current implementation, a base patient call is always made prior to the relationship ones.

### DIFF
--- a/src/js/entities-service/entities/patients.js
+++ b/src/js/entities-service/entities/patients.js
@@ -12,7 +12,12 @@ const TYPE = 'patients';
 
 const _Model = BaseModel.extend({
   type: TYPE,
-  urlRoot: '/api/patients',
+  url() {
+    if (this.isNew()) return '/api/patients';
+
+    const currentWorkspace = Radio.request('workspace', 'current');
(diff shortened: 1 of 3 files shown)


shadowhand (commented):
Do we really want to do this, or do we want to change endpoints?

paulfalgout:
this seems like no cost to me.  I'm not sure what changing endpoints necessarily buys us or how it could complicate all of the various `/relationships/foo` endpoints

nmajor25 (approved):
Code LGTM, assuming this is the route we want to go.

shadowhand (approved):

shadowhand:
We'll see if this causes all kinds of validation errors.

PR #1292, opened 2024-08-07
Create component tests for handlebars helpers

nmajor25:
Shortcut Story ID: [sc-53604]

### DIFF
--- a/src/js/base/helpers.cy.js
+++ b/src/js/base/helpers.cy.js
@@ -11,6 +11,7 @@ context('Handlebars helpers', function() {
       template: hbs`
         <div class="test-null">{{matchText "Patient Name" null}}</div>
         <div class="test-match">{{matchText "Patient Name" "Patient"}}</div>
+        <div class="test-noescape">{{matchText "<span style='color: green'>Patient</span> Name" "Patient" noEscape=true}}</div>
       `,
     });
 
@@ -29,6 +30,11 @@ context('Handlebars helpers', function() {
       .get('@root')
       .find('.test-match strong')
       .should('contain', 'Patient');
+
+    cy
+      .get('@root')
+      .find('.test-noescape strong')
+      .should('have.css', 'color', 'rgb(0, 128, 0)');
   });
 
   specify('Date time formatting', function() {
--- a/src/js/base/helpers.cy.js
+++ b/src/js/base/helpers.cy.js
@@ -62,9 +62,7 @@ context('Handlebars helpers', function() {
       .get('@root')
       .find('.test-element')
       .first()
-      .should(el => {
-        expect(el.text().trim()).equal('');
-      });
+      .should('be.empty');
 
     cy
       .get('@root')
@@ -112,9 +110,7 @@ context('Handlebars helpers', function() {
       .get('@root')
       .find('.test-element')
       .first()
-      .should(el => {
-        expect(el.text().trim()).equal('');
-      });
+      .should('be.empty');
 
     cy
       .get('@root')
@@ -126,9 +122,7 @@ context('Handlebars helpers', function() {
       .get('@root')
       .find('.test-element')
       .eq(2)
-      .should(el => {
-        expect(el.text().trim()).equal('');
-      });
+      .should('be.empty');
 
     cy
       .get('@root')
--- a/src/js/base/helpers.cy.js
+++ b/src/js/base/helpers.cy.js
@@ -1,4 +1,3 @@
-import Backbone from 'backbone';
 import { View } from 'marionette';
 
 import hbs from 'handlebars-inline-precompile';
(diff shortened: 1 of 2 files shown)


nmajor25:
Do we use this portion of the `matchText` helper anywhere: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/base/helpers.js#L13?

I couldn't find any places where it's used. Can it be removed? I had trouble figuring out how to test it.

src/js/base/helpers.cy.js
```diff
@@ -0,0 +1,173 @@
...
+        phone: '6155555551',
+        badPhone: 'UNKNOWN',
+      };
+    },
+  });
+
+
+  specify('Match text formatting', function() {
+    const matchTextView = new MatchTextView({
+      model: new Backbone.Model(),
```
paulfalgout: you don't need to pass it a model.. these can all just be `new MatchTextView` or the mount could just `return new FooView();`

You could probably move the View creation here since it's 1-to-1 with the specify.  so here would be `const FooView = View.extend` and then new it in the mount.

src/js/base/helpers.cy.js
```diff
@@ -0,0 +1,173 @@
...
+
+    cy
+      .mount(rootView => {
+        return matchTextView;
+      })
+      .as('root');
+
+    cy
+      .get('@root')
+      .find('.test-element')
```
paulfalgout: I think we might want to name these instead of filtering them so that in the future tests can be added without modifying other tests by having to change `last` or updated `eq(#)` below where I inserted one.  So explicitly this'd be `.find('.test-null')` or implicitly it's just next.. should... next... should...

src/js/base/helpers.cy.js
```diff
@@ -0,0 +1,173 @@
...
+        return dateTimeView;
+      })
+      .as('root');
+
+    cy
+      .get('@root')
+      .find('.test-element')
+      .first()
+      .should(el => {
+        expect(el.text().trim()).equal('');
```
paulfalgout: is this trim needed because the test is
```
<div>
  {{ value }}
</div>
```
and not `<div>{{ value }}</div>`?
`should('be.empty')` doesn't account for spaces I guess?

paulfalgout (changes requested):
This is a nice change.  I like component tests

paulfalgout:
> Do we use this portion of the `matchText` helper anywhere: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/base/helpers.js#L13?
> 
> I couldn't find any places where it's used. Can it be removed? I had trouble figuring out how to test it.

I think it's worth testing even though we aren't using it.  There might be a case where we want like style html in the text or something and we'd need it:
https://handlebarsjs.com/api-reference/utilities.html#handlebars-escapeexpression-string

test/integration/patients/patient/sidebar.js
```diff
@@ -48,11 +42,7 @@ context('patient sidebar', function() {
               'patientMRNIdentifier',
               'patientSSNIdentifier',
               'hbsWidget',
-              'phoneWidget1',
-              'phoneWidget2',
-              'phoneWidget3',
-              'dateTimeWidget1',
-              'dateTimeWidget2',
+              'hbsEmptyWidget',
```
paulfalgout, line 45: We could probably add a test here to fix that coverage issue.. I think it's where the field value is `{}` instead of `null`, so we could add an `emptyObjectField` test and I think that'd get us 💯

src/js/base/helpers.cy.js
nmajor25: Agreed, updated 👍

src/js/base/helpers.cy.js
nmajor25: Much cleaner, updated 👍

src/js/base/helpers.cy.js
nmajor25: Yep, updated and it now does `.should('be.empty')`.

src/js/base/helpers.cy.js
```diff
@@ -0,0 +1,127 @@
...
+import { testTs } from 'helpers/test-timestamp';
+import formatDate from 'helpers/format-date';
+
+context('Handlebars helpers', function() {
+  specify('Match text formatting', function() {
+    const MatchTextView = View.extend({
+      template: hbs`
+        <div class="test-null">{{matchText "Patient Name" null}}</div>
+        <div class="test-match">{{matchText "Patient Name" "Patient"}}</div>
+        <div class="test-noescape">{{matchText "<span style='color: green'>Patient</span> Name" "Patient" noEscape=true}}</div>
```
nmajor25, line 14: Is this how `noEscape=true` would potentially be used? I don't fully understand the use case for it.

test/integration/patients/patient/sidebar.js
nmajor25: I'll do this in a quick separate PR.

src/js/base/helpers.cy.js
paulfalgout: Yeah so I think perhaps a more relevant example might be something like `Patient Name <span class="role">Role</span>` we've done something like this in the past, and without noEscape the "<span>" html would end up as part of the text.

paulfalgout (approved):

test/integration/patients/patient/sidebar.js
nmajor25: This coverage drop was for the patient fields service. Which the widgets doesn't use anymore. It uses only the widget values service.

Now that `patientField.getValue()` is only used in [worklist sorting app](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/patients/worklist/worklist_sort.js#L9-L13) and [forms](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/services/forms.js#L132-L160) when fetching or updating a patient field.

So we'll need to test this in the form tests somehow I think.

At least that's my understanding of this.

PR #1295, opened 2024-08-14
Refactor `/patients/patient/sidebar.js` tests for json api

nmajor25:
Shortcut Story ID: [sc-49489]

## Summary by CodeRabbit

- **New Features**
  - Enhanced integration tests for the patient sidebar, improving the clarity and maintainability of patient data handling.
  - Introduced a standardized `testPatient` object for consistent patient attributes in test cases.

- **Improvements**
  - Updated the `getWorkspacePatient` function to support merging additional patient data, enhancing its flexibility and usability.
  - Modified API response data handling to reduce redundancy and streamline test setup.

- **Bug Fixes**
  - Corrected visit URLs for patient dashboard routes to ensure tests reference the correct patient context.

### DIFF
--- a/test/integration/patients/patient/sidebar.js
+++ b/test/integration/patients/patient/sidebar.js
@@ -3,13 +3,32 @@ import dayjs from 'dayjs';
 
 import formatDate from 'helpers/format-date';
 import { testDate, testDateSubtract } from 'helpers/test-date';
-import { getResource } from 'helpers/json-api';
+import { getResource, getRelationship, mergeJsonApi } from 'helpers/json-api';
 
 import { workspaceOne } from 'support/api/workspaces';
+import { getWorkspacePatient } from 'support/api/workspace-patients';
+import { getPatient } from 'support/api/patients';
+import { getCurrentClinician } from 'support/api/clinicians';
+import { roleAdmin, roleEmployee } from 'support/api/roles';
 
 context('patient sidebar', function() {
   specify('display patient data', function() {
     const dob = testDateSubtract(10, 'years');
+
+    const testPatient = getPatient({
+      attributes: {
+        first_name: 'First',
+        last_name: 'Last',
+        birth_date: dob,
+        sex: 'f',
+        identifiers: [
+          {
+            type: 'mrn',
+            value: 'A5432112345',
+          },
+        ],
+      },
+    });
 
     cy
       .routesForPatientDashboard()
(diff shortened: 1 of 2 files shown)


test/support/api/workspace-patients.js
```diff
@@ -1,12 +1,16 @@
...
 const TYPE = 'workspace-patients';
 
-export function getWorkspacePatient() {
-  return getResource(_.sample(fxWorkspacePatients), TYPE);
+export function getWorkspacePatient(data) {
+  const resource = getResource(_.sample(fxWorkspacePatients), TYPE);
+
+  if (!data) return resource;
+
+  return mergeJsonApi(resource, data);
```
nmajor25, line 11: More I need to add to this? I feel like I'm missing something. For example, some of these have a `VALID: ` portion added to the `mergeJsonApi()` parameters.

paulfalgout: that's for setting the possible valid keys of a relationship.  turning on debugging will show anywhere where a relationship is incorrectly used in the tests if the `VALID`s are all sett

paulfalgout: that said.. there don't look to be relationships here

test/support/api/workspace-patients.js
```diff
@@ -1,12 +1,16 @@
...
 import fxWorkspacePatients from 'fixtures/collections/workspace-patients';
 
 const TYPE = 'workspace-patients';
 
-export function getWorkspacePatient() {
-  return getResource(_.sample(fxWorkspacePatients), TYPE);
+export function getWorkspacePatient(data) {
+  const resource = getResource(_.sample(fxWorkspacePatients), TYPE);
+
+  if (!data) return resource;
```
paulfalgout: I don't think we want this guard here.  `mergeJSON` also clones the object so it can't be mutated for future tests

test/integration/patients/patient/sidebar.js
```diff
@@ -493,10 +518,18 @@ context('patient sidebar', function() {
 
     cy
       .url()
-      .should('contain', '/patient/dashboard/1');
+      .should('contain', `/patient/dashboard/${ testPatient.id }`);
   });
 
   specify('view patient modal', function() {
+    const testPatient = getPatient({
```
paulfalgout, line 521: I think this might be just `const testPatient = getPatient()`.   We should attempt to remove attributes from this test and those above that are not required for the tests.  We were previously over-setting because we didn't have mergejson to cover all the defaults

paulfalgout (changes requested):

test/integration/patients/patient/sidebar.js
nmajor25: I was able to remove the `sex` attribute, but the first/last name and dob attributes were needed for the test.

test/support/api/workspace-patients.js
nmajor25: Removed 👍

test/integration/patients/patient/sidebar.js
paulfalgout: Same with line 442?

nmajor25: Ah missed that one, looks like that one only needs to the `source: 'manual'` attribute. I just removed all the other attributes besides that one.

paulfalgout (approved):

PR #1308, opened 2024-08-29
Clinician tests refactoring/clean up

nmajor25:
## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced integration tests for the clinicians modal by replacing hardcoded IDs with dynamic references to current role and team data.
	- Updated clinician sidebar tests to use dynamically generated UUIDs for clinician IDs, improving test reliability.
	- Modified clinicians list tests to utilize UUIDs for clinician IDs, ensuring uniqueness and reducing conflicts.

- **Chores**
	- Removed unnecessary attributes from test data to simplify the testing framework.

### DIFF
--- a/test/integration/clinicians/clinician-sidebar.js
+++ b/test/integration/clinicians/clinician-sidebar.js
@@ -1,24 +1,22 @@
+import { v4 as uuid } from 'uuid';
+
 import { getErrors, getRelationship } from 'helpers/json-api';
 
-import { testTs } from 'helpers/test-timestamp';
 import stateColors from 'helpers/state-colors';
 
 import { teamCoordinator, teamNurse } from 'support/api/teams';
 import { roleAdmin, roleEmployee, roleManager } from 'support/api/roles';
-import { getWorkspaces, workspaceOne } from 'support/api/workspaces';
+import { workspaceOne } from 'support/api/workspaces';
 import { getClinician, getCurrentClinician } from 'support/api/clinicians';
 
 const testClinician = getClinician({
-  id: '1',
+  id: uuid(),
   attributes: {
     name: 'Test Clinician',
     email: 'test.clinician@roundingwell.com',
-    last_active_at: testTs(),
-    enabled: true,
   },
   relationships: {
     team: getRelationship(teamCoordinator),
-    workspaces: getRelationship(getWorkspaces()),
     role: getRelationship(roleEmployee),
   },
 });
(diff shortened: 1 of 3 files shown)


test/integration/clinicians/clinician-sidebar.js
```diff
@@ -1,24 +1,22 @@
...
 
 import { teamCoordinator, teamNurse } from 'support/api/teams';
 import { roleAdmin, roleEmployee, roleManager } from 'support/api/roles';
-import { getWorkspaces, workspaceOne } from 'support/api/workspaces';
+import { workspaceOne } from 'support/api/workspaces';
 import { getClinician, getCurrentClinician } from 'support/api/clinicians';
 
 const testClinician = getClinician({
-  id: '1',
+  id: uuid(),
```
nmajor25, line 13: I can't remember why it's required to include an `id` in `getClinician()`... is it so we don't interfere with the current clinician?

test/integration/clinicians/clinician-sidebar.js
```diff
@@ -1,24 +1,22 @@
...
   attributes: {
     name: 'Test Clinician',
     email: 'test.clinician@roundingwell.com',
-    last_active_at: testTs(),
-    enabled: true,
```
nmajor25, line 17: `enabled: true` is set for each clinician fixture, so we don't need to set it in `getClinician()`.

test/integration/clinicians/clinician-sidebar.js
paulfalgout: well the trick with clinicians is that they're assigned to global things.. so if you change the uuid here for just this instance then it won't necessarily be available in the owner component droplist..  so we need a "there's X number of test clinicians that work everywhere and when I modify it, it's going to be modified everywhere". I think the clinician part of the app is the only part where we can use generated `uuid`s without something else breaking elsewhere.

nmajor25: Gotcha. So should we keep this as `id: '1'` for now? Or maybe switch it to `22222` or something until we update it later?

paulfalgout:
Can you rebase this off of develop?

nmajor25:
@paulfalgout yep, rebased 👍

nmajor25:
@paulfalgout what is there still to do for this one?

paulfalgout:
> @paulfalgout what is there still to do for this one?

well I suppose we can get this in, but we should go ahead and work to remove the need to send an id to `getClinician` using the same patterns as the workspace/program work of pulling the first n fixtures needed for global entity testing.

paulfalgout (approved):

nmajor25:
@paulfalgout ah ya, okay I'll create a shortcut story for that

PR #1310, opened 2024-08-29
Move workspace and program related fixtures into generated but test-consistent fixtures

paulfalgout:
So looking at the patterns in your two PRs.. I think we can remove the need for generating uuids.  Not sure if this'll make sense, but I think it will affect clinicians.. essentially we need to pull specific generic fixtures for modification and then use the rest for sampling.

The main issue is that the route functions that are preloaded are somehow a separate runtime process from the tests themselves so local variables might be different between the two.  The only way to ensure exported test fixtures have the same ids is to hardcode them ie: `11111` or to pull them from the same source ie: the first n of the generated fixtures.

## Summary by CodeRabbit

- **New Features**
	- Introduced mock data generation for programs and workspaces, enhancing testing capabilities with realistic data structures.
	- Streamlined workspace data handling in integration tests, improving clarity and performance.

- **Bug Fixes**
	- Adjusted unique identifier generation for program entries to ensure consistency in test data.

- **Refactor**
	- Simplified functions for retrieving program and workspace data, reducing reliance on merging utilities.
	- Enhanced clarity in workspace handling by explicitly defining test workspaces instead of relying on random sampling.

- **Chores**
	- Updated import paths for fixture data to reflect organizational changes in the project structure.


paulfalgout:
still more thread to pull.. one thing unravels another, but I think this is the correct direction

test/support/api/programs.js
```diff
@@ -1,37 +1,55 @@
...
 
 import { getProgramFlows } from './program-flows';
 import { getProgramActions } from './program-actions';
-import { getWorkspaces } from './workspaces';
+import { workspaceOne, workspaceTwo } from './workspaces';
 
 const TYPE = 'programs';
 
-let programs;
+export const testPrograms = [];
```
nmajor25: Do we need to export this? Doesn't look like we're using them anywhere else besides this file.

nmajor25: Same comment for `testWorkspaces` in the other `/support/api/workspaces.js` file.

test/support/api/programs.js
```diff
@@ -1,37 +1,55 @@
...
+// Exporting only programs needed for testing variance
+export const programOne = getResource(_.extend(fxPrograms[0], {
+  name: 'Program One',
+}), TYPE);
+
+export const programTwo = getResource(_.extend(fxPrograms[1], {
+  name: 'Program Two',
+}), TYPE);
+
+testPrograms.push(programOne, programTwo);
```
nmajor25: So this is where the separate runtime issue comes into play? That's why we have to push to the empty `testPrograms` array instead of just doing `const testPrograms = [programOne, programTwo];` here?

test/support/api/programs.js
paulfalgout: good catch.. I don't think so.. this was part of an earlier though, but didn't end up using it... better without

test/support/api/programs.js
paulfalgout: nope the separate run time is.. if I add `uuid()` on the programOne.. then in the fixtures it would be one uuid but in the tests if I import it there it would be another.. currently we will get two instances, but that's ok because the data ends up identical.

nmajor25 (approved):

PR #1312, opened 2024-09-05
Refactor forms tests

nmajor25:
Changes to match the patterns in the recent refactoring we've done on other tests.

Use the generated `uuid` ids where we can. And reference fixture id's directly when possible (example: `'11111'` => `testForm.id`).

## Summary by CodeRabbit

- **New Features**
  - Introduced a new 'C Form' in the action sidebar for enhanced testing scenarios.
  - Updated test forms for improved variety and relevance in testing.

- **Refactor**
  - Streamlined form data management in the testing framework, focusing on a specific subset of forms.
  - Simplified retrieval of test forms, enhancing clarity and reducing redundancy in command definitions.


test/integration/forms/action.js
```diff
@@ -728,9 +757,9 @@ context('Patient Action Form', function() {
   });
 
   specify('submitting the form', function() {
-    const formResponses = [
+    const testFormResponses = [
       getFormResponse({
-        id: '1',
+        id: uuid(),
```
nmajor25: Every `getFormResponse()` returns a form response with `id: '11111'`. Which was causing issues in this test.

test/integration/forms/action.js
nmajor25, line 1: If these patterns work, I can apply them to the other forms tests as well.

test/support/api/forms.js
```diff
@@ -20,6 +20,15 @@ export function getForms() {
 
 // Exporting only form needed for testing variance
 export const testForm = getResource(_.find(fxTestForms, { id: '11111' }));
+export const testReadOnlyForm = getResource(_.find(fxTestForms, { id: '22222' }));
```
paulfalgout: I actually think we should probably remove 100% of these.

We should generate generic forms.. in collections instead of using tests..  probably with mostly no options.  And the test specifics should largely move to the tests.

Although maybe it's easier to have a single test form export?  If that's the case we should essentially do this:

```
const fxTestForms = [{ name: 'Test Form' }];
...
export const testForm = getForm({ name: 'Test Form' });
```

But we should remove all of the `formId = ` from the `routeFormX` stuff.  This maybe where having a default `testForm` or possibly two is useful..  But most of these exports are single use

paulfalgout (changes requested):

test/integration/forms/action.js
```diff
@@ -2014,25 +2188,36 @@ context('Patient Action Form', function() {
...
         return fx;
       })
-      .routeFormByAction(_.identity, '88888')
-      .routeFormResponse()
       .routeFormDefinition()
       .routeFormActionFields()
       .routeActionActivity()
       .routePatientByAction()
-      .visit('/patient-action/1/form/88888')
+      .visit(`/patient-action/${ testAction.id }/form/${ testSubmitHiddenForm }`)
```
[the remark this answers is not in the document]
nmajor25: Good bot

test/integration/forms/patient.js
```diff
@@ -803,16 +865,20 @@ context('Patient Form', function() {
 
   specify('form error', function() {
     cy
-      .routeForm(_.identity, '11111')
+      .routeForm(fx => {
+        fx.data = testForm;
+
+        return testForm;
+      })
```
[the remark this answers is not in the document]
nmajor25: Good bot

test/integration/patients/patient/sidebar.js
```diff
@@ -451,8 +469,8 @@ context('patient sidebar', function() {
           id: 'patient_creation_form',
           attributes: {
             value: {
-              form_id: '11111',
-              submit_text: 'Continue to Form 11111',
+              form_id: testForm.id,
+              submit_text: `Continue to Form ${ testForm.id }`,
```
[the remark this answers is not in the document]
nmajor25: Good bot

test/integration/patients/patient/sidebar.js
```diff
@@ -533,8 +551,8 @@ context('patient sidebar', function() {
...
```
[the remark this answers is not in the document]
nmajor25: Good bot

test/fixtures/config/forms.js
```diff
@@ -0,0 +1,9 @@
+import faker from '@roundingwellos/faker';
+
+export default () => {
+  return {
+    id: faker.datatype.uuid(),
+    name: `${ faker.company.bsBuzz() } ${ faker.company.catchPhraseNoun() }`,
```
paulfalgout, line 7: I think we need to add the `created_at` `updated_at` `submitted_at` timestamps here

test/integration/forms/action.js
```diff
@@ -39,9 +60,9 @@ context('Patient Action Form', function() {
       })
       .as('routeActionError')
       .routePatientByAction()
-      .routeFormByAction(_.identity, '11111')
+      .routeFormByAction()
       .routeLatestFormResponse()
-      .visit('/patient-action/1/form/11111')
+      .visit(`/patient-action/1/form/${ testForm.id }`)
```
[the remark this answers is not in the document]
paulfalgout: I think maybe we should make these ids something more obvious.. like `deletedActionId` which we could use in the intercept route.. and is just a `deletedActionId = uuid()`

test/integration/forms/action.js
```diff
@@ -251,19 +289,29 @@ context('Patient Action Form', function() {
...
-            'form': getRelationship('11111', 'forms'),
-            'form-responses': getRelationship([formResponse]),
-          },
-        });
+        fx.data = testAction;
 
         return fx;
       })
-      .routeFormByAction(_.identity, '11111')
+      .routeFormByAction(_.identity, testForm.id)
```
paulfalgout: I kinda think we could remove the 2nd argument.. it's a strange pattern now.. also we can probably default `routeFormByAction` to return the `testForm` and make everything else the exception

test/integration/outreach/outreach.js
```diff
@@ -598,7 +605,18 @@ context('Outreach', function() {
   specify('Read-only Form', function() {
     cy
       .routeOutreachStatus()
-      .routeFormByAction(_.identity, '22222')
+      .routeFormByAction(fx => {
+        fx.data = getForm({
+          attributes: {
+            name: 'Read Only Test Form',
```
paulfalgout, line 611: may not need to name it?

test/integration/patients/patient/patient-flow.js
```diff
@@ -1282,9 +1282,7 @@ context('patient flow page', function() {
       .routeFlowActions(fx => {
         fx.data = testFlowActions;
 
-        fx.included.push(
-          getForm({ id: '11111', attributes: { name: 'Test Form' } }),
-        );
+        fx.included.push(testForm);
```
paulfalgout: I think this could be removed.. we don't include forms on this endpoint

test/integration/forms/action.js
paulfalgout: actually it looks like you did remove it, so this isn't doing anything.. no need to pass the `_.identity`s

test/support/api/forms.js
```diff
@@ -1,62 +1,54 @@
 import _ from 'underscore';
-import { v4 as uuid } from 'uuid';
 import { getResource, mergeJsonApi } from 'helpers/json-api';
 
-import fxTestForms from 'fixtures/test/forms';
+import fxTestForms from 'fixtures/collections/forms';
+
+const fxSampleForms = _.rest(fxTestForms, 2);
```
paulfalgout: since we're only using `fxTextForms[0]` this should be `_.rest(..., 1)`

test/support/api/forms.js
```diff
@@ -1,62 +1,54 @@
...
-    form = _.clone(form);
-    form.options = {};
-    return form;
-  });
-
   cy
     .intercept('GET', '/api/forms', {
       body: mutator({
-        data: getResource(fxForms, TYPE),
+        data: getResource(fxTestForms, TYPE),
```
paulfalgout: I think this should be `data: [testForm, ...getResource(fxSampleForms, TYPE)],`

test/support/api/forms.js
```diff
@@ -1,62 +1,54 @@
...
+
+const fxSampleForms = _.rest(fxTestForms, 2);
 
 const TYPE = 'forms';
 
 export function getForm(data) {
-  const resource = getResource(_.sample(fxTestForms), TYPE);
-
-  data = _.extend({ id: uuid() }, data);
+  const resource = getResource(_.sample(fxSampleForms), TYPE);
```
paulfalgout: we don't really use a random form.. hardly ever.. I think this could be:
```
export function getForm(data) {
  return mergeJsonApi(testForm, data);
}
```

test/support/api/forms.js
paulfalgout: otherwise this the forms url will contain the `testForm.id` with different attributes than loaded elsewhere.

paulfalgout (changes requested):
looking good

test/integration/formservice/formservice.js
```diff
@@ -207,7 +207,7 @@ context('Formservice', function() {
       .should('contain', 'filter[action.tags]=foo-tag')
       .should('contain', `filter[flow]=${ testFlow.id }`)
       .should('contain', `filter[patient]=${ testPatient.id }`)
-      .should('contain', `filter[created]=<=${ createdAt }`);
+      .should('contain', `filter[submitted]=<=${ createdAt }`);
```
nmajor25, line 210: I messed up the merge conflict fix. But fixed it here.

test/integration/forms/action.js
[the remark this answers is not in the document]
nmajor25: Updated 👍

test/fixtures/config/forms.js
nmajor25: Added 👍

test/integration/forms/action.js
nmajor25: Ya I missed the removal of a couple of these. They're gone now 👋

test/integration/outreach/outreach.js
nmajor25: The name is tested in the form. I'm not sure why I added that originally. Doesn't seem super important.

test/integration/patients/patient/patient-flow.js
nmajor25: Removed 👋

test/support/api/forms.js
nmajor25: Yep, updated

test/support/api/forms.js
nmajor25: Yep, good call, updated 👍

test/support/api/forms.js
nmajor25: Updated 👍

test/support/api/forms.js
```diff
@@ -7,21 +7,19 @@ const fxSampleForms = _.rest(fxTestForms, 1);
...
-export function getForm(data) {
-  const resource = getResource(_.sample(fxSampleForms), TYPE);
+// Exporting only form needed for testing variance
+export const testForm = getResource(_.extend(fxTestForms[0], {
+  name: 'Test Form',
+}), TYPE);
 
-  return mergeJsonApi(resource, data);
+export function getForm(data) {
+  return mergeJsonApi(testForm, data);
```
nmajor25, line 16: This breaks this test: https://cloud.cypress.io/projects/ep9zr6/runs/6773/test-results?utm_source=github&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D

In that test an array of forms is created like `[testForm, getForm(), getForm()]`. And each form then has the same `id`. Which means only one shows up in the form dropdown.

test/support/api/forms.js
```diff
@@ -1,62 +1,52 @@
...
 }
 
 export function getForms() {
-  return getResource(fxTestForms, TYPE);
+  return getResource(fxSampleForms, TYPE);
```
paulfalgout: I think this needs to be `[testForm, ...getResource(fxSampleForms, TYPE)],` and `routeForms` should use it.

test/integration/programs/sidebar/action-sidebar.js
```diff
@@ -337,30 +337,35 @@ context('program action sidebar', function() {
...
       getForm(testForm),
       getForm({
+        id: uuid(),
         attributes: {
           name: 'C Form',
           published_at: null,
         },
       }),
       getForm({
+        id: uuid(),
```
paulfalgout, line 347: D&E can be removed entirely.. they aren't adding anything functionally as far as I can tell

paulfalgout (changes requested):

test/integration/programs/sidebar/action-sidebar.js
nmajor25: Nice, removed those

test/support/api/forms.js
nmajor25: Updated 👍

paulfalgout (approved):

PR #1313, opened 2024-09-10
Update spacing for worklist list items

nmajor25:
Shortcut Story ID: [sc-54584]

Screenshots of worklist after these updates were made:

Screenshot 2024-09-10 at 12 13 09 PM

Screenshot 2024-09-10 at 3 20 54 PM

## Summary by CodeRabbit

- **New Features**
	- Enhanced table layouts in the worklist for better visibility of timestamps and improved semantic structure.

- **Bug Fixes**
	- Adjusted padding and widths in various table components to improve visual presentation and alignment.

- **Style**
	- Modified CSS styles for table headers and cells to enhance readability and overall aesthetic appeal.

### DIFF
--- a/src/js/views/patients/worklist/worklist-list.scss
+++ b/src/js/views/patients/worklist/worklist-list.scss
@@ -73,7 +73,7 @@
 
 .worklist-list__item-name,
 .worklist-list__name-header {
-  width: 27%;
+  width: 22%;
 }
 
 .worklist-list__meta {
@@ -82,7 +82,7 @@
 }
 
 .worklist-list__meta-header {
-  padding-left: 48px;
+  padding-left: 40px;
   width: 43%;
 }
 
--- a/src/js/views/patients/worklist/worklist-list.scss
+++ b/src/js/views/patients/worklist/worklist-list.scss
@@ -81,8 +81,13 @@
   width: 43%;
 }
 
-.worklist-list__meta-header {
+.worklist-list__action-meta-header {
   padding-left: 40px;
+  width: 43%;
+}
+
+.worklist-list__flow-meta-header {
+  padding-left: 20px;
   width: 43%;
 }
 
--- a/src/scss/core/_layout.scss
+++ b/src/scss/core/_layout.scss
@@ -11,7 +11,6 @@
 // -------------------------------------
 //   Fluid Width classes
 // -------------------------------------
-.w-7-5 { width: 7.5% !important; }
 .w-10 { width: 10% !important; }
 .w-15 { width: 15% !important; }
 .w-20 { width: 20% !important; }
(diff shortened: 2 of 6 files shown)


src/scss/modules/table-list.scss
```diff
@@ -39,7 +39,7 @@
 
 .table-list__cell {
   overflow: hidden;
-  padding: 16px;
+  padding: 16px 8px;
```
nmajor25, line 42: I went with `8px` instead of `4px` here for the left/right padding. `8px` seems like enough to accommodate the updated/added date with the full year.

src/js/views/patients/worklist/worklist-list.scss
```diff
@@ -82,7 +82,7 @@
 }
 
 .worklist-list__meta-header {
-  padding-left: 48px;
+  padding-left: 40px;
```
nmajor25, line 85: For the flow version of the worklist, the alignment with the header and row items are off. In this PR and currently in production as well.

It seems to be because of the details element on the action worklist version. The flow version doesn't have that element, but uses this same classname and padding for the header.

If we use a different classname/padding on the flow version with ~16px less padding, the alignment should look better.

So I'd like to look into this more.

Screenshot 2024-09-10 at 12 36 29 PM

src/scss/core/_layout.scss
```diff
@@ -12,6 +12,7 @@
 //   Fluid Width classes
 // -------------------------------------
 .w-7-5 { width: 7.5% !important; }
+.w-9-5 { width: 9.5% !important; }
```
nmajor25: Note: `.w-7-5` is now no longer used anywhere.

src/js/views/patients/worklist/worklist-list.scss
nmajor25: On the flow version, using `16px` less padding for the header element would look like this:

Screenshot 2024-09-10 at 12 21 10 PM

src/scss/core/_layout.scss
paulfalgout: now that I'm seeing this.. maybe remove these two oddballs and just use w-10 and make the action name 1 less percentage

paulfalgout (changes requested):

src/js/views/patients/worklist/worklist-list.scss
nmajor25, line 1: Now looks like this:

Screenshot 2024-09-10 at 3 15 34 PM

nmajor25: Cleaner way to do the class names?

src/scss/core/_layout.scss
nmajor25: Ya that makes sense, updated to `w-10` and `23%` => `22%` width on the action/flow name column.

paulfalgout (approved):

PR #1316, opened 2024-09-16
Switch report forms to filter by submitted instead of created

paulfalgout:
Shortcut Story ID: [sc-54904]

@zfixler  I'm not entirely sure how to test this, but this is a pretty small change

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the handling of report forms to use the `submitted_at` timestamp instead of the `created_at` timestamp.
- **Tests**
  - Modified the integration tests for report forms to reflect the changes in the application's handling of the `submitted_at` timestamp.

### DIFF
--- a/src/js/formservice.js
+++ b/src/js/formservice.js
@@ -42,7 +42,7 @@ const ActionFormApp = App.extend({
 
     const isReport = form.isReport();
 
-    if (isReport) opts.created = `<=${ action.get('created_at') }`;
+    if (isReport) opts.submitted = `<=${ action.get('submitted_at') }`;
 
     const prefillActionTag = form.getPrefillActionTag();
 
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -210,7 +210,7 @@ export default App.extend({
 
     const isReport = form.isReport();
 
-    if (isReport) opts.created = `<=${ this.action.get('created_at') }`;
+    if (isReport) opts.submitted = `<=${ this.action.get('submitted_at') }`;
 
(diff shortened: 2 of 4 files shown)


paulfalgout:
@nmajor25 it's worth mentioning that our form test fixtures are lacking data like `created_at` and `updated_at` I think when we use form fixtures from config instead of test we should make sure all 3 timestamps are in there and match their status.

zfixler (approved):

nmajor25 (approved):

shadowhand:
hooray

src/js/services/forms.js
```diff
@@ -210,7 +210,7 @@ export default App.extend({
 
     const isReport = form.isReport();
 
-    if (isReport) opts.created = `<=${ this.action.get('created_at') }`;
+    if (isReport) opts.submitted = `<=${ this.action.get('submitted_at') }`;
```
zfixler, line 213: @nmajor25 , @paulfalgout , I think this is a bug. I think we still need to get the `created_at` timestamp here from the action, since this is the time _when the action was created_. I don't think `submitted_at` exists on the action, only on the form response.

src/js/formservice.js
```diff
@@ -42,7 +42,7 @@ const ActionFormApp = App.extend({
 
     const isReport = form.isReport();
 
-    if (isReport) opts.created = `<=${ action.get('created_at') }`;
+    if (isReport) opts.submitted = `<=${ action.get('submitted_at') }`;
```
zfixler, line 45: @nmajor25 , @paulfalgout , same thing here.

PR #1324, opened 2024-09-26
Refactor auth for consistency

paulfalgout:
Shortcut Story ID: [sc-55355]

I don't really know if the kinde refactor part works, but it doesn't really matter at the moment.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new authentication module for Kinde, enhancing user login and logout processes.
	- Added constants for standardized routing paths related to authentication.

- **Improvements**
	- Streamlined authentication logic for both Kinde and Auth0, improving maintainability and readability.
	- Enhanced handling of application state during authentication processes.
	- Improved management of authentication tokens and routing paths for better consistency.

- **Bug Fixes**
	- Resolved issues with hardcoded paths in the authentication flow, replacing them with defined constants.

### DIFF
--- a/src/js/auth.js
+++ b/src/js/auth.js
@@ -1,53 +1,35 @@
-import { extend } from 'underscore';
 import Radio from 'backbone.radio';
-import createKindeClient from '@kinde-oss/kinde-auth-pkce-js';
 
-import { kindeConfig as config, appConfig } from './config';
-
-import { auth as auth0, logout as auth0Logout, setToken as auth0SetToken, getToken as auth0GetToken } from './auth0';
+import * as auth0 from './auth/auth0';
+import * as none from './auth/none';
+import * as kinde from './auth/kinde';
 
 import 'scss/app-root.scss';
 
-import { LoginPromptView } from 'js/views/globals/prelogin/prelogin_views';
+let authAgent;
 
-const PATH_ROOT = '/';
-const PATH_RWELL = '/rw';
-const PATH_AUTHD = '/authenticated';
-const PATH_LOGIN = '/login';
-const PATH_LOGOUT = '/logout';
+function getAuthAgent() {
+  if (none.should()) return none;
 
-function shouldAuth0() {
-  return !config.createParams;
+  if (authAgent) return authAgent;
+
+  // These should be ordered by priority lowest to highest
+  if (auth0.should()) authAgent = auth0;
+  if (kinde.should()) authAgent = kinde;
+
+  return authAgent;
 }
 
-let kinde;
-let token;
-
-// Sets a token when not using auth0;
 function setToken(tokenString) {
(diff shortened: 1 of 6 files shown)


src/js/auth/kinde.js
```diff
@@ -0,0 +1,138 @@
...
+function getToken() {
+  if (token) return token;
+  if (!kinde || !navigator.onLine) return;
+
+  return kinde
+    .getToken()
+    .catch(() => {
+      logout();
+    });
+}
```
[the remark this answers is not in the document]
paulfalgout: this function is effectively a handler for a radio request which handles async or sync for us

src/js/auth/auth0.js
```diff
@@ -73,106 +74,96 @@
...
+      if (location.pathname === PATH_AUTHD) {
         authenticate(success);
         return;
       }
 
       if (!isAuthenticated) {
-        forceLogin(location.pathname);
+        login(location.pathname);
         return;
       }
```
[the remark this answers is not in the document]
paulfalgout: I can't think of any example of a pathname that is going to cause any issue.. the pathname is what it's redirected to after authentication..  that url is just read by the routers.. and only valid routes will match

src/js/auth/auth0.js
```diff
@@ -1,27 +1,27 @@
...
-
-import 'scss/app-root.scss';
+import { auth0Config as config, appConfig } from 'js/config';
 
 import { LoginPromptView } from 'js/views/globals/prelogin/prelogin_views';
 
-const RWELL_KEY = 'rw';
+import { PATH_ROOT, PATH_RWELL, PATH_AUTHD, PATH_LOGIN, PATH_LOGOUT } from './config';
+
 const RWELL_CONNECTION = 'google-oauth2';
```
nmajor25, line 11: Put in the `config.js` file?

src/js/auth.js
nmajor25, line 1: Is there a way to call the `.should()`'s only once for each auth type at the beginning of the file?

Something along the lines of this:

```js
let authAgent;

if (kinde.should()) authAgent = kinde;
if (auth0.should()) authAgent = auth0;
if (workos.should()) authAgent = workos;

// use authAgent throughout file:
// authAgent.setToken()
// authAgent.getToken()
// etc.
```

May be slightly cleaner possibly.

Would allow us to add some guards to some of those functions with `if (!authAgent) return ...`.

nmajor25 (approved):
Looks good 👍 , just a couple code organization thoughts as I reviewed.

src/js/auth/auth0.js
paulfalgout: It's auth0 specific

src/js/auth.js
paulfalgout: good suggestion.. check out the latest

nmajor25 (approved):

PR #1331, opened 2024-10-03
Test flow app websocket subscribe

paulfalgout:
Follow up test for #1330


paulfalgout:
The trick was to put the intercept directly after the last `wait` prior to the `onStart` where the `subscribe` occurs.

nmajor25 (approved):

PR #1334, opened 2024-10-04
Remove action from URL when action sidebar is closed 

nmajor25:
Shortcut Story ID: [sc-54866]

When a user is on a flow page with the action sidebar open (`/flow/:id/action/:id`). And then closes the sidebar via the `.js-close` icon in the UI, the URL should be changed to `/flow/:id` with the `/action/:id` portion removed.

We don't want this to happen when the action sidebar is closed because the user clicked on a different flow action in the list (this closes the current action sidebar and opens a new sidebar for the selected action) or navigated away from the patient flow page.

We only want this functionality on the patient flow page. The action sidebar is used in several different pages/apps. Those other versions should not have this functionality added.

## Summary by CodeRabbit

- **New Features**
	- Enhanced user activity tracking with improved navigation when closing the action sidebar.
	- Added functionality to the sidebar to communicate user activity events upon closure.

- **Bug Fixes**
	- Expanded integration tests for the patient flow page, improving action management and error handling scenarios.
	- Verified correct display and functionality of the flow action sidebar.

- **Tests**
	- Introduced new action definitions and expanded tests for bulk editing and socket notifications.
	- Added tests for multi-selection functionality and progress bar accuracy.

### DIFF
--- a/src/js/apps/patients/patient/flow/flow_app.js
+++ b/src/js/apps/patients/patient/flow/flow_app.js
@@ -16,6 +16,8 @@ import { LayoutView, ContextTrailView, HeaderView, ListView, SelectAllView } fro
 import { BulkEditButtonView, BulkEditActionsSuccessTemplate, BulkDeleteActionsSuccessTemplate } from 'js/views/patients/shared/bulk-edit/bulk-edit_views';
 import { AddButtonView, i18n } from 'js/views/patients/shared/add-workflow/add-workflow_views';
 
+const userActivityCh = Radio.channel('user-activity');
+
 export default SubRouterApp.extend({
   StateModel,
   routerAppName: 'FlowApp',
@@ -289,6 +291,10 @@ export default SubRouterApp.extend({
       'stop'() {
         this.setState('actionBeingEdited', null);
       },
+    });
+
+    this.listenTo(userActivityCh, 'close:actionSidebar', () => {
+      Radio.trigger('event-router', 'flow', this.flow.id);
     });
 
     this.startChildApp('action', { actionId });
--- a/src/js/apps/patients/patient/patient_app.js
+++ b/src/js/apps/patients/patient/patient_app.js
@@ -10,6 +10,8 @@ import PatientSidebarApp from 'js/apps/patients/patient/sidebar/sidebar_app';
 
 import { LayoutView } from 'js/views/patients/patient/patient_views';
 
+const userActivityCh = Radio.channel('user-activity');
+
 export default SubRouterApp.extend({
   eventRoutes() {
     return {
@@ -71,6 +73,15 @@ export default SubRouterApp.extend({
       'fail'() {
         this.startCurrent('dashboard');
       },
+    });
+
+    this.listenTo(userActivityCh, 'close:actionSidebar', () => {
+      if (this.getCurrent()._name === 'archive') {
+        Radio.trigger('event-router', 'patient:archive', this.patient.id);
+        return;
+      }
+
+      Radio.trigger('event-router', 'patient:dashboard', this.patient.id);
     });
 
     this.startChildApp('action', { actionId, patientId });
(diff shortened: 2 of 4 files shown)


src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -287,6 +287,7 @@ export default SubRouterApp.extend({
         action.trigger('editing', true);
       },
       'stop'() {
+        Radio.trigger('event-router', 'flow', this.flow.id);
```
nmajor25: This approach won't work. Think it's causing issues when navigating away from a `/flow/:id/action/:id` URL.

paulfalgout: even going from one action to another creates a long browser history

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -291,6 +293,10 @@ export default SubRouterApp.extend({
       },
     });
 
+    this.listenTo(userActivityCh, 'close:patient:action:sidebar', () => {
```
nmajor25: `'close:patient:action:sidebar'` seems like a long string message. But wanted to make it as specific as possible since this channel is global throughout the app.

src/js/views/patients/sidebar/action/action-sidebar_views.js
```diff
@@ -68,6 +69,9 @@ const LayoutView = View.extend({
   onAttach() {
     animSidebar(this.el);
   },
+  onClose() {
+    Radio.trigger('user-activity', 'close:patient:action:sidebar');
```
nmajor25: Doing a `this.triggerMethod()` doesn't work here. The flow app doesn't have access to it.

src/js/apps/patients/patient/flow/flow_app.js
paulfalgout: I think we can do `close:patientActionSidebar`. or maybe `close:actionSidebar` since this for instance is a flow action

paulfalgout (changes requested):
We need to do this same thing on the patient dashboard / archive

src/js/apps/patients/patient/flow/flow_app.js
nmajor25: Okay updated to `close:actionSidebar`

src/js/apps/patients/patient/patient_app.js
nmajor25, line 1: What this functionality looks like:

https://github.com/user-attachments/assets/de680845-d1c6-4b55-aaee-f0ced75a4f77

nmajor25: I don't like how the entire app is reloaded when the sidebar is closed. But maybe it'll look better in production. Since this screen recording is from local dev, which is abnormally slow.

src/js/apps/patients/patient/patient_app.js
```diff
@@ -73,6 +75,15 @@ export default SubRouterApp.extend({
       },
     });
 
+    this.listenTo(userActivityCh, 'close:actionSidebar', () => {
+      if (this.getCurrent()._name === 'archive') {
```
nmajor25, line 79: This app is shared by both the patient dashboard and archive. This is the only way I could think of to differentiate when the app is being used for the dashboard or archive.

src/js/apps/patients/patient/patient_app.js
nmajor25: If we stick with doing this, I still need to add tests.

src/js/apps/patients/patient/patient_app.js
paulfalgout: yeah.. hmm.. not a fan

src/js/apps/patients/patient/patient_app.js
paulfalgout: oh yeah.. don't love this either.. is this how the flow page works?

paulfalgout (changes requested):
I'm starting to think we are looking at this wrong and that this maybe is something we should handle with the SubRouterApp.. something where we make a request on the event-router for the current parent or something like that.  But as we apply this pattern elsewhere it's getting messier.

src/js/apps/patients/patient/patient_app.js
nmajor25: No, the flow page doesn't reload the app like it does here.

PR #1346, opened 2024-10-31
Handle `DetailsChanged` websocket notifications on flow page

nmajor25:
Shortcut Story ID: [sc-56329]

In the action sidebar, the details text area content will NOT be updated if the user is currently making edits.

### DIFF
--- a/src/js/views/patients/sidebar/action/action-sidebar-action_views.js
+++ b/src/js/views/patients/sidebar/action/action-sidebar-action_views.js
@@ -138,6 +138,8 @@ const ActionView = View.extend({
     'change:duration': 'onChangeDuration',
   },
   onChangeDetails() {
+    if (this.isEditingDetails) return;
+
     this.showEditForm();
   },
   onChangeActionState() {
@@ -177,22 +179,30 @@ const ActionView = View.extend({
     this.showDueTime();
     this.showDuration();
   },
-  showSave() {
-    this.showChildView('save', new SaveView({ model: this.clonedAction }));
-  },
   onSave() {
-    this.getRegion('save').empty();
+    this.stopEditingDetails();
   },
   onCancel() {
     this.showEditForm();
   },
   showEditForm() {
     this.cloneAction();
-    this.listenTo(this.clonedAction, 'change:details', this.showSave);
 
-    this.getRegion('save').empty();
+    this.listenTo(this.clonedAction, 'change:details', this.startEditingDetails);
+
+    this.stopEditingDetails();
 
     this.showDetails();
+  },
+  startEditingDetails() {
+    this.isEditingDetails = true;
+
+    this.showChildView('save', new SaveView({ model: this.clonedAction }));
+  },
+  stopEditingDetails() {
+    this.isEditingDetails = false;
+
+    this.getRegion('save').empty();
   },
   showDetails() {
     this.showChildView('details', new DetailsView({ model: this.clonedAction }));
--- a/src/js/views/patients/sidebar/action/action-sidebar-action_views.js
+++ b/src/js/views/patients/sidebar/action/action-sidebar-action_views.js
@@ -130,11 +130,15 @@ const ActionView = View.extend({
     duration: '[data-duration-region]',
   },
   modelEvents: {
+    'change:details': 'onChangeDetails',
     'change:_state': 'onChangeActionState',
     'change:_owner': 'onChangeOwner',
     'change:due_date': 'onChangeDue',
     'change:due_time': 'onChangeDue',
     'change:duration': 'onChangeDuration',
+  },
+  onChangeDetails() {
+    this.showEditForm();
   },
   onChangeActionState() {
     const isDone = this.model.isDone();
--- a/test/integration/patients/patient/patient-flow.js
+++ b/test/integration/patients/patient/patient-flow.js
@@ -143,12 +143,31 @@ context('patient flow page', function() {
       .wait('@routeFlow')
       .wait('@routePatientByFlow')
       .wait('@routeFlowActions')
-      .wait('@routeAction');
+      .wait('@routeAction')
+      .wait('@routeActionActivity')
+      .wait('@routeActionComments')
+      .wait('@routeActionFiles');
 
     cy
       .get('.sidebar')
       .find('[data-action-region] .action-sidebar__name')
       .should('contain', 'Test Action');
+
+    cy.sendWs({
+      category: 'DetailsChanged',
+      resource: {
+        type: 'patient-actions',
+        id: testFlowAction.id,
+      },
+      payload: {
+        details: 'New action details',
+      },
+    });
+
+    cy
+      .get('.sidebar')
+      .find('[data-details-region] .js-input')
+      .should('have.value', 'New action details');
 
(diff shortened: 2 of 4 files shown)


src/js/views/patients/sidebar/action/action-sidebar-action_views.js
```diff
@@ -130,12 +130,16 @@ const ActionView = View.extend({
...
+    'change:details': 'onChangeDetails',
     'change:_state': 'onChangeActionState',
     'change:_owner': 'onChangeOwner',
     'change:due_date': 'onChangeDue',
     'change:due_time': 'onChangeDue',
     'change:duration': 'onChangeDuration',
   },
+  onChangeDetails() {
+    this.showEditForm();
+  },
```
nmajor25, line 144: A user could be making edits to the action details in the action sidebar and have those edits wiped out and replaced if a `DetailsChanged` web socket message comes in.

Is that okay from a UI/user experience perspective?

src/js/views/patients/sidebar/action/action-sidebar-action_views.js
```diff
@@ -130,12 +130,16 @@ const ActionView = View.extend({
     duration: '[data-duration-region]',
   },
   modelEvents: {
+    'change:details': 'onChangeDetails',
```
nmajor25, line 133: We'll need to add something similar for the action name here in the sidebar.

Currently the action name won't be updated in the sidebar if a `NameChanged` ws message comes in.

src/js/views/patients/sidebar/action/action-sidebar-action_views.js
paulfalgout: So this might be tough.. but if the user is currently modifying the details we do not need to rerender the form.. we can just let them override them, but if they cancel the form, it should show the socket details.

paulfalgout (changes requested):

src/js/views/patients/sidebar/action/action-sidebar-action_views.js
nmajor25: Okay ya that seems like the way to go. See if I can get that working.

nmajor25: Just pushed a commit with a potential implementation.

src/js/views/patients/sidebar/action/action-sidebar-action_views.js
```diff
@@ -181,14 +183,22 @@ const ActionView = View.extend({
     this.showChildView('save', new SaveView({ model: this.clonedAction }));
   },
   onSave() {
+    this.isEditingDetails = false;
+
     this.getRegion('save').empty();
   },
```
nmajor25, line 184: Could this function just be:

```js
onSave() {
  this.showEditForm();
}
```

`showEditForm()` calls `this.getRegion('save').empty()`.

test/integration/patients/patient/patient-flow.js
nmajor25, line 1: So to test this I did:

1. User makes manual edit to details by typing in text area.
2. Web socket notification is received. Verify user's manual edits are NOT overridden.
3. Save the user's manual edited details. Now the user is no longer editing the details.
4. Web socket notification is received. Verify text area content is replaced with the socket details.

src/js/views/patients/sidebar/action/action-sidebar-action_views.js
paulfalgout: I actually think not doing this is a bug... but I don't think action names change much.  We have to handle events in the entire context of a page.

test/integration/patients/patient/patient-flow.js
paulfalgout: I'm not sure I understand 4.  The user will not get notifications for the modifications they make themselves.

src/js/views/patients/sidebar/action/action-sidebar-action_views.js
paulfalgout: so `showEditForm` immediately clones the action.. I think that would occur prior to the onSave in the app which copies the details from the clone to the model.. so saving would revert the data I think?

paulfalgout: _but_ it probably does make sense to put these two lines in some for of "stopEditing" that is called both here and within showEditForm

paulfalgout (changes requested):

test/integration/patients/patient/patient-flow.js
nmajor25: 1-3 is for when the current user is actively making edits in the details text area. To make sure that when a web socket notification comes in, their edits are not overridden.

For 3, I submitted the user's edits so the form won't be in the `isEditingDetails = true` state anywhere. To make testing 4 possible.

4 represents a web socket notification from another user. Where the details are not being edited by the current user and therefore the details text area contents are replaced.

If that makes sense...

src/js/views/patients/sidebar/action/action-sidebar-action_views.js
nmajor25: Agreed, I added a `stopEditingDetails()`

src/js/views/patients/sidebar/action/action-sidebar-action_views.js
nmajor25: Okay, I can tackle this in a separate story/PR.

https://app.shortcut.com/roundingwell/story/56480/in-action-sidebar-update-name-in-ui-when-that-action-s-model-name-changes

test/integration/patients/patient/patient-flow.js
nmajor25: I changed the test so the user cancel's their manually added edits. Instead of submitting them. I think that made the test confusing and overly complicated.

Now is this sequence:

1. User makes manual edit to details by typing in text area.
2. Web socket notification is received. Verify user's manual edits are NOT overridden.
3. **_User clicks cancel button to clear their changes._** Now the user is no longer editing the details.
4. Web socket notification is received. Verify text area content is replaced with the socket details.

src/js/views/patients/sidebar/action/action-sidebar-action_views.js
```diff
@@ -177,19 +183,28 @@ const ActionView = View.extend({
...
   onCancel() {
     this.showEditForm();
   },
   showEditForm() {
     this.cloneAction();
-    this.listenTo(this.clonedAction, 'change:details', this.showSave);
 
-    this.getRegion('save').empty();
+    this.listenTo(this.clonedAction, 'change:details', () => {
```
paulfalgout: can we make this line `this.listenTo(this.clonedAction, 'change:details', this.startEditingDetails);` that's the flag and the this.showChildView line from `showSave`?

paulfalgout (changes requested):
one nitpick but this is good

src/js/views/patients/sidebar/action/action-sidebar-action_views.js
nmajor25: Nice, I like it. Just pushed that change 👇

paulfalgout (approved):

PR #1348, opened 2024-11-05
Support `CommentAdded` web socket notification in patient action sidebar

nmajor25:
Shortcut Story ID: [sc-56328]

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -1,3 +1,4 @@
+import { invoke } from 'underscore';
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 import dayjs from 'dayjs';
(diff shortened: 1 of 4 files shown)


src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25, line 1: When we receive a `CommentAdded` ws notification, we need to make an api request to get the full comment model, right? And then add it to the activity collection (events & comments combined into one collection)?

That was my assumption with this.

To do that I think we need to do the `subscribe()` directly in the `action-sidebar_app`. Instead of the flow app. The flow app and views listen for changes to the action model. But this doesn't include a change to the model.

But that means the action sidebar will handle the `CommentAdded` ws notification outside of the patient flow page context.

test/integration/patients/patient/patient-flow.js
```diff
@@ -228,6 +241,33 @@ context('patient flow page', function() {
...
+        type: 'patient-actions',
+        id: testFlowAction.id,
+      },
+      payload: {
+        comment: {
+          type: 'comments',
+          id: testSocketComment.id,
+        },
+        attributes: {
+          message: 'New websocket comment.',
```
nmajor25, line 257: If we use this implementation, this portion of the ws notification message isn't used.

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: I would imagine we'll run into the same situation with the `AttachmentAdded` notification.

paulfalgout: nope.  subscribing within the sidebar will clear the subscription on the flow.  this action is already subscribed to.  Plus I think we should probably handle this on the action.. and if it passes the message itself do we need to fetch it from the API?  what data is missing?  If it's just a timestamp we can set it to now. Anything else we could probably just ask for it to be included on the comment.

paulfalgout (changes requested):

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: The websocket notification looks like this: https://github.com/RoundingWell/care-ops-backend/blob/develop/tests/Notifications/Data/Messages/comment-added-message-structure.json.

If we have the comment contents. So if we had the `created_at` and clinician name included, we wouldn't need to fetch it via the api.

nmajor25: I'm going to close this PR and start a new one.

PR #1349, opened 2024-11-06
Support `CommentAdded` web socket notification in patient action sidebar

nmajor25:
Shortcut Story ID: [sc-56328]

Implementation is on the patient action sidebar. But since the web socket subscription is currently only on the patient flow page, this message will only be handled there.

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -41,7 +41,9 @@ export default App.extend({
 
     const flow = this.action.getFlow();
     if (flow) this.listenTo(flow, 'change:_state', this.setAccess);
+
     this.listenTo(action, 'change:_owner', this.onChangeOwner);
+    this.listenTo(action, 'message', this.onWebsocketMessage);
 
     this.showMenu();
     this.showAction();
@@ -62,6 +64,21 @@ export default App.extend({
       Radio.request('entities', 'fetch:comments:collection:byAction', this.action.id),
       Radio.request('entities', 'fetch:files:collection:byAction', this.action.id),
     ];
+  },
+  onWebsocketMessage({ category, payload }) {
+    if (category !== 'CommentAdded') return;
+
+    const { comment, relationships, attributes } = payload;
+
+    const model = Radio.request('entities', 'comments:model', {
+      id: comment.id,
+      message: attributes.message,
+      created_at: attributes.created_at,
+      edited_at: attributes.edited_at,
+      _clinician: relationships._clinician,
+    });
+
+    this.activityCollection.add(model);
   },
   onChangeOwner() {
     this.setAccess();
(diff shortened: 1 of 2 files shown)


test/integration/patients/patient/patient-flow.js
```diff
@@ -228,6 +243,38 @@ context('patient flow page', function() {
...
+        attributes: {
+          message: 'New websocket comment.',
+          created_at: testCreatedAt,
+          edited_at: null,
+        },
+        relationships: {
+          _clinician: testClinician.id,
+        },
+      },
+    });
```
nmajor25, line 267: I'm not sure how this socket message data should be structured. But as long as we have the following, we should be able to just add the comment to the activity collection without doing an api request:

* Comment `id`
* Comment `message`
* Comment `created_at`
* Comment `edited_at`
* Author/clinician `id`

Socket message data is currently looks like this: https://github.com/RoundingWell/care-ops-backend/blob/develop/tests/Notifications/Data/Messages/comment-added-message-structure.json

Current ideas/questions:

1. Could we assume `edited_at` is always `null` for this? Since that will be `null` whenever a new comment is created.
2. Could we assume `created_at` is always `now()`? If that's not included in the socket message data. Can there be a delay behind the scenes between the comment being created and the socket notification being emitted?
3. Do we want a web socket notification for when a comment has been edited?

test/integration/patients/patient/patient-flow.js
nmajor25, line 1: I'll probably end up changing these tests a little bit. But wanted to get something created quick for this draft PR.

test/integration/patients/patient/patient-flow.js
paulfalgout: hmm this is getting complicated.  Can we move on to other stuff for a bit?

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -63,6 +65,21 @@ export default App.extend({
       Radio.request('entities', 'fetch:files:collection:byAction', this.action.id),
     ];
   },
+  onWebsocketMessage({ category, payload }) {
+    if (category !== 'CommentAdded') return;
+
+    const { comment, relationships, attributes } = payload;
```
paulfalgout, line 71: clinician wouldn't be included here it'd be the author https://github.com/RoundingWell/care-ops-backend/blob/2c937bc0180ffe7aefe6ba15449ad5d246cd1f6d/tests/Notifications/Data/Messages/comment-added-message-structure.json#L4

paulfalgout: we can set `created_at` to `dayjs().now()` and leave `edited_at` null.. this'd be fine, but until we can handle comment editing, we probably should handle adding, and that to me is much more complicated

paulfalgout (changes requested):

PR #1351, opened 2024-11-11
Support `SharingUpdated` web socket notification on patient flow page

nmajor25:
Shortcut Story ID: [sc-56333]

## Summary by CodeRabbit

- **New Features**
	- Introduced a new message handler for sharing attributes in the model.
  
- **Improvements**
	- Streamlined message handlers to accept a single `attributes` parameter, simplifying state updates.
	- Updated styling for the sharing form to enhance user interface consistency.

- **Tests**
	- Enhanced integration tests for patient flow, including new attributes for actions and improved websocket handling.

### DIFF
--- a/src/js/entities-service/entities/actions.js
+++ b/src/js/entities-service/entities/actions.js
@@ -39,6 +39,9 @@ const _Model = BaseModel.extend({
     DetailsChanged({ details, attributes = {} }) {
       this.set({ details, ...attributes });
     },
+    SharingUpdated({ attributes }) {
+      this.set(attributes);
+    },
   },
   urlRoot() {
     if (this.isNew()) {
--- a/src/js/entities-service/entities/actions.js
+++ b/src/js/entities-service/entities/actions.js
@@ -27,17 +27,17 @@ const _Model = BaseModel.extend({
     StateChanged({ state, attributes = {} }) {
       this.set({ _state: state.id, ...attributes });
     },
-    ActionDueChanged({ due_date, due_time, attributes = {} }) {
-      this.set({ due_date, due_time, ...attributes });
+    ActionDueChanged({ attributes }) {
+      this.set(attributes);
     },
-    ActionDurationChanged({ duration, attributes = {} }) {
-      this.set({ duration, ...attributes });
+    ActionDurationChanged({ attributes }) {
+      this.set(attributes);
     },
-    NameChanged({ name, attributes = {} }) {
-      this.set({ name, ...attributes });
+    NameChanged({ attributes }) {
+      this.set(attributes);
     },
-    DetailsChanged({ details, attributes = {} }) {
-      this.set({ details, ...attributes });
+    DetailsChanged({ attributes }) {
+      this.set(attributes);
     },
     SharingUpdated({ attributes }) {
       this.set(attributes);
--- a/src/js/views/patients/sidebar/action/action-sidebar-forms_views.js
+++ b/src/js/views/patients/sidebar/action/action-sidebar-forms_views.js
@@ -58,7 +58,7 @@ function getSharingOpts(sharing) {
 }
 
 const FormSharingView = View.extend({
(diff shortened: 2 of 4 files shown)


src/js/views/patients/sidebar/action/action-sidebar-forms_views.js
```diff
@@ -58,7 +58,7 @@ function getSharingOpts(sharing) {
 }
 
 const FormSharingView = View.extend({
-  className: 'sidebar__dialog',
+  className: 'sidebar__dialog u-margin--t-24',
```
nmajor25, line 61: So that the element has the correct top margin when being shown/hidden on the fly via a web socket notification.

test/integration/patients/patient/patient-flow.js
nmajor25, line 1: I think this is a simple implementation, unless I'm missing something. 

In the action sidebar tests, I did:

1. Start out action as having `sharing: 'disabled'`. i.e. sharing section is hidden in UI.
2. Socket notification comes in that sets `sharing: 'enabled'`. Verify UI elements are now shown.
3. Socket notification comes in tha sets `sharing: 'responded'`. Verify UI elements are updated.

In the action list, I did:

1. Start out action as having `outreach: 'disabled'`. i.e. icon for the action list item is the default version.
2. Socket notification comes in that sets `outreach: 'patient'`. Verify icon for action list item changes. (I'm not sure there's anything else that changes based on that `outreach` attribute).

src/js/entities-service/entities/actions.js
nmajor25, line 1: The notification from the BE contains data for the `recipient`: https://github.com/RoundingWell/care-ops-backend/blob/develop/tests/Notifications/Data/Messages/sharing-updated-message-structure.json.

Can we assume the recipient will always be the patient the action/flow belongs to? And therefore we don't need to do anything on the FE with that recipient data?

src/js/entities-service/entities/actions.js
```diff
@@ -39,6 +39,9 @@ const _Model = BaseModel.extend({
     DetailsChanged({ details, attributes = {} }) {
       this.set({ details, ...attributes });
     },
+    SharingUpdated({ sharing, outreach, attributes = {} }) {
```
paulfalgout: so a bunch of these are wrong.. the only thing doing any work here is attributes

src/js/views/patients/sidebar/action/action-sidebar-forms_views.js
```diff
@@ -102,7 +102,7 @@ const FormLayoutView = View.extend({
       {{#if hasForm}}<h4 class="sidebar__label u-margin--t-8">{{ @intl.patients.sidebar.action.actionSidebarFormsViews.formLayoutView.formLabel }}</h4>{{/if}}
       <div class="flex-grow" data-form-region></div>
     </div>
-    <div class="flex{{#if hasSharing}} u-margin--t-24{{/if}}">
+    <div class="flex">
```
paulfalgout, line 105: is the templateContext still in use?

src/js/entities-service/entities/actions.js
paulfalgout: I don't think FE needs to know the recipient.  FE needs to know all of the data such that if the user hard refreshed they'd see the same thing.

paulfalgout (changes requested):

src/js/views/patients/sidebar/action/action-sidebar-forms_views.js
nmajor25: The `hasSharing` value can be removed. But `hasForm` is still in use in that `templateContext()`.

src/js/entities-service/entities/actions.js
nmajor25: These have been updated 👍

paulfalgout (approved):

PR #1352, opened 2024-11-14
Change `updated_at` attribute when action/flow model is updated via web socket

nmajor25:
Shortcut Story ID: [sc-56745]

When we receive a web socket notification of any kind for a flow or action, its `updated_at` attribute will now be updated to the current date-time.

This change is applied on the base model level. But will only be put to use on the patient flow page since that's the only place we currently subscribe to web socket notifications.

On the patient flow page, that is displayed in these UI locations:

1. Each flow action list item.
2. Footer section of action sidebar.
3. Footer section of flow sidebar.

## Release Notes

- **New Features**
	- Enhanced message handling to include a timestamp (`updated_at`) for better tracking.
	- Introduced real-time updates in the UI for timestamp changes, improving responsiveness.

- **Bug Fixes**
	- Improved integration tests to ensure accurate display of updated timestamps and real-time changes in the patient flow management system.

- **Documentation**
	- Updated test cases to validate the new timestamp functionality and websocket event handling.

### DIFF
--- a/src/js/base/model.js
+++ b/src/js/base/model.js
@@ -1,5 +1,6 @@
 import { extend, isEmpty, isFunction, pick, reduce, result } from 'underscore';
 import Backbone from 'backbone';
+import dayjs from 'dayjs';
 import JsonApiMixin from './jsonapi-mixin';
 
 export default Backbone.Model.extend(extend({
@@ -84,6 +85,8 @@ export default Backbone.Model.extend(extend({
     return isFunction(messages[category]) ? messages[category] : this[messages[category]];
   },
   handleMessage({ category, payload }) {
+    payload.attributes = extend({}, payload.attributes, { updated_at: dayjs.utc().format() });
+
     const handler = this._getMessageHandler(category);
     if (handler) handler.call(this, payload);
 
--- a/src/js/views/patients/sidebar/flow/flow-sidebar_views.js
+++ b/src/js/views/patients/sidebar/flow/flow-sidebar_views.js
@@ -76,6 +76,7 @@ const LayoutView = View.extend({
   modelEvents: {
     'change:_state': 'showOwner',
     'change:_owner': 'showFlow',
+    'change:updated_at': 'showTimestamps',
   },
   onAttach() {
     animSidebar(this.el);
--- a/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
+++ b/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
@@ -264,6 +264,9 @@ const TimestampsView = View.extend({
       createdAt: this.getOption('createdEvent').get('date'),
     };
   },
+  modelEvents: {
+    'change:updated_at': 'render',
+  },
 });
 
 export {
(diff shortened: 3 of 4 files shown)


src/js/views/patients/sidebar/flow/flow-sidebar_views.js
```diff
@@ -76,6 +76,7 @@ const LayoutView = View.extend({
   modelEvents: {
     'change:_state': 'showOwner',
     'change:_owner': 'showFlow',
+    'change:updated_at': 'showTimestamps',
```
nmajor25, line 79: This doesn't do anything. But the `updated` timestamp in the flow sidebar is updating correctly on model change. 

So I'm not sure what's going on here yet.

src/js/base/model.js
```diff
@@ -84,6 +85,12 @@ export default Backbone.Model.extend(extend({
     return isFunction(messages[category]) ? messages[category] : this[messages[category]];
   },
   handleMessage({ category, payload }) {
+    if (payload.attributes) {
+      payload.attributes.updated_at = dayjs.utc().format();
+    } else {
+      payload.attributes = { updated_at: dayjs.utc().format() };
+    }
```
nmajor25: Probably a better way to do this. But just attaching `attributes.updated_at` to the payload data before its passed to the individual entity services.

src/js/views/patients/sidebar/flow/flow-sidebar_views.js
nmajor25: I'm noticing other weird things with the flow sidebar, in terms of updating on a web socket model change.

Name and owner update correctly. But details and state do not.

I'm going to investigate/address those in this story [here](https://app.shortcut.com/roundingwell/story/56894/update-flow-sidebar-on-web-socket-model-changes).

src/js/base/model.js
paulfalgout: `payload.attributes = extend(payload.attributes, { updated_at: day...`

paulfalgout (changes requested):

src/js/base/model.js
nmajor25: That would be much cleaner 👍

src/js/views/patients/sidebar/flow/flow-sidebar_views.js
nmajor25: For the first comment above, the `change:updated_at` model event is indeed working. I just had the test setup incorrectly. The flow sidebar needs to be open before the web socket notification comes in. I updated the test to do that.

I still plan on investigating the second comment in a separate story.

src/js/base/model.js
```diff
@@ -84,6 +85,8 @@ export default Backbone.Model.extend(extend({
     return isFunction(messages[category]) ? messages[category] : this[messages[category]];
   },
   handleMessage({ category, payload }) {
+    payload.attributes = extend(payload.attributes || {}, { updated_at: dayjs.utc().format() });
```
paulfalgout: nitpick but I think `extend({}, payload.attributes, updated_at` is probably better since it'll remove the logic branch and it doesn't mutate the `payload.attributes`

paulfalgout (changes requested):
One tiny thing then this is good

src/js/base/model.js
nmajor25: Agreed, I updated it to that 👍

paulfalgout (approved):

PR #1353, opened 2024-11-22
On ws notifications, update name, details, and state in flow sidebar

nmajor25:
Shortcut Story ID: [sc-56894]

## Summary by CodeRabbit

- **New Features**
	- Enhanced sidebar to dynamically display patient name and details using new data attributes.
	- Introduced `NameView` and `DetailsView` for improved rendering of patient information in the sidebar.

- **Bug Fixes**
	- Improved integration tests to ensure real-time updates are accurately reflected in the sidebar.

- **Documentation**
	- Updated test suite to verify websocket notifications and UI updates.

### DIFF
--- a/src/js/views/patients/sidebar/flow/flow-sidebar_views.js
+++ b/src/js/views/patients/sidebar/flow/flow-sidebar_views.js
@@ -60,6 +60,11 @@ const MenuView = View.extend({
   },
 });
 
+const NameView = View.extend({
+  className: 'patient-flow-sidebar__name',
+  template: hbs`{{ name }}`,
+});
+
 const PermissionView = View.extend({
   className: 'flex u-margin--t-8',
   template: hbs`
@@ -73,6 +78,7 @@ const PermissionView = View.extend({
 const SidebarView = View.extend({
   template: FlowSidebarTemplate,
   regions: {
+    name: '[data-name-region]',
     state: '[data-state-region]',
     owner: '[data-owner-region]',
     permission: '[data-permission-region]',
@@ -88,6 +94,7 @@ const SidebarView = View.extend({
     };
   },
   modelEvents: {
+    'change:name': 'showName',
     'change:_state': 'showOwner',
     'change:_owner': 'showFlow',
   },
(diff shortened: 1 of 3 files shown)


src/js/views/patients/sidebar/flow/flow-sidebar_views.js
```diff
@@ -107,7 +107,7 @@ const SidebarView = View.extend({
   modelEvents: {
     'change:name': 'showName',
     'change:details': 'showDetails',
-    'change:_state': 'showOwner',
```
nmajor25, line 91: Not sure why this was `showOwner`? May be missing something, but I think `showFlow` is what this should be.

src/js/views/patients/sidebar/flow/flow-sidebar_views.js
```diff
@@ -98,13 +117,23 @@ const SidebarView = View.extend({
   showFlow() {
     this.canEdit = this.model.canEdit();
 
+    this.showName();
```
paulfalgout, line 114: I don't think name or details need to be in this function..  they can just be in onRender.. this function appears to be needed if something changes that affects `canEdit`

src/js/views/patients/sidebar/flow/flow-sidebar_views.js
paulfalgout: canEdit is only affected by the owner. I think this should just be `'change:_state': 'showState'`

src/js/views/patients/sidebar/flow/flow-sidebar_views.js
paulfalgout: really I think this method did more a bit ago and it likely doesn't make any sense now..  `this.canEdit = this.model.canEdit();` should move into `showActions` and  changing the owner should do that.

paulfalgout (changes requested):

src/js/views/patients/sidebar/flow/flow-sidebar_views.js
nmajor25: I moved `showName()` and `showDetails()` up to the render function. And moved the `this.canEdit = ...` to `showActions()`.

src/js/views/patients/sidebar/flow/flow-sidebar_views.js
nmajor25: So this was originally `showOwner` because of this: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/views/patients/sidebar/flow/flow-sidebar_views.js#L133.

If the flow state is set to `Done`, the owner component becomes disabled.

So I have it set to `'change:_state': 'showActions'`. Since the `change:_state` model events listener needs to now update the state component and potentially the owner component if state is set to done. 

But in this scenario `this.canEdit = this.model.canEdit();` and `this.showPermission()` would be ran for no purpose.

paulfalgout (approved):

PR #1354, opened 2024-11-22
Handle `ResourceDeleted` websocket notification for actions on patient flow page

nmajor25:
Shortcut Story ID: [sc-56332]

When this happens, we want to remove the action from the list via `destroy()` and close the action sidebar if it's open.

We added some code to the base model `destroy()` method to stop the `DELETE` api request in this scenario.

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -43,7 +43,11 @@ export default App.extend(extend({
     this.action.trigger('editing', true);
     const flow = this.action.getFlow();
     if (flow) this.listenTo(flow, 'change:_state', this.setAccess);
-    this.listenTo(action, 'change:_owner', this.onChangeOwner);
+
+    this.listenTo(action, {
+      'change:_owner': this.onChangeOwner,
+      'destroy': this.stop,
+    });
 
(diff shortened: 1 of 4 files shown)


src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -44,6 +44,7 @@ export default App.extend(extend({
     const flow = this.action.getFlow();
     if (flow) this.listenTo(flow, 'change:_state', this.setAccess);
     this.listenTo(action, 'change:_owner', this.onChangeOwner);
+    this.listenTo(action, 'destroy', this.stop);
```
paulfalgout: I think the two action listenTos can be made into one with {}

paulfalgout (changes requested):

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: Updated, much cleaner 👍

paulfalgout (approved):

paulfalgout:
needs a rebase @nmajor25

nmajor25:
@paulfalgout rebased, should be good now.

PR #1356, opened 2024-11-26
Add an action name a-z default sort option to worklists

nmajor25:
Shortcut Story ID: [sc-57090]

Adds `Action: A - Z` & `Action: Z - A` sort options on the action versions of worklists, excluded from flow versions.

This is a default sort option shown for all environments/customers.


paulfalgout (approved):
We need to fix our coverage issues.. makes the "change in base" confusing on the coverage reports..

PR #1358, opened 2024-12-03
Handle comments web socket notifications on patient flow page

nmajor25:
Shortcut Story ID: [sc-56328] [sc-57310]

Coinciding with the backend PR: https://github.com/RoundingWell/care-ops-backend/pull/8796.

- [x] Support `CommentedAdded` notifications
- [x] Support `CommentEdited` notifications
- [x] Support `CommentRemoved` notifications

## Summary by CodeRabbit

- **New Features**
	- Enhanced functionality for adding, editing, and removing comments within the patient flow.
	- New event listeners for real-time updates in the comment and timestamp views.
	- Added support for attachments with error handling during uploads.

- **Bug Fixes**
	- Improved handling of comment timestamps and attachment uploads.

- **Tests**
	- Expanded test cases for comments, including adding, editing, and removing scenarios in the patient flow.

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -47,6 +47,7 @@ export default App.extend(extend({
     this.listenTo(action, {
       'change:_owner': this.onChangeOwner,
       'destroy': this.stop,
+      'add:comment': this.onCommentAdded,
     });
 
     this.showChildView('heading', new HeadingView({ model: this.action }));
@@ -72,6 +73,9 @@ export default App.extend(extend({
     this.setAccess();
     /* istanbul ignore else : Covers edge case when owner changes prior to beforeStart */
     if (this.isRunning()) this.showAttachments();
+  },
+  onCommentAdded(model) {
+    this.activityCollection.add(model);
   },
   onStart(options, activity, comments, attachments) {
     this.activityCollection = new Backbone.Collection([...activity.models, ...comments.models]);
--- a/src/js/base/model.js
+++ b/src/js/base/model.js
@@ -90,12 +90,13 @@ export default Backbone.Model.extend(extend({
 
     return isFunction(messages[category]) ? messages[category] : this[messages[category]];
   },
-  handleMessage({ category, payload }) {
+  handleMessage({ category, author, payload }) {
     payload.attributes = extend({}, payload.attributes, { updated_at: dayjs.utc().format() });
+    payload.author = author;
 
     const handler = this._getMessageHandler(category);
     if (handler) handler.call(this, payload);
 
-    this.trigger('message', { category, payload });
+    this.trigger('message', { category, author, payload });
   },
 }, JsonApiMixin));
--- a/src/js/entities-service/entities/actions.js
+++ b/src/js/entities-service/entities/actions.js
@@ -42,6 +42,17 @@ const _Model = BaseModel.extend({
     SharingUpdated({ attributes }) {
       this.set(attributes);
     },
+    CommentAdded({ author, comment, attributes }) {
+      const commentModel = Radio.request('entities', 'comments:model', {
+        id: comment.id,
+        message: attributes.message,
+        created_at: dayjs.utc().format(),
+        edited_at: null,
+        _clinician: author,
+      });
+
+      this.trigger('add:comment', commentModel);
+    },
     ResourceDeleted() {
       this.destroy({ isDeleted: true });
     },
--- a/src/js/entities-service/entities/comments.js
+++ b/src/js/entities-service/entities/comments.js
@@ -2,6 +2,7 @@ import Radio from 'backbone.radio';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
+import dayjs from 'dayjs';
 
 import trim from 'js/utils/formatting/trim';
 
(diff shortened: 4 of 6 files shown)


src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -68,6 +69,21 @@ export default App.extend(extend({
       Radio.request('entities', 'fetch:files:collection:byAction', this.action.id),
     ];
   },
+  onWebsocketMessage({ category, author, payload }) {
```
paulfalgout: I think this needs to be on the action somehow.  we're talking about having a comment count.. possibly having these show up in a notification bar.. this'd require the action sidebar be open to get a new comment.  I think we'll have to trigger some event on the action we can listento here for just the `this.activityCollection.add(model);`

paulfalgout (changes requested):

src/js/base/model.js
```diff
@@ -90,12 +90,13 @@ export default Backbone.Model.extend(extend({
 
     return isFunction(messages[category]) ? messages[category] : this[messages[category]];
   },
-  handleMessage({ category, payload }) {
+  handleMessage({ category, author, payload }) {
     payload.attributes = extend({}, payload.attributes, { updated_at: dayjs.utc().format() });
+    payload.author = author;
```
nmajor25, line 95: Fine with just attaching this to the `payload`?

src/js/base/model.js
```diff
@@ -90,12 +90,13 @@ export default Backbone.Model.extend(extend({
...
 
     const handler = this._getMessageHandler(category);
     if (handler) handler.call(this, payload);
 
-    this.trigger('message', { category, payload });
+    this.trigger('message', { category, author, payload });
```
nmajor25, line 100: Do we need to still include the `author` here? I don't think so...

src/js/entities-service/entities/actions.js
```diff
@@ -42,6 +42,17 @@ const _Model = BaseModel.extend({
...
+    CommentAdded({ author, comment, attributes }) {
+      const commentModel = Radio.request('entities', 'comments:model', {
+        id: comment.id,
+        message: attributes.message,
+        created_at: dayjs.utc().format(),
+        edited_at: null,
+        _clinician: author,
+      });
+
+      this.trigger('comment:added', commentModel);
```
nmajor25: Fine as `comment:added? Or should it be more specific like `ws:comment:added`? Wasn't sure.

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: Took a crack at implementing this.

src/js/entities-service/entities/actions.js
paulfalgout: should be `add:comment` present tense `verb:noun`

paulfalgout (changes requested):

src/js/entities-service/entities/actions.js
nmajor25: Ah right 👍 , okay switched it to that.

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -121,6 +121,9 @@ const CommentView = View.extend({
   regions: {
     comment: '[data-comment-activity-region]',
   },
+  modelEvents: {
+    'change:message': 'render',
+  },
```
nmajor25, line 126: Pretty sure doing a `render` of the individual comment in question is the correct way... because the comment may be in a "editing" state.

test/integration/patients/patient/patient-flow.js
```diff
@@ -340,10 +330,52 @@ context('patient flow page', function() {
...
+      .as('socketComment')
       .should('contain', formatDate(testTs(), 'AT_TIME'))
-      .should('contain', 'Test Clinician')
+      .should('contain', 'Clinician McTester')
       .should('contain', 'New websocket comment.');
 
+    cy
+      .get('@socketComment')
+      .find('.js-edit')
+      .click();
```
nmajor25, line 340: Not sure if this is 100% necessary. But I want to include something in the tests that signifies that if the comment is in a "editing" state, the `CommentEdited` ws notification will re-render the comment. And potentially wipe out the user's edits if they have one.

Since a comment can only be edited by it's author, this is an edge case that should only happen when a user is editing a form in two different browser tabs or devices at the same time.

src/js/entities-service/entities/comments.js
```diff
@@ -2,13 +2,19 @@ import Radio from 'backbone.radio';
...
 
 import trim from 'js/utils/formatting/trim';
 
 const TYPE = 'comments';
 
 const _Model = BaseModel.extend({
   type: TYPE,
+  messages: {
+    CommentEdited({ attributes }) {
+      this.set({ edited_at: dayjs.utc().format(), ...attributes });
```
nmajor25, line 15: Setting `edited_at` to the current dayjs timestamp so the comment will show `(Edited)` for the comment in the UI.

nmajor25:
@paulfalgout attachments ws stuff was merged in the BE (https://github.com/RoundingWell/care-ops-backend/pull/8787). But not the comments ws stuff. I was mistaken when I said that during the call.

So we still need to wait until https://github.com/RoundingWell/care-ops-backend/pull/8796 is merged.

nmajor25:
@paulfalgout looks like https://github.com/RoundingWell/care-ops-backend/pull/8796 was merged.

paulfalgout (approved):

PR #1361, opened 2024-12-10
Add action details tooltip to patient flow & dashboard/archive

nmajor25:
Shortcut Story ID: [sc-57533]

## Summary by CodeRabbit

- **New Features**
	- Introduced a tooltip feature for action items, displaying additional details when hovered over.
	- Added new elements for displaying action details in both the archive and dashboard views.
	- Added a new CSS class for styling display icons in tables.

- **Bug Fixes**
	- Enhanced test coverage for action details visibility and user permissions on the patient archive and dashboard pages.

- **Style**
	- Updated styles for margin adjustments in table elements to improve layout.
	- Increased width for patient list metadata to enhance visual presentation.

- **Tests**
	- Expanded integration tests to cover tooltip functionality and various clinician roles, ensuring proper error handling and visibility of actions.

### DIFF
--- a/src/js/views/patients/patient/dashboard/flow-item.hbs
+++ b/src/js/views/patients/patient/dashboard/flow-item.hbs
@@ -3,7 +3,7 @@
   <span class="patient__action-name">{{ name }}</span>
 </td>
 <td class="table-list__cell patient__list-meta">
-  <span class="table-list__meta">{{~ remove_whitespace ~}}
+  <span class="table-list__meta u-margin--l-32">{{~ remove_whitespace ~}}
     <span class="patient__flow-status action--{{ stateOptions.color }}">{{fa stateOptions.iconType stateOptions.icon}}</span>{{~ remove_whitespace ~}}
   </span>{{~ remove_whitespace ~}}
   <span class="table-list__meta" data-owner-region></span>{{~ remove_whitespace ~}}
--- a/src/js/views/patients/patient/flow/action-item.hbs
+++ b/src/js/views/patients/patient/flow/action-item.hbs
@@ -7,6 +7,7 @@
   </span>
 </td>
 <td class="table-list__cell patient__list-meta">
+  <span class="table-list__display-icon u-margin--r-16" data-details-region></span>{{~ remove_whitespace ~}}
   <span class="table-list__meta" data-state-region></span>{{~ remove_whitespace ~}}
   <span class="table-list__meta" data-owner-region></span>{{~ remove_whitespace ~}}
   <span class="table-list__meta">{{~ remove_whitespace ~}}
--- a/src/js/views/patients/patient/flow/header.hbs
+++ b/src/js/views/patients/patient/flow/header.hbs
@@ -2,7 +2,7 @@
 <table class="table-list">
   <td class="patient-flow__details">{{ details }}</td>{{~ remove_whitespace ~}}
   <td class="patient-flow__meta">
-    <span class="table-list__meta" data-state-region></span>
+    <span class="table-list__meta u-margin--l-32" data-state-region></span>
     <span class="table-list__meta" data-owner-region></span>
     <progress class="js-progress progress-bar patient-flow__progress" value={{ _progress.complete }} max={{ _progress.total }}></progress>{{~ remove_whitespace ~}}
   </td>
--- a/src/js/views/patients/patient/flow/patient-flow.scss
+++ b/src/js/views/patients/patient/flow/patient-flow.scss
@@ -82,7 +82,7 @@
 
 .patient-flow__meta {
   padding-left: 8px;
-  width: 478px;
+  width: 510px;
 }
 
 .patient-flow__actions {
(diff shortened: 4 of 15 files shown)


src/js/views/patients/patient/flow/patient-flow.scss
```diff
@@ -81,7 +81,7 @@
 }
 
 .patient-flow__meta {
-  padding-left: 16px;
+  padding-left: 8px;
```
nmajor25, line 84: On the patient flow page, the alignment is off for this currently in QA:

Screenshot 2024-12-10 at 12 36 15 PM

This is why I made this update here.

src/js/views/patients/patient/flow/header.hbs
```diff
@@ -2,7 +2,7 @@
 <table class="table-list">
   <td class="patient-flow__details">{{ details }}</td>{{~ remove_whitespace ~}}
   <td class="patient-flow__meta">
-    <span class="table-list__meta" data-state-region></span>
+    <span class="table-list__meta u-margin--l-32" data-state-region></span>
```
nmajor25, line 5: So the meta items for the flow in the header and action items align with each other.

src/js/views/patients/patient/flow/action-item.hbs
nmajor25, line 1: With these updates, there's some overflow issues when an action item is completely full of meta items:

Screenshot 2024-12-10 at 11 08 42 AM

So I still need to figure out how to fix that. Increasing the width by 32px should fix it. But there's side effects that need to be handled in order to keep everything aligned with the flow header.

nmajor25: Same thing needs to be fixed for the patient dashboard/archive. But that should be easier.

Screenshot 2024-12-10 at 11 03 45 AM

src/js/views/patients/patient/dashboard/flow-item.hbs
```diff
@@ -3,7 +3,7 @@
   <span class="patient__action-name">{{ name }}</span>
 </td>
 <td class="table-list__cell patient__list-meta">
-  <span class="table-list__meta">{{~ remove_whitespace ~}}
+  <span class="table-list__meta u-margin--l-32">{{~ remove_whitespace ~}}
```
nmajor25, line 6: There can be both actions and flows in the dashboard/arcive lists.

So I added this to keep the flows aligned with the actions. Which will now have the details icon that has 32px width.

nmajor25: Looks like this without this 32px margin:

Screenshot 2024-12-10 at 10 04 12 AM

src/js/views/patients/patient/flow/action-item.hbs
nmajor25: I fixed these via comit: https://github.com/RoundingWell/care-ops-frontend/pull/1361/commits/7a61240dd2c8467e2637f98577df4cc91aea54e6.

paulfalgout (changes requested):
This looks good, but I think we need to handle if details are added over websockets or changed in a sidebar

nmajor25:
@paulfalgout both of those things are happening already. The tooltip shows on the `DetailsChanged` ws notification and if details are added/removed in the action sidebar. Looks like the action list items are re-rendering on any model change.

I added a test for the websocket portion. But I'm not sure where to add tests for the other stuff. I can't find anywhere else where we're testing that updates in the sidebar are mirrored in the action list items.

paulfalgout:
@nmajor25 I'm not seeing it actually.  I see it rerendering on owner change, but not details..  I realize the tooltip data renders each time, but if there's no details.. then details are added the icon shows up?  And if the details are remove the icon goes away?  It might support this somewhere as this is how it should work on the worklists, but I don't see where this is happening.

nmajor25:
@paulfalgout yep, the details icon is shown/hidden when the user updates the details via the sidebar. I added a screen grab example of the patient dashboard.

The flow page action items have a `'change': 'render'` model event listener: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/views/patients/patient/flow/flow_views.js#L140

The dashboard/archive action items have a `RowBehavior` that does the `'change': 'render'` model event listener: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/views/patients/patient/dashboard/dashboard_views.js#L34-L45

https://github.com/user-attachments/assets/15e84a96-64cd-49bd-bdfa-d6f542f73e2e

paulfalgout:
@nmajor25 ah in that case the `'change:_owner': 'render',` modelEvents is redundant and should be removed..  it means when we change the owner we're rendering the row twice

nmajor25:
@paulfalgout good call, removed those in https://github.com/RoundingWell/care-ops-frontend/pull/1361/commits/c0e8f4bc3d57c635a31ea5acac13cf4af30f74f7.

paulfalgout (approved):

PR #1365, opened 2024-12-13
Handle attachments web socket notifications on patient flow page

nmajor25:
Shortcut Story ID: [sc-###]

Coinciding with the backend PR: https://github.com/RoundingWell/care-ops-backend/pull/8787.

- [x]  Support `AttachmentAdded` notifications
- [ ]  Support `FileReplaced` notifications
- [ ]  Support `FileRemoved` notifications

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -48,6 +48,7 @@ export default App.extend(extend({
       'change:_owner': this.onChangeOwner,
       'destroy': this.stop,
       'add:comment': this.onCommentAdded,
+      'add:attachment': this.onAttachmentAdded,
     });
 
     this.showChildView('heading', new HeadingView({ model: this.action }));
@@ -76,6 +77,9 @@ export default App.extend(extend({
   },
   onCommentAdded(model) {
     this.activityCollection.add(model);
+  },
+  onAttachmentAdded(model) {
+    this.attachments.add(model);
   },
   onStart(options, activity, comments, attachments) {
     this.activityCollection = new Backbone.Collection([...activity.models, ...comments.models]);
--- a/src/js/entities-service/entities/actions.js
+++ b/src/js/entities-service/entities/actions.js
@@ -53,6 +53,19 @@ const _Model = BaseModel.extend({
 
       this.trigger('add:comment', commentModel);
     },
+    AttachmentAdded({ file, attributes }) {
+      const attachmentModel = Radio.request('entities', 'files:model', {
+        id: file.id,
+        path: attributes.path,
+        created_at: dayjs.utc().format(),
+        _action: this.id,
+        _patient: this.getPatient().id,
+        _view: attributes.view,
+        _download: attributes.download,
+      });
+
+      this.trigger('add:attachment', attachmentModel);
+    },
     ResourceDeleted() {
       this.destroy({ isDeleted: true });
     },
--- a/src/js/entities-service/entities/actions.js
+++ b/src/js/entities-service/entities/actions.js
@@ -64,6 +64,9 @@ const _Model = BaseModel.extend({
         _download: attributes.download,
       });
 
+      const newFilesRelationship = [...this.get('_files'), { id: file.id, type: 'files' }];
+      this.set({ _files: newFilesRelationship });
+
       this.trigger('add:attachment', attachmentModel);
     },
     ResourceDeleted() {
--- a/test/integration/patients/patient/patient-flow.js
+++ b/test/integration/patients/patient/patient-flow.js
@@ -108,6 +108,7 @@ context('patient flow page', function() {
 
   specify('patient flow action sidebar', function() {
     const testCommentId = uuid();
+    const testOtherFileId = uuid();
 
     const testPatient = getPatient({
       attributes: {
@@ -141,6 +142,7 @@ context('patient flow page', function() {
         'form': getRelationship(testForm),
         'patient': getRelationship(testPatient),
         'program-action': getRelationship(testProgramAction),
+        'files': getRelationship([{ id: testOtherFileId }], 'files'),
       },
     });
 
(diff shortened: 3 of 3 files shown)


src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -77,6 +78,9 @@ export default App.extend(extend({
   onCommentAdded(model) {
     this.activityCollection.add(model);
   },
+  onAttachmentAdded(model) {
```
nmajor25, line 81: This function name is really similar to the existing `onAddAttachment()` one. Maybe we should rename these to something more specific like `onWsCommentAdded` and `onWsAttachmentAdded()` or something similar?

src/js/entities-service/entities/actions.js
```diff
@@ -53,6 +53,19 @@ const _Model = BaseModel.extend({
...
+    AttachmentAdded({ file, attributes }) {
+      const attachmentModel = Radio.request('entities', 'files:model', {
+        id: file.id,
+        path: attributes.path,
+        created_at: dayjs.utc().format(),
+        _action: this.id,
+        _patient: this.getPatient().id,
+        _view: attributes.view,
+        _download: attributes.download,
+      });
```
nmajor25, line 65: Maybe we don't need some of these? Not sure yet.

test/integration/patients/patient/patient-flow.js
```diff
@@ -389,6 +404,37 @@ context('patient flow page', function() {
...
+          type: 'files',
+          id: '70294032-539b-4dc0-93df-9e4ff2eee8c1',
+        },
+        attributes: {
+          path: `patients/${ testPatient.id }/HRA.pdf`,
+          view: `https://www.bucket_name.s3.amazonaws.com/patients/${ testPatient.id }/view/HRA.pdf`,
+          download: `https://www.bucket_name.s3.amazonaws.com/patients/${ testPatient.id }/download/HRA.pdf`,
+        },
+      },
+    });
```
nmajor25, line 447: The `attributes` object here isn't currently being included in what the BE sends. They're only sending `payload.clinician` and `payload.file`.

BE example message structure is [here](https://github.com/RoundingWell/care-ops-backend/blob/develop/tests/Notifications/Data/Messages/attachment-added-message-structure.json).

If we're given the `path`, `view`, and `download` attributes, we should be able to construct an attachment model ourselves and add it to the UI. We'd need the same data for the `FileReplaced` notification as well.

Otherwise we currently only have the file `id`. Meaning we'd have to make a request to the BE api to get the file model.

test/integration/patients/patient/patient-flow.js
```diff
@@ -389,6 +404,37 @@ context('patient flow page', function() {
...
+      category: 'AttachmentAdded',
+      resource: {
+        type: 'patient-actions',
+        id: testFlowAction.id,
+      },
+      payload: {
+        clinician: {
+          type: 'clinicians',
+          id: '95014c50-2401-4d95-83e1-4a39fb928c90',
+        },
```
nmajor25, line 436: I don't think `payload.clinician` is of any use to us.

test/integration/patients/patient/patient-flow.js
```diff
@@ -389,6 +404,37 @@ context('patient flow page', function() {
...
+      },
+    });
+
+    cy
+      .get('[data-attachments-files-region]')
+      .children()
+      .as('attachmentItems')
+      .should('have.length', 1)
+      .first()
+      .contains('HRA.pdf');
```
nmajor25, line 455: Note to self: ensure that the attachment icon becomes visible for the action list item in this scenario.

test/integration/patients/patient/patient-flow.js
```diff
@@ -389,6 +404,37 @@ context('patient flow page', function() {
...
+      payload: {
+        clinician: {
+          type: 'clinicians',
+          id: '95014c50-2401-4d95-83e1-4a39fb928c90',
+        },
+        file: {
+          type: 'files',
+          id: '70294032-539b-4dc0-93df-9e4ff2eee8c1',
+        },
+        attributes: {
```
paulfalgout, line 441: so there's attributes would be the `resource` attributes... but these are the file attributes so they'd need to be nested in file or else we need to make a request.

src/js/entities-service/entities/actions.js
```diff
@@ -64,6 +64,12 @@ const _Model = BaseModel.extend({
         _download: attributes.download,
       });
 
+      const newFilesRelationship = this.get('_files').push({
+        id: file.id,
+        type: 'files',
+      });
+      this.set({ _files: newFilesRelationship });
```
nmajor25, line 68: This sets a new `_files` relationship on the action model. Which will trigger the `modelEvent` listener on the `ActionItemView`([link](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/views/patients/patient/flow/flow_views.js#L140)) and show/hide the paperclip icon.

nmajor25: The paperclip icon showing or hiding is based on the `action.hasAttachments()` method, which checks the length of `action.get('_files')`.

So doing a `modelEvents: { 'add:comment': 'render' }` on the `ActionItemView` doesn't work.

nmajor25: The action may already have attachments, so we need to be aware of that when updating the `_files` relationship array.

test/integration/patients/patient/patient-flow.js
```diff
@@ -141,6 +142,7 @@ context('patient flow page', function() {
         'form': getRelationship(testForm),
         'patient': getRelationship(testPatient),
         'program-action': getRelationship(testProgramAction),
+        'files': getRelationship([{ id: testOtherFileId }], 'files'),
```
nmajor25, line 145: Adding an existing file for the action. To ensure that it isn't removed when we set the new `_files` relationship.

PR #1367, opened 2024-12-17
Delete action routing and error handling

paulfalgout:
This does two things.
When a action is deleted we should close the sidebar so the default routing can be handled. Also if routed to an action that is no longer cached we should alert the user the action is no longer available.

Shortcut Story ID: [sc-57747]

## Summary by CodeRabbit

- **New Features**
	- Introduced improved error handling for actions not found, enhancing user notifications.
	- Added internationalization support for patient flow views and various components.
	- Implemented a new method for centralized error alerts related to actions.

- **Bug Fixes**
	- Enhanced state management for action deletions and user notifications in the sidebar.

- **Tests**
	- Expanded test coverage for action deletion and real-time updates in the patient flow and sidebar functionalities.

### DIFF
--- a/src/js/apps/patients/patient/flow/flow_app.js
+++ b/src/js/apps/patients/patient/flow/flow_app.js
@@ -2,7 +2,7 @@ import { bind, invoke, get } from 'underscore';
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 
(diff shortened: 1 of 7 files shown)


nmajor25 (approved):
Code looks good. Tested it locally and works as expected 👍

PR #1368, opened 2024-12-18
Subscribe/unsubscribe to each comment in action sidebar

nmajor25:
Shortcut Story ID: [sc-57310]

Should subscribe to all comments when the action sidebar is opened. And should also subscribe to an individual comment when one is added via websocket or manual user action.

Subscriptions should be removed when the action sidebar is closed.

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -61,6 +61,8 @@ export default App.extend(extend({
     if (flow) this.stopListening(flow);
     this.stopListening(this.action);
     this.action.trigger('editing', false);
+
+    this.removeSubscriptions();
   },
   beforeStart() {
     return [
@@ -76,14 +78,20 @@ export default App.extend(extend({
   },
   onCommentAdded(model) {
     this.activityCollection.add(model);
+    this.comments.add(model);
+
+    Radio.request('ws', 'add', model);
   },
   onStart(options, activity, comments, attachments) {
     this.activityCollection = new Backbone.Collection([...activity.models, ...comments.models]);
+    this.comments = comments;
     this.attachments = attachments;
 
     this.showActivity();
     this.showNewCommentForm();
     this.showAttachments();
+
+    this.addSubscriptions();
   },
   showContent() {
     const sidebarView = new SidebarView({ model: this.action });
@@ -185,7 +193,12 @@ export default App.extend(extend({
   },
   onPostNewComment({ model }) {
     model.set({ created_at: dayjs.utc().format() }).save();
+
     this.activityCollection.add(model);
+    this.comments.add(model);
+
+    Radio.request('ws', 'add', model);
+
     this.showNewCommentForm();
   },
   onCancelNewComment() {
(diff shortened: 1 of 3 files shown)


src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -224,5 +232,26 @@ export default App.extend(extend({
   onRemoveAttachment(model) {
     model.destroy();
   },
+  addSubscriptions() {
+    const commentSubscriptions = [];
+
+    this.activityCollection.each(model => {
+      if (model.get('message')) {
```
nmajor25: The activity collection items don't contain a `type` attribute for comments. The only way to tell if it's a comment is if it contains the `message` attribute.

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -76,6 +78,7 @@ export default App.extend(extend({
   },
   onCommentAdded(model) {
     this.activityCollection.add(model);
+    Radio.request('ws', 'add', { id: model.id, type: 'comments' });
```
paulfalgout: `Radio.request('ws', 'add', model)` is ok here

src/js/apps/patients/sidebar/action-sidebar_app.js
paulfalgout: maybe `const commentSubscripts = this.activityCollection.filter({ type: 'comments' })`

paulfalgout: however we build this collection in the `onStart`.. we could probably just attach the comments separately rather than filter them each time.  We just need to add to both.

paulfalgout (changes requested):

src/js/apps/patients/sidebar/action-sidebar_app.js
paulfalgout: `model.type` I believe

test/integration/patients/worklist/reduced-schedule.js
```diff
@@ -254,11 +254,16 @@ context('reduced schedule page', function() {
       .routeFlow()
       .routeFlowActions()
       .routePatientByFlow()
+      .routeAction()
+      .routeActionComments()
+      .routeActionActivity()
```
nmajor25: Doesn't seem like I should have to add these, but it fails without it when using the `wait()`s I added below.

test/integration/patients/patient/patient-flow.js
```diff
@@ -330,7 +330,7 @@ context('patient flow page', function() {
       .find('.comment__item')
       .last()
       .as('socketComment')
-      .should('contain', formatDate(testTs(), 'AT_TIME'))
+      .should('contain', formatDate(dayjs.utc().format(), 'AT_TIME'))
```
nmajor25, line 333: Unrelated to this PR. I've noticed this has been causing flakiness for a little while now.

paulfalgout: yeah..  tests revolving things happening in real time and not on the api probably need to be like this.. _or_ we need `visitOnClock` where we're controlling what's happening, but yeah this matches that first line in `onPostNewComment`

paulfalgout (approved):

PR #1371, opened 2024-12-20
Handle attachments web socket notifications on patient flow page

nmajor25:
Shortcut Story ID: [sc-56327] [sc-57309]

Coinciding with the backend PR: https://github.com/RoundingWell/care-ops-backend/pull/8787.

- [x]  Support `AttachmentAdded` notifications
- [x]  Support `FileReplaced` notifications
- [x]  Support `FileRemoved` notifications
- [x] Show paperclip icon in action list items when file is added/removed via websocket

## Release Notes

- **New Features**
  - Enhanced WebSocket communication for comments and attachments
  - Improved file management functionality with real-time updates

- **Bug Fixes**
  - Refined event handling for adding and removing attachments
  - Updated message processing for better resource tracking

- **Improvements**
  - Streamlined integration of WebSocket notifications
  - Added support for file replacement and removal operations

- **Testing**
  - Expanded test coverage for patient flow and file attachment scenarios

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -48,6 +48,7 @@ export default App.extend(extend({
       'change:_owner': this.onChangeOwner,
       'destroy': this.onDestroy,
       'add:comment': this.onCommentAdded,
+      'add:attachment': this.onAttachmentAdded,
     });
 
     this.showChildView('heading', new HeadingView({ model: this.action }));
@@ -79,6 +80,11 @@ export default App.extend(extend({
   onCommentAdded(model) {
     this.activityCollection.add(model);
     this.comments.add(model);
+
+    Radio.request('ws', 'add', model);
+  },
+  onAttachmentAdded(model) {
+    this.attachments.add(model);
 
     Radio.request('ws', 'add', model);
   },
@@ -230,6 +236,8 @@ export default App.extend(extend({
     });
     attachment.upload(file);
 
+    Radio.request('ws', 'add', file);
+
     this.listenTo(attachment, 'upload:failed', () => {
       Radio.request('alert', 'show:error', intl.patients.sidebar.actionSidebarApp.uploadError);
     });
@@ -238,12 +246,12 @@ export default App.extend(extend({
     model.destroy();
   },
   addSubscriptions() {
-    Radio.request('ws', 'add', this.comments.models);
+    Radio.request('ws', 'add', [...this.comments.models, ...this.attachments.models]);
   },
   removeSubscriptions() {
     // for when sidebar is closed before comments api request is finished
-    if (!this.comments) return;
+    if (!this.comments || !this.attachments) return;
 
-    Radio.request('ws', 'unsubscribe', this.comments.models);
+    Radio.request('ws', 'unsubscribe', [...this.comments.models, ...this.attachments.models]);
   },
 }, SidebarMixin));
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -251,8 +251,8 @@ export default App.extend(extend({
     Radio.request('ws', 'add', [...this.comments.models, ...this.attachments.models]);
   },
   removeSubscriptions() {
-    // for when sidebar is closed before comments api request is finished
-    if (!this.comments || !this.attachments) return;
+    // for when the sidebar is closed prior to beforeStart being able to finish
+    if (!this.isRunning()) return;
 
     Radio.request('ws', 'unsubscribe', [...this.comments.models, ...this.attachments.models]);
   },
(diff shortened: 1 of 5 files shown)


src/js/entities-service/entities/actions.js
```diff
@@ -250,6 +250,14 @@ const _Model = BaseModel.extend({
...
     return !!size(programAction.get('allowed_uploads'));
   },
+  removeAttachment(resource) {
+    const files = this.get('_files');
+    const newFilesRelationship = files.filter(file => {
+      return file.id !== resource.id;
+    });
+
+    this.set({ _files: newFilesRelationship });
+  },
```
nmajor25, line 260: The paperclip attachments icon on action list items is shown or hidden based on the length of `action.get('_files')`.

We're currently not updating that when a file is removed manually by a user. And therefore it doesn't work out of the box for the `FileRemoved` ws notification.

So i added this method to the action entity service that filters out the file from `_files` by `id`. And then sets a new `_files` relationship array for the action.

This allows for the paperclip icon to be hidden after attachment deletion.

nmajor25: If this is how we want to implement this, we would just need to call `action.removeAttachment(model)` in the action sidebar app when a file is manually deleted by a user.

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -82,6 +83,11 @@ export default App.extend(extend({
 
     Radio.request('ws', 'add', model);
   },
+  onAttachmentAdded(model) {
+    this.attachments.add(model);
+
+    Radio.request('ws', 'add', model);
+  },
```
nmajor25, line 90: This `onAttachmentAdded()` function name is almost identical to the already existing `onAddAttachment()` in this file.

We should probably name this something else, like `OnWsAttachmentAdded()`.

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -238,12 +246,12 @@ export default App.extend(extend({
     model.destroy();
   },
   addSubscriptions() {
-    Radio.request('ws', 'add', this.comments.models);
+    Radio.request('ws', 'add', [...this.comments.models, ...this.attachments.models]);
   },
   removeSubscriptions() {
     // for when sidebar is closed before comments api request is finished
-    if (!this.comments) return;
+    if (!this.comments || !this.attachments) return;
```
nmajor25: This doesn't seem right to me. I don't know if there's a situation where only one of these is `undefined`. Since the `beforeStart()` requests are called together, right?

Is there another way of guarding against the sidebar being closed before `beforeStart()` can finish?

paulfalgout: I think this might be an `isRunning` or `isLoading` test of some sort

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -230,20 +236,24 @@ export default App.extend(extend({
     });
     attachment.upload(file);
 
+    Radio.request('ws', 'add', file);
```
nmajor25: `file` is going to be `undefined` here. I think we need to change `upload(file)` to a promise and then do something like:

```js
attachment.upload(file).then(file => {
  Radio.request('ws', 'add', file);
})
```

src/js/entities-service/entities/actions.js
nmajor25: Currently in the app, the `.paperclip` icon isn't shown when an attachment is added and isn't hidden when an attachment is removed manually by a user in the sidebar.

I created a separate story [here]( https://app.shortcut.com/roundingwell/story/58034/in-action-list-items-show-hide-paperclip-icon-when-an-attachment-is-manually-added-removed-via-action-sidebar), that I can tackle in a separate PR.

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: I switched it to use `this.isRunning()` 👍

src/js/apps/patients/sidebar/action-sidebar_app.js
paulfalgout: actually I think it's just `Radio.request('ws', 'add', attachment)`

src/js/apps/patients/sidebar/action-sidebar_app.js
paulfalgout: The events should possibly be: `ws:add:attachment`   Then `onWsAddAttachment` similar to `state:verb:subject` `onStateVerbSubject`

paulfalgout (changes requested):

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: Updated to this ☝️

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: Makes sense, updated to that 👍

paulfalgout (approved):

PR #1372, opened 2025-01-02
Show/hide action list item paperclip icon when attachment is manually added/removed by users

nmajor25:
Shortcut Story ID: [sc-58034]

## Summary by CodeRabbit

- **New Features**
	- Enhanced attachment management for actions.
	- Improved file upload and removal handling.
	- Added user feedback for upload success and failure.

- **Bug Fixes**
	- Corrected file relationship tracking in actions.
	- Ensured UI reflects accurate attachment status.

- **Tests**
	- Added integration test for action attachment functionality.
	- Verified attachment icon updates during file upload and removal.

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -236,14 +236,20 @@ export default App.extend(extend({
     });
     attachment.upload(file);
 
-    Radio.request('ws', 'add', attachment);
-
-    this.listenTo(attachment, 'upload:failed', () => {
-      Radio.request('alert', 'show:error', intl.patients.sidebar.actionSidebarApp.uploadError);
+    this.listenTo(attachment, {
+      'upload:success': uploadedAttachment => {
+        this.action.addFile(uploadedAttachment);
+        Radio.request('ws', 'add', uploadedAttachment);
+      },
+      'upload:failed': () => {
+        Radio.request('alert', 'show:error', intl.patients.sidebar.actionSidebarApp.uploadError);
+      },
     });
   },
   onRemoveAttachment(model) {
     model.destroy();
+
+    this.action.removeFile(model);
 
     Radio.request('ws', 'unsubscribe', model);
   },
--- a/src/js/entities-service/entities/actions.js
+++ b/src/js/entities-service/entities/actions.js
@@ -64,8 +64,7 @@ const _Model = BaseModel.extend({
         _download: attributes.urls.download,
       });
 
-      const newFilesRelationship = [...this.get('_files'), { id: file.id, type: 'files' }];
-      this.set({ _files: newFilesRelationship });
+      this.addFile(attachmentModel);
 
       this.trigger('ws:add:attachment', attachmentModel);
     },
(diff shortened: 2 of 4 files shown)


src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -236,6 +236,7 @@ export default App.extend(extend({
     });
     attachment.upload(file);
 
+    this.action.addAttachment(attachment);
```
nmajor25: When passing the attachment to the `addAttachment` function here, `attachment.id` is `undefined`. Which will cause the test I created in this PR to fail.

nmajor25: I think we need to wait until the upload is finished.

We could do something like this:

```js
this.listenTo(attachment, 'upload:success', uploadedAttachment => {
  this.action.addAttachment(uploadedAttachment);
  Radio.request('ws', 'add', uploadedAttachment);
});
```

Or:

```js
attachment.upload(file).then(uploadedFile => {
  this.action.addAttachment(uploadedAttachment);
  Radio.request('ws', 'add', uploadedAttachment);
});

Comment (nmajor25) [reply src/js/apps/patients/sidebar/action-sidebar_app.js]: And if an upload fails (`'upload:failed'` triggered), this would also stop the paperclip icon and websocket from being added when they shouldn't be.

File: test/integration/patients/sidebar/action-sidebar.js
Comment (nmajor25) [inline test/integration/patients/sidebar/action-sidebar.js:1]: Decided to create a new test. I couldn't figure out how to weave the testing of this feature into an existing one.

Since this test needs to have an action with this sequence:

* Start with zero attachments (icon hidden)
* Add an attachment (icon switched to shown)
* Remove that attachment (icon switch back to hidden)

File: src/js/entities-service/entities/actions.js [hunk 2]
```diff
@@ -250,6 +249,11 @@ const _Model = BaseModel.extend({
 
     return !!size(programAction.get('allowed_uploads'));
   },
+  addAttachment(resource) {
```
nmajor25: Are the names of these clear? Could cause confusion in the future (for me at least).

Maybe they should be `addAttachmentRelationship()`/`removeAttachmentRelationshop()` or `addFileRelationship()`/`removeFileRelationship()` 🤔

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -236,7 +236,10 @@ export default App.extend(extend({
     });
     attachment.upload(file);
 
-    Radio.request('ws', 'add', attachment);
+    this.listenTo(attachment, 'upload:success', uploadedAttachment => {
```
paulfalgout: we should combine this with 244 `this.listenTo(attachment, {}`

src/js/entities-service/entities/actions.js
```diff
@@ -250,7 +249,12 @@ const _Model = BaseModel.extend({
 
     return !!size(programAction.get('allowed_uploads'));
   },
-  removeAttachment(resource) {
+  addFileRelationship(resource) {
```
paulfalgout: I think we can remove the "Relationship" from the name  The other instances like this don't have it.. `addTag` `removeTag` etc

paulfalgout (changes requested):

src/js/entities-service/entities/actions.js
nmajor25: Makes sense, updated to that 👍

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: Yep good call, updated to that 👍

paulfalgout (approved):

PR #1374, opened 2025-01-06
When a new comment is created, add web socket subscription after api request finishes

nmajor25:
Shortcut Story ID: [sc-58035]

This handles an edge case situation where a user creates a comment in one tab and then edits it in another. In that scenario, the comment's `id` in the web socket notification would be `undefined` if we don't wait for the api request to come back.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved comment submission reliability by ensuring WebSocket notifications are sent only after successful comment save.

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -198,12 +198,14 @@ export default App.extend(extend({
     });
   },
   onPostNewComment({ model }) {
-    model.set({ created_at: dayjs.utc().format() }).save();
+    model.set({ created_at: dayjs.utc().format() });
+
+    model.save().then(() => {
+      Radio.request('ws', 'add', model);
+    });
 
     this.activityCollection.add(model);
     this.comments.add(model);
-
-    Radio.request('ws', 'add', model);
 
     this.showNewCommentForm();
   },


src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -198,13 +198,13 @@ export default App.extend(extend({
     });
   },
   onPostNewComment({ model }) {
-    model.set({ created_at: dayjs.utc().format() }).save();
+    model.set({ created_at: dayjs.utc().format() }).save().then(comment => {
+      Radio.request('ws', 'add', comment);
+    });
```
nmajor25, line 205: Not sure if we want or need to test this change specifically... the api requests in cypress are almost instant so we'd have to add a delay or something to the request.

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -198,13 +198,13 @@ export default App.extend(extend({
     });
   },
   onPostNewComment({ model }) {
-    model.set({ created_at: dayjs.utc().format() }).save();
+    model.set({ created_at: dayjs.utc().format() }).save().then(comment => {
```
paulfalgout: I think maybe just breaking up this line like:
model.set(..)
model.save().then

Also one question..  I'm not sure `then(comment` is the comment?  it's the result of the fetch I _think_.   But I'm not sure.. but `'ws', 'add', model` will always work within the `then`

paulfalgout (changes requested):

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: That makes sense. Updated to split them between lines.

nmajor25: Ya it's a response stream. So to access that data we'd have to read it via `.json()`.

Like:

```js
model.save().then(response => {
  response.json().then(data => console.log(data))
});
```

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25, line 1: I don't think we specifically test this scenario anywhere.

Where a comment is manually added by a user and then we verify that the ws subscription was added by doing a `CommentEdited` notification on that comment.

What we have now will have coverage. But we don't have a test that will fail if we were to remove this `Radio.request('ws', 'add', model);` code.

I'm having a hard time getting that to work in Cypress. So I need to investigate it more.

paulfalgout (approved):

PR #1376, opened 2025-01-08
Support websocket workspace

paulfalgout:
Shortcut Story ID: [sc-58278]

For the moment at least all websocket subscriptions.. similar to api requests can by default pass the current workspace.

We might run into issue in the future where we want to monitor across workspace, but we'll tackle that when we get there as it will likely coincide with changes to how we interact with the API as well.  (ie: get a notification from a different workspace.. make a request on that different workspace)

In addition to this is a bit of cleanup of unused features or flake-related issues

* I had added `getResource` to model in order to make it easier to feed the subscription service `{ id: type }` but later we just made the subscription service support an array of models so it's no longer needed.
* I removed the `connectionResponseMessage` from the mock websocket server as it was never used and likely never will be
* I removed wsIntercept for a stub for any handler.  wsIntercept was difficullt to manage timing wise.  Removing it uncovered issues for more consistent tests.
* websockets were closing at the end of the tests causing the app to resubscribe which that message did overlap tests causing some bleed into the new stub
* similarly we were sending an empty subscribe in the tests because `onOpen` sometimes passed data and sometimes didn't depending on timing.

## Release Notes

- **Bug Fixes**
  - Improved WebSocket service handling and connection management
  - Enhanced resource subscription logic

- **Refactor**
  - Simplified WebSocket message processing in testing environment
  - Streamlined WebSocket notification handling
  - Removed deprecated resource retrieval method

- **Tests**
  - Updated WebSocket testing support
  - Improved test assertions for WebSocket interactions

### DIFF
--- a/src/js/apps/patients/patient/flow/flow_app.js
+++ b/src/js/apps/patients/patient/flow/flow_app.js
@@ -1,4 +1,4 @@
-import { bind, invoke, get } from 'underscore';
+import { bind, get } from 'underscore';
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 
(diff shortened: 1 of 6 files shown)


nmajor25 (approved):
Looks good 👍

PR #1379, opened 2025-01-09
Fix sass deprecation warnings

nmajor25:
Shortcut Story ID: [sc-57510]

Main deprecation updates:

* Switch from deprecated `map-get()` to `map.get()`.
* Switch from using deprecated `@import`'s to `@use` & `@forward` instead.
  * More info on the deprecation [here](https://sass-lang.com/blog/import-is-deprecated/).


nmajor25:
I've got this PR set as a draft. I'd like to do more testing in my local dev environment. To make sure all the styles are working as before.

vite.config.js
```diff
@@ -32,7 +32,6 @@ const css = {
     scss: {
       additionalData: `
         @use 'sass:math';
-        @use 'sass:color';
```
nmajor25, line 35: I couldn't find anywhere this is being used.

src/scss/core/_base.scss
```diff
@@ -1,3 +1,5 @@
+@use '../provider-variables' as *;
```
nmajor25, line 1: The sass `preprocessorOptions` in the `vite.config.js` file adds `@use 'src/scss/provider-variables.scss' as *;` to the top of all our sass files? Or just the ones that are imported via javascript?

Because that didn't happen for these `/src/scss/core` sass files. I had to add these here for the `$` prefixed variables to work.

src/scss/provider-variables.scss
```diff
@@ -6,8 +6,10 @@
 //   Main Colors
 // -------------------------------------
 
+@use 'sass:map';
```
nmajor25: Eh this line should probably be placed above the `Main Color` comment...

src/scss/core/_base.scss
paulfalgout: imports never needed this because they were effectively combined into the provider-core file which would have the include but `@forward` compiles and namespaces an export rather than putting each partial into the same namespace

src/scss/provider-core.scss
```diff
@@ -1,13 +1,13 @@
...
-@import 'core/flex';
-@import 'core/fonts';
-@import 'core/icon';
-@import 'core/layout';
-@import 'core/overrides';
-@import 'core/scrollbar';
-@import 'core/space';
-@import 'core/text';
-@import 'core/typography';
+@forward 'core/reset';
```
paulfalgout, line 3: So.. I'm not certain, but what we might be able to do here is:

```scss
@use '../provider-variables' as *;

@use 'core/reset';
...
@use 'core/typography';

@forward 'core/base';
...
@forward 'core/typography';
```

I think that would put them all in the same namespace and the export them all with the shared namespacing..  but I'm not 💯  on that.

I'm also not sure that it's better..

vite.config.js
```diff
@@ -31,9 +31,7 @@ const css = {
   preprocessorOptions: {
     scss: {
       additionalData: `
-        @use 'sass:math';
-        @use 'sass:color';
-        @import 'src/scss/provider-variables.scss';
+        @use 'src/scss/provider-variables.scss' as *;
```
paulfalgout, line 36: It's possible that we should remove this altogether and break up the variables into separate `@use` namespaces imports.. at least that seems to be what scss wants us to do.. but.. I dunno.. this pattern of having a few global scss vars hasn't hurt us at all.. and the imports would be everywhere...

paulfalgout (commented):
Bah.. I dislike this change to sass..  but we gotta do something.. up for whichever TBH

src/scss/core/_base.scss
nmajor25: Makes sense 👍

vite.config.js
nmajor25: So to implement this we'd need to add `@use src/scss/provider-variables.scss` in every `.scss` file? With relative paths too? Like `@use ../../scss/provider-variables.scss`?

When working on this I thought about doing that. But I couldn't figure out another way besides having the `@use` in a ton of different files.

src/scss/provider-core.scss
nmajor25: Hmm interesting, that's worth investigating. Not sure if that will work either, but would be nice if it did!

vite.config.js
paulfalgout: yeah.. well I think that's sass intention in deprecating `@import`.  They want explicit and obvious dependencies.. and for people to only use them when necessary... so they'd probably suggest against using `as *` and rather using `as colors` or something where the sass ends up `color: colors.blue;`   In theory I agree... but in practice.. yeah.. it's so many @use when our global "if you see a $variable, you'll find it in our short list of provider-variables.scss" seems to work pretty well for us.  The `additionalData` was always a little hacky.. but in some ways the more we fight the standard the more hacky our solutions are starting to look..   but as you said.. the alternative is just so much `@use` which doesn't really seem to clarify anything IMO.  So I dunno. sass sucks.

src/scss/provider-core.scss
nmajor25: Couldn't get this to work 😞

paulfalgout: one other thought that might be a hair cleaner.. at least for now.. we could make a js file at the very top of index called `provider-core-styles.js`.. remove this file and all the `_`d file names.. or something and instead of using sass partials just import all of these at the top..  then they'd get the `additionalData` and we wouldn't have to add that to the partials.

nmajor25:
Closing this PR and switching to a new one: https://github.com/RoundingWell/care-ops-frontend/pull/1382. I messed up a rebase.

PR #1380, opened 2025-01-10
Add worklist subscription and filters

paulfalgout:
Shortcut Story ID: [sc-59124]

So to implement the filters messages we have to grab a model from the store.. rather than copy this pattern to a bunch of places I moved it to the entities channel.

Then I looked at other places where this pattern existed and changed it to `Model.getRelationship` which will coincide eventually with simplify the various relationships once https://github.com/RoundingWell/care-ops-frontend/pull/1320 is complete.

For not the filters implementation on the ws service is overly simplistic, but I don't know what kind of filter maintenance we're going to run into yet or if it's always a simple replace.

But I'm opening this as a draft as I think I need to real-world test the socket filters before adding tests in cypress etc.

## Summary by CodeRabbit

- **New Features**
  - Enhanced the patient worklist with real-time notifications that immediately reflect updates for flows and actions.
  - Introduced improved filtering for a more accurate and responsive display of data.
  - Streamlined subscription mechanism to accommodate filters during WebSocket communications.

- **Bug Fixes**
  - Strengthened error handling during live data communications, reducing the risk of processing incomplete updates.

- **Tests**
  - Added automated tests to ensure that real-time notifications trigger the correct UI updates for patient worklist changes.
  - Implemented specifications for socket notifications related to flows and actions to verify UI updates.

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -154,20 +154,9 @@ export default App.extend({
 
     this.showList();
   },
-  _getWsFetchOptions(filterKey) {
-    if (filterKey === 'flows') return {};
-
-    return {
-      data: {
-        fields: { flows: ['name', 'state'] },
-        include: ['program-action.program', 'flow.program-flow.program'].join(),
-      },
-    };
-  },
   subscribe() {
     const channel = Radio.channel('ws');
     const filterKey = this.getState().getType();
-    const flowStates = this.getState('flowStates');
 
     channel.request('subscribe', this.collection.models, { filters: { [filterKey]: this.filters } });
 
@@ -177,16 +166,37 @@ export default App.extend({
 
       if (this.collection.get(model) || model.isFetching()) return;
 
-      const options = this._getWsFetchOptions(filterKey);
+      if (filterKey === 'flows') {
+        this._fetchFlow(model);
 
-      model.fetch(options).then(() => {
-        // notifications do not fully filter our action flow_states
-        if (filterKey === 'actions' && !flowStates.includes(model.getFlow().getState().id)) return;
+        return;
+      }
 
-        this.collection.add(model);
+      this._fetchAction(model);
 
-        channel.request('add', model);
-      });
+      return;
+    });
+  },
+  _fetchFlow(model) {
+    const fetchFlow = Radio.request('entities', 'fetch:flows:model', model.id);
+
+    fetchFlow.then(() => {
+      this.collection.add(model);
+
+      Radio.request('ws', 'add', model);
+    });
+  },
+  _fetchAction(model) {
+    const flowStates = this.getState('flowStates');
+
+    const fetchAction = Radio.request('entities', 'fetch:actions:model', model.id);
+
+    fetchAction.then(() => {
+      if (!flowStates.includes(model.getFlow().getState().id)) return;
+
+      this.collection.add(model);
+
+      Radio.request('ws', 'add', model);
     });
   },
   // NOTE: Shows views dependent on getState().getType()
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -154,6 +154,16 @@ export default App.extend({
 
     this.showList();
   },
+  _getWsFetchOptions(filterKey) {
+    if (filterKey === 'flows') return {};
+
+    return {
+      data: {
+        fields: { flows: ['name', 'state'] },
+        include: ['program-action.program', 'flow.program-flow.program'].join(),
+      },
+    };
+  },
   subscribe() {
     const channel = Radio.channel('ws');
     const filterKey = this.getState().getType();
@@ -167,7 +177,9 @@ export default App.extend({
 
       if (this.collection.get(model) || model.isFetching()) return;
 
-      model.fetch().then(() => {
+      const options = this._getWsFetchOptions(filterKey);
+
+      model.fetch(options).then(() => {
         // notifications do not fully filter our action flow_states
         if (filterKey === 'actions' && !flowStates.includes(model.getFlow().getState().id)) return;
 
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -157,6 +157,7 @@ export default App.extend({
   subscribe() {
     const channel = Radio.channel('ws');
     const filterKey = this.getState().getType();
+    const flowStates = this.getState('flowStates');
 
     channel.request('subscribe', this.collection.models, { filters: { [filterKey]: this.filters } });
 
@@ -164,6 +165,9 @@ export default App.extend({
       if (this.collection.get(model)) return;
 
       model.fetch().then(() => {
+        // notifications do not fully filter our action flow_states
+        if (filterKey === 'actions' && !flowStates.includes(model.getFlow().getState().id)) return;
+
         this.collection.add(model);
       });
     });
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -162,8 +162,6 @@ export default App.extend({
     channel.request('subscribe', this.collection.models, { filters: { [filterKey]: this.filters } });
 
     this.listenTo(channel, 'message', ({ category, resource, payload }) => {
-      if (!category) return;
-
       const modelResource = category === 'ResourceCreated' ? payload.resource : resource;
       const model = Radio.request('entities', 'get:store', modelResource);
 
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -162,7 +162,7 @@ export default App.extend({
     channel.request('subscribe', this.collection.models, { filters: { [filterKey]: this.filters } });
 
     this.listenTo(channel, 'message', (message, model) => {
-      if (this.collection.get(model)) return;
+      if (this.collection.get(model) || model.isFetching()) return;
 
       model.fetch().then(() => {
         // notifications do not fully filter our action flow_states
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -161,7 +161,12 @@ export default App.extend({
 
     channel.request('subscribe', this.collection.models, { filters: { [filterKey]: this.filters } });
 
-    this.listenTo(channel, 'message', (message, model) => {
+    this.listenTo(channel, 'message', ({ category, resource, payload }) => {
+      if (!category) return;
+
+      const modelResource = category === 'ResourceCreated' ? payload.resource : resource;
+      const model = Radio.request('entities', 'get:store', modelResource);
+
       if (this.collection.get(model) || model.isFetching()) return;
 
       model.fetch().then(() => {
@@ -169,6 +174,8 @@ export default App.extend({
         if (filterKey === 'actions' && !flowStates.includes(model.getFlow().getState().id)) return;
 
         this.collection.add(model);
+
+        channel.request('add', model);
       });
     });
   },
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -126,9 +126,10 @@ export default App.extend({
   },
   beforeStart() {
     const listType = this.getState().getType();
+    this.filters = this.getState().getEntityFilter();
 
     const data = {
-      filter: this.getState().getEntityFilter(),
+      filter: this.filters,
       include: this.sortOptions.getInclude(),
     };
 
@@ -143,6 +144,8 @@ export default App.extend({
     this.filteredCollection = collection.clone();
     this.editableCollection = collection.clone();
 
+    this.subscribe();
+
     this.listenTo(this.filteredCollection, 'reset', this.showCountView);
     this.showCountView();
 
@@ -150,6 +153,20 @@ export default App.extend({
     this.toggleBulkSelect();
 
     this.showList();
+  },
+  subscribe() {
+    const channel = Radio.channel('ws');
+    const filterKey = this.getState().getType();
+
+    channel.request('subscribe', this.collection.models, { filters: { [filterKey]: this.filters } });
+
+    this.listenTo(channel, 'message', (message, model) => {
+      if (this.collection.get(model)) return;
+
+      model.fetch().then(() => {
+        this.collection.add(model);
+      });
+    });
   },
   // NOTE: Shows views dependent on getState().getType()
   showTypeViews() {
@@ -210,7 +227,7 @@ export default App.extend({
     const filtersState = this.getFiltersState();
 
     const sidebarApp = this.getChildApp('filtersSidebar');
-    
+
     Radio.request('sidebar', 'start', sidebarApp, { filtersState });
   },
   toggleBulkSelect() {
--- a/src/js/base/model.js
+++ b/src/js/base/model.js
@@ -1,4 +1,4 @@
-import { extend, isEmpty, isFunction, pick, reduce, result } from 'underscore';
+import { each, extend, isEmpty, isFunction, pick, reduce, result } from 'underscore';
 import Backbone from 'backbone';
 import dayjs from 'dayjs';
 import JsonApiMixin from './jsonapi-mixin';
@@ -41,6 +41,8 @@ export default Backbone.Model.extend(extend({
     // Resolve with entity if successful
     return fetcher.then(response => {
       this._isFetching = false;
+
+      each(this._notifications, this._handleMessage, this);
 
       if (!response || response.ok) return this;
 
@@ -105,12 +107,30 @@ export default Backbone.Model.extend(extend({
 
     return isFunction(messages[category]) ? messages[category] : this[messages[category]];
   },
-  handleMessage({ category, resource, author, payload }) {
-    payload.attributes = extend({}, payload.attributes, { updated_at: dayjs.utc().format() });
+  queueMessage(data) {
+    const category = data.category;
 
+    if (!this._notifications) {
+      this._notifications = { [category]: data };
+      return;
+    }
+
+    this._notifications[category] = data;
+  },
+  _handleMessage({ category, resource, author, payload }) {
     const handler = this._getMessageHandler(category);
     if (handler) handler.call(this, payload, { category, resource, author });
 
     this.trigger('message', { category, resource, author, payload });
+  },
+  handleMessage({ category, resource, author, payload }) {
+    payload.attributes = extend({}, payload.attributes, { updated_at: dayjs.utc().format() });
+
+    if (this.isFetching()) {
+      this.queueMessage({ category, resource, author, payload });
+      return;
+    }
+
+    this._handleMessage({ category, resource, author, payload });
   },
 }, JsonApiMixin));
--- a/src/js/base/model.js
+++ b/src/js/base/model.js
@@ -39,15 +39,21 @@ export default Backbone.Model.extend(extend({
     this._isFetching = true;
 
     // Resolve with entity if successful
-    return fetcher.then(response => {
-      this._isFetching = false;
+    return fetcher
+      .then(response => {
+        this._isFetching = false;
 
-      each(this._notifications, this._handleMessage, this);
+        each(this._notifications, this._handleMessage, this);
 
-      if (!response || response.ok) return this;
+        if (!response || response.ok) return this;
 
-      return response;
-    });
+        return response;
+      })
+      .catch(error => {
+        this._isFetching = false;
+
+        throw error;
+      });
   },
   isFetching() {
     return this._isFetching;
--- a/src/js/base/model.js
+++ b/src/js/base/model.js
@@ -36,12 +36,19 @@ export default Backbone.Model.extend(extend({
     // Model fetches default to aborting.
     const fetcher = Backbone.Model.prototype.fetch.call(this, extend({ abort: true }, options));
 
+    this._isFetching = true;
+
     // Resolve with entity if successful
     return fetcher.then(response => {
+      this._isFetching = false;
+
       if (!response || response.ok) return this;
 
       return response;
     });
+  },
+  isFetching() {
+    return this._isFetching;
   },
   parse(response) {
     /* istanbul ignore if */
--- a/src/js/services/ws.js
+++ b/src/js/services/ws.js
@@ -103,6 +103,9 @@ export default App.extend({
   },
 
   onMessage(event) {
+    /* istanbul ignore next: Can't test this bref functionality in node websockets */
+    if (!event.data) return;
+
     let model;
 
     const channel = this.getChannel();
--- a/src/js/services/ws.js
+++ b/src/js/services/ws.js
@@ -112,7 +112,7 @@ export default App.extend({
 
     const data = JSON.parse(event.data);
 
-    if (data.name === 'pong') return;
+    if (!data.category || data.name === 'pong') return;
 
     if (data.resource) {
       model = Radio.request('entities', 'get:store', data.resource);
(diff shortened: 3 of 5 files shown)


src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -161,14 +161,20 @@ export default App.extend({
 
     channel.request('subscribe', this.collection.models, { filters: { [filterKey]: this.filters } });
 
-    this.listenTo(channel, 'message', (message, model) => {
+    this.listenTo(channel, 'message', ({ category, payload }) => {
+      if (category !== 'ResourceCreated') return;
+
+      const model = Radio.request('entities', 'get:store', payload.resource);
```
nmajor25: I wanted to do this ☝️ in the `onMessage()` of the `/js/services/ws.js` [file](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/services/ws.js#L101-L116).

Specifically I wanted to do something like:

```js
if (data.category === 'ResourceCreated') {
  model = Radio.request('entities', 'get:store', data.payload.resource);
  channel.trigger('message', data, model);
  return;
}
```

But that would break our handling of the `ResourceCreated` ws notification for actions on the patient flow page.

Both of these `ResourceCreated`'s have both a `data.resource` and `data.payload.resource`. So I'm not sure how to differentiate between the two in that `onMessage()` function.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -161,14 +161,20 @@ export default App.extend({
...
+
       if (this.collection.get(model) || model.isFetching()) return;
 
       model.fetch().then(() => {
         // notifications do not fully filter our action flow_states
         if (filterKey === 'actions' && !flowStates.includes(model.getFlow().getState().id)) return;
 
         this.collection.add(model);
+
+        channel.request('add', model);
```
nmajor25, line 177: Tested this locally and it worked well 👍

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -161,14 +161,20 @@ export default App.extend({
...
+
+      const model = Radio.request('entities', 'get:store', payload.resource);
+
       if (this.collection.get(model) || model.isFetching()) return;
 
       model.fetch().then(() => {
         // notifications do not fully filter our action flow_states
         if (filterKey === 'actions' && !flowStates.includes(model.getFlow().getState().id)) return;
 
         this.collection.add(model);
```
nmajor25, line 175: Do we need to do anything with the `this.filteredCollection` and `this.editableCollection` here?

I haven't looked into the details yet to answer that question.

src/js/base/model.js
```diff
@@ -105,7 +105,22 @@ export default Backbone.Model.extend(extend({
 
     return isFunction(messages[category]) ? messages[category] : this[messages[category]];
   },
-  handleMessage({ category, resource, author, payload }) {
+  addMessageToQueue(data) {
+    if (!this.notifications) {
+      this.notifications = [data];
+      return;
+    };
```
nmajor25, line 112: Is there a way to set this automatically set `model.notifications: []` initially for each model? Do we want to do that?

If we did that we wouldn't have to check if `this.notifications` exists here. And just do the `push()`.

nmajor25: And wasn't sure what to use for naming. But went with `model.notifications` for now.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -161,14 +161,21 @@ export default App.extend({
 
     channel.request('subscribe', this.collection.models, { filters: { [filterKey]: this.filters } });
 
-    this.listenTo(channel, 'message', (message, model) => {
+    this.listenTo(channel, 'message', ({ category, resource, payload }) => {
+      if (!category) return;
```
nmajor25: In my local dev environment I'm getting a ws notification that contains a lambda error. When that happens this `listenTo` is called with `category`, `resource`, and `payload` as `undefined`.

I don't know if this is something that could happen in production or not.

Screenshot 2025-02-24 at 4 50 48 PM

src/js/base/model.js
```diff
@@ -105,7 +105,22 @@ export default Backbone.Model.extend(extend({
 
     return isFunction(messages[category]) ? messages[category] : this[messages[category]];
   },
-  handleMessage({ category, resource, author, payload }) {
+  queueMessage(data) {
+    if (!this.notifications) {
+      this.notifications = [data];
+      return;
+    }
```
nmajor25: Is there a way to set `model.notifications: []` initially for each model? Do we want to do that?

If we did that we wouldn't have to check if `this.notifications` exists here. And just do the `push()`.

nmajor25: And wasn't sure what to use for naming. But went with `model.notifications` for now.

src/js/services/ws.js
```diff
@@ -99,6 +103,9 @@ export default App.extend({
   },
 
   onMessage(event) {
+    /* istanbul ignore next: Can't test this bref functionality in node websockets */
+    if (!event.data) return;
+
```
paulfalgout, line 108: I think we should handle `event.data.error` near the pong return.. such that we trigger `error` instead of `message` and not have a `if (!category)` check in the message handler.

src/js/base/model.js
```diff
@@ -98,7 +105,22 @@ export default Backbone.Model.extend(extend({
...
+
+    this.notifications.push(data);
+  },
+  handleMessage(data) {
+    if (this.isFetching()) {
```
paulfalgout: I think extending the attributes with the updated_at should happen here so that it's stored with the queue.. that timestamp is when the action occurred, not when it was processed.

src/js/base/model.js
paulfalgout: there is not a good place to instantiate this, but it should probably be `_notifications` and it should be `{}`. where `this.notifications['StateChanged']` would always be the latest `StateChanged`

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -151,6 +155,36 @@ export default App.extend({
...
+      if (this.collection.get(model) || model.isFetching()) return;
+
+      model.fetch().then(() => {
+        // notifications do not fully filter our action flow_states
+        if (filterKey === 'actions' && !flowStates.includes(model.getFlow().getState().id)) return;
+
+        this.collection.add(model);
+        channel.request('add', model);
+
+        if (model.notifications) {
```
paulfalgout: we should either entirely handle this inside the base model, or entirely handle it here..  perhaps the base model is better..  I can't think of why we wouldn't always want this to happen

src/js/base/model.js
```diff
@@ -36,13 +36,20 @@ export default Backbone.Model.extend(extend({
     // Model fetches default to aborting.
     const fetcher = Backbone.Model.prototype.fetch.call(this, extend({ abort: true }, options));
 
+    this._isFetching = true;
+
     // Resolve with entity if successful
     return fetcher.then(response => {
+      this._isFetching = false;
+
```
paulfalgout, line 45: we should likely process the queue here.  and it's likely `_.each(this._notifications, this._handleMessage);` where `_handleMessage` is after the isFetching return

src/js/apps/patients/worklist/worklist_app.js
nmajor25: I think it would work to do this in the base model.

But does `this.collection.add(model)` need to happen before we handle the queued notifications? Or does that order not matter?

For example if the order of things happened like this, would the correct flow.state show in the UI?

1. A `StateChanged` notification is queued while a new flow is fetched.
2. Flow fetch finishes.
3. Queued `StateChanged` ws notification is handled and state is updated on the flow model.
4. Flow model is added to the worklist collection via `this.collection.add(model)`, which makes it show in the UI.

Does the order of 3 & 4 matter? My assumptions is that it doesn't matter. Either way the flow's state should show correctly in the UI.

paulfalgout: better to do data before render

src/js/base/model.js
nmajor25: `_notifications` naming makes sense.

But why an object `{}` instead of an array `[{}]`?

src/js/apps/patients/worklist/worklist_app.js
nmajor25: Yep 👍

src/js/base/model.js
paulfalgout: because we only need to run the most recent of a particular `category`

nmajor25: Ah okay, right. Using an object we can just replace the previous notification in the queue if one already exists with the same notification category (`StateChanged`, `OwnerChanged`, etc.).

src/js/base/model.js
```diff
@@ -36,13 +36,26 @@ export default Backbone.Model.extend(extend({
...
+      if (this._notifications) {
```
paulfalgout: no need for the if..  I think `each(this._notifications, this._handleMessage);` is fine here.

src/js/apps/patients/worklist/worklist_app.js
paulfalgout: we shouldn't need the `if` here.  should be able to handle those errors prior to triggering the message

nmajor25: I'm looking at that currently. Would be in the next commit. I have a `event.data.category` check added to the `onMessage()` [here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/services/ws.js#L101C3-L101C12). 

So at the top of that function it would be:

```js
if (!event.data || !event.data.category) return;
```

But still need to test it locally and think about it a little more.

src/js/base/model.js
nmajor25: Yep, not needed 👍

src/js/base/model.js
paulfalgout: why `, data => { this._handleMessage(data) });` over `, this._handleMessage);`?

src/js/base/model.js
```diff
@@ -98,12 +109,30 @@ export default Backbone.Model.extend(extend({
 
     return isFunction(messages[category]) ? messages[category] : this[messages[category]];
   },
-  handleMessage({ category, resource, author, payload }) {
-    payload.attributes = extend({}, payload.attributes, { updated_at: dayjs.utc().format() });
+  queueMessage(data) {
```
paulfalgout, line 116: So this function looks complex to me, but I get why.. but we could start using some of the newer operators to make this a bit simpler:

```js
queueMessage(data) {
 return (this._notifications ||= {})[data.category] = data;
}
```

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -151,6 +154,29 @@ export default App.extend({
...
+
+    channel.request('subscribe', this.collection.models, { filters: { [filterKey]: this.filters } });
+
+    this.listenTo(channel, 'message', ({ category, resource, payload }) => {
+      const modelResource = category === 'ResourceCreated' ? payload.resource : resource;
+      const model = Radio.request('entities', 'get:store', modelResource);
+
+      if (this.collection.get(model) || model.isFetching()) return;
+
+      model.fetch().then(() => {
```
paulfalgout: I think this'll be:

```
const include = [
      'program-action.program',
      'flow.program-flow.program',
    ].join()

const data = {
  fields: { flows: ['name', 'state'] },
  include,
};

model.fetch({ data }).then(...
```

Honestly that's probably enough logic to move it to a `Radio.request('entities', 'fetch:action..`

Or maybe just add the fields to `'fetch:actions:model'` and use that over model.fetch here.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -151,6 +154,29 @@ export default App.extend({
 
     this.showList();
   },
+  subscribe() {
```
paulfalgout, line 157: Alright I think this whole function needs refactoring..

we should probably just split based on `filterKey` and hardcode types for this one..

also somewhere we can call break this up into another function
```
subscribe() {
  this.listenTo(... () => {
    fetch().then(() => {. // <- one or both of these could be a separate handler..
```

But I think we still want to check the store for the resource.. but if we need to fetch it, as mentioned below.. switch to either the action or flow entity request rather than adding the fetch specifications here for include/fields.

src/js/base/model.js
nmajor25: I was getting an error where `this` is `undefined`, is why I switched to that.

`each(this._notifications, this._handleMessage);` results in that error. But I just tested this and it works:       `each(this._notifications, () => { this._handleMessage });`.

paulfalgout: ah yep, set the context `each(this._notifications, this._handleMessage, this);`

src/js/base/model.js
nmajor25: That's pretty cool, didn't know `||=` existed until now 👍 

But `vite-plugin-eslint` doesn't support it. It's using eslint `v8.21.0` ([link](https://github.com/gxmari007/vite-plugin-eslint/blob/main/package.json#L50)). Eslint added support logical assignment operators in `v8.24.0` ([source](https://eslint.org/docs/latest/rules/logical-assignment-operators#version)).

That package's repo seems to be abandoned, so we probably want to switch to something different.

src/js/base/model.js
nmajor25: Nice that works, updated to that.

src/js/apps/patients/worklist/worklist_app.js
nmajor25: Great thanks, I just got this working in my local dev environment. 

For now I'm just doing the following:

```js
  // js/apps/patients/worklist/worklist_app.js

  _getWsFetchOptions(filterKey) {
    if (filterKey === 'flows') return {};

    return {
      data: {
        fields: { flows: ['name', 'state'] },
        include: ['program-action.program', 'flow.program-flow.program'].join(),
      },
    };
  },
  subscribe() {

    . . .

    const options = this._getWsFetchOptions(filterKey);

    model.fetch(options).then(() => {

       . . .

    })
  },
```

nmajor25: Having this working now allows me to finish creating the initial tests for the action version of the worklist. After that I'll come back and do some refactoring on this and other parts too.

test/integration/patients/worklist/worklist.js
nmajor25, line 1: Summary of what I'm testing here.

**Flow list notifications**:

* Load owned-by worklist with one flow
    * Test each ws notification category that applies on preloaded flow
        * NameChanged
            * Verify updated_at date
        * OwnerChanged
        * StateChanged
    * Add flow to list via `ResourceCreated` ws notification (w/ trailing `StateChanged` msg)
        * Verify included is not in request URL string
        * Verify flow count
        * Verify flow was added to list
        * Verify trailing `StateChanged` was queued and processed after model fetch finishes
        * Verify new flow is subscribed to future ws notifications
    * Add new flow to list via `StateChanged` ws notification
    * Add new flow to list via `OwnerChanged` ws notification

**Action list notifications**:
* Load owned-by worklist with one action
    * Test each ws notification category that applies on preloaded action
        * NameChanged
            * Verify updated_at date
        * OwnerChanged
        * ActionDueChanged
        * DetailsChanged
        * StateChanged
        * CommentAdded
        * AttachmentAdded
        * SharingUpdated
    * Add action to list via `ResourceCreated` ws notification (w/ trailing `StateChanged` msg)
        * Verify included  data is in request URL string
        * Verify action count
        * Verify action was added to list
        * Verify trailing `StateChanged` was queued and processed after model fetch finishes
        * Verify new action is subscribed to future ws notifications
    * Add new action to list via `StateChanged` ws notification
    * Add new action to list via `OwnerChanged` ws notification
    * Handle ws notification for new action with a flow state not in the current filters
      * In this scenario, the action should not be added to list

src/js/base/model.js
paulfalgout: That's just the dev dependency.  We're using 9 and it's a peer dependency https://github.com/gxmari007/vite-plugin-eslint/blob/main/package.json#L60

paulfalgout: FWIW it looks like the vite community doesn't run eslint as part of the build process.  They're using it in pre-commit pre-build hooks..  I find that mildly annoying

test/integration/patients/worklist/worklist.js
```diff
@@ -303,6 +303,287 @@ context('worklist page', function() {
       .wait('@routeFlows');
   });
 
+  specify('flow list - socket notifications', function() {
+    localStorage.setItem(`owned-by_11111_${ workspaceOne.id }-${ STATE_VERSION }`, JSON.stringify({
+      id: 'owned-by',
+      listType: 'flows',
+      flowsSortId: 'sortCreatedDesc',
+      clinicianId: '11111',
```
paulfalgout: Don't love using new string ids.. is this `getCurrentClinician().id`?

test/integration/patients/worklist/worklist.js
```diff
@@ -303,6 +303,287 @@ context('worklist page', function() {
...
+      });
+
+    cy.sendWs({
+      category: 'ResourceCreated',
+      resource: {
+        type: testPatient1.type,
+        id: testPatient1.id,
+      },
+      payload: {
+        resource: {
```
paulfalgout, line 467: I do badly want destructing within an object like
```s
resource: { id, type } = testPatient1, 
```
but alas none of these types of proposals go anywhere right now in TC39 😿 

we could make some sort of generic helper. `function foo({ id, type ]) => { id, type };`
`resource: foo(testNewSocketFlow)`

thought for another day

test/integration/patients/worklist/worklist.js
```diff
@@ -848,6 +1129,445 @@ context('worklist page', function() {
...
+        due_date: testDate(),
+        due_time: '06:00:00',
+        duration: 10,
+        updated_at: testTsSubtract(4),
+        created_at: testTsSubtract(4),
+      },
+      relationships: {
+        state: getRelationship(stateTodo),
+        flow: getRelationship(testFlow),
+        files: getRelationship([]),
```
paulfalgout: it's ok to do so, but I suspect we don't need to set a lot of this for the tests?  details?  due_time? duration. files relationship?  If they're just random values from `getAction` does that potentially cause flake?  Ideally were setting whatever minimum test data we need and letting `getAction` do the rest

test/integration/patients/worklist/worklist.js
```diff
@@ -848,6 +1129,445 @@ context('worklist page', function() {
...
+      })
+      .routeAction(fx => {
+        fx.data = testSocketAction;
+
+        return fx;
+      })
+      .routeFormByAction()
+      .routeFormDefinition()
+      .routeLatestFormResponse()
+      .visitOnClock('/worklist/owned-by', { now: testTs() })
```
paulfalgout, line 1259: the visitOnClock is to set the time to now or why is this needed?

paulfalgout (commented):
looking pretty good.

src/js/base/model.js
shadowhand: FWIW backend runs static analysis (including lint) on commit and on test (CI).

test/integration/patients/worklist/worklist.js
nmajor25: It's to avoid flakiness when checking that the `updated_at` timestamp is updated correctly in the UI.

First introduced in this commit: https://github.com/RoundingWell/care-ops-frontend/commit/8eb7d817796c7852cc0ce7842fc676c38ae23017.

test/integration/patients/worklist/worklist.js
nmajor25: Agreed, I cleaned up a number of these: https://github.com/RoundingWell/care-ops-frontend/compare/003a41f45cc3b2abdda8b3aa33189122edb965a0..54708cca95f3c34880bfa2975ca936e7e887260c

test/integration/patients/worklist/worklist.js
nmajor25: Ya that would be pretty nice 👍

test/integration/patients/worklist/worklist.js
nmajor25: Yep, that's the current clinician (i.e. `getCurrentClinician()`). I updated the tests to reference that directly instead of using `11111`.

src/js/apps/patients/worklist/worklist_app.js
nmajor25: I took a crack at refactoring this in https://github.com/RoundingWell/care-ops-frontend/pull/1380/commits/836c7c01aa9d8cd6d86bfbacb59c0977853f5302, let me know what you think.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -151,6 +154,51 @@ export default App.extend({
...
+    const flowStates = this.getState('flowStates');
+
+    const fetchAction = Radio.request('entities', 'fetch:actions:model', model.id);
+
+    fetchAction.then(() => {
+      if (!flowStates.includes(model.getFlow().getState().id)) return;
+
+      this.collection.add(model);
+
+      Radio.channel('ws').request('add', model);
```
paulfalgout: `Radio.request('ws', 'add',...` here and above

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -151,6 +154,51 @@ export default App.extend({
...
+      const model = Radio.request('entities', 'get:store', modelResource);
+
+      if (this.collection.get(model) || model.isFetching()) return;
+
+      if (filterKey === 'flows') {
+        this._fetchFlow(model);
+        return;
+      }
+
+      if (filterKey === 'actions') {
```
paulfalgout: I think you'll need to remove the `if` here and have a default, otherwise we'll have an uncovered `else` branch

paulfalgout (commented):
couple of small things

src/js/apps/patients/worklist/worklist_app.js
nmajor25: Updated 👍

src/js/apps/patients/worklist/worklist_app.js
nmajor25: Updated 👍

nmajor25:
@paulfalgout coveralls coverage looks good 👍

src/js/base/model.js
```diff
@@ -36,13 +36,22 @@ export default Backbone.Model.extend(extend({
     // Model fetches default to aborting.
     const fetcher = Backbone.Model.prototype.fetch.call(this, extend({ abort: true }, options));
 
+    this._isFetching = true;
+
     // Resolve with entity if successful
     return fetcher.then(response => {
+      this._isFetching = false;
+
```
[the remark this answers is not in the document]
nmajor25: Correct me if I'm wrong here @paulfalgout, but I think we'd want `this._isFetching` to remain `true` in the scenario of a fetch error. Since we wouldn't want the model's queued ws notifications to be processed. So I think the code we current have is good.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -151,6 +154,51 @@ export default App.extend({
...
+        this._fetchFlow(model);
+
+        return;
+      }
+
+      this._fetchAction(model);
+
+      return;
+    });
+  },
```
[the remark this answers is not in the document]
nmajor25: In `src/js/apps/patients/worklist/worklist_app.js` we now check that the data contains a `category`. And we don't handle that web socket notification if it doesn't have that.

If there's no category, that's a good indication that the data isn't going to be a valid web socket notification.

src/js/base/model.js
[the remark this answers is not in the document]
paulfalgout: hmm no I think we should add this catch with the rethrow of the error.  It is more correct for the state of the base model, regardless of how it affects us downstream.  if a fetch fails though, we're not necessarily doing anyhing in order to process the queue correctly anyhow.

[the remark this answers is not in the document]
nmajor25: That's been added in commit 0222306fd0fa389b896351ffa29ceb1c20e35305.

Nice catch @coderabbitai 👍

paulfalgout (commented):
approved on my end

nmajor25 (approved):

paulfalgout:
The error is coveralls 504ing.

PR #1381, opened 2025-01-13
Update Vite version to v6

nmajor25:
Shortcut Story ID: [sc-57765]

## Summary by CodeRabbit

- **Dependencies**
  - Updated multiple package versions, including Vite, Cypress, PWA plugin, and Stylelint
  - Added `js-yaml` dependency
  - Removed `@modyfi/vite-plugin-yaml`

- **Configuration**
  - Introduced a custom YAML plugin for Vite
  - Updated Vite configuration to use the new local YAML plugin
  - Added legacy API option for SCSS preprocessor

- **Styles**
  - Removed the overrides stylesheet for form elements, eliminating specific browser adjustments for consistent styling.
  - Removed import statements for the overrides stylesheet in core styles.


src/scss/core/_overrides.scss
```diff
@@ -56,7 +56,7 @@ select {
 // -------------------------------------
 
 // remove bg image on selects from Firefox
-@-moz-document url-prefix('') {
+@document url-prefix('') {
```
nmajor25: Something is updating this automatically now in my ide. I think it's the new version of `vite-plugin-stylelint`? I'm not entirely sure.

nmajor25: The [mdn web doc](https://developer.mozilla.org/en-US/docs/Web/CSS/@document) for this rule seems to indicate that `@document` should be used. Which is only compatible with firefox: https://developer.mozilla.org/en-US/docs/Web/CSS/@document#browser_compatibility.

But I don't understand the context around this css. Looks like y'all added it back in the day on the initial build of the app.

config/vite-plugin-yaml.js
```diff
@@ -0,0 +1,24 @@
...
+      if (yamlExtension.test(id)) {
+        const yamlData = load(code, {
+          filename: id,
+        });
+
+        return {
+          code: `const data = ${ toSource(yamlData) };\nexport default data;`,
+          map: { mappings: '' },
+        };
+      }
```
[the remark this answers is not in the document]
nmajor25: Yep, this is probably a good idea 👍 we have similar `try...catch` error handling for our handlebars vite plugin.

src/scss/core/_overrides.scss
paulfalgout: this might not be used at all anymore.. I think it was related to an old dropdown plugin, but I'm not sure.

config/vite-plugin-yaml.js
```diff
@@ -0,0 +1,24 @@
+import { load } from 'js-yaml';
+import toSource from 'tosource';
```
paulfalgout: Let's remove `toSource` and just use `JSON.stringify`

src/scss/core/_overrides.scss
paulfalgout: I actually believe we can safely remove this entire file.  Had chatgpt analyze it along side our browserlist and it seems like it's all obsolete.

paulfalgout (changes requested):

config/vite-plugin-yaml.js
```diff
@@ -7,17 +7,20 @@ export default function() {
...
+            code: `const data = ${ JSON.stringify(yamlData) };\nexport default data;`,
+            map: { mappings: '' },
+          };
+        } catch (exception) {
+          console.error(`${ id } errored during yaml processing: `, exception);
+          return null;
+        }
       }
-
-      return null;
```
nmajor25: Do we need this `return null;` if `(yamlExtension.test(id))` is `false`? I think we are good without it. The handlebars vite plugin doesn't have it: https://github.com/RoundingWell/care-ops-frontend/blob/develop/config/vite-plugin-handlebars-loader.js

paulfalgout: ooh.. I think it does need it.. we should add that to the handlebars one too

nmajor25: Added 👍

paulfalgout (approved):

PR #1382, opened 2025-01-15
Sass deprecation warning fixes

nmajor25:
Shortcut Story ID: [sc-57510]

Original PR is: https://github.com/RoundingWell/care-ops-frontend/pull/1379. I messed up a rebase so it was recreated it in this PR. There's some discussion in that PR that will be carried over here.

Main deprecation updates:

- Switch from deprecated `map-get()` to `map.get()`.
- Switch from using deprecated `@import`'s to `@use` & `@forward` instead.
  - More info on the deprecation [here](https://sass-lang.com/blog/import-is-deprecated/).

### DIFF
--- a/src/js/components/datepicker/datepicker.scss
+++ b/src/js/components/datepicker/datepicker.scss
@@ -1,3 +1,5 @@
+@use 'sass:math';
+
 .datepicker {
   font-size: 12px;
   user-select: none;
--- a/src/scss/provider-core.scss
+++ b/src/scss/provider-core.scss
@@ -1,12 +1,12 @@
 // Core Styles
 // These are the core styles shared between mobile and desktop
-@import 'core/reset';
-@import 'core/base';
-@import 'core/flex';
-@import 'core/fonts';
-@import 'core/icon';
-@import 'core/layout';
-@import 'core/scrollbar';
-@import 'core/space';
-@import 'core/text';
-@import 'core/typography';
+@forward 'core/reset';
+@forward 'core/base';
+@forward 'core/flex';
+@forward 'core/fonts';
+@forward 'core/icon';
+@forward 'core/layout';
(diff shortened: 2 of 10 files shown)


vite.config.js
```diff
@@ -33,9 +33,7 @@ const css = {
       // remove this when we switch sass to use modern js api
       api: 'legacy',
       additionalData: `
-        @use 'sass:math';
-        @use 'sass:color';
-        @import 'src/scss/provider-variables.scss';
+        @use 'src/scss/provider-variables.scss' as *;
```
nmajor25, line 36: From what I'm reading in the sass docs, doing a `as *` is fine, as long as it's a stylesheets written by us: https://sass-lang.com/documentation/at-rules/use/#choosing-a-namespace. Since we are then in control of the namespacing.

src/scss/provider-core.scss
nmajor25, line 1: Not sure where to go with this file. I've tried a few alternatives to this that I couldn't get to work.

1. Put them all in the same namespace and then export them like this:

```scss
@use '../provider-variables' as *;

@use 'core/reset';
...
@use 'core/typography';

@forward 'core/base';
...
@forward 'core/typography';
```

2. Switch to a `provider-core-styles.js` file that is imported in to the index of the app.

src/js/components/datepicker/datepicker.scss
```diff
@@ -1,3 +1,5 @@
+@use 'sass:math';
```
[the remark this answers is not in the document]
nmajor25: That's added to the file by vite when the application is built. Via the `additionalData` scss option in the `vite.config.js` file.

paulfalgout (approved):
I still dislike what sass is doing to us, but I can't find a better alternative that's future-proof.

PR #1385, opened 2025-01-23
Show comment count on action list items

nmajor25:
Shortcut Story ID: [sc-58537]

This is based on the `action.relationships.comments` array. The Icon is shown if `length > 0` and hidden if `length === 0`.

The visibility of the icon and it's count will change when an action's comments change via manual user action in the UI or via websocket nofitications.

Comment icon is shown or incremented when a comment is added. And hidden or decremented when a comment is deleted.

Pages where icon is added:

* Worklist
* Patient flow
* Patient dashboard
* Patient archive

Screenshot example of new icon:

Screenshot 2025-01-23 at 5 26 39 PM

## Summary by CodeRabbit

- **New Features**
  - Introduced enhanced comment management across multiple views, with real-time updates for added and removed comments.
  - New comment icon and count display have been integrated to provide clear visual feedback on actions.

- **Style**
  - Improved spacing and layout adjustments for attachments and comment icons across patient, dashboard, worklist, and sidebar views.

- **Tests**
  - Updated integration tests to validate the proper display and behavior of comments and associated icons.
  - Added new tests to ensure comment icons reflect the correct state in action lists.

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -201,6 +201,7 @@ export default App.extend(extend({
     model.set({ created_at: dayjs.utc().format() });
 
     model.save().then(() => {
+      this.action.addComment(model);
       Radio.request('ws', 'add', model);
     });
 
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -174,12 +174,17 @@ export default App.extend(extend({
     this.action.save({ sharing: ACTION_SHARING.PENDING });
   },
   showActivity() {
-    this.showContentView('activity', new ActivitiesView({
+    const activitiesView = new ActivitiesView({
       collection: this.activityCollection,
       model: this.action,
-    }));
+    });
     const createdEvent = this.activityCollection.find({ event_type: 'ActionCreated' });
 
+    this.listenTo(activitiesView, {
+      'remove:comment': this.onRemoveComment,
+    });
+
+    this.showContentView('activity', activitiesView);
     this.showFooterView('timestamps', new TimestampsView({ model: this.action, createdEvent }));
   },
   showNewCommentForm() {
@@ -212,6 +217,11 @@ export default App.extend(extend({
   },
   onCancelNewComment() {
     this.showNewCommentForm();
+  },
+  onRemoveComment(model) {
+    this.action.removeComment(model);
+
+    Radio.request('ws', 'unsubscribe', model);
   },
   showAttachments() {
     const canUploadAttachments = !!Radio.request('settings', 'get', 'upload_attachments') && this.action.hasAllowedUploads();
--- a/src/js/views/patients/patient/patient.scss
+++ b/src/js/views/patients/patient/patient.scss
@@ -145,7 +145,7 @@
 
 .patient__action-attachment,
 .patient__action-comment {
-  color: #999;
+  color: $black-60;
   display: inline-block;
   vertical-align: middle;
   width: 16px;
--- a/src/js/views/patients/patient/patient.scss
+++ b/src/js/views/patients/patient/patient.scss
@@ -140,14 +140,22 @@
 }
 
 .patient__list-meta {
-  width: 510px;
+  width: 562px;
 }
 
-.patient__action-attachment {
+.patient__action-attachment,
+.patient__action-comment {
   color: #999;
   display: inline-block;
-  margin-left: 8px;
   vertical-align: middle;
+  width: 16px;
+}
+
+.patient__action-comment-text {
+  font-size: 12px;
+  font-weight: 600;
+  margin-left: 2px;
+  vertical-align: top;
 }
 
 .patient__flow-icon .icon {
--- a/src/js/views/patients/patient/patient.scss
+++ b/src/js/views/patients/patient/patient.scss
@@ -143,15 +143,14 @@
   width: 562px;
 }
 
-.patient__action-attachment,
-.patient__action-comment {
+.patient__action-display-icon {
   color: $black-60;
   display: inline-block;
   vertical-align: middle;
   width: 16px;
 }
 
-.patient__action-comment-text {
+.patient__action-display-icon-text {
   font-size: 12px;
   font-weight: 600;
   margin-left: 2px;
--- a/src/js/views/patients/worklist/worklist-list.scss
+++ b/src/js/views/patients/worklist/worklist-list.scss
@@ -24,7 +24,7 @@
 }
 
 .worklist-list__display-icon {
-  color: #999;
+  color: $black-60;
   display: inline-block;
   vertical-align: middle;
   width: 16px;
--- a/src/js/views/patients/worklist/worklist-list.scss
+++ b/src/js/views/patients/worklist/worklist-list.scss
@@ -30,6 +30,13 @@
   width: 16px;
 }
 
+.worklist-list__display-icon-text {
+  font-size: 12px;
+  font-weight: 600;
+  margin-left: 2px;
+  vertical-align: top;
+}
+
 .work-list__item:hover {
   .fa-square {
     color: $black-20;
@@ -73,22 +80,22 @@
 
 .worklist-list__item-name,
 .worklist-list__name-header {
-  width: 22%;
+  width: 17%;
 }
 
 .worklist-list__meta {
   text-overflow: inherit;
-  width: 43%;
+  width: 48%;
 }
 
 .worklist-list__action-meta-header {
   padding-left: 40px;
-  width: 43%;
+  width: 48%;
 }
 
 .worklist-list__flow-meta-header {
   padding-left: 20px;
-  width: 43%;
+  width: 48%;
 }
 
 .worklist-list__count {
--- a/test/integration/patients/patient/archive.js
+++ b/test/integration/patients/patient/archive.js
@@ -1,5 +1,4 @@
 import dayjs from 'dayjs';
-import { v4 as uuid } from 'uuid';
 
 import { testTs, testTsSubtract } from 'helpers/test-timestamp';
 import { testDate, testDateSubtract } from 'helpers/test-date';
@@ -14,6 +13,7 @@ import { stateDone, stateInProgress, stateTodo } from 'support/api/states';
 import { getClinician, getCurrentClinician } from 'support/api/clinicians';
 import { roleNoFilterEmployee, roleTeamEmployee } from 'support/api/roles';
 import { teamCoordinator, teamNurse } from 'support/api/teams';
+import { getComment } from 'support/api/comments';
 
 context('patient archive page', function() {
   const currentClinican = getCurrentClinician({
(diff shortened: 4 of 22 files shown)


src/js/views/patients/patient/patient.scss
```diff
@@ -140,14 +140,22 @@
...
 
 .patient__list-meta {
-  width: 510px;
+  width: 562px;
 }
 
-.patient__action-attachment {
+.patient__action-attachment,
+.patient__action-comment {
   color: #999;
```
nmajor25: Can use `$black-60` for these I believe...

src/js/views/patients/patient/patient.scss
```diff
@@ -140,14 +140,22 @@
 }
 
 .patient__list-meta {
-  width: 510px;
+  width: 562px;
```
nmajor25, line 143: I'm going to play around with these widths a little more on the patient flow page. When this meta section is at it's maximum potential width, things look a little funky to me.

src/js/views/patients/worklist/worklist-list.scss
nmajor25, line 1: We're really reaching our limits on the worklist when it comes to horizontal spacing for these action list items.

src/js/views/patients/patient/patient.scss
```diff
@@ -140,14 +140,22 @@
...
 }
 
-.patient__action-attachment {
+.patient__action-attachment,
+.patient__action-comment {
```
nmajor25: Looking at this now, I think these two could be consolidated into one class.

test/integration/patients/sidebar/action-sidebar.js
nmajor25, line 1: So in this I'm testing the following when a user manually adds or removes comments via the action sidebar:

* The comment icon is either shown or incremented when a comment is created.
* The comment icon is either hidden or decremented when a comment is deleted.

I wanted the test sequence to be this:

1. Initiate test with an action that has no comments.
2. Add comment and verify icon shows.
3. Add second comment and verify icon count increments.
4. Delete a comment and verify icon decrements.
5. Delete remaining comment and verify icon becomes hidden.

But in our tests, newly added comments remain as `.isNew() === true` because their model isn't assigned an `id` after creation. And therefore, their `Delete` button remains hidden in the UI. I couldn't figure out how to fix this.

So I switched the test sequence around to be this:

1. Initiate test with an action that has two comments.
2. Delete a comment and verify icon decrements.
3. Delete remaining comment and verify icon becomes hidden.
4. Add comment and verify icon shows.
5. Add second comment and verify icon count increments.

Doesn't seem like a big deal to me. But it may be a head scratcher for someone in the future when they look at this test.

src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -212,6 +218,11 @@ export default App.extend(extend({
   onCancelNewComment() {
     this.showNewCommentForm();
   },
+  onRemoveComment(model) {
+    this.action.removeComment(model);
+
+    Radio.request('ws', 'unsubscribe', model);
+  },
```
nmajor25, line 225: Do we need a `modal.destroy()` here? I don't think we do. It already happens in the view.

But the `onRemoveAttachment()` function in this file has it, which makes me unsure.

src/js/views/patients/patient/patient.scss
paulfalgout: could be a min-width?

test/integration/patients/sidebar/action-sidebar.js
paulfalgout: I don't understand the `isNew` issue.  We should resolve this.  The comment should get an id when the `POST` resolves.  This is in our FE code... it should `POST` with the `id` and have the `id` after the save is complete.

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -1515,6 +1515,184 @@ context('action sidebar', function() {
...
+
+    cy
+      .get('.sidebar')
+      .find('[data-comment-region]')
+      .as('postCommentRegion')
+      .find('.js-input')
+      .type('Test comment');
+
+    cy
+      .intercept('POST', '/api/actions/*/relationships/comments', {
```
paulfalgout: It's possible I suppose that the 204 is preventing the id, but I don't think it does..  we could check that though.

paulfalgout (changes requested):
Looks close

src/js/views/patients/patient/patient.scss
nmajor25: Min-width breaks the layout on smaller width browser windows. 

I think what we have is fine right now, but I'll make a note on the card for Mark so he can take a look when this is up on QA. And see if it looks good to him and we can make adjustments if needed.

src/js/views/patients/patient/patient.scss
nmajor25: Updated these to use `$black-60`.

src/js/views/patients/patient/patient.scss
nmajor25: I consolidated these into one `patient__action-display-icon` class.

test/integration/patients/sidebar/action-sidebar.js
nmajor25: Yep, I confirmed this works as expected on QA/production. It's only an issue in the tests.

test/integration/patients/sidebar/action-sidebar.js
nmajor25: The `204` code was the problem. Switched them to `201` and the response body worked 👍 thanks!

test/integration/patients/sidebar/action-sidebar.js
nmajor25: This test is no longer in reverse sequence. In a3061855c907ad21af2bebad79a434285cf43833, the `isNew()` problem is fixed and the test is refactored to be in the desired order.

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -1515,6 +1515,183 @@ context('action sidebar', function() {
...
+        statusCode: 201,
+        body: {
+          data: {
```
paulfalgout: I think ideally we use a getComment here instead of making this a pojo

test/integration/patients/worklist/worklist.js
```diff
@@ -389,6 +389,7 @@ context('worklist page', function() {
           files: getRelationship([{ id: '1' }], 'files'),
           owner: getRelationship(teamCoordinator),
           patient: getRelationship(testPatient1),
+          comments: getRelationship([{ id: uuid() }], 'comments'),
```
paulfalgout: even this ideally uses `getRelationship([getComment()])`

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -1515,6 +1515,183 @@ context('action sidebar', function() {
...
   });
 
+  specify('action comments - show/hide icon in action list items', function() {
+    const testPatient = getPatient();
+
+    const testAction = getAction({
+      relationships: {
+        patient: getRelationship(testPatient),
+        state: getRelationship(stateTodo),
+        comments: getRelationship([], 'comments'),
```
paulfalgout: I don't think `, comments'` is needed

test/integration/patients/patient/patient-flow.js
```diff
@@ -2983,6 +2990,45 @@ context('patient flow page', function() {
...
 
+    cy.sendWs({
+      category: 'CommentAdded',
+      author: getCurrentClinician().id,
+      resource: {
+        type: 'patient-actions',
+        id: testSocketAction.id,
+      },
+      payload: {
+        comment: {
```
paulfalgout, line 3004: So I even think we should be using `getComment` here instead of `uuid` and then doing 
```
comment: {
  type: testComment.type,
  id: testComment.id,
}
```
we nearly have no instances of `type: 'patient-actions'` except for in this file for ws tests..  we probably want to change any instance of `type: '`

paulfalgout (changes requested):

test/integration/patients/patient/patient-flow.js
```diff
@@ -2638,6 +2638,8 @@ context('patient flow page', function() {
     const testSocketFileId = uuid();
     const testComment = getComment();
 
+    const testClinician = getClinician({ id: uuid() });
```
nmajor25, line 2641: I guess we could just do `getCurrentClinician()` here, not sure it matters...

test/integration/patients/sidebar/action-sidebar.js
nmajor25: Correct, removed instances of these in https://github.com/RoundingWell/care-ops-frontend/pull/1385/commits/d9e8c49a663caa3d5f4c03a80da4d6105d6e6d6a.

test/integration/patients/worklist/worklist.js
nmajor25: Updated in https://github.com/RoundingWell/care-ops-frontend/pull/1385/commits/d9e8c49a663caa3d5f4c03a80da4d6105d6e6d6a.

test/integration/patients/sidebar/action-sidebar.js
nmajor25: Updated in https://github.com/RoundingWell/care-ops-frontend/pull/1385/commits/4dfeaeb0b7c7dfcac78f72243b843409a68577c3.

test/integration/patients/patient/patient-flow.js
nmajor25: Made these updates in https://github.com/RoundingWell/care-ops-frontend/pull/1385/commits/b790f093818b64d3ad19f08e35c147fedd3b44a9 & https://github.com/RoundingWell/care-ops-frontend/pull/1385/commits/da01569ecbe3351e44c519c8b27776eb6308767d.

I updated all of these `type: ''`'s in the file. With the exception of `type: 'files'`. We don't currently have a `getFile()` to make that happen. Do we want one?

test/integration/patients/patient/archive.js
```diff
@@ -56,6 +57,7 @@ context('patient archive page', function() {
               state: getRelationship(stateDone),
               form: getRelationship(testForm),
               files: getRelationship([{ id: '1' }], 'files'),
+              comments: getRelationship(getComment()),
```
paulfalgout: seems like this should either be `comments: getRelationship([])` or `comments: getRelationship([getComment()])`

paulfalgout (changes requested):
I don't think `comments: getRelationship(getComment()),` is valid.. `comments` needs to be an array

test/integration/patients/patient/archive.js
nmajor25: Yep, that's invalid. Fixed those. Not sure why that didn't cause the tests to fail though...

nmajor25: And should we add the `'comments'` type?

Like `comments: getRelationship([getComment()], 'comments')`.

paulfalgout (approved):
Cool.  I'll wait to merge until the hotfix goes out.

PR #1386, opened 2025-01-29
Add fields request to reduced schedule

paulfalgout:
When we added fields to the schedule, it looks like we missed doing it on the reduced schedule.. not the end of the world, but it used to fail silently.. now it doesn’t

Shortcut Story ID: [sc-58838]

## Summary by CodeRabbit

- **New Features**
	- Enhanced data retrieval for patient schedule actions by specifying which flow fields to include in the request.

- **Tests**
	- Updated integration test to verify new URL query parameters for flow fields.

### DIFF
--- a/src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
+++ b/src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
@@ -98,7 +98,8 @@ export default App.extend({
   },
   beforeStart() {
     const filter = this.getState().getEntityFilter();
-    return Radio.request('entities', 'fetch:actions:collection', { data: { filter } });
+    const fields = { flows: ['name', 'state'] };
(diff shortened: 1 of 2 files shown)


src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
```diff
@@ -98,7 +98,8 @@ export default App.extend({
   },
   beforeStart() {
     const filter = this.getState().getEntityFilter();
-    return Radio.request('entities', 'fetch:actions:collection', { data: { filter } });
+    const fields = { flows: ['name', 'state'] };
```
paulfalgout, line 101: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/patients/schedule/schedule_app.js#L117

The pattern we're matching is here

sofiaroundingwell (approved):

PR #1391, opened 2025-02-05
Move patient-field id to construction

paulfalgout:
We had added an id accomodation for saving new patient fields prior to having a method for it.  Fields later changed to v5 uuids and a stable id, so it makes more sense to construct them with the eventual id, so that backbone.store caches the correct model

Shortcut Story ID: [sc-58951]

## Summary by CodeRabbit

- **New Features**
  - Enhanced patient record management with more reliable patient field updates, ensuring improved data consistency for a smoother experience.

- **Refactor**
  - Streamlined the processing of patient field updates by simplifying the underlying logic, resulting in more efficient operations.

### DIFF
--- a/src/js/entities-service/entities/patient-fields.js
+++ b/src/js/entities-service/entities/patient-fields.js
@@ -1,5 +1,4 @@
 import { extend } from 'underscore';
-import { v5 as uuid } from 'uuid';
 import Store from 'backbone.store';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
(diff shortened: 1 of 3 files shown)


src/js/entities-service/entities/patient-fields.js
```diff
@@ -15,22 +14,12 @@ const _Model = BaseModel.extend({
...
-    return false;
-  },
   saveAll(attrs) {
     attrs = extend({}, this.attributes, attrs);
-    const patient = this.toRelation(attrs._patient);
-
-    // NOTE: sets the id instead of attrs.id due to how backbone's save works
-    /* istanbul ignore next: Currently not saving new fields, but would be important if we do */
-    if (!attrs.id) {
-      this.set({ id: uuid(`resource:field:${ attrs.name.toLowerCase() }`, patient.id) });
```
paulfalgout, line 30: this issue here was that `toRelation` returns `{ data: { id, type }`
I was a little surprised this wasn't caught by a test, but I missed the above ignore.  It turns out because of our model caching, setting the id really messed with tests..  but because we _can_ always determine an id with v5 uuid we don't have use for a bit of the workaround logic in here anyhow

shadowhand (approved):

nmajor25 (approved):

PR #1393, opened 2025-02-06
Outreach: switch opt-in dob to `<select>` elements

nmajor25:
Shortcut Story ID: [sc-58453]

Screenshot 2025-02-06 at 3 33 16 PM

## Summary by CodeRabbit

- **New Features**
  - Enhanced the Date of Birth input by replacing the single field with three dropdown menus (month, day, and year) to streamline data entry.
  
- **Style**
  - Updated the visual design of the dropdown controls for a cleaner and more accessible user interface.
  
- **Tests**
  - Modified test cases to validate the new dropdown selection process for the Date of Birth input, ensuring the submit button's state is correctly managed based on user selections.

### DIFF
--- a/src/js/outreach/views/opt-in.scss
+++ b/src/js/outreach/views/opt-in.scss
@@ -21,29 +21,6 @@
   width: 100%;
 }
 
-.opt-in__field-select {
-  border: 1px solid $black-90;
-  border-radius: 4px;
-  display: inline-block;
-
-  &:has(:focus-visible) {
-    outline: 1px solid color(blue);
-  }
-}
-
-.opt-in__field-select select {
-  border: none;
-  border-radius: 4px;
-  color: $black-40;
-  font-size: 16px;
-  font-weight: 500;
-  padding: 12px 20px 12px 12px;
-
-  &:focus-visible {
-    outline: 0;
-  }
-}
-
 .opt-in__disclaimer {
   font-size: 14px;
   margin-top: 32px;
--- a/src/js/outreach/views/opt-in.scss
+++ b/src/js/outreach/views/opt-in.scss
@@ -34,6 +34,10 @@
   font-size: 16px;
   font-weight: 500;
   padding: 12px 20px 12px 12px;
+
+  &:focus {
+    outline: none;
+  }
 }
 
 .opt-in__disclaimer {
--- a/src/js/outreach/views/opt-in.scss
+++ b/src/js/outreach/views/opt-in.scss
@@ -21,6 +21,21 @@
   width: 100%;
 }
 
+.opt-in__field-select {
+  border: 1px solid #E6E6E6;
+  border-radius: 4px;
+  display: inline-block;
+}
+
+.opt-in__field-select select {
+  border: none;
+  border-radius: 4px;
+  color: $black-40;
+  font-size: 16px;
+  font-weight: 500;
+  padding: 12px 20px 12px 12px;
+}
+
 .opt-in__disclaimer {
   font-size: 14px;
   margin-top: 32px;
--- a/src/js/outreach/views/opt-in.scss
+++ b/src/js/outreach/views/opt-in.scss
@@ -22,9 +22,13 @@
 }
 
 .opt-in__field-select {
-  border: 1px solid #E6E6E6;
+  border: 1px solid $black-90;
   border-radius: 4px;
   display: inline-block;
+
+  &:has(:focus-visible) {
+    outline: 1px solid color(blue);
+  }
 }
 
 .opt-in__field-select select {
@@ -35,8 +39,8 @@
   font-weight: 500;
   padding: 12px 20px 12px 12px;
 
-  &:focus {
-    outline: none;
+  &:focus-visible {
+    outline: 0;
   }
 }
 
--- a/src/js/outreach/views/opt-in_views.js
+++ b/src/js/outreach/views/opt-in_views.js
@@ -1,4 +1,5 @@
 import { range } from 'underscore';
+import dayjs from 'dayjs';
 import hbs from 'handlebars-inline-precompile';
 import { View } from 'marionette';
 import parsePhoneNumber from 'libphonenumber-js/min';
@@ -52,7 +53,7 @@ const OptInView = View.extend({
       </div>
       <div class="opt-in__field-select">
         <select name="dob-year" class="u-margin--r-8 js-dob-year">
-          <option value="Year">Year</option>
+          <option value="">Year</option>
           {{#each dobYearRange}}
             <option value={{this}}>{{this}}</option>
           {{/each}}
@@ -114,16 +115,18 @@ const OptInView = View.extend({
     this.model.set({ last_name: trim(this.ui.lastName.val()) });
   },
   onChangeDobInput() {
-    const dobMonth = this.ui.dobMonth.val();
-    const dobDay = this.ui.dobDay.val();
-    const dobYear = this.ui.dobYear.val();
+    const month = this.ui.dobMonth.val();
+    const day = this.ui.dobDay.val();
+    const year = this.ui.dobYear.val();
 
-    if (!dobMonth || !dobDay || !dobYear) {
+    if (!month || !day || !year) {
       this.model.set({ birth_date: '' });
       return;
     }
 
-    this.model.set({ birth_date: `${ dobYear }-${ dobMonth }-${ dobDay }` });
+    this.model.set({
+      birth_date: dayjs(`${ year }-${ month }-${ day }`).format('YYYY-MM-DD'),
+    });
   },
   onChangePhone() {
     const phone = parsePhoneNumber(this.ui.phone.val(), 'US');
--- a/src/js/outreach/views/opt-in_views.js
+++ b/src/js/outreach/views/opt-in_views.js
@@ -6,6 +6,7 @@ import parsePhoneNumber from 'libphonenumber-js/min';
 
 import 'scss/modules/buttons.scss';
 import 'scss/modules/forms.scss';
+import 'scss/modules/select.scss';
 
 import './opt-in.scss';
 
@@ -35,24 +36,24 @@ const OptInView = View.extend({
     </div>
     <div class="opt-in__field">
       <label class="opt-in__field-label">Your date of birth</label>
-      <div class="opt-in__field-select u-margin--r-4">
-        <select name="dob-month" class="u-margin--r-8 js-dob-month ">
+      <div class="select-wrapper u-margin--r-4">
+        <select name="dob-month" class="select u-margin--r-8 js-dob-month ">
           <option value="">Month</option>
           {{#each dobMonthRange}}
             <option value={{this}}>{{this}}</option>
           {{/each}}
         </select>
       </div>
-      <div class="opt-in__field-select u-margin--r-4">
-        <select name="dob-day" class="u-margin--r-8 js-dob-day">
+      <div class="select-wrapper u-margin--r-4">
+        <select name="dob-day" class="select u-margin--r-8 js-dob-day">
           <option value="">Day</option>
           {{#each dobDayRange}}
             <option value={{this}}>{{this}}</option>
           {{/each}}
         </select>
       </div>
-      <div class="opt-in__field-select">
-        <select name="dob-year" class="u-margin--r-8 js-dob-year">
+      <div class="select-wrapper">
+        <select name="dob-year" class="select u-margin--r-8 js-dob-year">
           <option value="">Year</option>
           {{#each dobYearRange}}
             <option value={{this}}>{{this}}</option>
--- a/src/js/outreach/views/opt-in_views.js
+++ b/src/js/outreach/views/opt-in_views.js
@@ -76,7 +76,7 @@ const OptInView = View.extend({
   templateContext() {
     const dobMonthRange = range(1, 13);
     const dobDayRange = range(1, 32);
-    const dobYearRange = range(1900, 2015).reverse();
+    const dobYearRange = range(1908, dayjs().add(1, 'year').year()).reverse();
 
     return {
       dobMonthRange,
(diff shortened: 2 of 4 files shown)


src/js/outreach/views/opt-in_views.js
```diff
@@ -33,12 +34,30 @@ const OptInView = View.extend({
...
     <div class="opt-in__field">
       <label class="opt-in__field-label">Your date of birth</label>
-      <input
-        type="date"
-        class="input-primary opt-in__field-input js-birth-date"
-        placeholder="Enter your date of birth"
-        value="{{ birth_date }}"
-      />
+      <div class="opt-in__field-select u-margin--r-4">
+        <select name="dob-month" class="u-margin--r-8 js-dob-month ">
```
nmajor25: My thinking on this is that we use a `<select>` element for these and let the user's mobile device os deal with the rest.

src/js/outreach/views/opt-in_views.js
```diff
@@ -79,8 +113,17 @@ const OptInView = View.extend({
...
-    this.model.set({ birth_date: trim(this.ui.birthDate.val()) });
+  onChangeDobInput() {
+    const dobMonth = this.ui.dobMonth.val();
+    const dobDay = this.ui.dobDay.val();
+    const dobYear = this.ui.dobYear.val();
+
+    if (!dobMonth || !dobDay || !dobYear) {
+      this.model.set({ birth_date: '' });
+      return;
+    }
```
nmajor25, line 126: This switches the `Submit` button to disabled if the birth date is no longer valid.

src/js/outreach/views/opt-in_views.js
```diff
@@ -53,20 +72,35 @@ const OptInView = View.extend({
     <p class="opt-in__disclaimer">By clicking Submit you agree to receive SMS text message notifications. You may opt out at any time.</p>
     <button class="opt-in__submit button--green w-100 js-submit" disabled>Submit</button>
   `,
+  templateContext() {
+    const dobMonthRange = range(1, 13);
+    const dobDayRange = range(1, 32);
+    const dobYearRange = range(1900, 2015).reverse();
```
nmajor25: Did this so we can do `{{#each arrayOfValues}}`'s in the handlebars template. I implemented something quick to get something working initially.

Probably want to do this different somehow.

nmajor25: Also not sure what year range we want to use. And if we want to display the month in a numerical or name format (1 vs. "Jan")

src/js/outreach/views/opt-in_views.js
```diff
@@ -33,12 +34,30 @@ const OptInView = View.extend({
     </div>
     <div class="opt-in__field">
       <label class="opt-in__field-label">Your date of birth</label>
-      <input
-        type="date"
-        class="input-primary opt-in__field-input js-birth-date"
-        placeholder="Enter your date of birth"
-        value="{{ birth_date }}"
-      />
+      <div class="opt-in__field-select u-margin--r-4">
```
nmajor25: This wrapping `.opt-in__field-select` is here so that I could add spacing between the `<select>`'s down arrow icon and the right side border.

You can't add padding to that directly. So I added a `<div>` wrapper around the `<select>`, added the border styling to the `<div>` along with right margin to add that spacing.

Screenshot 2025-02-06 at 3 39 05 PM

src/js/outreach/views/opt-in_views.js
```diff
@@ -79,8 +113,17 @@ const OptInView = View.extend({
...
+    const dobMonth = this.ui.dobMonth.val();
+    const dobDay = this.ui.dobDay.val();
+    const dobYear = this.ui.dobYear.val();
+
+    if (!dobMonth || !dobDay || !dobYear) {
+      this.model.set({ birth_date: '' });
+      return;
+    }
+
+    this.model.set({ birth_date: `${ dobYear }-${ dobMonth }-${ dobDay }` });
```
nmajor25: We probably need to add leading zeros to any single digit values, right? For it to be valid when sending the dob string to the BE.

Like this: `2025-5-5` => `2025-05-05`

src/js/outreach/views/opt-in_views.js
paulfalgout: I suspect the year range should be chosen programatically..

src/js/outreach/views/opt-in_views.js
paulfalgout: yep, but we should probably stick it in dayjs to make sure it's event valid.  Like the 31st of Feb.. the dayjs can format `YYYY-MM-DD`

test/integration/outreach/outreach.js
```diff
@@ -33,8 +33,16 @@ context('Outreach', function() {
       .type('Patient');
 
     cy
-      .get('.js-birth-date')
-      .type('1990-10-01');
+      .get('.js-dob-month')
+      .select('10');
```
paulfalgout: maybe select a single digit just to make sure we're getting the leading zeros

nmajor25: Good idea, updated the test to select a single digit for both month and day.

src/js/outreach/views/opt-in_views.js
nmajor25: Validating this date while also keeping the submit button in the correct enabled/disabled state was tricky.

Do we need to validate the date? If I remember correctly, this date string is being sent to the BE and compared to what's in the db. It's not saving/inserting any new data into the db, right? 

So an invalid date like `2025-2-31` being sent wouldn't really cause any issues, right?

nmajor25: When generating the `dayjs()` date obj and then using `.isValid()` against it, there were a lot of edge cases not covered.

Like both of these are being considered a valid date by `dayjs().isValid()`:

```js
dayjs('-10-10');
dayjs('null-10-10');
```

Since `isValid()` simply checks if a date object could be created.

You can do a strict check like this:

```js
dayjs('1990-02-15', 'YYYY-MM-DD', true).isValid();
```

But then we run into issues because the originally inputted values may or may not have the leading zeros.

src/js/outreach/views/opt-in.scss
```diff
@@ -34,6 +34,10 @@
   font-size: 16px;
   font-weight: 500;
   padding: 12px 20px 12px 12px;
+
+  &:focus {
+    outline: none;
+  }
```
nmajor25: The browser adds a border around the `<select>` element when it's open. Which looks weird because I have a border applied to the parent `<div>`, not the `<select>` itself.

I simply removed it here. If we want to keep that, we'll have to do something creative or rethink how we're doing the border styling.

Example:

Screenshot 2025-02-07 at 3 51 17 PM

shadowhand: The `outline` is also an accessibility hint. Whenever possible, we should find another solution than removing it.

paulfalgout: Yeah so back in the day, rather than adding a wrapper we did:
```css
    appearance: none;
    background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAACQAAAAaCAYAAADfcP5FAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAA2hpVFh0WE1MOmNvbS5hZG9iZS54bXAAAAAAADw/eHBhY2tldCBiZWdpbj0i77u/IiBpZD0iVzVNME1wQ2VoaUh6cmVTek5UY3prYzlkIj8+IDx4OnhtcG1ldGEgeG1sbnM6eD0iYWRvYmU6bnM6bWV0YS8iIHg6eG1wdGs9IkFkb2JlIFhNUCBDb3JlIDUuMy1jMDExIDY2LjE0NTY2MSwgMjAxMi8wMi8wNi0xNDo1NjoyNyAgICAgICAgIj4gPHJkZjpSREYgeG1sbnM6cmRmPSJodHRwOi8vd3d3LnczLm9yZy8xOTk5LzAyLzIyLXJkZi1zeW50YXgtbnMjIj4gPHJkZjpEZXNjcmlwdGlvbiByZGY6YWJvdXQ9IiIgeG1sbnM6eG1wTU09Imh0dHA6Ly9ucy5hZG9iZS5jb20veGFwLzEuMC9tbS8iIHhtbG5zOnN0UmVmPSJodHRwOi8vbnMuYWRvYmUuY29tL3hhcC8xLjAvc1R5cGUvUmVzb3VyY2VSZWYjIiB4bWxuczp4bXA9Imh0dHA6Ly9ucy5hZG9iZS5jb20veGFwLzEuMC8iIHhtcE1NOk9yaWdpbmFsRG9jdW1lbnRJRD0ieG1wLmRpZDowMzgwMTE3NDA3MjA2ODExODIyQUU5OEMwNzQzMzg5RSIgeG1wTU06RG9jdW1lbnRJRD0ieG1wLmRpZDpENjMxMUU5QjU2QjMxMUUyOEZDRkVCRjEzQUQ0MDczRiIgeG1wTU06SW5zdGFuY2VJRD0ieG1wLmlpZDpENjMxMUU5QTU2QjMxMUUyOEZDRkVCRjEzQUQ0MDczRiIgeG1wOkNyZWF0b3JUb29sPSJBZG9iZSBQaG90b3Nob3AgQ1M2IChNYWNpbnRvc2gpIj4gPHhtcE1NOkRlcml2ZWRGcm9tIHN0UmVmOmluc3RhbmNlSUQ9InhtcC5paWQ6MDc4MDExNzQwNzIwNjgxMTgyMkFFOThDMDc0MzM4OUUiIHN0UmVmOmRvY3VtZW50SUQ9InhtcC5kaWQ6MDM4MDExNzQwNzIwNjgxMTgyMkFFOThDMDc0MzM4OUUiLz4gPC9yZGY6RGVzY3JpcHRpb24+IDwvcmRmOlJERj4gPC94OnhtcG1ldGE+IDw/eHBhY2tldCBlbmQ9InIiPz70MjdUAAABcUlEQVR42uyWPUvDUBRAEysFu+jiIOpSNxF0EBcnP7tYxyIO4mLiJvoL/AOCmwR0EAcRnPyCKtJZN8FBEQTBrYuCiBRKPQ/uFGiTNu+9InjhDC8v93LCu7mJ63meExGDcAZL8BzeDILA0RkdEft9UIIxuIWsYzgaCfXCNQzJul/WA+0Q6oYrGAldV3JFkbUmlIFLGK+TMww3Im1cKC0NPBmRNyrHlzEp1AmnMBMzdwIufN/vMiGUgkPIN5k/BSdIpXUKubAHyy3WUA9xhFRKl9AOrCWsU4B9HUJujEndMGxPauvxL/TnhFRT9+gqRoN/JK2hpvM2bGjw2YVNHUemihwkrKNm0JauHqqBrz4BLdY4hnWOq6azqauwAudN5qv7V5GpmnjLKvLfXIqZq+4rIFMx+dp/wyLcR+TdwQL82JhDXzAPj3VyHiAn8tYG4ydMw0vo+hPMyb71SV0WqTdZv4pMuZ2fjneYlZ7Kydpo/AowAB4YSiejVMMzAAAAAElFTkSuQmCC');
    background-position: right 2px top 50%;
    background-repeat: no-repeat;
    background-size: auto 14px;
```

then you can keep the outline.  Otherwise you have to add the outline with a `:focus-within` selector.. this wasn't something we could use at the time.. but either way works for me.

_but_ this is a pretty generic fix.. styles that aren't specific to how the select is positioned or sized within this page should probably be move to a select.scss in scss/modules.  We wouldn't want to reproduce this for every instance of a select we add

paulfalgout (changes requested):

src/js/outreach/views/opt-in.scss
nmajor25: In a new [commit](be870e523f7f96c73baf07e50a21818622cf67fb), I updated the styling to add an outline when the child `<select>` element has `:focus-visible`, via `.opt-in__field-select > &:has(:focus-visible)`.

`:focus-visible` is what we should be targeting on mobile. Which will be when the user has the select menu open.

Found a great article that explained it well in detail: https://larsmagnus.co/blog/focus-visible-within-the-missing-pseudo-class

nmajor25: This way we can go the parent `<div>` route I have implemented currently. Which I think is the simpler option compared to doing a custom background image.

shadowhand: @nmajor25 love that solution.

src/js/outreach/views/opt-in.scss
```diff
@@ -21,6 +21,29 @@
   width: 100%;
 }
 
+.opt-in__field-select {
```
paulfalgout: these styles can be removed.. they're down in the select module which looks good to me

paulfalgout (changes requested):
one clean up and :shipit:

src/js/outreach/views/opt-in.scss
nmajor25: Ooops, missed that. That's been removed now 👍

paulfalgout (approved):

src/js/outreach/views/opt-in_views.js
```diff
@@ -53,20 +74,35 @@ const OptInView = View.extend({
...
     'change': 'setSubmitButtonState',
   },
   ui: {
     firstName: '.js-first-name',
     lastName: '.js-last-name',
-    birthDate: '.js-birth-date',
+    dobMonth: '.js-dob-month',
+    dobDay: '.js-dob-day',
+    dobYear: '.js-dob-year',
     phone: '.js-phone',
```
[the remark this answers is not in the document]
paulfalgout: @nmajor25 hmm.. is this a thing?  might be worth looking into it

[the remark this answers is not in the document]
nmajor25: From what I'm reading in the mdn web docs, the `input` and `change` events do the same thing on `<select>` elements. I'm not seeing anything that says you shouldn't use the `input` event for `<select>`'s. 

But `change` does seem to be the standard/make more sense for `<select>`'s. So I think it's a good suggestion to use that instead.

paulfalgout (approved):

PR #1395, opened 2025-02-11
Allow for customizing the wrapper template of any widget

paulfalgout:
Adds `definition.wrapperTemplate` that is passed the widget attributes and the widget values.

Returning “” from the wrapperTemplate should `display:none` the widget

Shortcut Story ID: [sc-58659]

I think the readme modifications should cover the change, but this should allow any handlebars logic with any requested widget data to determine if the widget should show or if something entirely different should show based on the data... this should work for any widget.  I've only tested simple cases so far, but I'm pretty sure it works for all the use cases.

## Summary by CodeRabbit

- **New Features**
	- Enhanced widget configuration with optional template support to improve customization and conditional content rendering.
	- Introduced sidebar widgets that handle scenarios where a content region is not defined or where content is conditionally displayed.
- **Style**
	- Updated styles to hide empty sections in forms and patient sidebars, resulting in a cleaner layout.
- **Documentation**
	- Expanded instructions for widget template customization with clear examples and usage guidelines.

### DIFF
--- a/test/integration/patients/patient/sidebar.js
+++ b/test/integration/patients/patient/sidebar.js
@@ -82,6 +82,8 @@ context('patient sidebar', function() {
                 'patientSSNIdentifier',
                 'hbsWidget',
                 'hbsEmptyWidget',
+                'hbsNoRegionWidget',
+                'hbsEmptyTemplateWidget',
               ],
             },
           },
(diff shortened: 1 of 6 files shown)


test/integration/patients/patient/sidebar.js
```diff
@@ -177,6 +178,15 @@ context('patient sidebar', function() {
...
           }),
+          addWidget({
+            slug: 'hbsNoRegionWidget',
+            category: 'widget',
+            definition: {
+              template: 'Content that will not appear because the region is missing',
+              // Missing <div data-content-region> on purpose
+              wrapperTemplate: '<div class="no-region-wrapper">No region defined</div>',
+            },
+          }),
```
nmajor25, line 190: Do we need to add another widget to the test that's similar to the one you added in the README?

Where the `wrapperTemplate` has logic that shows/hides the widget based on a value. And also verifies that it renders correctly when the `<div data-content-region>` region is included in the `wrapperTemplate`.

nmajor25 (commented):
Looks good, just one comment/question on the tests.

nmajor25 (approved):
🚢

PR #1396, opened 2025-02-13
Display `ActionFileAttached` events in action sidebar activity log

nmajor25:
Shortcut Story ID: [sc-59190]

## Summary by CodeRabbit

- **New Features**
  - File attachment notifications are now displayed in the activity logs, showing details when a file is added by a clinician or the system.
  - A new display template has been introduced to enhance how file attachment events appear in the patient sidebar.
  - Tests have been updated to ensure the proper display of these new file attachment notifications.


test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -122,6 +122,13 @@ context('action sidebar', function() {
             previous: 0,
             value: null,
           }),
+          getActivity({
+            event_type: 'ActionFileAttached',
+            source: 'api',
+            value: 'user-created-file.pdf',
```
nmajor25, line 128: `value` here is currently not in the BE data provided to the FE.

This is what's provided currently:

```js
{
  "type": "events",
  "id": "49cb92aa-e982-11ef-b943-b7c15233f8c3",
  "attributes": { "event_type": "ActionFileAttached", "date": "2025-02-12T20:45:26+00:00", "source": "api" },
  "relationships": {
    "editor": {
      "data": { "type": "clinicians", "id": "323a0b8a-a498-11ec-a067-3db09696fbc9" }
    },
    "action": {
      "data": { "type": "patient-actions", "id": "caf52702-ed26-4915-b40c-70dd6b01b9cd" }
    }
  }
}
```

nmajor25: If they passed us the full file path, we could do a `path.split('/').pop()` to get the filename.

nmajor25: There’s also no file relationship data provided for the event. But I don’t think that would work even if we had it. 

Since we’d have to do something like `actionModel.getFileById(file.id).getFilename()` to get the filename. But then what if that file is deleted in the future and no longer resides on the action model?

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -122,6 +122,13 @@ context('action sidebar', function() {
...
+          }, {
+            editor: getRelationship(getCurrentClinician()),
```
nmajor25: Eh we don't need this here. `getActivity()` adds this already.

PR #1397, opened 2025-02-13
For file entities, switch from using `relationship.action: {}` => `relationship.actions: []`

nmajor25:
Shortcut Story ID: [sc-59178]

## Summary by CodeRabbit

- **New Features**
  - Attachments and file management now support handling multiple actions, enhancing user flexibility.

- **Refactor**
  - Improved the underlying process for associating and processing actions with attachments and files for a more robust experience.

### DIFF
--- a/src/js/entities-service/entities/files.js
+++ b/src/js/entities-service/entities/files.js
@@ -1,5 +1,6 @@
 import { get, first } from 'underscore';
 import Store from 'backbone.store';
+import Radio from 'backbone.radio';
 import BaseCollection from 'js/base/collection';
 import BaseModel from 'js/base/model';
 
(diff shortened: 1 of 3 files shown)


src/js/entities-service/entities/files.js
```diff
@@ -42,7 +43,9 @@ const _Model = BaseModel.extend({
     return '/api/files';
   },
   getAction() {
-    return this.getRelationship('_action');
+    const actionRelationship = this.get('_actions')[0];
+
+    return Radio.request('entities', 'actions:model', actionRelationship.id);
```
nmajor25: This is the only way I could get this working.

Doing `this.getRelationship('_actions')` here doesn't work. `getRelationship()` ([here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/base/model.js#L9-L13)) calls `getStore()` ([here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/base/entity-service.js#L9-L12)), which is expected an object in `{id: '', type: ''}` format. Therefore `this.getRelationship('_actions')` would returnS `undefined`.

paulfalgout: Well this would be `getActions` and Radio.request('entities', 'actions:collection'...`. 

I think we need to iterate over all of the actions in order to handle the removeFile notification?

where is getAction used?  just the `isNew` correct?  that can assume `getActions()[0]` because on creation it can't be anything else.

paulfalgout (changes requested):

src/js/entities-service/entities/files.js
nmajor25: Okay this has been updated to `Radio.request` the collection of actions. 

`urlRoot()` uses the action at `[0]` and on the `FileRemoved` ws notification the `removeFile()` function is called on each action model in the collection returned by `getActions()`.

paulfalgout (approved):

PR #1400, opened 2025-03-04
Handle web socket notifications on patient dashboard & archive

nmajor25:
Shortcut Story ID: [sc-59837]

- [x] Handle ws notifications on patient dashboard
- [x] Handle ws notifications on patient archive
- [x] Use `NIL_UUID` for action.flow filter
- [x] Manually test functionality/edge cases in local dev environment
- [x] Add dashboard cypress tests
- [x] Add archive cypress tests
- [x] Fix failing tests - web socket messages intermittently not coming through at beginning of test
- [x] Rebase branch & use new service request implemented in https://github.com/RoundingWell/care-ops-frontend/pull/1417

## Summary by CodeRabbit

- **New Features**
	- Enabled real-time updates for patient data on the archive and dashboard, ensuring that changes to patient flows and actions are reflected live.
	- Improved the visual handling of completed items with smooth fade-out effects for a better user experience.
- **Tests**
	- Expanded test coverage to validate that real-time notifications update the UI consistently for both flows and actions.
	- Added new test specifications for socket notifications related to flows and actions in both the patient archive and dashboard contexts.

### DIFF
--- a/src/js/apps/patients/patient/archive/archive_app.js
+++ b/src/js/apps/patients/patient/archive/archive_app.js
@@ -33,24 +33,15 @@ export default App.extend({
     this.showChildView('content', new ListView({ collection: this.collection }));
   },
   subscribe() {
-    const channel = Radio.channel('ws');
-
     const filters = {
       states: this.states,
       patient: this.patient.id,
     };
 
-    channel.request('subscribe', this.collection.models, {
+    Radio.request('ws', 'subscribe', this.collection.models, {
       filters: { actions: { ...filters, flow: NIL_UUID }, flows: filters },
     });
-
-    this.listenTo(channel, 'message', (data, model) => {
-      if (this.collection.get(model) || model.isFetching()) return;
-
-      model.fetch().then(() => {
-        this.collection.add(model);
-        Radio.request('ws', 'add', model);
-      });
-    });
+    Radio.request('ws', 'manage:add', this, this.collection, 'flows');
+    Radio.request('ws', 'manage:add', this, this.collection, 'patient-actions');
   },
 });
--- a/src/js/apps/patients/patient/archive/archive_app.js
+++ b/src/js/apps/patients/patient/archive/archive_app.js
@@ -7,13 +7,15 @@ import { LayoutView, ListView } from 'js/views/patients/patient/archive/archive_
 
 export default App.extend({
   onBeforeStart({ patient }) {
+    this.patient = patient;
     this.showView(new LayoutView({ model: patient }));
     this.getRegion('content').startPreloader();
   },
   beforeStart({ patient }) {
     const currentWorkspace = Radio.request('workspace', 'current');
-    const states = currentWorkspace.getStates();
-    const filter = { states: states.groupByDone().done.getFilterIds() };
+    this.states = currentWorkspace.getStates();
+
+    const filter = { states: this.states.groupByDone().done.getFilterIds() };
 
     return [
       Radio.request('entities', 'fetch:actions:collection:byPatient', { patientId: patient.id, filter }),
@@ -22,6 +24,32 @@ export default App.extend({
   },
   onStart({ patient }, actions, flows) {
     this.collection = new Backbone.Collection([...actions.models, ...flows.models]);
+
+    this.subscribe();
+
     this.showChildView('content', new ListView({ collection: this.collection }));
+  },
+  subscribe() {
+    const channel = Radio.channel('ws');
+
+    const filter = {
+      states: this.states.groupByDone().done.getFilterIds(),
+      patient: this.patient.id,
+    };
+
+    channel.request('subscribe', this.collection.models, {
+      filters: { actions: filter, flows: filter },
+    });
+
+    this.listenTo(channel, 'message', (data, model) => {
+      if (this.collection.get(model) || model.isFetching()) return;
+
+      model.fetch().then(() => {
+        if (model.type === 'patient-actions' && model.getFlow()) return;
+
+        this.collection.add(model);
+        Radio.request('ws', 'add', model);
+      });
+    });
   },
 });
--- a/src/js/apps/patients/patient/archive/archive_app.js
+++ b/src/js/apps/patients/patient/archive/archive_app.js
@@ -1,5 +1,6 @@
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
+import { NIL as NIL_UUID } from 'uuid';
 
 import App from 'js/base/app';
 
@@ -32,21 +33,19 @@ export default App.extend({
   subscribe() {
     const channel = Radio.channel('ws');
 
-    const filter = {
+    const filters = {
       states: this.states.groupByDone().done.getFilterIds(),
       patient: this.patient.id,
     };
 
     channel.request('subscribe', this.collection.models, {
-      filters: { actions: filter, flows: filter },
+      filters: { actions: { ...filters, flow: NIL_UUID }, flows: filters },
     });
 
     this.listenTo(channel, 'message', (data, model) => {
       if (this.collection.get(model) || model.isFetching()) return;
 
       model.fetch().then(() => {
-        if (model.type === 'patient-actions' && model.getFlow()) return;
-
         this.collection.add(model);
         Radio.request('ws', 'add', model);
       });
--- a/src/js/apps/patients/patient/archive/archive_app.js
+++ b/src/js/apps/patients/patient/archive/archive_app.js
@@ -8,15 +8,17 @@ import { LayoutView, ListView } from 'js/views/patients/patient/archive/archive_
 
 export default App.extend({
   onBeforeStart({ patient }) {
+    const currentWorkspace = Radio.request('workspace', 'current');
+    const { done } = currentWorkspace.getStates().groupByDone();
+    this.states = done.getFilterIds();
+
     this.patient = patient;
+
     this.showView(new LayoutView({ model: patient }));
     this.getRegion('content').startPreloader();
   },
   beforeStart({ patient }) {
-    const currentWorkspace = Radio.request('workspace', 'current');
-    this.states = currentWorkspace.getStates();
-
-    const filter = { states: this.states.groupByDone().done.getFilterIds() };
+    const filter = { states: this.states };
 
     return [
       Radio.request('entities', 'fetch:actions:collection:byPatient', { patientId: patient.id, filter }),
@@ -34,7 +36,7 @@ export default App.extend({
     const channel = Radio.channel('ws');
 
     const filters = {
-      states: this.states.groupByDone().done.getFilterIds(),
+      states: this.states,
       patient: this.patient.id,
     };
 
--- a/src/js/apps/patients/patient/dashboard/dashboard_app.js
+++ b/src/js/apps/patients/patient/dashboard/dashboard_app.js
@@ -1,5 +1,6 @@
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
+import { NIL as NIL_UUID } from 'uuid';
 
 import App from 'js/base/app';
 
@@ -46,21 +47,19 @@ export default App.extend({
   subscribe() {
     const channel = Radio.channel('ws');
 
-    const filter = {
+    const filters = {
       states: this.states.groupByDone().notDone.getFilterIds(),
       patient: this.patient.id,
     };
 
     channel.request('subscribe', this.collection.models, {
-      filters: { actions: filter, flows: filter },
+      filters: { actions: { ...filters, flow: NIL_UUID }, flows: filters },
     });
 
     this.listenTo(channel, 'message', (data, model) => {
       if (this.collection.get(model) || model.isFetching()) return;
 
       model.fetch().then(() => {
-        if (model.type === 'patient-actions' && model.getFlow()) return;
-
         this.collection.add(model);
         Radio.request('ws', 'add', model);
       });
--- a/src/js/apps/patients/patient/dashboard/dashboard_app.js
+++ b/src/js/apps/patients/patient/dashboard/dashboard_app.js
@@ -14,20 +14,24 @@ export default App.extend({
   },
 
   onBeforeStart({ patient }) {
+    const currentWorkspace = Radio.request('workspace', 'current');
+    const { notDone } = currentWorkspace.getStates().groupByDone();
+    this.states = notDone.getFilterIds();
+
     this.currentUser = Radio.request('bootstrap', 'currentUser');
     this.patient = patient;
+
     this.showView(new LayoutView({ model: patient }));
+
     if (!this.currentUser.can('work:own')) {
       this.getRegion('addWorkflow').empty();
     }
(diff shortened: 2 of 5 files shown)


src/js/apps/patients/patient/dashboard/dashboard_app.js
```diff
@@ -25,21 +25,68 @@ export default App.extend({
...
+      Radio.request('entities', 'fetch:actions:collection:byPatient', { patientId: patient.id, filter: this.filter }),
+      Radio.request('entities', 'fetch:flows:collection:byPatient', { patientId: patient.id, filter: this.filter }),
     ];
   },
 
   onStart(options, actions, flows) {
-    this.collection = new Backbone.Collection([...actions.models, ...flows.models]);
+    // remove these when BE api endpoint `filter[state]=` is fixed
+    const doneActions = actions.filter(action => !action.isDone());
+    const doneFlows = flows.filter(flow => !flow.isDone());
```
nmajor25: Without this, we'll need to rethink how we determine if a model needs to be fetched or not.

We currently do a `if (this.collection.get(model))`. But without this filtering ☝️, that check won't work in all scenarios.

src/js/apps/patients/patient/archive/archive_app.js
```diff
@@ -13,15 +13,63 @@ export default App.extend({
...
   },
+  subscribe() {
+    const channel = Radio.channel('ws');
+
+    channel.request('subscribe', this.collection.models, {
+      filters: { actions: this.filter, flows: this.filter },
+    });
+
+    this.listenTo(channel, 'message', ({ category, resource, payload }) => {
+      const modelResource = category === 'ResourceCreated' ? payload.resource : resource;
```
nmajor25: This is technically possible? For a flow or action to be created with a `Done` state? Does that ever happen in the wild? But we should probably handle it either way, right?

src/js/views/patients/patient/dashboard/dashboard_views.js
```diff
@@ -200,7 +200,7 @@ const FlowItemView = View.extend({
   modelEvents: {
     'change:_owner': 'render',
   },
-  behaviors: [RowBehavior],
+  behaviors: [RowBehavior, DoneBehavior],
```
nmajor25, line 203: `DoneBehavior` is when the `model.state` changes to `isDone === true`. The action/flow is removed from the list with a fade-out animation.

This was never added for flow list items because it wasn't possible for that to change before. The flow state button is disabled. But now it can change via web socket notification.

src/js/apps/patients/patient/dashboard/dashboard_app.js
```diff
@@ -25,21 +25,68 @@ export default App.extend({
...
+
+      if (this.collection.get(model) || model.isFetching()) return;
+
+      if (modelResource.type === 'flows') {
+        this._fetchFlow(model);
+
+        return;
+      }
+
+      this._fetchAction(model);
```
nmajor25: Maybe we can consolidate this to one `this._fetchModel(model)` via `Radio.request('entities', 'fetch:${ type }:model', model.id);`...

src/js/apps/patients/patient/dashboard/dashboard_app.js
paulfalgout: Why? If it is already in the collection you should never need to fetch it because you are maintaining the data with notifications even if it is not visible

src/js/apps/patients/patient/archive/archive_app.js
```diff
@@ -13,15 +13,63 @@ export default App.extend({
...
-      Radio.request('entities', 'fetch:actions:collection:byPatient', { patientId: patient.id, filter }),
-      Radio.request('entities', 'fetch:flows:collection:byPatient', { patientId: patient.id, filter }),
+      Radio.request('entities', 'fetch:actions:collection:byPatient', { patientId: patient.id, filter: this.filter }),
+      Radio.request('entities', 'fetch:flows:collection:byPatient', { patientId: patient.id, filter: this.filter }),
     ];
   },
   onStart({ patient }, actions, flows) {
-    this.collection = new Backbone.Collection([...actions.models, ...flows.models]);
+    // remove these when BE api endpoint `filter[state]=` is fixed
+    const doneActions = actions.filter(action => action.isDone());
```
paulfalgout: I don't think we need to do this, the view is handling the filtering enough

src/js/apps/patients/patient/archive/archive_app.js
```diff
@@ -13,15 +13,63 @@ export default App.extend({
...
+
+    this.subscribe();
+
     this.showChildView('content', new ListView({ collection: this.collection }));
   },
+  subscribe() {
+    const channel = Radio.channel('ws');
+
+    channel.request('subscribe', this.collection.models, {
+      filters: { actions: this.filter, flows: this.filter },
```
paulfalgout: I think both of these filters must include `patientId`

src/js/apps/patients/patient/archive/archive_app.js
paulfalgout: Well we can probably avoid it for now because it probably doesn't happen and we won't need this after the BE change to `ResourceCreated`.   Doing this now just mean we have to test for it.

src/js/apps/patients/patient/archive/archive_app.js
```diff
@@ -13,15 +13,63 @@ export default App.extend({
...
+    });
+
+    this.listenTo(channel, 'message', ({ category, resource, payload }) => {
+      const modelResource = category === 'ResourceCreated' ? payload.resource : resource;
+      const model = Radio.request('entities', 'get:store', modelResource);
+
+      if (this.collection.get(model) || model.isFetching()) return;
+
+      if (modelResource.type === 'flows') {
```
paulfalgout: yeah I'm perfectly fine combining these with `fetch:${ model.type }:model`   we can probably recombine these methods now since they're simpler than the worklist

paulfalgout: in fact..  I think we will _always_ have the context needed that this can be
```js
const model = Radio.request('entities', 'get:store', modelResource);

if (this.collection.get(model) || model.isFetching()) return;
model.fetch().then(() => { 
  this.collection.add(model);

});
```

unless I'm missing something?

src/js/views/patients/patient/dashboard/dashboard_views.js
paulfalgout: this is nice how easy that is to implement

paulfalgout (changes requested):

src/js/apps/patients/patient/dashboard/dashboard_app.js
nmajor25: Ya I agree, don't think it's needed.

src/js/apps/patients/patient/archive/archive_app.js
nmajor25: That seems right. 

But that will subscribe us to each flow's child actions. So when we receive a ws message for one of those, how will we know if it's an action that should be fetched or not? We'd have to check if that action belonged to a flow, but we won't know that until after we fetch it, right?

Or maybe I'm overthinking it? I need to investigate it more next week.

nmajor25: I think we'll have to check if the action has a parent flow after fetching it. And if it does, don't add it to the collection.

 I don't think we can determine that until after the action has been fetched.

```js
model.fetch().then(() => {
  if (model.type === 'patient-actions' && model.getFlow()) return;

  this.collection.add(model);

});
```

src/js/apps/patients/patient/archive/archive_app.js
nmajor25: Yup that works 👍

src/js/apps/patients/patient/archive/archive_app.js
paulfalgout: @DrechslerDerek we need to have a way to indicate we want to subscribe to actions for a patient that _aren't_ related to a flow.. ideas?

shadowhand: @paulfalgout I wonder if we need to have two different `types` for patient/flow actions? If we went that route, we might have `patient-actions` and `flow-actions` and then `actions` is a union of them?

shadowhand: Actually, I think it should be `actions-of-flow` and `actions-of-patient`. Then sorting places them together so there is a visual relationship.

paulfalgout: Hmm that's a pretty big change and there's been some discussion about decoupling actions from patients at some point, so I'd personally want to avoid it.  Really I wish the current type was just `actions`   I'm wondering if we can specify `filters.actions.flow = null`, or maybe `filters.patientActions` instead of `filters.actions`, to indicate we want patient actions that aren't associated with a flow..  then if there's some reason to split the types we can do so later?

shadowhand: @paulfalgout if that's not low hanging fruit, then shelve it for now. We can certainly make the current type `actions` with minimal effort, it just requires coordination, right?

In any case... regarding the filter. 😅 I don't like `filter.actions.flow = null` (because PHP can't differentiate between `null` and `undefined` like JS can), but I think `filters.actions.only = 'patient'` (implied enum `patient|flow`) would work?

shadowhand: But I am also open to `filters.patientActions`. Let's see what makes more sense to @DrechslerDerek.

paulfalgout: Worth mentioning we'd never use `only = 'flow'` as for that we just set flow to a particular id.  this is just an issue of filtering out flow actions for a patient filter

DrechslerDerek: I think there could be 2 easy fixes, the first one would be having a flag that says, no flow or flow = null or something, that would be an extra scan condition for dynamo db that would be added dynamically if this flag was set. The other option I see is to do some trick with the current flow filter, where you explicitly pass like 
```
{
    filters: {
        actions: {
            flows: 'null' <or maybe null (at some point in BE this will be a string anyway)>
        }
    }
}
```

which you could actually do right now and it would be a one line change to fix the `$action->flowId ??`. I might be missing how this would cause other issues, and it is more of a bandaid/nuanced solution, but we could do it easily.

shadowhand: If we're going to do some shenanigans like that, it should use the constant value UUID::NIL.

DrechslerDerek: Now I don't think this will work, we need to add another DDB `and` or another filter, meaning the flow on the action is dynamic.

test/integration/patients/patient/archive.js
```diff
@@ -652,7 +655,10 @@ context('patient archive page', function() {
       })
       .visitOnClock(`/patient/archive/${ testPatient.id }`, { now: testTs() })
       .wait('@routePatient')
-      .wait('@routePatientActions');
+      .wait('@routePatientActions')
+      .wait('@routePatientFlows');
+
+    cy.wait(200);
```
nmajor25, line 661: Pushing this up to see if it fixes things when cypress runs in circleci. 

Something is running out of sync. I think it's receiving the first ws notification before the subscription is created. So the app isn't ready to handle it yet. But only sometimes. When I run them locally, it's 50/50 if the test passes or not.

Is there something we're not `.wait()`ing for properly? I can't tell yet.

test/integration/patients/patient/archive.js
```diff
@@ -463,7 +463,9 @@ context('patient archive page', function() {
       })
       .visitOnClock(`/patient/archive/${ testPatient.id }`, { now: testTs() })
       .wait('@routePatient')
-      .wait('@routePatientFlows');
+      .wait('@routePatientActions')
+      .wait('@routePatientFlows')
+      .wait(200); // for initial ws connection to be created
```
nmajor25: Without this `cy.wait()`, these tests fail ~50% of the time. I think what's happening is the `mockWs()` command isn't always finished before we send the first ws message in the test via `cy.sendWs()`.

Adding a `cy.wait(200)` seems to fix the issue, where the test passes 100% of the time. I don't think this does much harm. 

But I would prefer to not do this if there was another way to fix it. The `mockWs` would have to be reconfigured somehow, to somehow let us know when the initial socket connection is ready.

src/js/apps/patients/patient/archive/archive_app.js
```diff
@@ -1,19 +1,22 @@
...
   onBeforeStart({ patient }) {
+    this.patient = patient;
     this.showView(new LayoutView({ model: patient }));
     this.getRegion('content').startPreloader();
   },
   beforeStart({ patient }) {
     const currentWorkspace = Radio.request('workspace', 'current');
-    const states = currentWorkspace.getStates();
-    const filter = { states: states.groupByDone().done.getFilterIds() };
+    this.states = currentWorkspace.getStates();
```
paulfalgout: I think it'd be better to set this in the `onBeforeStart` and not attach things in the `beforeStart`.. and you can probably save the `states` since it's reused in the `subscribe`

paulfalgout (changes requested):

src/js/apps/patients/patient/archive/archive_app.js
nmajor25: Like this: https://github.com/RoundingWell/care-ops-frontend/pull/1400/commits/af63bdd3d766db9857e159ffa369e8e08fe22252?

paulfalgout:
@nmajor25 I think this could be rebased and modified with the new service request and maybe it'd work?

nmajor25:
@paulfalgout think so, I'll give it a try 👍

src/js/apps/patients/patient/archive/archive_app.js
```diff
@@ -22,6 +27,21 @@ export default App.extend({
...
+    const filters = {
+      states: this.states,
+      patient: this.patient.id,
+    };
+
+    Radio.request('ws', 'subscribe', this.collection.models, {
+      filters: { actions: { ...filters, flow: NIL_UUID }, flows: filters },
+    });
+    Radio.request('ws', 'manage:add', this, this.collection, 'flows');
+    Radio.request('ws', 'manage:add', this, this.collection, 'patient-actions');
```
nmajor25, line 45: Two separate service requests here, one for flows and another for actions.

nmajor25: I tested this locally for both patient dashboard and archive, it worked well.

test/integration/patients/patient/archive.js
paulfalgout:
```js
    cy
      .get('@wsHandleMessage')
      .should('have.been.calledOnce');
```
will do it.

The issue was that the `send` is happening in Cypress before the `subscribe` happens from the app.. so we have to wait until the subscribe is called.

paulfalgout (approved):

PR #1401, opened 2025-03-05
Modify `ResourceCreated` socket notification handling

nmajor25:
Shortcut Story ID: [sc-59992] [sc-59950]

`ResourceCreated` socket messages now have an empty `payload`. And the flow/action data is in the `resource` object. A change made in this BE PR: https://github.com/RoundingWell/care-ops-backend/pull/9167/files.

This PR also updates the flow page subscription to use a `filters.actions.flow` filter. And some flow page app code needed to be re-worked a result of that.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the real-time update mechanism on patient flows and worklist views for smoother and more reliable notifications.
	- Simplified WebSocket message handling and improved the structure of resource updates.
- **Tests**
	- Updated validations to ensure resource updates and state changes are accurately handled, improving overall system reliability.
	- Adjusted test cases to reflect changes in WebSocket message structures and relationships.

### DIFF
--- a/src/js/apps/patients/patient/flow/flow_app.js
+++ b/src/js/apps/patients/patient/flow/flow_app.js
@@ -88,7 +88,6 @@ export default SubRouterApp.extend({
 
     this.listenTo(this.flow, {
       'change:_owner': this.onFlowChangeOwner,
-      'message': this.onFlowMessage,
     });
 
     this.startRoute(currentRoute);
@@ -98,7 +97,16 @@ export default SubRouterApp.extend({
     this.stopChildApp('actionSidebar');
   },
   subscribe() {
-    Radio.request('ws', 'subscribe', [this.flow, ...this.actions.models]);
+    const channel = Radio.channel('ws');
+    const filters = { actions: { flow: this.flow.id } };
+
+    channel.request('subscribe', [this.flow, ...this.actions.models], { filters });
+
+    this.listenTo(channel, 'message', ({ category }, model) => {
+      if (category !== 'ResourceCreated') return;
+
+      model.fetch().then(bind(this._addAction, this));
+    });
   },
   _setFlowProgress() {
     const complete = this.actions.filter(action => action.isDone()).length;
@@ -110,7 +118,7 @@ export default SubRouterApp.extend({
     this.actions.add(action);
     this.editableCollection.add(action);
 
-    this.subscribe();
+    Radio.request('ws', 'add', action);
 
     this._setFlowProgress();
   },
@@ -119,13 +127,6 @@ export default SubRouterApp.extend({
   },
   onActionDestroy() {
     this._setFlowProgress();
-  },
-  onFlowMessage({ category, payload }) {
-    if (category !== 'ResourceCreated') return;
-    const { resource } = payload;
-
-    const fetchAction = Radio.request('entities', 'fetch:actions:model', resource.id);
-    fetchAction.then(bind(this._addAction, this));
   },
   onFlowChangeOwner(flow, _owner) {
     if (_owner.type === 'teams') return;
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -160,9 +160,8 @@ export default App.extend({
 
     channel.request('subscribe', this.collection.models, { filters: { [filterKey]: this.filters } });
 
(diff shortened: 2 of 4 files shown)


src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -98,7 +97,24 @@ export default SubRouterApp.extend({
...
   },
   subscribe() {
-    Radio.request('ws', 'subscribe', [this.flow, ...this.actions.models]);
+    const channel = Radio.channel('ws');
+    const filters = { actions: { flow: this.flow.id } };
+
+    channel.request('subscribe', [this.flow, ...this.actions.models], { filters });
+
+    this.listenTo(channel, 'message', ({ category, resource }) => {
+      if (category !== 'ResourceCreated') return;
```
nmajor25, line 106: I think we should be safe just doing this check. On the flow page, we only need to handle a `ResourceCreated` category message. Which would only happen for an action, not a flow.

test/integration/patients/patient/patient-flow.js
```diff
@@ -1099,41 +1098,63 @@ context('patient flow page', function() {
...
       .routeAction(fx => {
         fx.data = otherAction;
 
+        fx.included.push(testFlow);
+
         return fx;
       });
 
     cy.sendWs({
       category: 'ResourceCreated',
```
nmajor25, line 1114: I'm thinking about moving this down to the `'socket notifications'` test at the bottom of this file.

src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -160,9 +160,8 @@ export default App.extend({
 
     channel.request('subscribe', this.collection.models, { filters: { [filterKey]: this.filters } });
 
-    this.listenTo(channel, 'message', ({ category, resource, payload }) => {
-      const modelResource = category === 'ResourceCreated' ? payload.resource : resource;
-      const model = Radio.request('entities', 'get:store', modelResource);
+    this.listenTo(channel, 'message', ({ resource }) => {
```
paulfalgout, line 163: actually this should be `this.listenTo(channel, 'message', (data, model) => {`

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -98,7 +97,24 @@ export default SubRouterApp.extend({
     this.stopChildApp('actionSidebar');
   },
   subscribe() {
-    Radio.request('ws', 'subscribe', [this.flow, ...this.actions.models]);
+    const channel = Radio.channel('ws');
+    const filters = { actions: { flow: this.flow.id } };
+
+    channel.request('subscribe', [this.flow, ...this.actions.models], { filters });
+
+    this.listenTo(channel, 'message', ({ category, resource }) => {
```
paulfalgout: I think because of the nature of the flow page this whole function is just:
```js
this.listenTo(channel, 'message', ({ category }, model) => {
  if (category !== 'ResourceCreated') return;
  model.fetch().then(bind(this._addAction, this));
});
```

paulfalgout (changes requested):

src/js/apps/patients/patient/flow/flow_app.js
nmajor25: I could simplify it down to this:

```js
this.listenTo(channel, 'message', ({ category }, model) => {
  if (category !== 'ResourceCreated') return;

  const fetchAction = Radio.request('entities', 'fetch:actions:model', model.id);

  fetchAction.then(() => {
    this._addAction(model);
  });
});
```

Think we still need the `'fetch:actions:model'` radio request, so the `include` is on the api request.

nmajor25: Or maybe we don't need the include on the request here on the flow page 🤔

paulfalgout: yep I don't think the includes are necessary as we already have all of the context

nmajor25: Think so too, I tested it locally just to be 100% sure. Works without it.

test/integration/patients/patient/patient-flow.js
nmajor25: I'll leave it here for now.

paulfalgout (approved):

PR #1403, opened 2025-03-11
Refactor widget values to base widget

paulfalgout:
Shortcut Story ID: [sc-60162]

I also centralized the widget build, which was kind of all over the place

## Summary by CodeRabbit

- **Refactor**
  - Revised the widget initialization process to simplify how display values are retrieved and applied. These updates streamline the configuration and handling of widget data, resulting in a more consistent and reliable experience for end-users.

### DIFF
--- a/src/js/views/patients/widgets/widgets.js
+++ b/src/js/views/patients/widgets/widgets.js
@@ -17,15 +17,21 @@ function getWrapperTemplate(definition) {
   return { wrapperTemplate: Handlebars.compile(template) };
 }
 
-// NOTE: widget is a view or view definition
-export function buildWidget(widget, patient, widgetModel, options) {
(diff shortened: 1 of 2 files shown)


src/js/views/patients/widgets/widgets.js
```diff
@@ -34,14 +40,10 @@ const widgets = {
   widget: View.extend({
     className: 'widgets-value',
     initialize() {
-      const slug = this.getOption('slug');
-      const widgetValues = Radio.request('entities', 'get:widgetValues:model', slug, this.model.id);
-      this.values = widgetValues.get('values');
```
paulfalgout, line 39: values was only being attached here, but now it's up in `buildWidget` so if it works for one it works for all now

nmajor25 (approved):

PR #1407, opened 2025-03-19
Handle web socket notifications on schedule & reduced schedule

nmajor25:
Shortcut Story ID: [sc-59125]

### DIFF
--- a/src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
+++ b/src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
@@ -97,18 +97,42 @@ export default App.extend({
     if (!this.isRestarting()) this.stopChildApp('filters');
   },
   beforeStart() {
-    const filter = this.getState().getEntityFilter();
+    this.filters = this.getState().getEntityFilter();
     const fields = { flows: ['name', 'state'] };
-    return Radio.request('entities', 'fetch:actions:collection', { data: { filter, fields } });
+
+    return Radio.request('entities', 'fetch:actions:collection', { data: { filter: this.filters, fields } });
   },
   onStart(options, collection) {
     this.collection = collection;
     this.filteredCollection = collection.clone();
 
+    this.subscribe();
+
     this.listenTo(this.filteredCollection, 'reset', this.showCountView);
     this.showCountView();
 
     this.showList();
+  },
+  subscribe() {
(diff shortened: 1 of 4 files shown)


src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
```diff
@@ -97,19 +97,43 @@ export default App.extend({
...
+        return;
+      }
+
+      if (this.collection.get(model) || model.isFetching()) return;
+
+      model.fetch().then(() => {
+        this.collection.add(model);
+        Radio.request('ws', 'add', model);
+
+        this.showList();
```
nmajor25, line 133: I'm not sure how else to do this yet. 

When I test this locally, re-rendering the entire list doesn't seem to cause any issues. But I'm worried about what happens in production when the list is 100's of actions long. And what if we get multiple/many of these types of ws notifications at once?

src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
```diff
@@ -97,19 +97,43 @@ export default App.extend({
...
     this.showList();
   },
+  subscribe() {
+    const channel = Radio.channel('ws');
+
+    channel.request('subscribe', this.collection.models, { filters: { actions: this.filters } });
+
+    this.listenTo(channel, 'message', ({ category }, model) => {
+      if (category === 'ActionDueChanged') {
+        this.showList();
```
nmajor25, line 123: An `ActionDueChanged` ws notification can mean the action's due date or time changed. 

If the date changed, we need to move it to a different day list group. If the time changed, we only need to update the action's time and it doesn't need to be moved.

I'm not sure how to differentiate between those two scenarios yet.

But unlike the worklist & patient dashboard/flow pages, we need to handle this `ActionDueChanged` ws notification in a special way compared to the others.

src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
paulfalgout: correct this doesn't seem like it will scale.. plus it will completely blow away any action the user is currently taking anywhere on the list

paulfalgout (changes requested):
not sure what yet, but we can't just rerender the whole page when a due date changes

PR #1408, opened 2025-03-21
Add `searchSanitize(query)` to `build-matchers-array` utility

nmajor25:
Shortcut Story ID: [sc-60320]

## Summary by CodeRabbit

- **Refactor**
  - Improved the search functionality to ensure more robust and accurate query matching.
- **Bug Fixes**
  - Updated the schedule action label to “Second Action - Dash in Name” for clearer, more consistent display.

### DIFF
--- a/src/js/utils/formatting/build-matchers-array.js
+++ b/src/js/utils/formatting/build-matchers-array.js
@@ -1,9 +1,10 @@
 import { map } from 'underscore';
 
 import words from 'js/utils/formatting/words';
+import searchSanitize from 'js/utils/formatting/search-sanitize';
 
 export default query => {
-  const searchWords = map(words(query), RegExp.escape);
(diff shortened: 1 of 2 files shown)


test/integration/patients/worklist/worklist.js
```diff
@@ -3917,7 +3917,7 @@ context('worklist page', function() {
 
     cy
       .get('@listSearch')
-      .type('Flow Coordinator');
+      .type('Flow - Coordinator');
```
paulfalgout: We can remove this. This is testing a behavior that if it didn't happen would be fine

paulfalgout (changes requested):

test/integration/patients/worklist/worklist.js
nmajor25: Okay ya makes sense, that's been removed.

paulfalgout (approved):

PR #1412, opened 2025-03-25
Set a height of `36px` on `.list-page__filters`

nmajor25:
Shortcut Story ID: [sc-60431]

To stop layout from shifting when bulk edit button visibility is toggled on worklists & schedule.

Screengrab of the layout shifting, which this PR fixes:

https://github.com/user-attachments/assets/ad6233ec-bbfe-4e98-94f0-bf65e73f373d

## Summary by CodeRabbit

- **Style**
  - Updated the filter panel styling for clearer element alignment.
  - Adjusted button text styling by removing a fixed line height, enhancing layout consistency.

### DIFF
--- a/src/js/views/patients/sidebar/filters/filters-sidebar.scss
+++ b/src/js/views/patients/sidebar/filters/filters-sidebar.scss
@@ -2,4 +2,5 @@
   color: color(blue);
   margin-right: 16px;
   padding: 0;
+  vertical-align: baseline;
 }
--- a/src/scss/modules/buttons.scss
+++ b/src/scss/modules/buttons.scss
@@ -228,7 +228,6 @@
 
   border-color: transparent;
   color: $black-60;
-  line-height: normal;
 
   &:hover,
   &:focus {


src/scss/modules/list-pages.scss
```diff
@@ -57,5 +57,6 @@
 }
 
 .list-page__filters {
+  height: 34px;
```
nmajor25: Well this should probably be `36px`... a px divisible by 4.

paulfalgout (changes requested):
I think this is working around the issue.  I think the issue is `.button--text` on the cancel button which needs the `line-height: normal` removed.  But I think the reason this was added was to fix the clear filters button on the filters sidebar.  after removing the line height I think adding `vertical-align: baseline;` to `.filters-sidebar__clear-filters` will fix that..

nmajor25:
@paulfalgout your suggested change works well 👍

paulfalgout (approved):

PR #1413, opened 2025-03-25
Use `line-height: normal` for `.list__title-filter-button`

nmajor25:
Shortcut Story ID: [sc-55237]

To stop bottom of text being cut off because of the `overflow: hidden` property, for example:

Screenshot 2025-03-25 at 11 17 27 AM

## Summary by CodeRabbit

- **Style**
	- Adjusted the vertical spacing for button text in the patient list view, resulting in a subtly updated look for better readability.

### DIFF
--- a/src/js/views/patients/shared/list.scss
+++ b/src/js/views/patients/shared/list.scss
@@ -3,7 +3,6 @@
   align-items: center;
   color: color(light_blue);
   display: flex;
-  line-height: 1;
   margin-left: 8px;
 
   .icon.fa-angle-down {


src/js/views/patients/shared/list.scss
```diff
@@ -3,7 +3,6 @@
   align-items: center;
   color: color(light_blue);
   display: flex;
-  line-height: 1;
```
nmajor25, line 6: Removing this so the browser uses `line-height: normal` fixes the issue. And I couldn't find any negative side effects because of it.

paulfalgout: I like removing it, but it was added here: https://github.com/RoundingWell/care-ops-frontend/pull/837. Sure this is not breaking something elsewhere?

paulfalgout (approved):

paulfalgout:
Didn't realize this is only used in one place.  Not sure why we're just seeing this now.

PR #1417, opened 2025-03-31
Websocket manage create notifications from the service

paulfalgout:
Shortcut Story ID: [sc-60612]

## Summary by CodeRabbit

- **New Features**
  - Enhanced real-time updates and improved handling of user actions in patient workflows and worklist management.
  - Introduced a new `AdderApp` class for better app lifecycle management and message handling.

- **Refactor**
  - Streamlined internal subscription and messaging logic for smoother and faster interactions.
  - Improved action management process and event subscription logic.

- **Tests**
  - Updated automated tests to validate the new event handling, ensuring consistent performance and robust user experiences.

### DIFF
--- a/src/js/services/ws.js
+++ b/src/js/services/ws.js
@@ -4,6 +4,16 @@ import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
+const AdderApp = App.extend({
+  restartWithParent: false,
+  beforeStart({ model, dataParams }) {
+    return model.fetch({ data: dataParams });
+  },
+  onStart({ model, collection }) {
+    collection.add(model);
+  },
+});
+
 export default App.extend({
   HEART_BEAT_INTERVAL: 50000,
   channelName: 'ws',
@@ -14,6 +24,7 @@ export default App.extend({
     'add': 'add',
     'unsubscribe': 'unsubscribe',
     'trigger': 'triggerMessage',
+    'manage:add': 'manageAdd',
   },
 
   initialize({ url }) {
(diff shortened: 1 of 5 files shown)


src/js/services/ws.js
```diff
@@ -102,24 +103,28 @@ export default App.extend({
     if (!_TEST_ && this.resources.length) this._subscribe();
   },
 
+  triggerMessage(data, model) {
+    const channel = this.getChannel();
+    if (model) model.trigger('message', data);
+    if (data.resource) channel.trigger(`message:${ data.resource.type }`, data, model);
+    channel.trigger('message', data, model);
+  },
```
nmajor25, line 122: These don't need `return`s? Or `triggerMessage()` can do multiple `triggers` in one call?

src/js/services/ws.js
```diff
@@ -110,6 +121,21 @@ export default App.extend({
...
+
+    app.listenTo(channel, `message:${ type }`, (data, model) => {
+      if (collection.get(model)) return;
+
+      const appName = `${ model.type }-${ model.id }`;
+
+      if (app.isRunning() && app.getChildApp(appName)) return;
+
+      const adderApp = app.addChildApp(appName, AdderApp);
+      adderApp.start({ model, collection, dataParams });
```
nmajor25, line 135: I'm not understanding the need for creating the child app? Instead of just doing the `await fetch()` and then `collection.add(model)` here without the child app?

src/js/services/ws.js
paulfalgout: yep triggering multiple things on one call.  one on the model.. two on the service.. one that's generic and one that's resource specific.. this is similar to events on a model.. or model events on a collection.. model has change events but also has like `change:foo` events

src/js/services/ws.js
paulfalgout: the child app does a few jobs.  the naming makes it unique.  It handles whether or not the app is still fetching.. It also follows the lifecycle of the parent without issues.  it handles the parent app stopping or restarting.

src/js/services/ws.js
nmajor25: Gotcha 👍

nmajor25 (approved):

PR #1418, opened 2025-04-01
Support read only & report forms on the patient sidebar

nmajor25:
Shortcut Story ID: [sc-60600]

## Summary by CodeRabbit

- **New Features**
	- Introduced a view-only modal form with a clear "Done" action, providing a distinct, non-editable interface for users.
	- Enhanced patient sidebar functionality with a new read-only form and associated modal widgets.

- **Tests**
	- Expanded testing to validate that the view-only and report forms display correctly with disabled inputs and appropriate footer actions.

### DIFF
--- a/test/integration/patients/patient/sidebar.js
+++ b/test/integration/patients/patient/sidebar.js
@@ -44,6 +44,22 @@ context('patient sidebar', function() {
       },
     });
 
+    const testReadOnlyForm = getForm({
+      attributes: {
+        options: {
+          read_only: true,
+        },
+      },
+    });
+
+    const testReportForm = getForm({
+      attributes: {
+        options: {
+          is_report: true,
+        },
+      },
+    });
+
     cy
       .routesForPatientDashboard()
       .routeFormDefinition()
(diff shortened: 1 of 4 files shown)


test/integration/patients/patient/sidebar.js
```diff
@@ -386,6 +434,76 @@ context('patient sidebar', function() {
       .get('.modal')
       .should('not.exist');
 
+    cy
+      .routeForm(fx => {
+        fx.data = testReadOnlyForm;
+
+        return fx;
+      });
```
nmajor25: This doesn't work and causes the test to fail.

The first `cy.routeForm()` in this test loads a script reducer form. And then here I want to override that and route the `testReadOnlyForm`.

We can do that with the other intercepts like `routeAction`, `routeFlow`, etc, but it doesn't seem to work for `routeForm`. So still trying to figure that out.

paulfalgout: I believe all of the forms are bootstrapped.  We only load them individually when rendering a single form.

paulfalgout (approved):

PR #1419, opened 2025-04-03
Update cypress `settings.json` fixtures to contain all used org settings

nmajor25:
Shortcut Story ID: [sc-59938]

## Summary by CodeRabbit

- **New Features**
  - Introduced new customer-facing configuration options including a help link, manual patient creation toggle, and patient creation form settings.
  - Added new sorting options for improved worklist organization.
  - Enhanced patient sidebar widget configurations for a clearer display.

- **Bug Fixes**
  - Refined settings behavior to explicitly distinguish missing values, ensuring more reliable functionality.

- **Refactor**
  - Streamlined settings handling across patient and program interfaces for improved maintainability and performance.

### DIFF
--- a/src/js/services/settings.js
+++ b/src/js/services/settings.js
@@ -1,4 +1,4 @@
-import { get } from 'underscore';
+import { get, isUndefined } from 'underscore';
 import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
@@ -16,7 +16,7 @@ export default App.extend({
     const workspaceSettings = currentWorkspace.get('settings');
     const workspaceSetting = get(workspaceSettings, settingName);
 
-    if (workspaceSetting) return workspaceSetting;
+    if (!isUndefined(workspaceSetting)) return workspaceSetting;
 
     const setting = this.settings.get(settingName);
 
--- a/test/support/api/settings.js
+++ b/test/support/api/settings.js
@@ -5,12 +5,29 @@ import fxTestSettings from 'fixtures/test/settings';
 
 const TYPE = 'settings';
 
-Cypress.Commands.add('routeSettings', (mutator = _.identity) => {
-  const data = getResource(fxTestSettings, TYPE);
+function buildSettings(fixture, settingKey, settingValue) {
+  if (!settingKey) return fixture;
+
+  const doesSettingExist = _.find(fixture, { id: settingKey });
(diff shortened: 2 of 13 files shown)


test/support/api/settings.js
```diff
@@ -5,12 +5,25 @@ import fxTestSettings from 'fixtures/test/settings';
...
+    if (!doesSettingExist) throw new Error(`Setting with the ${ setting_key } key was not found in settings.json fixture.`);
+
+    data = _.map(data, item => {
+      if (item.id === setting_key) {
+        return { ...item, attributes: { value: setting_value } };
+      }
+
+      return item;
+    });
+  }
```
nmajor25: This will probably need some refactoring and/or be split off into a separate function. In a "get it working" state.

src/js/services/settings.js
```diff
@@ -16,7 +16,7 @@ export default App.extend({
     const workspaceSettings = currentWorkspace.get('settings');
     const workspaceSetting = get(workspaceSettings, settingName);
 
-    if (workspaceSetting) return workspaceSetting;
+    if (!isUndefined(workspaceSetting)) return workspaceSetting;
```
nmajor25, line 19: I think this is a bug I found where a workspace setting is sometimes set but doesn't take precedent as it should.

So in this settings scenario:

1. `manual_patient_creation: false` setting on the workspace level.
2. `manual_patient_creation: true` setting on the org level.

The workspace setting should be returned (`false`). 

But instead this `getSetting()` returns the org setting value because it's checking if the workspace setting is truthy. I think it should check if it's `undefined` instead, checking if the workspace setting exists or not.

test/support/api/settings.js
paulfalgout: yeah I think just some sort of `buildSettings(fixtures, settingKey, settingValue)` would work.  Is there a reason for the snake_case here?

Then `buildSettings` can just `if (!settingKey) return fixtures;`

paulfalgout (changes requested):
All makes sense to me.. just that one refactor and LGTM.  Makes the tests cleaner too, so wins all around

test/support/api/settings.js
nmajor25: No particular reason for the snake case, removed it.

I created a `buildSettings()` function and pushed that up 👍

paulfalgout (approved):

PR #1421, opened 2025-04-09
Normalize api query between the list and single action/flow gets

paulfalgout:
Previously we were relying on default API settings on the action/flow list GETs.  fields only worked with includes, but they were auto-applied includes.

If we’re explicit, then it is the same request for the list and the request we need to make to add a newly notified entity.

This is much cleaner and clearer than before.  So after this is released for a cycle we can likely safely remove the defaults for the action/flow lists.  @DrechslerDerek 

Shortcut Story ID: [sc-60405]

I'd like to get this into the release so I needed 👀 while Nick is out.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the worklist page’s data retrieval to provide more efficient and consistent information handling.
  - Adjusted the sorting logic to ensure field parameters are represented in a unified format.

- **Tests**
  - Updated URL parameter validations to align with the refined data and sorting logic for improved reliability.

### DIFF
--- a/src/js/apps/patients/worklist/worklist_app.js
+++ b/src/js/apps/patients/worklist/worklist_app.js
@@ -125,19 +125,25 @@ export default App.extend({
     this.showView();
   },
   beforeStart() {
-    const listType = this.getState().getType();
+    const isFlowType = this.getState().isFlowType();
+    const entityRequest = isFlowType ? 'fetch:flows:collection' : 'fetch:actions:collection';
+
+    const includes = ['patient', ...this.sortOptions.getInclude()];
+    const fields = { patients: ['first_name', 'last_name', 'patient-fields'] };
     this.filters = this.getState().getEntityFilter();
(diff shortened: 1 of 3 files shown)


src/js/apps/patients/worklist/worklist_app.js
```diff
@@ -125,19 +125,25 @@ export default App.extend({
...
+      fields,
+      include: includes.join(','),
     };
 
-    if (listType === 'actions') {
-      data.fields = { flows: ['name', 'state'] };
-    }
-
-    return Radio.request('entities', `fetch:${ listType }:collection`, { data });
+    return Radio.request('entities', entityRequest, { data: this.query });
```
shadowhand, line 146: Does this mean that FE will now be sending explicit `include` and `fields` ?! 😮

zfixler (approved):

PR #1425, opened 2025-04-15
Remove top/bottom padding for forms rendered inside modals

nmajor25:
Shortcut Story ID: [sc-61035] [sc-60670]

## Summary by CodeRabbit

- **New Features**
	- The form app now automatically detects when it is opened in a modal and adjusts its appearance accordingly.
- **Style**
	- Improved layout for modal views by updating padding when displayed in a modal window.

### DIFF
--- a/src/scss/formapp-core.scss
+++ b/src/scss/formapp-core.scss
@@ -4,4 +4,8 @@
 
 body {
   padding: 24px;
+
+  &.is-modal {
+    padding: 0 24px;
+  }
 }
--- a/src/js/formapp.js
+++ b/src/js/formapp.js
@@ -340,6 +340,10 @@ function startFormApp() {
     </div>
     <div class="loader__text js-loading">${ intl.regions.preload.loading }</div>
   `);
+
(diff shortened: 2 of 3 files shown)


src/scss/formapp-core.scss
```diff
@@ -4,4 +4,8 @@
 
 body {
   padding: 24px;
+
+  &.is-modal {
+    padding: 0 24px;
+  }
```
nmajor25, line 10: This removes only the top/bottom padding. I'm assuming we still want the left/right padding?

src/scss/formapp-core.scss
nmajor25, line 1: @zfixler how do we ensure this doesn't break any other current form modals?

paulfalgout (approved):
If @zfixler likes it :shipit:

src/scss/formapp-core.scss
zfixler: I can't think of a reason it would break anything. There are also almost no form modals in use.

zfixler (approved):

PR #1426, opened 2025-04-16
Handle flow 'ResourceDeleted' web socket notifications on lists

nmajor25:
Shortcut Story ID: [sc-61078]

And update tests to include 'ResourceDeleted' ws notifications for both flows & actions in lists.

## Summary by CodeRabbit

- **New Features**
  - The application now responds to real-time deletion notifications, automatically removing deleted items from lists when notified via WebSocket.
- **Tests**
  - Enhanced integration tests to verify that lists update correctly when resources are deleted through WebSocket notifications.

### DIFF
--- a/src/js/entities-service/entities/flows.js
+++ b/src/js/entities-service/entities/flows.js
@@ -20,6 +20,9 @@ const _Model = BaseModel.extend({
     DetailsChanged({ attributes }) {
       this.set(attributes);
     },
+    ResourceDeleted() {
+      this.destroy({ isDeleted: true });
+    },
   },
   urlRoot() {
     if (this.isNew()) {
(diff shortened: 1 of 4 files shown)


src/js/entities-service/entities/flows.js
```diff
@@ -20,6 +20,9 @@ const _Model = BaseModel.extend({
     DetailsChanged({ attributes }) {
       this.set(attributes);
     },
+    ResourceDeleted() {
+      this.destroy({ isDeleted: true });
+    },
```
nmajor25, line 25: Like we currently do for actions, flows will be removed from the list when we get a `ResourceDeleted` ws notification.

nmajor25: When a flow `ResourceDeleted` happens on the patient flow page, we'll also get `ResourceDeleted` notifications for each of the flow's child actions. 

So in that scenario, each action in the list is removed showing `No Actions` and the flow remains visible to the user. And any changes made to that flow will be reflected in the UI but the api request will fail silently.

paulfalgout (approved):

PR #1427, opened 2025-04-17
Handle `FlowProgressChanged` ws notifications for flows in lists

nmajor25:
Shortcut Story ID: [sc-61108]

A flow's progress is calculated as `child actions completed / total child actions = flow progress %`. So this calculation can change if a child action's state changes or if a child action is created or deleted.

Corresponding BE PR: https://github.com/RoundingWell/care-ops-backend/pull/9338.

- [x] Handle `FlowProgressChanged` ws notifications
- [x] Add cypress tests
- [x] Test in local dev environment

### DIFF
--- a/src/js/entities-service/entities/flows.js
+++ b/src/js/entities-service/entities/flows.js
@@ -20,6 +20,9 @@ const _Model = BaseModel.extend({
     DetailsChanged({ attributes }) {
       this.set(attributes);
     },
+    FlowProgressChanged({ attributes }) {
+      this.set({ _progress: attributes.progress });
+    },
     ResourceDeleted() {
       this.destroy({ isDeleted: true });
     },
--- a/test/integration/patients/patient/archive.js
+++ b/test/integration/patients/patient/archive.js
@@ -442,6 +442,12 @@ context('patient archive page', function() {
         patient: getRelationship(testPatient),
         owner: getRelationship(teamCoordinator),
       },
+      meta: {
+        progress: {
+          complete: 0,
+          total: 2,
+        },
+      },
     });
 
     cy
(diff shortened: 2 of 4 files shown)


src/js/entities-service/entities/flows.js
```diff
@@ -20,6 +20,9 @@ const _Model = BaseModel.extend({
     DetailsChanged({ attributes }) {
       this.set(attributes);
     },
+    FlowProgressChanged({ attributes }) {
+      this.set({ _progress: attributes.progress });
```
nmajor25, line 24: I don't want the flow's `updated_at` value to change here. So I'm ignoring the rest of the attributes and only updating `_progress`.

paulfalgout (approved):

src/js/entities-service/entities/flows.js
nmajor25: @DrechslerDerek @shadowhand  is this correct? Does a flow's `updated_at` timestamp change when its `meta.progress` changes?

DrechslerDerek: It shouldn't because it's not updating the flow, it's updating the action. Same for adding/deleting an action that has that `flow_id` but the flow in the database doesn't change or get a new `updated_at`

DrechslerDerek: But in that WS message it should just have actionsTotal and actionsComplete, no other data unless you fetch the Flow, which I guess you are?

nmajor25: @DrechslerDerek great, my assumption was that the `updated_at` should not be updated when the FE gets this notification. But wanted to make sure. Thanks for confirming.

For other ws notifications, we automatically update the flow's/action's `updated_at` timestamp. But for the `FlowProgressChanged` notification, we don't want to do that.

@paulfalgout should be good to go with this PR 👍

PR #1429, opened 2025-04-22
Refactor schedule tests to use json-api & uuid

nmajor25:
Shortcut Story ID: [sc-61177]

## Summary by CodeRabbit

- **Refactor**
  - Updated tests to use dynamic clinician references instead of hardcoded IDs for improved consistency.
  - Standardized action creation in tests using a helper function.
  - Improved test setup for team-based permissions with explicit clinician, team, and role relationships.
  - Simplified test data handling for better maintainability.

### DIFF
--- a/test/integration/patients/worklist/schedule.js
+++ b/test/integration/patients/worklist/schedule.js
@@ -1,6 +1,6 @@
 import _ from 'underscore';
 import dayjs from 'dayjs';
-import { v5 as uuid } from 'uuid';
+import { v4 as uuidv4, v5 as uuidv5 } from 'uuid';
 
 import formatDate from 'helpers/format-date';
 import { testDate, testDateAdd, testDateSubtract } from 'helpers/test-date';
@@ -146,7 +146,7 @@ context('schedule page', function() {
       })
       .routeWorkspacePatient(fx => {
         fx.data = getWorkspacePatient({
-          id: uuid(testPatient1.id, workspaceOne.id),
+          id: uuidv5(testPatient1.id, workspaceOne.id),
         });
 
         return fx;
@@ -747,7 +747,7 @@ context('schedule page', function() {
       },
       actionsSelected: {
         [testActions[0].id]: true,
-        '4444': true,
+        [uuidv4()]: true,
       },
     }));
 
@@ -1716,7 +1716,7 @@ context('schedule page', function() {
     });
 
     const testNonTeamMemberClincian = getClinician({
-      id: '3',
+      id: uuidv4(),
       attributes: {
         name: 'Non Team Member',
       },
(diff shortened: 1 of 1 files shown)


test/integration/patients/worklist/schedule.js
```diff
@@ -1705,10 +1705,27 @@ context('schedule page', function() {
...
   specify('bulk editing with work:team:manage permission', function() {
+    const testCurrentClinician = getCurrentClinician({
+      relationships: {
+        role: getRelationship(roleTeamEmployee),
+        team: getRelationship(teamNurse),
+      },
+    });
+
+    const testNonTeamMemberClincian = getClinician({
+      id: '3',
```
nmajor25: Leaving these and others in this test. I'm hoping we can just remove those as part of this work: https://app.shortcut.com/roundingwell/story/54617

test/integration/patients/worklist/schedule.js
nmajor25, line 1: The `'bulk editing with work:team:manage permission'` verifies that users with the `work:team:manage` permission can't bulk edit actions owned by another team or clinician on another team.

The clinician with `id: 2` in `routeWorkspaceClinicians()` wasn't being used for anything so I removed it. And refactored the rest of the test as well.

test/integration/patients/worklist/schedule.js
```diff
@@ -745,28 +727,33 @@ context('schedule page', function() {
...
       dateFilters: {
         dateType: 'due_date',
         selectedDate: null,
         selectedMonth: null,
         relativeDate: null,
       },
-      actionsSelected: { '1': true, '4444': true },
+      actionsSelected: {
+        [testActions[0].id]: true,
+        '4444': true,
```
paulfalgout: 4444 is also some sort of uuid?

test/integration/patients/worklist/schedule.js
paulfalgout: agreed though this could be `uuid()` or no?

paulfalgout (changes requested):
small stuff

test/integration/patients/worklist/schedule.js
nmajor25: Yes, I believe it's there to test the scenario where the user had it selected previously but the action no longer exists or was completed. And I switched to it be a `uuid()`.

test/integration/patients/worklist/schedule.js
nmajor25: Yes, I switched it to a `uuid()`.

test/integration/patients/worklist/schedule.js
```diff
@@ -1,6 +1,6 @@
 import _ from 'underscore';
 import dayjs from 'dayjs';
-import { v5 as uuid } from 'uuid';
+import { v4 as uuidv4, v5 as uuidv5 } from 'uuid';
```
nmajor25, line 3: We don't do this anywhere else, where both v4 & v5 imported with different naming. I'm not sure how else to do it...

paulfalgout (approved):

PR #1430, opened 2025-04-22
Refactor worklist tests to use uuid & reference ids directly when possible

nmajor25:
Shortcut Story ID: [sc-61178]

## Summary by CodeRabbit

- **Tests**
  - Improved test reliability by replacing hardcoded IDs with dynamic values and reusable test data.
  - Updated state filter tests to include new states and dynamic assertions.
  - Enhanced team-based permission tests with more realistic clinician scenarios.

- **Chores**
  - Added a new test constant for an additional state to support expanded test coverage.

### DIFF
--- a/test/integration/patients/worklist/worklist.js
+++ b/test/integration/patients/worklist/worklist.js
@@ -14,7 +14,7 @@ import { getPatient } from 'support/api/patients';
 import { getPatientField } from 'support/api/patient-fields';
 import { getFlow, getFlows } from 'support/api/flows';
 import { getCurrentClinician, getClinician } from 'support/api/clinicians';
-import { stateTodo, stateInProgress, stateDone } from 'support/api/states';
+import { stateTodo, stateInProgress, stateDone, stateUnableToComplete, stateThmgTransfered } from 'support/api/states';
 import { getWidget } from 'support/api/widgets';
 import { roleAdmin, roleEmployee, roleNoFilterEmployee, roleTeamEmployee } from 'support/api/roles';
 import { teamCoordinator, teamNurse } from 'support/api/teams';
@@ -22,8 +22,9 @@ import { workspaceOne } from 'support/api/workspaces';
 import { testForm } from 'support/api/forms';
 import { getComment } from 'support/api/comments';
 
+const currentClinician = getCurrentClinician();
+
 const testPatient1 = getPatient({
-  id: '1',
   attributes: {
     first_name: 'Test',
     last_name: 'Patient',
@@ -31,7 +32,6 @@ const testPatient1 = getPatient({
 });
 
 const testPatient2 = getPatient({
-  id: '2',
   attributes: {
     first_name: 'Other',
     last_name: 'Patient',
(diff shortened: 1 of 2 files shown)


test/support/api/states.js
```diff
@@ -20,6 +20,7 @@ export const stateTodo = _.find(states, { id: '22222' });
 export const stateInProgress = _.find(states, { id: '33333' });
 export const stateDone = _.find(states, { id: '55555' });
 export const stateUnableToComplete = _.find(states, { id: '66666' });
+export const stateThmgTransfered = _.find(states, { id: '77777' });
```
paulfalgout, line 23: I wonder if this whole state could be removed.    Not for this PR, but I think it might be done

test/integration/patients/worklist/worklist.js
```diff
@@ -4804,6 +4805,23 @@ context('worklist page', function() {
...
 
   specify('flows with work:team:manage permission', function() {
+    const testCurrentClinician = getCurrentClinician({
+      relationships: {
+        role: getRelationship(roleTeamEmployee),
+        team: getRelationship(teamCoordinator),
+      },
+    });
+
+    const testNonTeamMemberClinician = getCurrentClinician({
```
paulfalgout: why use `getCurrentClinician` and not `getClinician`?  We might want to make `getCurrentClinician` error if `id` is passed as a key

test/integration/patients/worklist/worklist.js
```diff
@@ -1964,38 +1959,40 @@ context('worklist page', function() {
...
-      .routeWorkspaceClinicians(fx => {
-        const relationships = {
+    const relationships = {
+      role: getRelationship(roleEmployee),
+    };
+
+    const testClinicians = [
+      getCurrentClinician({ relationships }),
+      getClinician({
+        id: 'test-clinician',
```
paulfalgout: test-clinician?

test/integration/patients/worklist/worklist.js
```diff
@@ -1823,32 +1816,34 @@ context('worklist page', function() {
...
-          getCurrentClinician({ relationships }),
+          testCurrentClinician,
           getClinician({
             id: 'test-clinician',
             attributes: { name: 'Test Clinician' },
             relationships,
           }),
           getClinician({
-            id: '1',
+            id: uuid(),
```
paulfalgout, line 1833: not sure, but it seems like part of a refactor will be just making `getClinician` auto use `uuid`..  or perhaps it's `getNewClinician` or something..  I understand that with clinicians because they're bootstrapped, many places will need to `getClinician` and pull from the list of ids in the workspace.  We should be able to accommodate that, and also not require passing a `uuid` to every helper when we're setting the bootstrapped clinicians.

test/integration/patients/worklist/worklist.js
```diff
@@ -1823,32 +1816,34 @@ context('worklist page', function() {
       role: getRelationship(roleEmployee),
     };
 
+    const testCurrentClinician = getCurrentClinician({ relationships });
+
     cy
       .routeWorkspaceClinicians(fx => {
         fx.data = [
-          getCurrentClinician({ relationships }),
+          testCurrentClinician,
```
paulfalgout, line 1830: `id: 'test-clinician'`

paulfalgout (changes requested):
couple of small things.  in theory once all of these are done we can replace the ids in `api/states` and probably roles etc with `uuid`s instead of `77777` etc. That could be a decent way to see where we still have hardcoded ids.

test/support/api/states.js
nmajor25: I made a new story for using uuids for the state & roles fixtures [here](https://app.shortcut.com/roundingwell/story/61215/in-tests-have-roles-states-fixtures-use-uuid-ids). And I included a note to look into removing this.

There's also a `Evernorth` done state with `id: '88888'` that could potentially be removed.

test/integration/patients/worklist/worklist.js
nmajor25: I updated this to use `uuid()`.

test/integration/patients/worklist/worklist.js
nmajor25: I updated this to use `uuid()`.

test/integration/patients/worklist/worklist.js
nmajor25: It's supposed to be `getClinician()`, my mistake. I updated it just now.

paulfalgout (approved):

PR #1433, opened 2025-04-29
Use uuids for roles & states fixtures

nmajor25:
Shortcut Story ID: [sc-61215]

## Summary by CodeRabbit

- **Chores**
  - Updated test data to use UUIDs instead of numeric string IDs for roles and states.
  - Added a "slug" field to each state in test data.
  - Corrected spelling errors in role and state names for consistency.
  - Changed test logic to reference roles and states by name or slug instead of hardcoded IDs for improved reliability and clarity.

### DIFF
--- a/test/support/api/roles.js
+++ b/test/support/api/roles.js
@@ -15,12 +15,18 @@ export function getRoles() {
 
 const roles = getRoles();
 
-export const roleAdmin = _.find(roles, { id: '22222' });
-export const roleManager = _.find(roles, { id: '11111' });
-export const roleEmployee = _.find(roles, { id: '33333' });
-export const roleReducedEmployee = _.find(roles, { id: '44444' });
-export const roleNoFilterEmployee = _.find(roles, { id: '66666' });
-export const roleTeamEmployee = _.find(roles, { id: '77777' });
+function getRoleByName(name) {
+  return _.find(roles, role => {
+    return role.attributes.name === name;
+  });
+}
+
+export const roleAdmin = getRoleByName('admin');
+export const roleManager = getRoleByName('manager');
+export const roleEmployee = getRoleByName('employee');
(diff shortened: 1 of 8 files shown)


test/support/api/roles.js
nmajor25, line 1: This seems like a simple way to do it. Hardcode a uuid for each state/role fixture. And then when we need to export individual resources, we just find them by their name attribute.

But maybe there is a way to do this without hardcoding the uuid in the fixture files.

nmajor25:
We need to do the same thing for the `teams` fixtures, but that one is more complicated. So I'm going to do that in a separate [story](https://app.shortcut.com/roundingwell/story/61276/use-uuids-for-teams-fixtures-in-cypress-tests).

test/support/api/states.js
```diff
@@ -15,12 +15,17 @@ export function getState() {
...
 
-// Exporting one of each state.status
-export const stateTodo = _.find(states, { id: '22222' });
-export const stateInProgress = _.find(states, { id: '33333' });
-export const stateDone = _.find(states, { id: '55555' });
-export const stateUnableToComplete = _.find(states, { id: '66666' });
-export const stateThmgTransfered = _.find(states, { id: '77777' });
+function getStateByName(name) {
+  return _.find(states, state => {
+    return state.attributes.name === name;
```
shadowhand: The equivalent of `role.name` is actually `state.slug`. Annoying discrepancy, but there it is.

paulfalgout: yeah switching to state.slug (and adding that value to the fixtures) is better

shadowhand: Realized my typo, corrected. State slugs are enforced as unique, just like role name.

paulfalgout (approved):

nmajor25:
Looks like tests passed, but upload to coveralls failed...

PR #1434, opened 2025-04-30
Use explicit query for action collection fetch

paulfalgout:
This should cover all cases currently

Shortcut Story ID: [sc-61313]

## Summary by CodeRabbit

- **New Features**
  - Patient details (first name and last name) are now included when viewing schedules, providing more comprehensive information alongside flow data.

- **Tests**
  - Enhanced tests to verify that patient information is correctly requested and included in schedule-related API calls.

### DIFF
--- a/test/integration/patients/worklist/reduced-schedule.js
+++ b/test/integration/patients/worklist/reduced-schedule.js
@@ -100,6 +100,7 @@ context('reduced schedule page', function() {
       .wait('@routeActions')
       .itsUrl()
       .its('search')
+      .should('contain', 'include=patient')
       .should('contain', `filter[clinicians]=${ currentClinician.id }`)
       .should('contain', `filter[states]=${ stateTodo.id },${ stateInProgress.id }`)
       .should('contain', 'fields[flows]=name,state');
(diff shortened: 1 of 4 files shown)


test/integration/patients/worklist/reduced-schedule.js
```diff
@@ -100,6 +100,7 @@ context('reduced schedule page', function() {
       .wait('@routeActions')
       .itsUrl()
       .its('search')
+      .should('contain', 'include=patient')
       .should('contain', `filter[clinicians]=${ currentClinician.id }`)
       .should('contain', `filter[states]=${ stateTodo.id },${ stateInProgress.id }`)
       .should('contain', 'fields[flows]=name,state');
```
nmajor25, line 106: Should we check `.should('contain', 'fields[patients]=first_name,last_name')` as well?

paulfalgout: we could.. I'm on the fence.. I'm not even convinced testing the include makes a ton of sense.  I think the `fields[flow]` check here comes from copy-paste of the worklist.. where it exists for actions but not flows.. but these things are just static.. no logic determines if these things show up..  which isn't necessarily a reason not to test it.. but might be overkill.. thoughts?

nmajor25: I usually like to put things in the tests to stop us from deleting something in the future and having unintended consequences. But that doesn't apply to this. It's just static without any logic. 

So ya... I'm on the fence now too haha. Probably fine without it?

nmajor25 (approved):

PR #1436, opened 2025-04-30
Use uuids for team fixtures

nmajor25:
Shortcut Story ID: [sc-61276]

## Summary by CodeRabbit

- **Chores**
	- Updated team identifiers in test data to use UUIDs instead of numeric strings.
	- Improved test utilities to support team lookup by abbreviation as well as by identifier.
	- Adjusted test constants to use abbreviation-based lookup for teams.
	- Fixed form data source to correctly pass team identifiers as strings.
	- Removed editor relationship data from event action test fixtures.
	- Enhanced test assertions to dynamically verify dashboard URLs using UUIDs.

### DIFF
--- a/test/fixtures/test/events-actions.json
+++ b/test/fixtures/test/events-actions.json
@@ -1,14 +1,8 @@
 [
   {
-    "id": "11111",
+    "id": "108cbc05-a26f-4f8e-91ed-13662ba4d322",
     "date": "2019-07-06T17:45:25.360Z",
     "event_type": "ActionCreated",
-    "source": "system",
-    "_relationships": {
-      "editor": {
-        "id": "11111",
-        "type": "clinicians"
-      }
-    }
+    "source": "system"
   }
 ]
--- a/test/support/api/teams.js
+++ b/test/support/api/teams.js
@@ -10,7 +10,7 @@ const TYPE = 'teams';
 function getFixture(data) {
   if (!data) return _.sample(fxTestTeams);
 
-  if (data.name) return _.find(fxTestTeams, { name: data.name });
+  if (data.abbr) return _.find(fxTestTeams, { abbr: data.abbr });
 
   if (!data.id) data.id = _.sample(fxTestTeams).id;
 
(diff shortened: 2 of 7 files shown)


nmajor25:
The `getClinicians({ teamId: id })` form method is failing in the tests for some reason. Need to fix that.

test/fixtures/test/events-actions.json
```diff
@@ -1,14 +1,8 @@
...
-    "id": "11111",
+    "id": "108cbc05-a26f-4f8e-91ed-13662ba4d322",
     "date": "2019-07-06T17:45:25.360Z",
     "event_type": "ActionCreated",
-    "source": "system",
-    "_relationships": {
-      "editor": {
-        "id": "11111",
-        "type": "clinicians"
-      }
```
nmajor25, line 11: This is already set to the current clinician in the `getActivities()` api helper function.

nmajor25:
I snuck in a small change to use uuids in the event and action event fixture files.

Only thing left now is the clinician ids. The final boss... which I'll do in a separate PR.

test/support/api/teams.js
```diff
@@ -38,19 +40,19 @@ export function getTeam(data, { depth = 0 } = {}) {
...
+  return teams;
 }
 
 // Exporting only teams needed for testing variance
-export const teamCoordinator = getTeamResource({ id: '11111' });
-export const teamNurse = getTeamResource({ id: '22222' });
-export const teamOther = getTeamResource({ id: '77777' });
+export const teamCoordinator = getTeamResource({ name: 'Coordinator' });
+export const teamNurse = getTeamResource({ name: 'Nurse' });
+export const teamOther = getTeamResource({ name: 'Other' });
```
shadowhand: Like states, the `name` is not a unique value here. It would be better to use `abbr`, which is treated as a secondary unique ID.

nmajor25: Ah okay good call, updated to that.

paulfalgout (approved):

PR #1437, opened 2025-05-01
Refactor clinician fixture data & `getClinician()` helper function in tests

nmajor25:
Shortcut Story ID: [sc-54617]

Two overarching changes to the tests:

1. Use hardcoded uuids in the clinician json fixture file.
2. Make it so `getClinician()` no longer requires that an `id` be sent with the data. It will automatically generate a `id: uuid()` if it's not included. 

## Summary by CodeRabbit

- **Chores**
  - Simplified test data setup by removing explicit ID assignments for clinicians in multiple test suites. Tests now rely on default or internally generated IDs.
  - Updated clinician fixture data to use UUIDs instead of numeric string IDs.
  - Improved test data helpers to automatically generate UUIDs for clinicians when not provided, reducing manual setup.
  - No changes to application logic or user-facing features.

### DIFF
--- a/src/js/views/patients/shared/components/owner_component.cy.js
+++ b/src/js/views/patients/shared/components/owner_component.cy.js
@@ -183,7 +183,7 @@ context('Owner Component', function() {
   });
 
   specify('Current User is not in workspaces', function() {
-    const notInWorkspace = getClinician({ id: '12345', attributes: { name: 'Not in Workspace' } });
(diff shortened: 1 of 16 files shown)


paulfalgout (approved):

paulfalgout:
the e2e tests did pass, but the check didn't update.  you can see the success in the test-build

PR #1438, opened 2025-05-01
Use the QuickSight AWS SDK to load dashboard iframe content

nmajor25:
Shortcut Story ID: [sc-61175]

## Summary by CodeRabbit

- **New Features**
  - Dashboards are now embedded using Amazon QuickSight's official embedding SDK, providing a more dynamic and integrated experience.
- **Bug Fixes**
  - Ensured dashboard IDs and embed URLs are consistently matched in test fixtures.
- **Tests**
  - Updated integration tests to accommodate the new embedding method and mock external QuickSight requests for reliable testing.

### DIFF
--- a/src/js/views/dashboards/dashboard_views.js
+++ b/src/js/views/dashboards/dashboard_views.js
@@ -1,5 +1,6 @@
 import Radio from 'backbone.radio';
 import hbs from 'handlebars-inline-precompile';
+import * as QuicksightEmbedding from 'amazon-quicksight-embedding-sdk';
 
 import { View } from 'marionette';
 
@@ -25,7 +26,20 @@ const ContextTrailView = View.extend({
 
 const IframeView = View.extend({
   className: 'flex-grow',
-  template: hbs`<iframe src="{{ embed_url }}"></iframe>`,
+  template: hbs`<div class="js-dashboard-container"></div>`,
+  ui: {
+    dashboardContainer: '.js-dashboard-container',
+  },
+  onRender() {
+    QuicksightEmbedding.createEmbeddingContext().then((embeddingContext => {
+      embeddingContext.embedDashboard({
+        url: this.model.get('embed_url'),
+        container: this.el,
+        height: '100%',
+        width: '100%',
+      });
+    }));
+  },
 });
 
 const LayoutView = View.extend({
--- a/src/js/views/dashboards/dashboard_views.js
+++ b/src/js/views/dashboards/dashboard_views.js
@@ -26,11 +26,8 @@ const ContextTrailView = View.extend({
 
 const IframeView = View.extend({
   className: 'flex-grow',
-  template: hbs`<div class="js-dashboard-container"></div>`,
-  ui: {
-    dashboardContainer: '.js-dashboard-container',
-  },
-  onRender() {
+  template: false,
+  initialize() {
     QuicksightEmbedding.createEmbeddingContext().then((embeddingContext => {
       embeddingContext.embedDashboard({
         url: this.model.get('embed_url'),
--- a/test/integration/dashboards/dashboard.js
+++ b/test/integration/dashboards/dashboard.js
@@ -6,10 +6,13 @@ import { getDashboard } from 'support/api/dashboards';
 
 context('dashboard', function() {
   specify('display dashboard', function() {
+    const testDashboardId = uuid();
+
     const testDashboard = getDashboard({
+      id: testDashboardId,
       attributes: {
         name: 'Test Dashboard',
-        embed_url: '/test_dashboard',
+        embed_url: `https://us-west-2.quicksight.aws.amazon.com/embed/embed_id/dashboards/${ testDashboardId }?identityprovider=quicksight`,
       },
     });
 
@@ -19,6 +22,9 @@ context('dashboard', function() {
         fx.data = testDashboard;
 
         return fx;
+      })
+      .intercept('GET', 'https://*.quicksight.aws.amazon.com/**', req => {
+        req.reply('<html><body class="test-iframe-content">Test Iframe Content</body></html>');
       })
       .visit(`/dashboards/${ testDashboard.id }`)
       .wait('@routeDashboard');
@@ -31,7 +37,8 @@ context('dashboard', function() {
     cy
       .get('.dashboard__frame')
       .find('.dashboard__iframe iframe')
-      .should('have.attr', 'src', '/test_dashboard');
+      .should('have.attr', 'src')
+      .and('include', `https://us-west-2.quicksight.aws.amazon.com/embed/embed_id/dashboards/${ testDashboardId }?`);
 
     cy
       .get('.dashboard__frame')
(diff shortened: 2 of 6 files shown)


src/js/views/dashboards/dashboard_views.js
nmajor25, line 1: Not sure if this will work until it's on QA or some other environment. The local dev environment has one dashboard, but it doesn't have an embed URL.

Maybe we could create a QuickSight dashboard with dummy data (or no data) or something that doesn't require authentication to view? Otherwise we can't test it locally.

nmajor25, line 1: I'm not sure what to do about the tests yet.

Before we could just use any dummy `embed_url` and just check that the `iframe` had the correct url. Even though the `iframe` would be empty.

But we can't do that when using this sdk. It throws an error if the embed url is invalid.

Screenshot 2025-05-01 at 4 30 52 PM

src/js/views/dashboards/dashboard_views.js
```diff
@@ -24,8 +25,21 @@ const ContextTrailView = View.extend({
...
+  onRender() {
+    QuicksightEmbedding.createEmbeddingContext().then((embeddingContext => {
+      embeddingContext.embedDashboard({
+        url: this.model.get('embed_url'),
+        container: this.ui.dashboardContainer,
+        height: '100%',
+        width: '100%',
+      });
+    }));
+  },
```
nmajor25, line 39: Not really sure what `createEmbeddingContext()` does, but it needs to be initiated first before the dashboard embed can happen...

src/js/views/dashboards/dashboard_views.js
```diff
@@ -24,8 +25,21 @@ const ContextTrailView = View.extend({
...
+  className: 'flex-grow quicksight-dashboard',
+  template: hbs`<div class="js-dashboard-container"></div>`,
+  ui: {
+    dashboardContainer: '.js-dashboard-container',
+  },
+  onRender() {
+    QuicksightEmbedding.createEmbeddingContext().then((embeddingContext => {
+      embeddingContext.embedDashboard({
+        url: this.model.get('embed_url'),
+        container: this.ui.dashboardContainer,
```
paulfalgout: you might be able to just use `this.el` and set template to `false`

src/js/views/dashboards/dashboard_views.js
```diff
@@ -24,8 +25,21 @@ const ContextTrailView = View.extend({
 });
 
 const IframeView = View.extend({
-  className: 'flex-grow',
-  template: hbs`<iframe src="{{ embed_url }}"></iframe>`,
+  className: 'flex-grow quicksight-dashboard',
```
paulfalgout: needed added class?

src/js/views/dashboards/dashboard_views.js
paulfalgout: I think this'll work.  good ole chatgpt..  but either way stubbing the library API is fine here.  
```js
cy
  .intercept('https://*.quicksight.aws.amazon.com/**', req => {
    req.reply('<html><body class="test-iframe">test iframe</body></html>');
  });

cy
  .window()
  .then(win => {
    // simulate the postMessage handshake QuickSight would send
    win.postMessage({ event: 'QS_EVENT_HANDSHAKE' }, '*');
    win.postMessage({ event: 'QS_EVENT_READY' }, '*');
  });
```

paulfalgout (changes requested):

src/js/views/dashboards/dashboard_views.js
```diff
@@ -25,7 +26,20 @@ const ContextTrailView = View.extend({
 
 const IframeView = View.extend({
   className: 'flex-grow',
-  template: hbs`<iframe src="{{ embed_url }}"></iframe>`,
+  template: hbs`<div class="js-dashboard-container"></div>`,
```
paulfalgout: I think template can be false and no ui necessary?

paulfalgout (changes requested):

src/js/views/dashboards/dashboard_views.js
nmajor25: The intercept worked well 👍 . I don't think the postmessage simulation is necessary though...

src/js/views/dashboards/dashboard_views.js
nmajor25: Nope, removed it 👍

src/js/views/dashboards/dashboard_views.js
nmajor25: `template: false` didn't work. But using an empty `hbs''` worked?

test/integration/dashboards/dashboard.js
```diff
@@ -31,7 +31,8 @@ context('dashboard', function() {
     cy
       .get('.dashboard__frame')
       .find('.dashboard__iframe iframe')
-      .should('have.attr', 'src', '/test_dashboard');
+      .should('have.attr', 'src')
+      .and('include', `https://us-west-2.quicksight.aws.amazon.com/embed/embed_id/dashboards/${ testDashboard.id }?`);
```
nmajor25, line 36: I want to also test that the `<body>` content in the iframe is successfully loaded. But I'm having a hard time getting cypress to see/grab it. So I'm still working on that.

src/js/views/dashboards/dashboard_views.js
```diff
@@ -26,7 +26,7 @@ const ContextTrailView = View.extend({
 
 const IframeView = View.extend({
   className: 'flex-grow',
-  template: hbs`<div class="js-dashboard-container"></div>`,
+  template: hbs``,
   ui: {
```
paulfalgout: there's no content for the `ui` here.  Also `template: false` should work here, but views without templates aren't "rendered" so you'd just change `onRender` to `initialize`

paulfalgout (changes requested):

src/js/views/dashboards/dashboard_views.js
nmajor25: Okay yep, updated to that and it works 👍

test/integration/dashboards/dashboard.js
nmajor25: I still am not sure how to do this. In cypress, the `<body>` inside the iframe is always empty.

Are we fine with what we have currently? Just testing that the iframe loads with the correct `src` attribute? If not, I can come back to this after we test the implementation on QA.

paulfalgout: I'm not worried about testing the content too much..  do you know if it does actually show it?

if so I think it's just that `cy.iframe()` helper we added for testing forms?

nmajor25: I can see the body content in the test ([screenshot](https://github.com/user-attachments/assets/f6fe580a-d428-49ee-b515-c9252b80e289)).

Yep, I tried `cy.iframe()`. I have no trouble getting the iframe. It's just that the `<body>` content comes back as empty. Cypress can't see it for some reason.

Interestingly, `cy.iframe()` was returning two iframes.

paulfalgout: ah.. I bet the sdk adds an iframe that we're picking up... fun.. well okie

paulfalgout (approved):

test/integration/dashboards/dashboard.js
nmajor25: Ya I'm also thinking it's something along those lines...

PR #1441, opened 2025-05-09
Use `scrollbar-gutter` css property for action/flow/patient sidebars

nmajor25:
Shortcut Story ID: [sc-61623]

To fix layout shift that happens as a result of the browser scrollbar.

Video screengrab of the fix, where layout no longer shifts in these sidebars:

https://github.com/user-attachments/assets/3451b1ba-d031-4c15-a8c7-cbf82670fe1d

## Summary by CodeRabbit

- **Style**
	- Reduced horizontal padding in sidebars for a more compact layout.
	- Improved sidebar stability by reserving space for scrollbars, preventing layout shifts when scrollbars appear or disappear.

### DIFF
--- a/src/js/views/patients/sidebar/patient/patient-sidebar.scss
+++ b/src/js/views/patients/sidebar/patient/patient-sidebar.scss
@@ -1,7 +1,8 @@
 .worklist-patient-sidebar {
   border-left: 1px solid #{$black-90};
   overflow-y: auto;
-  padding: 16px 20px;
+  padding: 16px 8px;
+  scrollbar-gutter: stable both-edges;
   width: 256px;
 }
 
--- a/src/scss/modules/sidebar.scss
+++ b/src/scss/modules/sidebar.scss
@@ -1,7 +1,8 @@
 .sidebar {
   border-left: 1px solid #{$black-90};
   overflow-y: auto;
-  padding: 16px 20px;
+  padding: 16px 12px;
+  scrollbar-gutter: stable both-edges;
   width: 400px;
 }
 


src/js/views/patients/sidebar/patient/patient-sidebar.scss
```diff
@@ -1,7 +1,8 @@
 .worklist-patient-sidebar {
   border-left: 1px solid #{$black-90};
   overflow-y: auto;
-  padding: 16px 20px;
+  padding: 16px 8px;
+  scrollbar-gutter: stable both-edges;
```
nmajor25, line 5: Doing just `scrollbar-gutter: stable` here will add extra padding to only the right side of the element (if no scrollbar is currently showing). 

So to make things evenly spaced on both sides when no scrollbar is showing, I reduced the right/left padding and added `both-edges`.

src/js/views/patients/sidebar/patient/patient-sidebar.scss
nmajor25, line 1: This is a relative easy fix to the layout shift in the action/flow/patient sidebars. But fixing the layout shift for the flow/action lists is more complicated (flow and patient dashboard pages).

We'd need to apply `scrollbar-gutter: stable both-edges;` to `.app-frame__content`. And then adjust the right/left padding for each of it's children elements, on each page. Or have `.app-frame__content` control the right/left padding instead of the children elements...

But maybe those broad changes won't be necessary if we switch to having the list scroll independently on the flow/patient-dashboard pages.

src/scss/modules/sidebar.scss
```diff
@@ -1,7 +1,8 @@
 .sidebar {
   border-left: 1px solid #{$black-90};
   overflow-y: auto;
-  padding: 16px 20px;
+  padding: 16px 12px;
+  scrollbar-gutter: stable both-edges;
```
nmajor25, line 5: Also this property only started to be supported by Safari around 6 months ago. So there could be a good percentage of users who won't have this supported.

Nothing would break for them. But the right/left padding would be tight at `12px`.

nmajor25:
@paulfalgout  good to merge this one on it's own? And tackle the other related stuff in a different PR? I think this one should be good on its own.

paulfalgout (approved):

PR #1442, opened 2025-05-13
Remove form preview from the app

nmajor25:
Shortcut Story ID: [sc-61636]

Currently form preview is only used in the programs admin (which doesn't get much use). And most forms are too complex to preview.

Removes all form preview related code besides what's in the `formapp.js` file. That file is under active development so that code will be removed at a later time.

This PR also includes a few miscellaneous test updates.

## Summary by CodeRabbit

- **Bug Fixes**
  - Disabled form preview buttons throughout the application, making them non-interactive.
  - Removed the ability to preview forms from program flows, workflows, and sidebars.

- **Tests**
  - Deleted all tests related to form preview functionality.
  - Updated integration tests to reflect the removal of form preview and to check for disabled form preview buttons.

- **Documentation**
  - Removed localization strings related to form preview.

- **Chores**
  - Cleaned up code by removing all components, event handlers, service methods, and logic associated with form preview features.


src/js/views/programs/program/flow/action-item.hbs
```diff
@@ -12,7 +12,7 @@
     <span class="table-list__meta" data-due-region></span>{{~ remove_whitespace ~}}
     <span class="table-list__meta">
       {{#if hasForm}}
-        <button class="js-form button-secondary--compact">{{far "square-poll-horizontal"}}</button>
+        <button class="button-secondary--compact" disabled>{{far "square-poll-horizontal"}}</button>
```
nmajor25, line 15: I switched the form icon for action list items to be `disabled`.

Instead we could also just show a form icon without button styling or keep the button enabled and open the action sidebar when the form button is clicked.

paulfalgout (changes requested):
the     'fetch:form': 'fetchForm', from the form service won't be used anymore.. so we need to strip that too.

paulfalgout (approved):
Nice bit of clean up

PR #1443, opened 2025-05-14
Scroll patient sidebar independently

nmajor25:
- Shortcut Story ID: [sc-61626] [sc-61625]

Needs to be done on:

- [x] Form page
- [x] Patient dashboard page
- [x] Patient archive page
- [x] Flow page

## Summary by CodeRabbit

- **Style**
  - Improved layout consistency and spacing across patient-related views and sidebars.
  - Added new container classes to enhance structure and scrolling behavior for patient lists and actions.
  - Set key frames and layouts to fill the viewport height for a more consistent appearance.
  - Adjusted sidebar padding and introduced stable scrollbar gutters to prevent layout shifts during scrolling.

### DIFF
--- a/src/js/views/patients/patient/dashboard/layout.hbs
+++ b/src/js/views/patients/patient/dashboard/layout.hbs
@@ -1,4 +1,4 @@
-<div>
+<div class="patient__tabs">
   <span class="patient__tab--selected">
     {{~ @intl.patients.patient.dashboard.dashboardViews.dashboardBtn ~}}
   </span>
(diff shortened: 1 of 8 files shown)


src/js/views/patients/patient/dashboard/layout.hbs
nmajor25, line 1: Screengrab of when both the list and sidebar are scrollable:

Screenshot 2025-05-14 at 4 08 23 PM

nmajor25: I still need to figure out how to align the header sections and list section when the list scrollbar is visible. It's a tricky one. 

We handle a similar situation in the worklist via a [fixWidth()](https://github.com/RoundingWell/care-ops-frontend/blob/123bf45c02c734db5844f648fac031ca371848ed/src/js/views/patients/worklist/worklist_views.js#L65-L75) function that runs whenever a list item is attached/rendered or the browser window is resized.

nmajor25: And there should probably be some padding to the right of the list scrollbar too when it's visible.

paulfalgout: we should remove fixWidth entirely.  the gutter css is made to fix this hack

https://chatgpt.com/share/682584a3-76c0-8005-b92e-a63b2bb0ef84

nmajor25: Interesting, putting the header and list in the same scroll box. Put scrollbar-gutter on that. And then making the header elements sticky... that could work...

nmajor25: This works functionally, but the alignment/spacing will be off when either the list scrollbar is shown or hidden.

The parent container has a scrollbar gutter that's applied to the header and list. But then the list child element still needs to scroll independently (i.e. extra padding added). I don't know how to make it so the padding is the same across the header and list whether the list is scrollable or not.

nmajor25, line 1: New screenshot of the patient dashboard with both list and pt sidebar scrollable. With padding/layout maintained. And no more layout shifts on load or when list or sidebar becomes scrollable.

Screenshot 2025-05-16 at 4 57 08 PM

paulfalgout (approved):

PR #1454, opened 2025-05-30
Remove babel inline hbs transform

nmajor25:
Shortcut Story ID: [sc-61861]

## Summary by CodeRabbit

- **Chores**
	- Updated build and test configuration to use a new Handlebars template inlining plugin for Vite.
	- Removed an unused Babel plugin and its dependency.
	- Improved dependency management for template compilation during development and testing.


config/hbs-stub.js
nmajor25, line 1: This is the alias path file for whenever `import hbs from 'handlebars-inline-precompile'` is added to an app file.

nmajor25: And I'm hoping there is a better way to do this...

paulfalgout: what caused this to be needed?  I can imagine, but with the exact issue maybe I can help find an alternate solution?  the plugin doesn't currently solve for this?

package.json
```diff
@@ -163,22 +163,16 @@
...
-    "@babel/core": "^7.15.5",
     "@cypress/mount-utils": "^4.0.0",
     "@cypress/vite-dev-server": "^6.0.0",
     "@eslint/eslintrc": "^3.1.0",
     "@eslint/js": "^9.5.0",
     "@faker-js/faker": "^9.7.0",
     "@rollup/plugin-node-resolve": "^16.0.0",
-    "@roundingwellos/babel-plugin-handlebars-inline-precompile": "^3.0.1",
     "@stylistic/eslint-plugin-js": "^4.2.0",
-    "babel-plugin-istanbul": "^7.0.0",
```
paulfalgout, line 181: it might be better to just leave this and only replace the hbs stuff..  if that completely works, we could revisit removing babel and swapping the istanbul solution?

paulfalgout (changes requested):

config/hbs-stub.js
nmajor25: Without this, the app doesn't know what the `import hbs from 'handlebars-inline-precompile';` is.

Screenshot 2025-06-02 at 3 58 49 PM

nmajor25: I don't fully understand how [babel-plugin-handlebars-inline-precompile](https://github.com/jamiebuilds/babel-plugin-handlebars-inline-precompile/blob/master/src/index.js) works (what we are currently using in `develop`).

But I think it looks for the `import hbs from 'handlebars-inline-precompile';` import in a file, adds a `import Handlebars from 'handlebars/runtime'` import, and then pre-compiles each template it finds inside every `hbs'template'`.

nmajor25: But I need to rework the `/config/vite-plugin-inline-handlebars.js` plugin. I think the `import hbs from 'handlebars-inline-precompile';` needs to be removed from files on build, I'm currently not doing that.

nmajor25: Got something working, but lots of refactoring is needed still.

nmajor25: Everything is working locally, but the component tests are failing in circleci for some reason.

package.json
nmajor25: That's a good idea, the istanbul stuff can be in a future PR.

vite.config.js
```diff
@@ -7,11 +7,12 @@ import browserslistToEsbuild from 'browserslist-to-esbuild';
 
 import babel from 'vite-plugin-babel';
 import eslint from 'vite-plugin-eslint';
-import handlebars from './config/vite-plugin-handlebars-loader.js';
 import { nodeResolve } from '@rollup/plugin-node-resolve';
 import stylelint from 'vite-plugin-stylelint';
 import { VitePWA } from 'vite-plugin-pwa';
 import yaml from './config/vite-plugin-yaml.js';
+import handlebars from './config/vite-plugin-handlebars-loader.js';
+import inlineHbsCompile from './config/vite-plugin-inline-handlebars.js';
```
nmajor25, line 15: Should we rename these so they are more clear? One compiles `.hbs` files and the other compiles inline `hbs''` strings in `.js` files.

vite.config.js
```diff
@@ -110,6 +113,9 @@ export default defineConfig(({ mode }) => {
         },
       }),
     ],
+    optimizeDeps: {
+      exclude: ['handlebars-inline-precompile'],
+    },
```
nmajor25, line 118: For future reference, vite does [dependency pre bundling](https://vite.dev/guide/dep-pre-bundling.html). And it was causing issues when that process tried to resolve the `handlebars-inline-precompile` import before the inline hbs plugin could replace it. Since `handlebars-inline-precompile` is not an actual npm package.

paulfalgout (approved):

PR #1459, opened 2025-06-09
Switch to vite istanbul plugin

paulfalgout:
And remove babel

Shortcut Story ID: [sc-62084]

Much more difficult than expected, but this seems to work.

I'm not seeing significant perf improvements on CI.. I think I'm seeing better performance locally, but it could be in my head.

## Summary by CodeRabbit

- **Chores**
  - Updated code coverage tooling by replacing Babel-based instrumentation with Istanbul-based plugins and configurations.
  - Improved and streamlined test and build processes in the continuous integration workflow, including enhanced caching and expanded nightly browser test coverage.
  - Adjusted development dependencies and configuration files to support the new coverage setup.
- **Tests**
  - Refined code coverage inclusion and exclusion patterns for more targeted reporting.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,7 @@ executors:
     environment:
       DISPLAY: :99
       CYPRESS_DISPLAY: :99
+      CYPRESS_COVERAGE: true
       CYPRESS_CACHE_FOLDER: ~/.cache/Cypress
       CYPRESS_watchForFileChanges: false
 
@@ -25,22 +26,22 @@ jobs:
     parameters:
       cypress_command:
         type: string
-      start_command:
-        type: string
-        default: "npx vite preview -m test"
+      should_build:
+        type: boolean
+        default: true
       parallelism:
         type: integer
-        default: 10
+        default: 12
       upload_coverage:
         type: boolean
         default: true
     executor: cypress-all-browsers
     parallelism: << parameters.parallelism >>
     steps:
-      # 1 Checkout
+      # Checkout
       - checkout
 
-      # 2 Install & start virtual display
+      # Install & start virtual display
       - run:
           name: Install & start Xvfb
           command: |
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -164,6 +183,17 @@ workflows:
             --group firefox-e2e
             --config retries=4
 
+      - run-cypress:
```
paulfalgout, line 186: this also add component tests back to the nightly

.circleci/config.yml
```diff
@@ -25,12 +26,12 @@ jobs:
...
-      start_command:
-        type: string
-        default: "npx vite preview -m test"
+      should_build:
+        type: boolean
+        default: true
       parallelism:
         type: integer
-        default: 10
+        default: 12
```
paulfalgout, line 34: default parallel is now 12.. seems to be a sweet spot right now..

vite.config.js
```diff
@@ -42,14 +43,20 @@ const cypressConfig = defineConfig({
...
-    babel(),
     nodeResolve({
       modulePaths: [
         path.resolve('./node_modules'),
         path.resolve('./src'),
         path.resolve('./test'),
       ],
     }),
+    istanbul({
+      requireEnv: true,
```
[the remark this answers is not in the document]
paulfalgout: `requireEnv` is a boolean.  The code is `CYPRESS_COVERAGE` if `cypress: true` which it is.  We don't technically need to do this, but it aligns with the cypress coverage plugin.. might be a nice to have, although currently we get a little more our of our current plugin.

.circleci/config.yml
```diff
@@ -55,36 +56,39 @@
...
-      - restore_cache:
-          keys:
-            - vite-dist-{{ checksum "vite.config.js" }}
-      - run: npx vite build -m test
       - save_cache:
-          key: vite-dist-{{ checksum "vite.config.js" }}
-          paths: dist
+          key: node-modules-{{ checksum "package-lock.json" }}
+          paths:
+            - node_modules
```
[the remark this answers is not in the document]
paulfalgout: cypress/install runs npm ci

.circleci/config.yml
```diff
@@ -55,36 +56,39 @@ jobs:
...
 
-      # 5 Start Server
+      # 4 Build app (cached)
```
nmajor25: These numbers are off, since `#3` was deleted.

vite.config.js
```diff
@@ -80,9 +88,6 @@ export default defineConfig(({ mode }) => {
   return {
     plugins: [
       inlineHbsCompile(),
-      babel({
-        exclude: '**/formio.form.min*',
-      }),
```
nmajor25, line 85: We don't need this exclude anywhere anymore? For `formio.form.min`? I'm not sure either way, but just checking to be sure.

vite.config.js
```diff
@@ -5,11 +5,12 @@ import utcPlugin from 'dayjs/plugin/utc.js';
...
-import babel from 'vite-plugin-babel';
 import eslint from 'vite-plugin-eslint';
 import { nodeResolve } from '@rollup/plugin-node-resolve';
 import stylelint from 'vite-plugin-stylelint';
 import { VitePWA } from 'vite-plugin-pwa';
+import istanbul from 'vite-plugin-istanbul';
+import { COVER_INCLUDE, COVER_EXCLUDE } from './config/coverage.js';
 import yaml from './config/vite-plugin-yaml.js';
 import handlebars from './config/vite-plugin-handlebars-loader.js';
 import inlineHbsCompile from './config/vite-plugin-inline-handlebars.js';
```
nmajor25, line 16: This is unrelated to the changes in this PR.

But should we rename these handlebars plugins? To make it more clear that one is for compiling `.hbs` files & the other is for inline hbs? 

Like `hbsFileCompile` & `hbsInlineCompile`. Or something like that.

vite.config.js
paulfalgout: nope.  the vite plugin ignores node_modules by default

vite.config.js
paulfalgout: maybe?  It seems implied to me.. like I can't figure out what else a handlebars vite plugin might do?

nmajor25 (approved):
🚢

PR #1461, opened 2025-06-11
Remove scrollbar CSS styling overrides

nmajor25:
Shortcut Story ID: [sc-61940]

## Summary by CodeRabbit

- **New Features**
  - Lists now automatically adjust their width to ensure proper alignment and improved appearance across various views.

- **Style**
  - Increased horizontal padding for list containers and sidebars for a more consistent layout.
  - Updated form and sidebar overflow and padding for better content containment and spacing.
  - Removed custom scrollbar styles for a cleaner and more uniform look.

- **Bug Fixes**
  - Improved list alignment and responsiveness when resizing windows or updating list content.


src/scss/modules/sidebar.scss
```diff
@@ -1,7 +1,7 @@
 .sidebar {
   border-left: 1px solid #{$black-90};
   overflow-y: auto;
-  padding: 16px 12px;
```
nmajor25, line 4: If we switch to `scrollbar-width: thin`, we may not need this extra padding.

src/js/views/patients/patient/archive/archive_views.js
```diff
@@ -267,25 +274,42 @@ const LayoutView = View.extend({
...
     </div>
   `,
+  ui: {
+    listHeader: '.js-list-header',
+    list: '.js-list',
+  },
   triggers: {
     'click .js-dashboard': 'click:dashboard',
   },
+  initialize() {
```
paulfalgout: not thrilled with having to do this.  But if we are we should make it a behavior

paulfalgout (changes requested):
I suspect making this into a behavior will prevent needing to increase tests for coverage

src/js/views/patients/patient/archive/archive_views.js
nmajor25: I like it from a user perspective. But ya I'm not a huge fan either of the code implementation. I just don't know of another way to do it.

src/js/views/forms/form/form.scss
```diff
@@ -26,6 +26,7 @@
   flex: 1;
   flex-direction: column;
   min-width: 744px;
+  overflow: hidden;
```
nmajor25, line 29: Not having this was causing a double scrollbar to show when the form page height is very short:

Screenshot 2025-06-12 at 2 41 17 PM

nmajor25: But the overall form layout doesn't need to scroll, since the actual form content scrolls independently.

src/js/views/patients/patient/archive/archive_views.js
nmajor25: I just pushed a commit with the behavior added.

nmajor25: Coverage miss was on the archive views. Because I forgot the `childViewEvents: { 'update:listDom': 'fixListWidth' }`.

Which I added in the [last commit](https://github.com/RoundingWell/care-ops-frontend/pull/1461/files#diff-517afe962d13585a95fd46396c804969978a82f1d71412ed0d3ff4227dd3c1b5R301-R303). So coverage should be good now 🤞

src/js/views/patients/schedule/schedule_views.js
```diff
@@ -26,6 +28,7 @@ import './schedule-list.scss';
 const LayoutView = View.extend({
   className: 'flex-region',
   template: LayoutTemplate,
+  behaviors: [FixListWidthBehavior],
```
paulfalgout, line 31: The only thing that needs to be in these files is:
```js
behaviors: [{
  behaviorClass: FixListWidthBehavior,
  listHeader: '.js-list-header',
  list: '.js-list',
}]
```
and really the listHeader/list classes could be defaulted to those values

src/js/behaviors/fix-list-width.js
```diff
@@ -0,0 +1,22 @@
+import Radio from 'backbone.radio';
+import { Behavior } from 'marionette';
+
+export default Behavior.extend({
+  initialize() {
+    const userActivityCh = Radio.channel('user-activity');
+    this.listenTo(userActivityCh, 'window:resize', this.fixWidth);
+
+    this.listenTo(this.view, 'fix:list:width', this.fixWidth);
```
paulfalgout: behaviors has `triggers`

src/js/behaviors/fix-list-width.js
```diff
@@ -0,0 +1,22 @@
...
+  fixWidth() {
+    /* istanbul ignore if */
+    if (!this.view.isRendered()) return;
+
+    const headerWidth = this.view.ui.listHeader.width();
+    const listWidth = this.view.ui.list.contents().width();
+    const listPadding = parseInt(this.view.ui.list.css('paddingRight'), 10);
+    const scrollbarWidth = headerWidth - listWidth;
+
+    this.view.ui.list.css({ paddingRight: `${ listPadding - scrollbarWidth }px` });
```
paulfalgout: behaviors have `ui` that also contain the view ui
https://github.com/marionettejs/backbone.marionette/blob/master/src/behavior.js#L42

src/js/views/patients/patient/dashboard/dashboard_views.js
```diff
@@ -266,20 +268,32 @@ const ListView = CollectionView.extend({
   viewFilter({ model }) {
     return !model.isDone();
   },
+  onAttach() {
```
paulfalgout: this event is needed it's the same as `dom:refresh`

paulfalgout (changes requested):
I think the only keys that need to be added to the view class are:
```js
behaviors: [FixListWidthBehavior],
childViewTriggers: {
  'dom:refresh': 'childview:dom:refresh',
}
```
and everything else is in the behavior itself.

src/js/behaviors/fix-list-width.js
paulfalgout: also in the initialize we can probably `this.fixWidth = debounce(this.fixWidth, 50)` or something

src/js/views/patients/patient/dashboard/dashboard_views.js
nmajor25: Didn't know `dom:refresh` existed before now, nice.

But we'd still need the `update:listDom` in the `onRenderChildren()` correct?

```js
  // no longer needed, same as `dom:refresh`
  onAttach() {
  },
  // still needed?
  onRenderChildren() {
    if (!this.isAttached()) return;
  },
```

Since `dom:refresh` won't be triggered when the list changes due to find in list or actions/flows are added/removed via websockets or other means.

src/js/behaviors/fix-list-width.js
nmajor25: Ya doing a debounce is a good idea 👍

src/js/views/patients/schedule/schedule_views.js
nmajor25: What do you mean by `"the listHeader/list classes could be defaulted to those values"`?

paulfalgout: you could just allow for these as overrides on the behavior, but have them defined on the behavior as these classes since they're fairly generic

src/js/views/patients/patient/dashboard/dashboard_views.js
paulfalgout: oh right.. well still I think in the view the behavior is added to we can still just add:
```js
childViewTriggers: {
  'attach': 'childView:attach',
  'render:children', 'childView:render:children',
},
```

Then in the behavior `triggers` the render children handler can check this.view.attached.. or maybe just change the isRendered check to isAttached in fixWidth.

src/js/behaviors/fix-list-width.js
```diff
@@ -8,15 +8,27 @@ export default Behavior.extend({
...
   },
+  ui: {
+    listHeader: '.js-list-header',
+    list: '.js-list',
+  },
+  onChildViewAttach() {
+    this.fixWidth();
+  },
+  onChildViewRenderChildren() {
+    if (!this.view.isAttached()) return;
```
nmajor25: Is this correct? Or should it be a check if the child view attached? Do we even need this?

src/js/behaviors/fix-list-width.js
nmajor25: Adding a debounce is causing odd behavior when the browser window is resized.

src/js/views/patients/patient/dashboard/dashboard_views.js
nmajor25: Okay I've updated the views and behavior to use this pattern 👍

src/js/behaviors/fix-list-width.js
paulfalgout: The children are rendered the first time prior to dom attachment when width measurements can't be made. Really though the is attached check can just replace the is rendered check in fixWidth.

src/js/behaviors/fix-list-width.js
paulfalgout: The debounce is needed for the resize.  Possibly just lower the time. Otherwise we could get thousands of fixWidth calculations per second on the resize event.. Though I don't remember if the event is throttled.. Maybe so

src/js/behaviors/fix-list-width.js
nmajor25: Okay makes sense, I updated it to that.

src/js/behaviors/fix-list-width.js
nmajor25: I lowered it to `10` and still got weird behavior. I'm not really sure on this one.

We weren't debouncing it previously and never ran into any issues that I know of. 

And if I do a `console.log` when `fixWidth()` is called, it looks like it gets ran a handful of times when I drag-resize the browser manually. Which doesn't seem like a huge deal.

nmajor25: But this is based on the jquery `$(window).on('resize.app')` event, right? I could look into if it's being throttled by default somehow.

paulfalgout: yeah maybe we don't worry about it.. we don't throttle it in one other case either

paulfalgout (approved):

PR #1470, opened 2025-06-24
Merge stylistic eslint rules with our custom rules

nmajor25:
## Summary by CodeRabbit

- **Chores**
  - Updated ESLint configuration to improve integration of stylistic rules. No impact on application features or user interface.
  - Cleaned up code formatting by removing extraneous blank lines and trailing whitespace across multiple files to maintain code quality and consistency.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,54 +1,21 @@
 import globals from 'globals';
 import js from '@eslint/js';
 import mocha from 'eslint-plugin-mocha';
-import cypress from 'eslint-plugin-cypress';
 import chaiFriendly from 'eslint-plugin-chai-friendly';
+import cypress from 'eslint-plugin-cypress';
 import stylistic from '@stylistic/eslint-plugin';
 
-/* ───────── Stylistic customisation ───────── */
-const stylisticRules = stylistic.configs.customize({
-  semi: true,
-  templateCurlySpacing: 'always',
-  arrowParens: 'as-needed',
-});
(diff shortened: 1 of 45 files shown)


eslint.config.mjs
```diff
@@ -59,8 +59,8 @@ export default [
         _NOW_: 'readonly',
       },
     },
-    ...stylisticRules,
     rules: {
+      ...stylisticRules.rules,
```
nmajor25: Eslint enforcing rules for commas, semis, etc. stopped working in my local dev environment. Looks like the stylistic rules and our manual rules were conflicting with each other? 

This update seems to fix the issue and linting works as before.

nmajor25: But I'm not sure if this is the right approach. For instance, do we need the `stylistic.configs.customize()` still or can those rules go here?

paulfalgout (approved):

eslint.config.mjs
```diff
@@ -60,30 +27,108 @@ export default [
...
       'cypress/unsafe-to-chain-command': 'off',
-      'mocha/no-setup-in-describe': 'off',
+    },
+  },
+  {
+    name: 'Generic test overrides',
+    files: TEST_FILES,
+    rules: {
       'no-console': 'warn',
       'no-debugger': 'warn',
```
nmajor25: Do we still need this section if these rules are included in the rules for `Base globals & quality rules`?

PR #1473, opened 2025-07-02
Support nested values array for field history

paulfalgout:
Shortcut Story ID: [sc-###]

@zfixler this ought to support the nested values array for history

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted field history data to return multiple historical values instead of a single value when fetching field history.

* **Tests**
  * Updated integration tests to handle and verify the new field history data structure that includes multiple values.

### DIFF
--- a/src/js/services/forms.js
+++ b/src/js/services/forms.js
@@ -156,7 +156,7 @@ export default App.extend({
       _patient: this.patient.getResource(),
     }, { limit, sort })
       .then(field => {
-        this.send(message, { value: field.get('value') }, requestId);
+        this.send(message, { value: field.get('values') }, requestId);
       })
       .catch(
         /* istanbul ignore next: Don't test BE errors */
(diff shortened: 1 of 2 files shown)


src/js/services/forms.js
```diff
@@ -156,7 +156,7 @@ export default App.extend({
       _patient: this.patient.getResource(),
     }, { limit, sort })
       .then(field => {
-        this.send(message, { value: field.get('value') }, requestId);
+        this.send(message, { value: field.get('values') }, requestId);
```
shadowhand, line 159: @zfixler this would be the reason you were getting `undefined` results when using the history API, the code was trying to pull a non-existent key.

shadowhand (approved):

zfixler (approved):

src/js/services/forms.js
paulfalgout: well all keys were non-existent since a model can't parse an array

PR #1475, opened 2025-07-14
Ability to define wider sizes for form modals

nmajor25:
Shortcut Story ID: [sc-62490]

## Summary by CodeRabbit

* **Style**
  * Updated modal styling for form-related dialogs with a flexible minimum width and improved iframe sizing.

* **Bug Fixes**
  * Ensured consistent large modal styling is applied to all form modals.

* **Tests**
  * Updated integration tests to reflect changes in modal styling selectors.

### DIFF
--- a/src/scss/modules/modals.scss
+++ b/src/scss/modules/modals.scss
@@ -16,7 +16,8 @@
 .modal--large {
   @extend .modal;
 
-  width: 600px;
+  min-width: 600px;
+  width: auto;
 }
 
 .modal--sidebar {
@@ -127,6 +128,7 @@
   @extend .modal__form-iframe;
 
   height: 75vh;
+  width: 75vw;
 }
 
 .modal__form-label {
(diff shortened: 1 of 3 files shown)


src/scss/modules/modals.scss
```diff
@@ -19,6 +19,13 @@
   width: 600px;
 }
 
+.modal--form {
+  @extend .modal;
+
+  width: auto;
+  min-width: 544px;
+}
```
nmajor25, line 21: Use `min-width` instead of `width`.

Then the modal should be set to whatever width is set for `modal__form-iframe`, `.modal__form-iframe--small` or `.modal__form-iframe--large`. And if not set, the iframe width should fit the `min-width: 544px`.

paulfalgout (changes requested):
I'm not sure about the naming of `--form`  the modifier for `--sidebar` makes more sense because it's modifying a modal to be an iframe.. in this case we're not making the modal a form.. we're?  making it wider?  modal--large is already width:600px. maybe we should just be modifying that to match this min-width/width:auto?

paulfalgout (approved):

PR #1476, opened 2025-07-15
Support patient field queries setting for patient search

nmajor25:
Shortcut Story ID: [sc-62563]

- [x] Support patient field queries setting
- [x] Add test for search by patient phone number
- [x] Split search by identifier testing into its own test

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying field query results (e.g., phone numbers) alongside identifiers in patient search results.
  * Patient search UI now shows additional search examples and a new column for field query results when available.

* **Style**
  * Updated styles to ensure consistent appearance for new field query result elements in search results.

* **Documentation**
  * Added a new translation key for field query result headers.

* **Tests**
  * Expanded and updated tests to cover searching by patient identifiers and fields such as phone numbers.
  * Adjusted test data and assertions to reflect UI changes and improved test reliability.

### DIFF
--- a/src/js/entities-service/entities/patient-search-results.js
+++ b/src/js/entities-service/entities/patient-search-results.js
@@ -22,6 +22,7 @@ const Collection = BaseCollection.extend({
     if (search.length < 3) {
       if (!search.length || !this.prevSearch.includes(search)) {
         delete this._hasIdentifiers;
+        delete this._hasFieldQueryResults;
         this.reset();
         this.prevSearch = '';
       }
@@ -33,6 +34,13 @@ const Collection = BaseCollection.extend({
     this.prevSearch = search;
     this.isSearching = true;
     this._debouncedSearch(search);
+  },
+  hasFieldQueryResults() {
+    if (isBoolean(this._hasFieldQueryResults)) return this._hasFieldQueryResults;
+
+    this._hasFieldQueryResults = this.some(model => model.has('value'));
+
+    return this._hasFieldQueryResults;
   },
   hasIdentifiers() {
     if (isBoolean(this._hasIdentifiers)) return this._hasIdentifiers;
@@ -47,6 +55,7 @@ const Collection = BaseCollection.extend({
     const filter = { search };
 
     delete this._hasIdentifiers;
+    delete this._hasFieldQueryResults;
     this.controller.abort();
     this.controller = new AbortController();
 
--- a/src/js/views/globals/search/patient-search_views.js
+++ b/src/js/views/globals/search/patient-search_views.js
@@ -28,6 +28,12 @@ const TipTemplate = hbs`
         {{ @intl.globals.search.patientSearchViews.emptyView.exampleId }} {{this.example}}
       </div>
     {{/each}}
+    {{#each settings.queries}}
+      <div class="u-margin--t-8 qa-search-option">
+        <span class="patient-search__search-by-label">{{this.label}}</span>
+        {{ @intl.globals.search.patientSearchViews.emptyView.exampleId }} {{this.example}}
+      </div>
+    {{/each}}
   </div>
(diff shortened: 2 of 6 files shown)


src/js/views/globals/search/patient-search_views.js
```diff
@@ -99,6 +105,11 @@ const PicklistItem = View.extend({
           {{#if identifiers.0.value}}{{matchText identifiers.0.value search}}{{else}}&ndash;{{/if}}
         </div>
       {{/if}}
+      {{#if hasFieldQueryResults}}
+        <div class="patient-search__picklist-item-identifier u-text--overflow">
```
nmajor25: I'd like to use this same class for both the identifier and field query column. But maybe it should be renamed to something generic.

src/js/views/globals/search/patient-search_views.js
```diff
@@ -99,6 +105,11 @@ const PicklistItem = View.extend({
...
+          {{#if value}}{{matchText value search}}{{else}}&ndash;{{/if}}
```
nmajor25: For both the identifier and field query columns, we are showing a dash (`&ndash;`) if the value is empty. But is that possible? it's either going to have a value or it wouldn't be included in the search results, right?

src/js/entities-service/entities/patient-search-results.js
```diff
@@ -34,6 +35,15 @@ const Collection = BaseCollection.extend({
     this.isSearching = true;
     this._debouncedSearch(search);
   },
+  hasFieldQueryResults() {
+    if (isBoolean(this._hasFieldQueryResults)) return this._hasFieldQueryResults;
+
+    this._hasFieldQueryResults = !!this.find(model => {
```
paulfalgout: `this.some(model => model.has('value'))`
 really the other one should be `some` as well, but `has` is better here in case the value is 0 or `false`

src/js/views/globals/search/patient-search_views.js
paulfalgout: yep something generic.. always better for css to describe what it looks like and not it's domain

src/js/views/globals/search/patient-search_views.js
paulfalgout: yeah I don't think this column shows up unless the user is matching these values.  we probably did this assuming typing "jan" could match "jane" as well as some identifier, but I guess somehow it's one or the other only.

src/js/views/globals/search/patient-search_views.js
```diff
@@ -151,6 +163,11 @@ const HeaderView = View.extend({
           {{ @intl.globals.search.patientSearchViews.headerView.id }}
         </div>
       {{/if}}
+      {{#if hasFieldQueryResults}}
+        <div class="patient-search__picklist-header-identifier">
+          Results
```
paulfalgout: i18n

paulfalgout (changes requested):

src/js/views/globals/search/patient-search_views.js
nmajor25: I think that was the original reason too. I removed the `&ndash;` conditional for these.

src/js/views/globals/search/patient-search_views.js
nmajor25: I agree, but looking at the class names more, I don't know if it's worth the refactor. We'd need to add `u-margin--l-16` to the templates and then handle small differences with some of the columns. Seems more complicated then just adding an additional class name.

src/js/entities-service/entities/patient-search-results.js
nmajor25: Updated to use `some`.

src/js/views/globals/search/patient-search_views.js
nmajor25: Updated to use i18n.

paulfalgout (approved):

PR #1477, opened 2025-07-21
In patient search, update UI when search results collection is updated

nmajor25:
Shortcut Story ID: [sc-62702]

To fix a bug where the UI doesn't always update properly when copy/pasting search queries.

Video example of the bug this PR fixes:

https://github.com/user-attachments/assets/25c674bd-d661-42ec-87d5-54c466c9ffd0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the responsiveness of patient search views, ensuring that updates to the collection are immediately reflected in the interface.
* **Tests**
  * Enhanced patient search tests to cover additional input scenarios and verify correct UI updates during searches.

### DIFF
--- a/src/js/views/globals/search/patient-search_views.js
+++ b/src/js/views/globals/search/patient-search_views.js
@@ -39,9 +39,6 @@ const TipTemplate = hbs`
 `;
 
 const EmptyView = View.extend({
-  collectionEvents: {
-    'search': 'render',
-  },
   tagName: 'li',
   className: 'patient-search__no-results',
   initialize({ state }) {
@@ -184,7 +181,7 @@ const HeaderView = View.extend({
 const DialogView = View.extend({
   className: 'patient-search__picklist',
   collectionEvents: {
-    'search': 'showHeader',
+    'search': 'onSearchComplete',
   },
   modelEvents: {
     'change:search': 'showHeader',
@@ -222,10 +219,11 @@ const DialogView = View.extend({
   `,
   onRender() {
     this.showHeader();
-    this.showChildView('list', new ListView({
-      collection: this.collection,
-      model: this.model,
-    }));
+    this.showList();
+  },
+  onSearchComplete() {
+    this.showHeader();
+    this.showList();
   },
   showHeader() {
     if (!this.collection.length) {
@@ -234,6 +232,12 @@ const DialogView = View.extend({
     }
 
     this.showChildView('header', new HeaderView({ collection: this.collection }));
+  },
+  showList() {
+    this.showChildView('list', new ListView({
+      collection: this.collection,
+      model: this.model,
+    }));
   },
 });
 
--- a/test/integration/services/patient-quick-search.js
+++ b/test/integration/services/patient-quick-search.js
@@ -260,7 +260,7 @@ context('Patient Quick Search', function() {
 
     const { first_name, last_name, birth_date, status } = testPatient.attributes;
 
-    const searchResultData = [{
+    const searchResult = {
       id: testPatient.id,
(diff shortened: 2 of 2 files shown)


src/js/views/globals/search/patient-search_views.js
```diff
@@ -129,6 +129,9 @@ const ListView = CollectionView.extend({
   tagName: 'ul',
   serializeCollection: noop,
   childView: PicklistItem,
+  collectionEvents: {
+    'sync update reset': 'render',
```
nmajor25: Without `sync` the picklist items render twice. Not sure if we need `reset`.

If I log these like:

```js
'sync': function() {console.log('sync');},
'update': function() {console.log('update');},
'reset': function() {console.log('reset');},
```

* `reset` is called when the search query length is less than 3 or the search input is cleared.
* `update` and then `sync` are called when a search api request is made. I can't tell what the difference is between those two.

src/js/views/globals/search/patient-search_views.js
nmajor25, line 1: We may be able to remove `this.listenTo(this.state, 'change:search', this.render);` from the `PicklistItem` view. I think that's redundant now...

src/js/views/globals/search/patient-search_views.js
nmajor25: Maybe only the `sync` event is needed here? Are `update` and `reset` re-renders already handled automatically behind the scenes?

src/js/views/globals/search/patient-search_views.js
paulfalgout: I think that might have been more related to when we were filtering the list on the FE

src/js/views/globals/search/patient-search_views.js
paulfalgout: yeah update and reset are the default:
https://github.com/marionettejs/backbone.marionette/blob/master/src/collection-view.js#L92

when a collection.fetch is called internally collection.set is called with the data from the API which triggers `update` for any changes..  granted no update if the results are the same. then at the end of the fetch's successful set it triggers `sync`   so `sync` should always be redundant to an `update` which is already handled internally..  I think something else is going on.

src/js/views/globals/search/patient-search_views.js
```diff
@@ -184,7 +184,7 @@ const HeaderView = View.extend({
 const DialogView = View.extend({
   className: 'patient-search__picklist',
   collectionEvents: {
-    'search': 'showHeader',
+    'search': 'onSearchComplete',
```
nmajor25, line 184: I feel much better about this way to fix it. The header was being updated properly, just not the list items. So now whenever a search api fetch is completed, both the header and list will be re-rendered. Before only the header was being re-rendered.

src/js/views/globals/search/patient-search_views.js
nmajor25: I agree, something else is going on. I found a different way to solve it that makes more sense I think.

src/js/views/globals/search/patient-search_views.js
nmajor25: Okay we can just leave that untouched.

src/js/views/globals/search/patient-search_views.js
```diff
@@ -39,9 +39,6 @@ const TipTemplate = hbs`
 `;
 
 const EmptyView = View.extend({
-  collectionEvents: {
-    'search': 'render',
-  },
```
nmajor25, line 44: We should be able to remove this. This view will always re-render alongside the `ListView`.

paulfalgout (approved):

PR #1480, opened 2025-07-24
On pt dashboard/archive, subscribe via `Radio.request('ws', 'add', action)` when flow/action is added to collection

nmajor25:
Shortcut Story ID: [sc-62143]

This PR is to fix a bug where actions/flows added to the patient dashboard/archive do not receive ws notifications for `Done` vs. `Non-Done` states. i.e. action/flow on the pt dashboard changed to a done state or switched to a non-done state on the pt archive.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of real-time updates, ensuring new flows and actions are correctly added to WebSocket subscriptions and reflected in the UI.

* **Tests**
  * Enhanced integration tests for patient archive, dashboard, flow, and worklist pages to verify that WebSocket subscription resources update accurately when new flows or actions are created remotely.

### DIFF
--- a/src/js/apps/patients/patient/flow/flow_app.js
+++ b/src/js/apps/patients/patient/flow/flow_app.js
@@ -110,8 +110,6 @@ export default SubRouterApp.extend({
   onAddAction(action) {
     this.editableCollection.add(action);
 
-    Radio.request('ws', 'add', action);
-
     this._setFlowProgress();
   },
   onActionChangeState() {
--- a/src/js/services/ws.js
+++ b/src/js/services/ws.js
@@ -11,6 +11,7 @@ const AdderApp = App.extend({
   },
   onStart({ model, collection }) {
     collection.add(model);
+    Radio.request('ws', 'add', model);
   },
 });
 
--- a/test/integration/patients/patient/archive.js
+++ b/test/integration/patients/patient/archive.js
@@ -532,6 +532,20 @@ context('patient archive page', function() {
       .its('request.url')
       .should('contain', testNewSocketFlow.id);
 
+    // verify the new flow is added to the ws subscription resources
+    cy
+      .get('@wsHandleMessage')
+      .should('have.been.calledTwice')
+      .then(stub => {
+        const secondCallData = stub.getCall(1).args[0].data;
+        const { resources } = secondCallData;
+
+        expect(resources).to.deep.include({
+          id: testNewSocketFlow.id,
+          type: testNewSocketFlow.type,
+        });
+      });
+
     cy
       .get('.app-frame__content')
       .find('.table-list__item')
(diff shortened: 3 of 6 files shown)


src/js/apps/patients/patient/dashboard/dashboard_app.js
```diff
@@ -50,7 +50,6 @@ export default App.extend({
   },
   subscribe() {
     const filters = {
-      states: this.states,
```
nmajor25: On the pt dashboard, this currently stops us from receiving `Done` state changes on actions/flows added via web sockets. Because in that scenario the ws filters excludes done states and the action/flow isn't in the ws subscription resources list.

There are steps to reproduce it on the shortcut story.

So removing this from the ws filters means we'll receive ws notifications for flows/actions with any state. Which makes sense to me. The UI list only shows non-done state actions/flows, but we want to get ws notifications for all states.

But maybe i'm missing something. I can't remember why we added this ws filter in the first place.

paulfalgout (changes requested):
we definitely want to keep the state filter so that changes to actions that aren't done, don't start showing up on archive.  What seems to be missing somehow is manage:add isn't resubscribing to the created elements id?

nmajor25:
The `viewFilter` filters out the done vs non-done for the dashboard and archive. So I don't think changes would show up? But we probably don't want to rely on that.

Maybe its a backend thing? Where the action/flow is switched to `Done` and therefore doesn't send the ws notification because it no longer fits the filters?

I tried adding a `Radio.request('ws', 'add', action);` & `Radio.request('ws', 'add', flow);` after actions/flows are created on the dashboard. But that didn't fix it.

src/js/apps/patients/patient/dashboard/dashboard_app.js
```diff
@@ -44,6 +44,10 @@ export default App.extend({
 
     this.subscribe();
 
+    this.listenTo(this.collection, 'add', model => {
+      Radio.request('ws', 'add', model);
+    });
```
nmajor25: This fixes the issue. And ensures the flow/action is added to the ws subscription resource list when one is added to the collection.

We do this in the flow app [here](https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/js/apps/patients/patient/flow/flow_app.js#L110-L116).

nmajor25: We should probably test this somehow too. When a flow/action is added via ws in a test, we could check that the flow/action is in the ws subscription resource list.

paulfalgout (approved):
We've probably tested this one line of code in enough contexts.  I'm not sure how to test it right now from the service, but we probably don't need to test this for every instance that there's an add.

PR #1485, opened 2025-08-06
Implement `match` object in patient search api response data

nmajor25:
Shortcut Story ID: [sc-62877]

Corresponding BE PR: https://github.com/RoundingWell/care-ops-backend/pull/957301

Screenshot 2025-08-06 at 10 32 02 AM

## Summary by CodeRabbit

* **New Features**
  * Patient search results now consistently display a unified result value and label for each entry, improving clarity in the search picklist.

* **Style**
  * Increased the height of the patient search picklist and updated spacing and styling for headers and result labels.
  * Enlarged the minimum width of large modals for better content display.

* **Bug Fixes**
  * Improved consistency in displaying search results by removing conditional rendering of identifiers and field query results.

* **Documentation**
  * Updated localization strings to reflect new result labeling and simplified status indicators.

* **Tests**
  * Updated test data and assertions to align with the new unified match structure in search results.

### DIFF
--- a/src/js/views/globals/search/patient-search.scss
+++ b/src/js/views/globals/search/patient-search.scss
@@ -1,3 +1,7 @@
+.patient-search__modal {
+  width: 700px;
+}
+
 .patient-search__picklist {
   display: flex;
   flex-direction: column;
@@ -50,14 +54,13 @@
   }
 }
 
-.patient-search__picklist-header-name {
-  display: flex;
-  flex: 2;
-}
-
 .patient-search__picklist-header-name-icon {
   margin-right: 16px;
   width: 28px;
+}
+
+.patient-search__picklist-header-name-text {
+  flex: 1;
 }
 
 .patient-search__picklist-header-meta {
@@ -114,19 +117,19 @@
   content: '–';
 }
 
-.patient-search__picklist-item-name {
-  display: flex;
-  flex: 2;
-  font-size: 16px;
-  padding-right: 4px;
-}
-
 .patient-search__picklist-item-name-icon {
   align-items: center;
   display: flex;
   justify-content: center;
   margin-right: 16px;
   width: 28px;
+}
+
+.patient-search__picklist-item-name-text {
+  display: flex;
+  flex: 1;
+  margin-right: 8px;
+  min-width: 0;
 }
 
 .patient-search__picklist-item-meta {
--- a/src/js/views/globals/search/patient-search.scss
+++ b/src/js/views/globals/search/patient-search.scss
@@ -1,7 +1,7 @@
 .patient-search__picklist {
   display: flex;
   flex-direction: column;
-  height: 420px;
+  height: 460px;
 }
 
 .patient-search__close {
@@ -69,10 +69,10 @@
   width: 100px;
 }
 
-.patient-search__picklist-header-identifier,
-.patient-search__picklist-header-query,
+.patient-search__picklist-header-results,
+.patient-search__picklist-header-results-label,
 .patient-search__picklist-header-status {
-  margin-left: 16px;
+  margin-left: 12px;
   width: 100px;
 }
 
@@ -88,6 +88,10 @@
   &.is-inactive {
     color: #{$black-60};
     font-style: italic;
+
+    .patient-search__picklist-item-result-label {
+      font-style: normal;
+    }
   }
 
   &.is-highlighted,
@@ -96,7 +100,8 @@
     color: #{$white};
 
     .patient-search__picklist-item,
-    .patient-search__picklist-item-meta {
+    .patient-search__picklist-item-meta,
+    .patient-search__picklist-item-result-label {
       color: #{$white};
     }
   }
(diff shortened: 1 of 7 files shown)


src/js/views/globals/search/patient-search_views.js
```diff
@@ -90,34 +90,30 @@ const PicklistItem = View.extend({
...
     </div>
   `,
   templateContext() {
     return {
       name: `${ this.model.get('first_name') } ${ this.model.get('last_name') }`,
       search: this.state.get('search'),
-      hasIdentifiers: this.collection.hasIdentifiers(),
-      hasFieldQueryResults: this.collection.hasFieldQueryResults(),
+      resultValue: this.model.get('match').value,
+      resultLabel: this.model.get('match').label,
```
nmajor25: Do we need to make this safer? Like `this.model.get('match')?.value` or something similar?

paulfalgout: just use {{ match.value}} in the template

src/js/views/globals/search/patient-search.scss
```diff
@@ -127,22 +132,26 @@
...
+.patient-search__picklist-item-result-value {
+  margin-left: 12px;
   width: 100px;
 }
 
-.patient-search__picklist-item-status {
-  margin-left: 16px;
-  text-transform: capitalize;
+.patient-search__picklist-item-result-label {
+  color: #999;
```
paulfalgout: theres a variable for this

src/js/views/globals/search/patient-search.scss
```diff
@@ -127,22 +132,26 @@
...
   width: 100px;
 }
 
-.patient-search__picklist-item-identifier,
-.patient-search__picklist-item-query {
-  margin-left: 16px;
+.patient-search__picklist-item-status {
+  margin-left: 12px;
+  text-transform: capitalize;
+  width: 100px;
```
nmajor25, line 147: We could make the `DOB` and `Status` columns smaller. Say `80px`. Since their content widths are more predictable. And it would give more room for the patient name.

src/js/views/globals/search/patient-search_views.js
```diff
@@ -90,34 +90,30 @@ const PicklistItem = View.extend({
         {{far "address-card"}}
       </div>
       <div class="patient-search__picklist-item-name-text">
-        <span>{{matchText name search}}{{~ remove_whitespace ~}}</span>
```
paulfalgout, line 93: do we need to remove this?   kinda wonder if we should add this to the match value as well

paulfalgout (changes requested):

src/js/views/globals/search/patient-search_views.js
paulfalgout: oh or it already is.. are we removing it here because the name shows up as the result?

nmajor25: Yep, that's the reason I removed it. So that `matchText` is only applied to the match value.

Since say a search for `Jane` will have the patient name show in the result.

nmajor25: That's an assumption I made. There isn't an example of that scenario in Mark's design.

src/js/views/globals/search/patient-search.scss
nmajor25: Updated 👍

src/js/views/globals/search/patient-search_views.js
nmajor25: Good call, updated to that

src/scss/modules/modals.scss
```diff
@@ -16,7 +16,7 @@
 .modal--large {
   @extend .modal;
 
-  min-width: 600px;
+  min-width: 700px;
   width: auto;
 }
```
nmajor25, line 21: Just realized this will effect the form modals. A change made in https://github.com/RoundingWell/care-ops-frontend/pull/1475.

I think we should just create a new classname: `.modal--patient-search` and use that:

```css
.modal--large {
  @extend .modal;

  width: 700px;
}
```

nmajor25: And the overflow hidden/ellipsis for the patient name column isn't currently working. So I'm looking into that as well.

paulfalgout: just make a `.patient-search__modal`.  `.modal--`  should be as generic as possible

nmajor25: Okay yep makes sense. I've got the overflow ellipsis thing figured out I think. Just need to test it out some more.

nmajor25: The overflow ellipsis functionality is working now in the latest commit.

Screenshot 2025-08-06 at 1 48 02 PM

src/js/views/globals/search/patient-search_views.js
```diff
@@ -85,39 +85,31 @@ const PicklistItem = View.extend({
...
+        {{matchText match.value search}}
+      </div>
+      <div class="patient-search__picklist-item-result-label u-text--overflow">
+        {{match.label}}
+      </div>
     </div>
   `,
   templateContext() {
     return {
       name: `${ this.model.get('first_name') } ${ this.model.get('last_name') }`,
```
paulfalgout: now that we don't need matchText we should just use. `{{ first_name }} {{ last_name }}` in the template.. no need for the name templateContext

paulfalgout (changes requested):
looks good.. one small clean up

src/js/views/globals/search/patient-search_views.js
nmajor25: Good call, updated

paulfalgout (approved):

PR #1489, opened 2025-08-11
In search results, don't show match column data if `match.label` equals `Name` or `Birth Date`

nmajor25:
Shortcut Story ID: [sc-62877]

## Summary by CodeRabbit

* **New Features**
  * Improved patient search picklist: patient names are highlighted; redundant labels (Name, Birth Date) are hidden for clarity.
  * Match metadata (identifiers, phone numbers) is shown only when meaningful.
  * Added localization entries for the picklist item labels to control visibility.

* **Tests**
  * Updated integration tests to verify active picklist behavior and that identifiers and phone numbers display with their labels.

### DIFF
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -241,6 +241,9 @@ careOptsFrontend:
           patient: Patient
           results: Results
           workspaceStatus: Status
+        picklistItemView:
+          labelDob: Birth Date
+          labelName: Name
   patients:
     patient:
       archive:
--- a/src/js/views/globals/search/patient-search_views.js
+++ b/src/js/views/globals/search/patient-search_views.js
@@ -6,6 +6,8 @@ import { View, CollectionView, Region } from 'marionette';
 import 'scss/modules/buttons.scss';
 import 'scss/modules/modals.scss';
 
+import intl from 'js/i18n';
+
 import Component from 'js/base/component';
 
 import InputFocusBehavior from 'js/behaviors/input-focus';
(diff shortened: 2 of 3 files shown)


src/js/i18n/en-US.yml
```diff
@@ -241,6 +241,8 @@ careOptsFrontend:
           patient: Patient
           results: Results
           workspaceStatus: Status
+        itemView:
+          showMatchData: "{label, select, Name {} Birth_Date {} other {true}}"
```
nmajor25: Ideally we could just do something like this:

```js
showMatchData: "{label, select, Name {} Birth_Date {} other {value}}"
```

But I don't know of a way to have the UI display nothing if the label is `Name` or `Birth_Date`. for `Name {}`/`Birth_Date {}`, `undefined` is displayed in the UI. And doing `{''}` shows the actual quotes in the UI.

So instead I have this returning either `undefined` or `"true"`. And then the view template determines if the `match.value`/`match.label` should be displayed.

nmajor25: For the `Birth_Date` underscore. I had to normalize the spacing in the labels. Since having the `Birth Date` in there with the space broke things. And putting quotes around the keys doesn't work either.

src/js/views/globals/search/patient-search_views.js
```diff
@@ -99,16 +99,24 @@ const PicklistItem = View.extend({
...
       </div>
     </div>
   `,
   templateContext() {
+    const match = this.model.get('match');
+
     return {
+      name: `${ this.model.get('first_name') } ${ this.model.get('last_name') }`,
       search: this.state.get('search'),
+      matchLabelNormalized: match.label.replace(/ /g, '_'),
```
paulfalgout: I think the simpler way to accomplish this will be
```
shouldShowMatch() {
  const label = this.model.get('label');
  const i18n = intl.globals.search.patientSearchViews;
  return label !== i18n.labelName && label !== i18n.labelDob;
}
```
and just `{{#if shouldShowMatch}}`

paulfalgout (changes requested):

paulfalgout:
not sure what the coverage report is about here

src/js/views/globals/search/patient-search_views.js
nmajor25: That works well, updated to that 👍

nmajor25: I also don't understand coveralls. I can't find where the miss is in the report.

paulfalgout (approved):

PR #1492, opened 2025-08-13
Fix createClient pattern

paulfalgout:
onRedirectCallback is called when  `pathName === AuthProvider.PATH_AUTHD`

but the auth -> redirect to PATH_AUTHD happens down at `if (!isAuthenticated) {`

so the client needs to resolve so that the login prompt can occur

Shortcut Story ID: [sc-63142]

## Summary by CodeRabbit

- New Features
  - More reliable redirect completion so sign-in finishes automatically after returning to the app.

- Bug Fixes
  - Resolved scenarios where authentication could hang or require a manual refresh.
  - Better handling when offline or already-authenticated to avoid unnecessary prompts.

- Refactor
  - Consolidated auth initialization and callback flow for more consistent, faster login behavior.

### DIFF
--- a/packages/care-ops-auth/providers/workos.js
+++ b/packages/care-ops-auth/providers/workos.js
@@ -35,29 +35,24 @@ export class WorkosAuthProvider extends AuthProvider {
     return clientId;
   }
 
-  async _initClient(clientId) {
-    return new Promise(resolve => {
-      const clientConfig = {
-        redirectUri: location.origin + AuthProvider.PATH_AUTHD,
-        onRedirectCallback: ({ user, state }) => {
-          const path = state;
-          if (!user) {
-            this.loginPrompt(path);
-            return;
-          }
+  async _initClient(clientId, success) {
+    const clientConfig = {
+      redirectUri: location.origin + AuthProvider.PATH_AUTHD,
+      onRedirectCallback: ({ user, state }) => {
+        const path = state;
+        if (!user) {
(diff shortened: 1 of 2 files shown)


packages/care-ops-auth/providers/workos.js
```diff
@@ -35,29 +35,24 @@ export class WorkosAuthProvider extends AuthProvider {
...
+      onRedirectCallback: ({ user, state }) => {
+        const path = state;
+        if (!user) {
+          this.loginPrompt(path);
+          return;
+        }
+
+        this.handleAuthedPath(path);
+
+        success();
```
paulfalgout, line 50: the only change here is unwrapping the `new Promise` and changing the resolve to success

packages/care-ops-auth/providers/workos.js
```diff
@@ -72,12 +67,9 @@ export class WorkosAuthProvider extends AuthProvider {
...
     const clientId = this._getClientId(pathName);
 
-    await this._initClient(clientId);
+    this.client = await this._initClient(clientId, success);
 
-    if (pathName === AuthProvider.PATH_AUTHD) {
-      success();
-      return;
-    }
+    if (pathName === AuthProvider.PATH_AUTHD) return;
```
paulfalgout, line 73: The reason this is returned here is because if this is true the success will get called in the `onRedirectCallback`

nmajor25 (approved):

PR #1495, opened 2025-08-14
Support matching text substrings for patient search

nmajor25:
Shortcut Story ID: [sc-63161]

So we can highlight matches that are not at the beginning of a word or string in patient search results.

## Summary by CodeRabbit

- **New Features**
  - Search results now highlight substring matches within patient names and matched values, making partial query matches visibly emphasized.

- **Style**
  - Updated display format for MRNs and phone numbers in Patient Quick Search results for clearer, more consistent presentation.

### DIFF
--- a/src/js/base/helpers.cy.js
+++ b/src/js/base/helpers.cy.js
@@ -11,6 +11,7 @@ context('Handlebars helpers', function() {
       template: hbs`
         <div class="test-null">{{matchText "Patient Name" null}}</div>
         <div class="test-match">{{matchText "Patient Name" "Patient"}}</div>
+        <div class="test-match-substring">{{matchText "TestPatient Name" "patient" includeSubstrings=true}}</div>
         <div class="test-noescape">{{matchText "<span style='color: green'>Patient</span> Name" "Patient" noEscape=true}}</div>
       `,
     });
@@ -29,6 +30,11 @@ context('Handlebars helpers', function() {
     cy
       .get('@root')
       .find('.test-match strong')
+      .should('contain', 'Patient');
+
+    cy
+      .get('@root')
+      .find('.test-match-substring strong')
       .should('contain', 'Patient');
 
     cy
--- a/src/js/utils/formatting/build-matcher-substrings.js
+++ b/src/js/utils/formatting/build-matcher-substrings.js
@@ -1,0 +1,11 @@
+// takes a search string and builds a search for each word without the match tag
+import { map, memoize } from 'underscore';
+
+import words from 'js/utils/formatting/words';
+import searchSanitize from 'js/utils/formatting/search-sanitize';
+
+export default memoize(function(query) {
+  const searchWords = map(words(searchSanitize(query)), RegExp.escape);
+
+  return new RegExp(searchWords.join('|'), 'gi');
+});
--- a/src/js/utils/formatting/formatting.cy.js
+++ b/src/js/utils/formatting/formatting.cy.js
@@ -1,4 +1,5 @@
 import buildMatcher from './build-matcher';
+import buildMatcherSubstrings from './build-matcher-substrings';
 import buildMatchersArray from './build-matchers-array';
 import collectionOf from './collection-of';
 import hasAllText from './has-all-text';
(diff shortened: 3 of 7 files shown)


src/js/utils/formatting/build-matcher.js
```diff
@@ -7,5 +7,5 @@ import searchSanitize from 'js/utils/formatting/search-sanitize';
 export default memoize(function(query) {
   const searchWords = map(words(searchSanitize(query)), RegExp.escape);
 
-  return new RegExp(`\\b${ searchWords.join('|') }`, 'gi');
+  return new RegExp(searchWords.join('|'), 'gi');
```
nmajor25: The failing test is for the team component.

Screenshot 2025-08-14 at 4 30 43 PM

This is returning more results now. 4 instead of 2.

I can update the test for this to pass. But is this what we want? I would lean towards maintaining how this currently works, only matching the beginning of the name in the team/owner components.

nmajor25: If we only want to match the beginning of the names, we could add a function parameter to `build-matcher.js` that uses a different regular expression to include the `\b` or not.

paulfalgout: oh yep.. we should probably add a second utlity for just the patient search for now

src/js/utils/formatting/build-matcher-substrings.js
nmajor25, line 1: I'd like to do just do this in the existing `build-matcher.js` file:

```js
export default memoize(function(query, includeSubstrings) {
  const searchWords = map(words(searchSanitize(query)), RegExp.escape);

  if (includeSubstrings) new RegExp(searchWords.join('|'), 'gi');

  return new RegExp(`\\b${ searchWords.join('|') }`, 'gi');
});
```

But I'm not sure if that will cause issues with the `memoize`...

src/js/utils/formatting/match-text.js
```diff
@@ -2,14 +2,16 @@
...
+import buildMatcherSubstrings from 'js/utils/formatting/build-matcher-substrings';
 
-export default (text, query, pretag, posttag) => {
+export default (text, query, options = {}) => {
   if (!text) return;
 
-  pretag = pretag || 'strong';
-  posttag = posttag || pretag;
+  const pretag = options.pretag || 'strong';
+  const posttag = options.posttag || pretag;
```
nmajor25, line 11: Note: don't think either the `pretag` or `posttag` are used anywhere. Besides in the tests.

src/js/base/helpers.cy.js
```diff
@@ -11,6 +11,7 @@ context('Handlebars helpers', function() {
       template: hbs`
         <div class="test-null">{{matchText "Patient Name" null}}</div>
         <div class="test-match">{{matchText "Patient Name" "Patient"}}</div>
+        <div class="test-match-substring">{{matchText "Testpatient Name" "patient" includeSubstrings=true}}</div>
```
nmajor25: Doing `{{matchText "TestPatient Name" "Patient" includeSubstrings=true}}` does not work interestingly. Having the  capitalized `Patient`.

src/js/utils/formatting/build-matcher-substrings.js
nmajor25: In the component test it was causing issues. I'd need to do `buildMatcher.cache = {}` or something like that to reset it so I could pass a new `includeSubstrings` parameter value. Otherwise it wouldn't update between tests.

But I couldn't find any issues when testing it in local dev.

src/js/base/helpers.cy.js
paulfalgout: I don't see how this is possible.  We should test for this and fix it.

Because 
```js
/Patient/gi.test("TestPatient Name")
&& /Patient/gi.test("Testpatient Name")
&& /patient/gi.test("TestPatient Name")
&&  /patient/gi.test("Testpatient Name")
```
is `true`

paulfalgout (changes requested):

src/js/base/helpers.cy.js
nmajor25: It does work, I was mistaken. I updated the test 👍

src/js/utils/formatting/formatting.cy.js
```diff
@@ -13,10 +14,14 @@ import words from './words';
 context('formatting', function() {
   specify('buildMatcher', function() {
     const matcher = buildMatcher('test string');
-
     expect(matcher).to.eql(/\btest|string/gi);
   });
 
+  specify('buildMatcher - include substrings', function() {
+    const match = buildMatcherSubstrings('test string', true);
```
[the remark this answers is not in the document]
nmajor25: This is a good find, thank you.

paulfalgout (approved):

PR #1498, opened 2025-08-21
Fix bug where `AttachmentAdded` not working in action sidebar in some scenarios

nmajor25:
Shortcut Story ID: [sc-63256]

Scenario where the bug was occuring:

1. Action sidebar open on a patient flow page.
2. Manual uploading of attachments for the action is not allowed. (due to `org_settings.upload_attachments` or action `canEdit` data).
3. Action has no attachments added to it yet.
4. `AttachmentAdded` ws notification is received for the action.

In this scenario the Attachments UI section is not shown at all in the sidebar. So when we get a `AttachmentAdded` ws notification, we update the collection but the view doesn't exist so it doesn't update and nothing becomes visible to the user.

## Summary by CodeRabbit

- New Features
  - Attachments panel now displays automatically when the first attachment is added in real time.
  - Attachment list updates immediately on add, replace, and remove events, ensuring the sidebar reflects the current state.

- Tests
  - Updated patient flow tests to cover real-time attachment additions, replacements, and removals.
  - Adjusted scenarios to start without pre-existing attachments and validate dynamic count changes.

### DIFF
--- a/src/js/apps/patients/sidebar/action-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/action-sidebar_app.js
@@ -87,6 +87,9 @@ export default App.extend(extend({
     this.attachments.add(model);
 
     Radio.request('ws', 'add', model);
+
+    // to cover the scenario where the AttachmentsView hasn't been shown yet
+    if (this.attachments.length === 1) this.showAttachments();
   },
   onStart(options, activity, comments, attachments) {
     this.activityCollection = new Backbone.Collection([...activity.models, ...comments.models]);
--- a/test/integration/patients/patient/patient-flow.js
+++ b/test/integration/patients/patient/patient-flow.js
@@ -113,7 +113,6 @@ context('patient flow page', function() {
 
   specify('patient flow action sidebar', function() {
     const testFileId = uuid();
-    const testOtherFile = getFile();
     const testComment = getComment();
 
     const testPatient = getPatient({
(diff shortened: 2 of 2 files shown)


src/js/apps/patients/sidebar/action-sidebar_app.js
```diff
@@ -87,6 +87,9 @@ export default App.extend(extend({
     this.attachments.add(model);
 
     Radio.request('ws', 'add', model);
+
+    // to cover the scenario where the AttachmentsView hasn't been shown yet
+    if (this.attachments.length === 1) this.showAttachments();
```
nmajor25, line 92: The two scenarios I can think of where this will be called:

1. Attachments UI section not visible on initial load of action sidebar. Then a `AttachmentAdded` ws notification is received.
2. Attachments UI section is shown and has one attachment on initial load. Then a `FileRemoved` ws notifciaton is received for the action, clearing the attachments. Then a `AttachmentAdded` ws notification is received.

test/integration/patients/patient/patient-flow.js
nmajor25, line 1: Originally the test did this:

1. Load action sidebar with existing file attachment.
2. Add second attachment via ws and do stuff with it.

Now it does this:

1. Load action sidebar with no existing file attachments. Attachments UI section is hidden.
2. Add first attachment via ws to confirm the attachment UI section becomes visible in that scenario and do stuff with it. To test the `this.attachments.length === 1` is `true` branch.
3. Add a second attachment via ws. To test the `this.attachments.length === 1` is `false` branch.

src/js/apps/patients/sidebar/action-sidebar_app.js
nmajor25: First one is the bug scenario. The second one isn't where the bug happens, since the attachments UI section exists the entire time in that scenario.

paulfalgout (approved):

PR #1499, opened 2025-08-22
In patient search results, bold dob column value when `results.match.label === 'Birth Date'`

nmajor25:
Shortcut Story ID: [sc-63163]

A dob match in search results only happens on an exact match. So we don't need to handle partial matches like the other columns, where partial strings are highlighted.

So for the dob column we can simply check `results.match.label === 'Birth Date'`. If `true`, wrap dob value in `<strong></strong>` tags.

## Summary by CodeRabbit

- New Features
  - Patient quick search results now bold the Birth Date when your query matches a patient’s DOB, improving visual clarity.
  - Picklist items more accurately reflect the matched field (e.g., Birth Date vs. Name) in the results.

- Tests
  - Expanded quick search coverage to validate DOB-based and Name-based matches, including realistic patient data and UI assertions for both scenarios.

### DIFF
--- a/test/integration/services/patient-quick-search.js
+++ b/test/integration/services/patient-quick-search.js
@@ -393,7 +393,7 @@ context('Patient Quick Search', function() {
       body: {
         data: [{
           ...searchResult,
-          attributes: { ...searchResult.attributes, match: { label: 'DOB', value: '2008-01-16' } },
+          attributes: { ...searchResult.attributes, match: { label: 'Name', value: `${ first_name } ${ last_name }` } },
         }],
         included: [getResource(testPatient, 'patients')],
       },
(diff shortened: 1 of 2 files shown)


test/integration/services/patient-quick-search.js
```diff
@@ -393,7 +393,7 @@ context('Patient Quick Search', function() {
       body: {
         data: [{
           ...searchResult,
-          attributes: { ...searchResult.attributes, match: { label: 'DOB', value: '2008-01-16' } },
```
nmajor25, line 396: `label: 'DOB'` was incorrect. Should have been `label: 'Birth Date'`. But I ended up just switching this to use the patient name as the search match value.

And then below I added a couple tests to ensure the UI was re-rendered on the copy/paste. By verifying the results and label column are emptied.

paulfalgout (approved):

PR #1500, opened 2025-08-22
Handle long strings in custom filters sidebar dropdown buttons

nmajor25:
Shortcut Story ID: [sc-63304]

Screenshot 2025-08-22 at 3 26 54 PM

### DIFF
--- a/src/js/views/patients/sidebar/filters/filters-sidebar_views.js
+++ b/src/js/views/patients/sidebar/filters/filters-sidebar_views.js
@@ -35,7 +35,7 @@ const CustomFilterView = View.extend({
   className: 'flex flex-align-center u-margin--b-8',
   template: hbs`
     <h4 class="sidebar__label">{{ name }}</h4>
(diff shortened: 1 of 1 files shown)


src/js/views/patients/sidebar/filters/filters-sidebar_views.js
```diff
@@ -35,7 +35,7 @@ const CustomFilterView = View.extend({
   className: 'flex flex-align-center u-margin--b-8',
   template: hbs`
     <h4 class="sidebar__label">{{ name }}</h4>
-    <div class="flex-grow" data-filter-button></div>
+    <div class="flex-grow u-text--overflow" data-filter-button></div>
```
nmajor25, line 38: `overflow: hidden` is all that's needed. `.button-secondary` is what's actually applying the ellipsis and nowrap to the text inside the button.

But otherwise we need to create a classname to do just `overflow: hidden`. Or inline styles I guess.

paulfalgout (approved):

PR #1501, opened 2025-08-29
Add custom filter total

paulfalgout:
Shortcut Story ID: [sc-63433]

TODO:
- [x] Tests for new feature

## Summary by CodeRabbit

- **New Features**
  - Worklist now initializes from server metadata and syncs across Actions, Flows, Schedule, and Reduced Schedule with per-type persistence.
  - Added a custom filters entity so filters load scoped to the active worklist.

- **UI/Style**
  - Filters sidebar: improved styling, clearer dropdown labels, clearable selections, preload state, and no-results text.

- **Refactor**
  - Removed legacy directories-based filters; workspace startup no longer depends on directories.

- **Tests**
  - E2E tests updated with new filter helpers/fixtures and added worklist persistence assertions.

### DIFF
--- a/src/js/apps/patients/sidebar/filters-sidebar_app.js
+++ b/src/js/apps/patients/sidebar/filters-sidebar_app.js
@@ -10,6 +10,7 @@ import { LayoutView, HeadingView, MenuView, CustomFiltersView, StatesFiltersView
 export default App.extend(extend({
   onStart({ filtersState }) {
     this.filtersState = filtersState;
+    this.filters = Radio.request('entities', 'filters:customFilters');
 
     this.showHeadingView();
     this.showMenu();
@@ -18,7 +19,15 @@ export default App.extend(extend({
     this.showStatesFiltersView();
     this.showFlowStatesFiltersView();
 
-    this.listenTo(filtersState, 'change:listType', this.showFlowStatesFiltersView);
+    this.listenTo(filtersState, {
+      'change:listType'() {
+        this.showFlowStatesFiltersView();
+        this.fetchFilters();
+      },
+      'change:worklist'() {
+        this.fetchFilters();
+      },
+    });
   },
   onClose() {
     this.stop();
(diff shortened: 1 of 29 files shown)


src/js/entities-service/entities/filters.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+  getValues() {
+    return new FilterValues(this.get('values'));
+  },
+});
+
+const Collection = BaseCollection.extend({
+  model: Model,
+  invokeFetch({ entityType, worklist }) {
+    if (!worklist) return;
```
shadowhand: What will happen if this request fails with a 400 or 410? Will the UI handle it gracefully or will it show an oops screen?

test/integration/globals/app-nav.js
```diff
@@ -235,23 +235,10 @@ context('App Nav', function() {
 
         return fx;
       })
-      .routeDirectories(fx => {
```
shadowhand, line 238: Does this mean that the only usage of directories in frontend is via formapp?

src/js/entities-service/entities/filters.js
paulfalgout: The user won't see a failure, that filter droplist just won't show and we'll get the regular datadog 410 log.

test/integration/globals/app-nav.js
paulfalgout: yep

src/js/entities-service/entities/filters.js
shadowhand: For a 400 as well?

paulfalgout: yep.  5xx will uh-oh.  4xx will log the error and just not display the filter

paulfalgout: err well a 401 will redirect to login

shadowhand: A 401 won't happen here, all good. 👍🏼

src/js/apps/patients/sidebar/filters-sidebar_app.js
```diff
@@ -37,21 +46,21 @@ export default App.extend(extend({
 
     this.showChildView('menu', menuView);
   },
+  fetchFilters() {
```
nmajor25, line 49: Need to debounce these api calls?

For example, if a user clicks the `All Filters` button 5 times in quick succession, what happens? In local dev, the app borks. But that's because of a `DEV_MAX_REQUESTS_IN_PARALLEL` setting set by the BE.

nmajor25: On QA the app doesn't bork. But it does create a lot of unnecessary network calls. And my understanding is some of these requests will contain a lot of data for the BE to retrieve.

Or maybe it's a non-issue since that all gets cached?

paulfalgout: I'll just keep the all filters button from reopening the sidebar.

paulfalgout: I added this.  This means if we do want to re-start a sidebar app we need to stop it first, which ironically is already happening in a few places.

shadowhand: @nmajor25 

> And my understanding is some of these requests will contain a lot of data for the BE to retrieve.

The loading on the database is actually less intensive with new filters, since we do more work up in the background to cache the data.

nmajor25 (approved):

PR #1504, opened 2025-09-08
Update fontawesome to v7

nmajor25:
Shortcut Story ID: [sc-62263]

## Summary by CodeRabbit

* **Style**
  * Exposed FontAwesome styles in outreach and provider Sass bundles.
  * Adjusted SVG icon rendering to prevent clipping.
  * Removed several icon-size utility classes; refined check icon size/spacing which may change list icon appearance.
* **Chores**
  * Upgraded FontAwesome dependencies to v7.0.1.


src/js/views/globals/root_views.js
```diff
@@ -6,6 +6,7 @@ import { View, CollectionView, Region } from 'marionette';
 import px from 'js/utils/formatting/px';
 
 import 'scss/modules/fill-window.scss';
+import 'scss/modules/fontawesome.scss';
```
nmajor25: This should apply to all our views, right?

nmajor25: Well not outreach...

src/scss/modules/fontawesome.scss
```diff
@@ -0,0 +1,3 @@
+svg symbol {
```
nmajor25: We could get more specific with this selector by `svg symbol[data-prefix]`. Which should only include fontawesome svg's. And should exclude any non-fontawesome svgs. Which I think is only the `/rwell-logo.svg`.

paulfalgout (changes requested):
we should not update the formio package

src/js/views/globals/root_views.js
paulfalgout: why not make it a core module? a `scss/module` does essentially indicate we'll use it in some places but maybe not in others.

src/scss/modules/fontawesome.scss
paulfalgout: we probably should do this..  or what about `svg.icon symbol`?

paulfalgout (changes requested):

src/scss/modules/fontawesome.scss
nmajor25: `svg.icon symbol` doesn't work. `svg symbol[data-icon]` does though?

src/js/views/globals/root_views.js
nmajor25: Good call, updated it to that. And removed the v7 upgrade in `/packages/care-ops-formio` as well.

nmajor25: Also I just discovered we have this core module: https://github.com/RoundingWell/care-ops-frontend/blob/develop/src/scss/core/_icon.scss.

Only one we currently use in the app is `.u-icon--12`, on a fontawesome icon. But we could move those styles to the `fontawesome.scss` core module file.

nmajor25: My vote is to remove it entirely: https://github.com/RoundingWell/care-ops-frontend/pull/1504/commits/e0ed19ac0ba20a26bcd0487271f22d55ddc40ddf

src/js/components/picklist/index.js
```diff
@@ -59,7 +59,7 @@ const PicklistItem = View.extend({
   tagName: 'li',
   itemTemplate: hbs`
     <div class="flex-grow">{{#if icon}}{{fa icon.type icon.icon classes=icon.classes}}{{/if}}<span>{{matchText text query}}</span></div>
-    {{#if isChecked}}{{fas "check" classes="u-icon--12 u-margin--l-16"}}{{/if}}
+    {{#if isChecked}}{{fas "check"}}{{/if}}
```
nmajor25, line 62: This is used only in the form submit/submit + go back dropdown. 

Screenshot 2025-09-08 at 3 02 41 PM

nmajor25: I confirmed these changes didn't break anything with the styling.

src/scss/provider-core.scss
```diff
@@ -4,8 +4,8 @@
 @forward 'core/base';
 @forward 'core/flex';
 @forward 'core/fonts';
-@forward 'core/icon';
 @forward 'core/layout';
 @forward 'core/space';
 @forward 'core/text';
 @forward 'core/typography';
+@forward 'core/fontawesome';
```
paulfalgout: these are in alphabetical order

paulfalgout (changes requested):

src/scss/provider-core.scss
nmajor25: Does `@forward 'core/reset';` need to remain in the first spot?

paulfalgout (approved):

PR #1510, opened 2025-09-10
On `GET /api/actions` & `GET /api/flows` 400 errors, reset to default filters state

nmajor25:
Shortcut Story ID: [sc-63625]

## Summary by CodeRabbit

* **Bug Fixes**
  * Filters in Schedule, Reduced Schedule, and Worklist now reset to safe defaults on specific 400 responses, clearing incompatible custom filters to avoid stuck or inconsistent views.

* **Refactor**
  * Filter state was consolidated onto the app instance for more consistent, persistent behavior across pages.

* **Tests**
  * Added end-to-end tests verifying 400-error recovery and local preference reset (workspace- and version-scoped).

### DIFF
--- a/src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
+++ b/src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
@@ -1,3 +1,4 @@
+import { get } from 'underscore';
 import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
@@ -39,15 +40,15 @@ export default App.extend({
   startFiltersApp({ setDefaults } = {}) {
     const filtersApp = this.startChildApp('filters', { state: this.getState().getFiltersState() });
 
-    const filterState = filtersApp.getState();
+    this.filterState = filtersApp.getState();
 
-    filtersApp.listenTo(filterState, 'change', () => {
-      this.setState(filterState.getFiltersState());
+    filtersApp.listenTo(this.filterState, 'change', () => {
+      this.setState(this.filterState.getFiltersState());
     });
 
-    if (setDefaults) filterState.setDefaultFilterStates();
+    if (setDefaults) this.filterState.setDefaultFilterStates();
 
-    this.setState(filterState.getFiltersState());
+    this.setState(this.filterState.getFiltersState());
   },
   onChangeSearchQuery(state) {
     this.currentSearchQuery = state.get('searchQuery');
@@ -112,6 +113,12 @@ export default App.extend({
     this.showCountView();
 
     this.showList();
+  },
+  /* istanbul ignore next: error handling */
+  onFail(options, error) {
+    if (get(error, ['response', 'status']) === 400) {
+      this.filterState.setDefaultFilterStates();
+    }
   },
   setWorklist(worklist) {
     this.getState().setWorklist(worklist);
--- a/src/js/apps/patients/schedule/schedule_app.js
+++ b/src/js/apps/patients/schedule/schedule_app.js
@@ -1,3 +1,4 @@
+import { get } from 'underscore';
 import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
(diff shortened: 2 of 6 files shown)


test/integration/patients/worklist/worklist.js
```diff
@@ -5020,6 +5020,58 @@ context('worklist page', function() {
...
+      .should('contain', '3')
+      .click();
+
+    cy
+      .get('.app-frame__sidebar .sidebar')
+      .find('[data-states-filters-region]')
+      .find('[data-check-region]')
+      .eq(0)
+      .click()
+      .wait('@routeActions');
```
nmajor25: This uses the intercepted `/api/actions` request with a `400` response. Which triggers the `setDefaultFilterStates()`, which will trigger another `/api/actions` request.

I can't get that second request to not use the `400` intercept. Ideally that second one would be `200` that shows the actions in the list. But both requests use the `400` intercept.

But it doesn't effect the objective of the test, since we just need to test that the filters state is reset to defaults. The actions showing in the list doesn't really matter.

Screenshot 2025-09-10 at 12 46 12 PM

paulfalgout: the way to do this is for the intercept to also match part of the query that changes

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -335,4 +335,9 @@ export default App.extend({
 
     this.showChildView('search', searchComponent);
   },
+  onFail(options, { response }) {
+    if (response.status === 400) {
+      this.filterState.setDefaultFilterStates();
+    }
+  },
```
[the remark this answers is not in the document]
paulfalgout: I do think in a few places we use get(error, ['response', 'status']) === for this reason

test/integration/patients/worklist/worklist.js
```diff
@@ -5020,6 +5020,58 @@ context('worklist page', function() {
...
+    }));
+
+    cy
+      .routesForPatientAction()
+      .routeActions()
+      .visit('/worklist/owned-by')
+      .wait('@routeActions');
+
+    cy
+      .intercept('GET', '/api/actions?*', {
```
paulfalgout: I think if this intercept is a regex it can match for the customFilter which won't overwrite the typical routeActions allowing for going back to a 200

paulfalgout (changes requested):

test/integration/patients/worklist/worklist.js
nmajor25: Yep, that was the way to go. Just pushed an update using a regex.

paulfalgout (changes requested):
We need to use a safer get for response.status`get(error, ['response', 'status']) === 400`

And it needs an istanbul ignore.. on at least the else since we're not testing non-400 errors.

src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
```diff
@@ -191,4 +191,9 @@ export default App.extend({
 
     this.showChildView('search', searchComponent);
   },
+  onFail(options, { response }) {
```
paulfalgout: can we also move these `onFails` up under the `onStart`s

paulfalgout (changes requested):

src/js/apps/patients/schedule/reduced-schedule/reduced_schedule_app.js
nmajor25: Changes are pushed 👍

paulfalgout (approved):

src/js/apps/patients/schedule/schedule_app.js
```diff
@@ -1,3 +1,4 @@
+import { get } from 'underscore';
```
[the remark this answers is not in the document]
paulfalgout: nope

paulfalgout (approved):

PR #1511, opened 2025-09-11
Modernize fill-window and z-index

paulfalgout:
Shortcut Story ID: [sc-63656]

## Summary by CodeRabbit

- New Features
  - Full-screen overlays now reliably cover the viewport.

- Bug Fixes
  - Resolved layering so sidebars, context trails, flows, and the error page display above other content.
  - Patient search close button consistently sits above surrounding elements.

- Style
  - Standardized stacking using theme variables for predictable z-index across components.

- Chores
  - Centralized z-index values into a shared scheme for easier maintenance.

### DIFF
--- a/src/scss/provider-core.scss
+++ b/src/scss/provider-core.scss
@@ -9,3 +9,4 @@
 @forward 'core/space';
 @forward 'core/text';
 @forward 'core/typography';
+@forward 'core/z-index';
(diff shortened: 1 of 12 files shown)


src/scss/provider-core.scss
```diff
@@ -9,3 +9,4 @@
 @forward 'core/space';
 @forward 'core/text';
 @forward 'core/typography';
+@forward 'core/z-index';
```
nmajor25, line 12: Do we need this forwarded in the `src/scss/outreach-core.scss` file too? I don't think z-index is used there. But it may be in the future...?

nmajor25 (approved):

PR #1512, opened 2025-09-11
Fix intermittently failing patient archive websocket tests

nmajor25:
Shortcut Story ID: [sc-63584]

Example cypress run fail: https://cloud.cypress.io/projects/ep9zr6/runs/8006/overview?roarHideRunsWithDiffGroupsAndTags=1

## Summary by CodeRabbit

* **Bug Fixes**
  * Real-time name changes now display instantly and consistently across relevant lists.
  * Timestamp displays are more consistent and reliable, including after live updates.
* **Tests**
  * Stabilized time-dependent scenarios to reduce flakiness in real-time update flows.

### DIFF
--- a/test/integration/patients/patient/archive.js
+++ b/test/integration/patients/patient/archive.js
@@ -399,6 +399,8 @@ context('patient archive page', function() {
   });
 
   specify('flow list - socket notifications', function() {
+    const testDateTime = testTs();
+
     const testPatient = getPatient({
       relationships: {
         workspaces: getRelationship(workspaceOne),
@@ -408,7 +410,7 @@ context('patient archive page', function() {
     const testSocketFlow = getFlow({
       attributes: {
         name: 'Test Flow - Subscribed on Page Load',
-        updated_at: testTs(),
+        updated_at: testTsSubtract(1),
       },
       relationships: {
         state: getRelationship(stateDone),
@@ -420,7 +422,6 @@ context('patient archive page', function() {
     const testNewSocketFlow = getFlow({
       attributes: {
         name: 'New Flow - Created Elsewhere',
-        updated_at: testTsSubtract(2),
       },
       relationships: {
         state: getRelationship(stateDone),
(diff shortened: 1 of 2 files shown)


test/integration/patients/patient/archive.js
```diff
@@ -881,11 +920,6 @@ context('patient archive page', function() {
       .get('@firstRow')
       .should('contain', 'New Name Via Websocket');
 
-    cy
-      .get('@firstRow')
-      .find('.patient__action-ts')
-      .should('contain', formatDate(testTs(), 'TIME_OR_DAY'));
```
nmajor25, line 887: This wasn't the proper way of testing this. It was being tested on a action/flow that had been added via web sockets and then received a subsequent ws message. So the updated_at for that action/flow had already been updated multiple times prior to that.

It should instead be tested via:

1. Load page with action/flow that has updated_at timestamp from day previous.
2. Send ws `NameChange` notification.
3. Then ensure the `updated_at` timestamp is `testDate()`.

Which is how the pt dashboard does it.

test/integration/patients/patient/dashboard.js
```diff
@@ -1027,7 +1028,6 @@ context('patient dashboard page', function() {
     const testNewStateSocketFlow = getFlow({
       attributes: {
         name: 'New Flow - State Updated to Match Current Filter',
-        updated_at: testTsSubtract(1),
```
nmajor25, line 1030: These aren't needed. I think I originally put these here to ensure the order of list items in the UI. So I could do a `cy.get('@firstRow')` and get the correct item. But these will automatically be pushed to the top of the list based on the updated_at view sort.

test/integration/patients/patient/archive.js
```diff
@@ -399,6 +399,8 @@ context('patient archive page', function() {
   });
 
   specify('flow list - socket notifications', function() {
+    const testDateTime = testTs();
```
nmajor25, line 402: I'm not entirely sure if this is what's solving the issue. But it seems like the correct thing to do. To use a consistent `testTs()` throughout the test.

nmajor25: I tested the test locally. I couldn't get the test to error when starting the test at `:58` or `:59` second mark of the current minute. Which was causing the test to fail previously.

paulfalgout: oh yep.. an improvement at least

paulfalgout (approved):
I like removing attributes from the tests too.. minimal attribute setup is a win

PR #1515, opened 2025-09-17
Monkey patch old five9 umd wrapper

paulfalgout:
Shortcut Story ID: [sc-###]

## Summary by CodeRabbit

- New Features
  - Added named exports for key SDK APIs, enums and status/error codes alongside the default SDK export.
  - Exposed the SDK version via a new named export.

- Bug Fixes
  - Updated UMD wrapper to use globalThis for environment-agnostic global access and fixed wrapper binding.
  - Removed previous global cleanup behavior; improved robustness when loading the SDK.
  - Added local fallback when SDK download fails.

- Chores
  - Removed an internal ESLint configuration specific to the SDK package.


packages/five9/update-sdk.js
```diff
@@ -126,31 +138,31 @@ async function main() {
...
 
 // Export the main SDK
 export default Five9SDK;
 
 // Export individual APIs for convenience
-${ apiExports }
+${apiExports}
 
 // Export error codes and status codes
-${ enumExports }
```
nmajor25: These eslint code styles should stay, right? Same for line 141 & 142 above.

nmajor25: And line 64 has no trailing comma.

nmajor25 (approved):

PR #1522, opened 2025-09-25
Add phone dialer component to the action sidebar

nmajor25:
Shortcut Story ID: [sc-63637]

- [x] Implement in app UI
- [x] Disabled when action is in read only state
- [x] Cypress tests
- [x] Code cleanup/refactoring

## Summary by CodeRabbit

- **New Features**
  - Patient sidebar adds a Dialer for quick access to patient phone numbers and one‑click calling (when enabled); new dialer component and view integrated.

- **UI**
  - Picklist shows an animated loading state with an icon; sidebar has a dedicated dialer area.

- **Style**
  - New styles for dialer button, labels, highlighted states, and loading animation.

- **Localization**
  - Added English dialer labels and texts.

- **Tests**
  - Integration specs covering the dialer workflow and UI interactions.

- **Bug Fixes**
  - Option list selection now safely handles missing selection callbacks; added two new icons: circle and phone.

### DIFF
--- a/src/js/components/picklist/index.js
+++ b/src/js/components/picklist/index.js
@@ -1,5 +1,4 @@
 import { debounce, each, extend, noop, pick, result, size } from 'underscore';
-import { animate } from 'animejs';
 import hbs from 'handlebars-inline-precompile';
 import { View, CollectionView } from 'marionette';
 
@@ -54,34 +53,15 @@ const PicklistEmpty = View.extend({
   className: 'picklist__message',
   template: hbs`
     {{#if isLoading}}
-      <div class="picklist__message-loading js-loading">
+      <div class="picklist__message-loading">
         {{fas "circle"}}{{ loadingText }}
       </div>
     {{else}}
       {{ noResultsText }}
     {{/if}}
   `,
-  ui: {
-    loading: '.js-loading',
-  },
   serializeData() {
     return this.options;
-  },
-  onRender() {
-    if (this.getOption('isLoading')) {
-      this._loadingAnimation = animate(this.ui.loading[0], {
-        opacity: { to: 1, duration: 400 },
-        loop: Infinity,
-        ease: 'inOutSine',
-        alternate: true,
-      });
-    }
-  },
-  onDestroy() {
-    if (this._loadingAnimation) {
-      this._loadingAnimation.cancel();
-      this._loadingAnimation = null;
-    }
   },
 });
 
--- a/src/js/components/picklist/index.js
+++ b/src/js/components/picklist/index.js
@@ -50,14 +50,15 @@ const isSelectList = false;
 
 const PicklistEmpty = View.extend({
   tagName: 'li',
-  className: 'picklist__message',
   template: hbs`
     {{#if isLoading}}
       <div class="picklist__message-loading">
         {{fas "circle"}}{{ loadingText }}
       </div>
     {{else}}
-      {{ noResultsText }}
+      <div class="picklist__message">
+        {{ noResultsText }}
+      </div>
     {{/if}}
   `,
   serializeData() {
--- a/src/js/components/picklist/index.js
+++ b/src/js/components/picklist/index.js
@@ -1,4 +1,5 @@
 import { debounce, each, extend, noop, pick, result, size } from 'underscore';
+import { animate } from 'animejs';
 import hbs from 'handlebars-inline-precompile';
 import { View, CollectionView } from 'marionette';
 
@@ -53,13 +54,26 @@ const PicklistEmpty = View.extend({
   className: 'picklist__message',
   template: hbs`
     {{#if isLoading}}
-      {{ loadingText }}
+      <div class="picklist__message-loading js-loading">
+        {{fas "circle"}}{{ loadingText }}
+      </div>
     {{else}}
       {{ noResultsText }}
     {{/if}}
   `,
+  ui: {
+    loading: '.js-loading',
+  },
   serializeData() {
     return this.options;
+  },
+  onRender() {
+    animate(this.ui.loading[0], {
+      opacity: { to: 1, duration: 400 },
+      loop: Infinity,
+      ease: 'inOutSine',
+      alternate: true,
+    });
   },
 });
 
--- a/src/js/components/picklist/index.js
+++ b/src/js/components/picklist/index.js
@@ -68,12 +68,20 @@ const PicklistEmpty = View.extend({
     return this.options;
   },
   onRender() {
-    animate(this.ui.loading[0], {
-      opacity: { to: 1, duration: 400 },
-      loop: Infinity,
-      ease: 'inOutSine',
-      alternate: true,
-    });
+    if (this.getOption('isLoading')) {
+      this._loadingAnimation = animate(this.ui.loading[0], {
+        opacity: { to: 1, duration: 400 },
+        loop: Infinity,
+        ease: 'inOutSine',
+        alternate: true,
+      });
+    }
+  },
+  onDestroy() {
+    if (this._loadingAnimation) {
+      this._loadingAnimation.cancel();
+      this._loadingAnimation = null;
+    }
   },
 });
 
--- a/src/js/components/picklist/picklist.scss
+++ b/src/js/components/picklist/picklist.scss
@@ -69,8 +69,19 @@
   padding: 4px 16px;
 }
 
+@keyframes picklist-pulse {
+  from {
+    opacity: 0;
+  }
+
+  to {
+    opacity: 1;
+  }
+}
+
 .picklist__message-loading {
   align-items: center;
+  animation: picklist-pulse 400ms ease-in-out infinite alternate;
   display: flex;
   font-size: 12px;
   font-style: normal;
--- a/src/js/components/picklist/picklist.scss
+++ b/src/js/components/picklist/picklist.scss
@@ -82,11 +82,11 @@
 .picklist__message-loading {
   align-items: center;
   animation: picklist-pulse 400ms ease-in-out infinite alternate;
+  color: #{$black-60};
   display: flex;
   font-size: 12px;
-  font-style: normal;
   justify-content: center;
-  margin: 8px 0;
+  margin: 16px 0;
 
   .icon {
     font-size: 8px;
--- a/src/js/components/picklist/picklist.scss
+++ b/src/js/components/picklist/picklist.scss
@@ -67,8 +67,18 @@
   font-size: 14px;
   font-style: italic;
   padding: 4px 16px;
+}
 
-  .icon:first-child {
+.picklist__message-loading {
+  align-items: center;
+  display: flex;
+  font-size: 12px;
+  font-style: normal;
+  justify-content: center;
+  margin: 8px 0;
+
+  .icon {
+    font-size: 8px;
     margin-right: 8px;
   }
 }
--- a/src/js/views/patients/shared/components/dialer-component.scss
+++ b/src/js/views/patients/shared/components/dialer-component.scss
@@ -17,7 +17,7 @@
     display: none;
 
     .icon {
-      font-size: 11px;
+      font-size: 12px;
       margin: 0 0 2px 4px;
     }
   }
--- a/src/js/views/patients/shared/components/dialer-component.scss
+++ b/src/js/views/patients/shared/components/dialer-component.scss
@@ -1,13 +1,13 @@
 .dialer-component__phone-icon {
-  color: #0582DA;
+  color: color(blue);
 
   .picklist__item.is-highlighted & {
-    color: white;
+    color: #{$white};
   }
 }
 
 .dialer-component__phone-label {
-  color: #999;
+  color: #{$black-60};
   font-size: 12px;
   margin-left: auto;
   text-transform: capitalize;
@@ -23,7 +23,7 @@
   }
 
   .picklist__item.is-highlighted & {
-    color: white;
+    color: #{$white};
 
     .dialer-component__phone-label-default {
       display: none;
--- a/src/js/views/patients/shared/components/dialer-component.scss
+++ b/src/js/views/patients/shared/components/dialer-component.scss
@@ -1,0 +1,36 @@
+.dialer-component__phone-icon {
+  color: #0582DA;
+
+  .picklist__item.is-highlighted & {
+    color: white;
+  }
+}
+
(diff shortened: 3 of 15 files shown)


src/js/components/picklist/picklist.scss
```diff
@@ -67,8 +67,18 @@
   font-size: 14px;
   font-style: italic;
   padding: 4px 16px;
+}
 
-  .icon:first-child {
```
nmajor25, line 71: I couldn't find anywhere where this is used. No results text that has an icon as well.

src/js/components/picklist/index.js
```diff
@@ -53,14 +54,27 @@ const PicklistEmpty = View.extend({
   className: 'picklist__message',
   template: hbs`
     {{#if isLoading}}
-      {{ loadingText }}
+      <div class="picklist__message-loading js-loading">
+        {{fas "circle"}}{{ loadingText }}
+      </div>
```
nmajor25, line 57: This makes this the loading ui/styling for any picklist with the `isListsAsync: true` option. I figured this is fine for now. And we can figure out a way to customize via an options item it in the future if the need arises. Like if we want to use a different icon, spacing, or something across different picklists.

nmajor25: Other way would be to pass a `loadingTemplate`option or something similar that's a handlebars template.

src/js/views/patients/shared/components/dialer-component.scss
```diff
@@ -0,0 +1,36 @@
+.dialer-component__phone-icon {
+  color: #0582DA;
```
paulfalgout: color vars

src/js/views/patients/shared/components/dialer_component.js
```diff
@@ -0,0 +1,96 @@
...
+
+    return phones;
+  },
+  viewEvents: {
+    'click': 'onClick',
+  },
+  viewOptions() {
+    return {
+      tagName: 'button',
+      className: 'button-secondary w-100 action-sidebar__form',
```
paulfalgout: should probably make action-sidebar__form something more generic if it should be applied here as well

src/js/views/patients/shared/components/dialer_component.js
```diff
@@ -0,0 +1,96 @@
...
+    };
+  },
+  onClick() {
+    const view = this.getView();
+    view.$el.blur();
+
+    const lists = this.getLists();
+
+    const optionlist = new Optionlist({
+      attr: 'number',
```
paulfalgout: probably don't need this now

paulfalgout (changes requested):

src/js/views/patients/shared/components/dialer_component.js
nmajor25: We might not need that. Looks like it's just applying the blue coloring. 

I'm going to go clean some of the code up in a final commit. I'll include this change.

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -22,7 +22,6 @@ import { roleNoFilterEmployee, roleTeamEmployee } from 'support/api/roles';
 import { getFlow } from 'support/api/flows';
 import { getFile } from 'support/api/files';
 
-// TODO: Update to mergejson api
```
nmajor25, line 25: Don't think we need this comment anymore. These have all been switched over.

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -628,6 +629,11 @@ context('action sidebar', function() {
       .next()
       .should('contain', formatDate(testTs(), 'AT_TIME'));
 
+    cy
+      .get('.sidebar')
+      .find('[data-dialer-region]')
+      .should('be.empty');
```
nmajor25, line 635: Ensures dialer is not shown when 'settings.dialer' is not present or null.

test/integration/patients/sidebar/action-sidebar.js
```diff
@@ -712,6 +718,158 @@ context('action sidebar', function() {
       .should('contain', 'The Action you requested does not exist');
   });
 
+  specify('action phone dialer', function() {
```
nmajor25, line 721: Tests everything up until a phone number is clicked, which will open the five9 dialer UI when that's implemented.

Quick summary of test:

1. Open dialer picklist and verify phone numbers are shown properly and are ordered based on `preferred` value.
2. Close dialer picklist and open it again. Verify phone numbers are displayed via cached data. Api request shouldn't be made again.
3. Change action to done and verify dialer button is set to `disabled`.
4. Change parent flow to done and verify dialer button is set to `disabled`.

nmajor25: Potentially move some of this to a component test on the `dialer_component.js`.

src/js/views/patients/shared/components/dialer_component.js
```diff
@@ -0,0 +1,101 @@
...
+    return this._lists;
+  },
+  getPhones() {
+    const phones = new Backbone.Collection(this.field.get('value'), {
+      comparator(model) {
+        return model.get('preferred') ? 0 : 1;
+      },
+    });
+
+    phones.each(model => {
```
paulfalgout: I think I'd like to move away from setting onSelect on the model.  at least in this case it makes less sense because every handler is the same.  

We maybe need to modify the optionlist to only call onSelect if it's set, but we should just be able to listen to the `select` event on the optionlist with the same handler

src/js/components/picklist/index.js
```diff
@@ -53,14 +54,27 @@ const PicklistEmpty = View.extend({
...
     return this.options;
   },
+  onRender() {
+    animate(this.ui.loading[0], {
+      opacity: { to: 1, duration: 400 },
+      loop: Infinity,
+      ease: 'inOutSine',
+      alternate: true,
+    });
+  },
```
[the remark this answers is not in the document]
paulfalgout: This is true.  animations that end effectively clean up themselves, but not looping animations.

test/integration/patients/sidebar/action-sidebar.js
paulfalgout: yep we don't need to test anything specific to the component here

paulfalgout (changes requested):

src/js/views/patients/shared/components/dialer_component.js
```diff
@@ -0,0 +1,101 @@
...
+      model.set('onSelect', () => {
+        Radio.request('dialer', 'call', model.get('number'));
```
paulfalgout: ah one other thing the 4th argument for this request needs to be the action model @nmajor25

src/js/components/picklist/index.js
```diff
@@ -53,14 +54,35 @@ const PicklistEmpty = View.extend({
...
     {{/if}}
   `,
+  ui: {
+    loading: '.js-loading',
+  },
   serializeData() {
     return this.options;
   },
+  onRender() {
+    if (this.ui.loading[0]) {
```
paulfalgout: I think we should check the data and not the dom for this check

src/js/views/patients/shared/components/dialer-component.scss
```diff
@@ -0,0 +1,36 @@
...
+  color: #{$black-60};
+  font-size: 12px;
+  margin-left: auto;
+  text-transform: capitalize;
+
+  .dialer-component__phone-label-call {
+    align-items: center;
+    display: none;
+
+    .icon {
```
paulfalgout, line 19: this icon styles look very unique.  don't see 11px anywhere and don't love the 2px either..

src/js/views/patients/sidebar/action/action-sidebar.scss
```diff
@@ -5,7 +5,8 @@
   padding: 8px 0;
 }
 
-.action-sidebar__form {
+.action-sidebar__form-button,
+.action-sidebar__dialer-button {
```
paulfalgout: I don't think it being blue is related to form or dialer.. this should be something like action-sidebar__function

paulfalgout (changes requested):

src/js/components/picklist/index.js
nmajor25: This is updated to check `this.getOption('isLoading')` instead of the dom.

nmajor25: Although I do think this would be simpler not using animejs and just doing it in the CSS instead.

Via:

```css
.picklist__message-loading {
  animation: pulse 400ms ease-in-out infinite alternate;

  . . .
}

@keyframes pulse {
  from {
    opacity: 0;
  }

  to {
    opacity: 1;
  }
}
```

src/js/views/patients/sidebar/action/action-sidebar.scss
nmajor25: ~Updated this to one `.action-sidebar__blue-text` class~

src/js/views/patients/shared/components/dialer-component.scss
nmajor25: Updated font-size to `12px`

nmajor25: The `2px` makes the alignment appear a little better, is why I added it. The icon is vertically centered via flex and align items center. But the design of the icon next to the text makes it look off (at least to me).

With `margin-bottom: 2px`:

Screenshot 2025-09-30 at 11 28 50 AM

Without:

Screenshot 2025-09-30 at 11 29 00 AM

src/js/components/picklist/index.js
nmajor25: But we can look into that in a future task if we want to do that instead.

paulfalgout: css animation seems fine to me.. cleaner

src/js/views/patients/shared/components/dialer-component.scss
paulfalgout: does vertical-align on the icon itself do anything?  might not with line-height.. but like baseline or middle might improve it as well.

nmajor25: Vertical align doesn't seem to have any effect...

src/js/views/patients/sidebar/action/action-sidebar.scss
nmajor25: Updated this to one `.action-sidebar__button` class: https://github.com/RoundingWell/care-ops-frontend/pull/1522/commits/7dba12e9dadf79d88736cb82d91c2bf87b4eb652

src/js/components/picklist/index.js
nmajor25: I just switched this over to handle the animation in css instead of animejs: https://github.com/RoundingWell/care-ops-frontend/pull/1522/commits/f2ad84769485f80e99b94eb73ae58ce088c6a755

src/js/views/patients/shared/components/dialer-component.scss
paulfalgout: well we should at least add a comment so I don't remove it later

src/js/components/picklist/picklist.scss
```diff
@@ -67,8 +67,29 @@
...
+@keyframes pulse {
+  from {
+    opacity: 0;
+  }
+
+  to {
+    opacity: 1;
+  }
+}
```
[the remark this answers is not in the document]
paulfalgout: @nmajor25 what do you think about this.. maybe .picklist prefixing the keyframes is a good idea to prevent collisions

src/js/components/picklist/index.js
```diff
@@ -53,11 +53,16 @@ const PicklistEmpty = View.extend({
...
     {{#if isLoading}}
-      {{ loadingText }}
+      <div class="picklist__message-loading js-loading">
+        {{fas "circle"}}{{ loadingText }}
+      </div>
     {{else}}
       {{ noResultsText }}
     {{/if}}
   `,
+  ui: {
```
paulfalgout: looks like we can remove the selector and ui element?

src/js/components/picklist/index.js
```diff
@@ -53,11 +53,16 @@ const PicklistEmpty = View.extend({
   className: 'picklist__message',
   template: hbs`
     {{#if isLoading}}
-      {{ loadingText }}
+      <div class="picklist__message-loading js-loading">
+        {{fas "circle"}}{{ loadingText }}
+      </div>
     {{else}}
       {{ noResultsText }}
```
paulfalgout: what if we wrapped this in a div class="picklist__message"  and removed it from the view className?

paulfalgout (changes requested):
just down to nitpicks I think

src/js/components/picklist/picklist.scss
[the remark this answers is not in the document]
nmajor25: It wouldn't hurt 👍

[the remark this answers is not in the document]
shadowhand: The 🐇 did good with that one. Especially like the accessibility recommendation for reduced animation.

src/js/components/picklist/index.js
```diff
@@ -53,11 +53,16 @@ const PicklistEmpty = View.extend({
   className: 'picklist__message',
   template: hbs`
     {{#if isLoading}}
-      {{ loadingText }}
+      <div class="picklist__message-loading js-loading">
```
shadowhand: Is `js-loading` still needed, now that the animation is in CSS?

src/js/components/picklist/index.js
nmajor25: Yep, not needed anymore

src/js/components/picklist/index.js
nmajor25: Nope, not needed 👍

src/js/components/picklist/index.js
nmajor25: That's been removed now 👍

src/js/components/picklist/index.js
nmajor25: That works, updated in https://github.com/RoundingWell/care-ops-frontend/pull/1522/commits/4ebbc7f44da91e3106c38450c25409560014a497.

nmajor25: But broke some tests...

paulfalgout (approved):
Looks good. can refactor later if necessary

PR #1524, opened 2025-10-01
In `patient_app`, clear action `isEditing` state in `startRoute()`

nmajor25:
Shortcut Story ID: [sc-63858]

To fix bug where isEditing (adds `.is-selected` to action list items) persists when navigating between patient dashboard & archive routes.

To do:

- [x] Add section to cypress test that fails without clearing `this.action` cache between route changes.

## Summary by CodeRabbit

- New Features
  - None

- Bug Fixes
  - Editing mode now reliably exits when navigating within the patient app.
  - Selection state is cleared after switching between Dashboard and Archive tabs and returning to Dashboard.
  - Smoother sidebar behavior reduces flicker during tab transitions.

- Tests
  - Expanded integration tests to cover tab switching (Dashboard ↔ Archive), selection reset, and animation timing to ensure stable UI behavior.

### DIFF
--- a/src/js/apps/patients/patient/patient_app.js
+++ b/src/js/apps/patients/patient/patient_app.js
@@ -36,6 +36,13 @@ export default SubRouterApp.extend({
     };
   },
 
+  onBeforeStartRoute() {
+    if (this.action) {
+      this.action.trigger('editing', false);
+      delete this.action;
+    }
+  },
+
   onBeforeStart() {
     this.getRegion().startPreloader();
   },
--- a/test/integration/patients/patient/dashboard.js
+++ b/test/integration/patients/patient/dashboard.js
@@ -238,7 +238,8 @@ context('patient dashboard page', function() {
     cy
       .get('.list-page__list')
       .contains('First In List')
-      .click();
+      .click()
+      .tick(350); // this test uses visitOnClock, so needed for the sidebar animation
 
     cy
       .get('.list-page__list')
@@ -320,6 +321,27 @@ context('patient dashboard page', function() {
       .should(({ data }) => {
         expect(data.attributes.due_time).to.equal('09:45:00');
       });
+
+    cy
+      .get('.patient__tabs')
+      .find('.js-archive')
+      .click();
+
+    cy
+      .get('.patient__tabs')
+      .find('.js-dashboard')
+      .click();
+
+    cy
+      .get('.list-page__list')
+      .find('.is-selected')
+      .should('not.exist');
+
+    cy
+      .get('.list-page__list')
+      .contains('First In List')
+      .click()
+      .tick(350); // this test uses visitOnClock, so needed for the sidebar animation
 
     cy
       .get('.list-page__list')
--- a/src/js/base/subrouterapp.js
+++ b/src/js/base/subrouterapp.js
@@ -20,12 +20,17 @@ export default App.extend({
   },
 
   // Start a route from a currentRoute passed from a routerapp
-  startRoute({ event, eventArgs }) {
+  startRoute(currentRoute) {
+    this.triggerMethod('before:startRoute', currentRoute);
+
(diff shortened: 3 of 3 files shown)


src/js/apps/patients/patient/patient_app.js
```diff
@@ -29,6 +29,15 @@ export default SubRouterApp.extend({
...
   },
 
+  startRoute(currentRoute) {
+    if (this.action) {
+      this.action.trigger('editing', false);
+      delete this.action;
+    }
+
+    SubRouterApp.prototype.startRoute.call(this, currentRoute);
+  },
```
nmajor25: This fixes the issue without breaking anything else, but not sure if it's the correct way to do it yet. Seems like the simplest solution to me though. I think updating `startList()` and `startPatientAction()` directly will get complicated.

nmajor25: Problem is `this.action` is being cached between route changes between patient dashboard and archive. Which means that the action gets an `isEditing = true` set on it regardless of the sidebar being open or not when the route is loaded.

There's an `onStop()` that does `delete this.action`. But this app isn't being stopped when navigating between patient dashboard and archive routes.

test/integration/patients/patient/dashboard.js
```diff
@@ -321,6 +322,27 @@ context('patient dashboard page', function() {
...
+    cy
+      .get('.list-page__list')
+      .find('.is-selected')
+      .should('not.exist');
+
+    cy
+      .get('.list-page__list')
+      .contains('First In List')
+      .click()
+      .tick(350); // this test uses visitOnClock, so needed for the sidebar animation
```
nmajor25, line 344: At this point in the test, cypress is on the pt dashboard with the action sidebar open.

I added this:

1. Navigate to the patient archive.
2. Navigate back to the patient dashboard.
3. Verify the `.is-selected` class didn't remain on the action list item.
4. Open the acton sidebar back up. Other parts of the test rely on that.

src/js/apps/patients/patient/patient_app.js
paulfalgout: I think the pattern we should use is adding a `before:startRoute` and a `startRoute` in the subrouterapp event and we'd delete it in the `before:startRoute` rather than calling the prototype.

paulfalgout (changes requested):

src/js/apps/patients/patient/patient_app.js
nmajor25: I'm not 100% sure how to implement that.

Like this?

Trigger `'before:startRoute'` in the `startRoute` (`/js/base/subrouterapp.js`):

```js
startRoute(currentRoute) {
  this.trigger('before:startRoute', currentRoute);

  const { event, eventArgs } = currentRoute;
  const action = this._eventRoutes[event];

  if (!action) return;

  action.apply(this, eventArgs);
},
```

And then add a constructor in the `patient_app.js`:

```js
constructor: function() {
  this.on('before:startRoute', () => {
    if (this.action) {
      this.action.trigger('editing', false);
      delete this.action;
    }
  });

  SubRouterApp.apply(this, arguments);
},
```

paulfalgout:
```js
startRoute(currentRoute) {
  this.triggerMethod('before:startRoute', currentRoute);

  const { event, eventArgs } = currentRoute;
  const action = this._eventRoutes[event];

  if (!action) return;

  action.apply(this, eventArgs);

  this.triggerMethod('startRoute', currentRoute);
},
```
then in patientApp it's just adding an `onBeforeStartRoute() { ... },`

nmajor25: Gotcha, thanks 👍 . Update is made.

paulfalgout (approved):

PR #1532, opened 2025-10-07
Refactor optionlist onSelect handlers

nmajor25:
Shortcut Story ID: [sc-64123]

- [x] Relatively simple refactors
- [x] Refactor add workflow app/views optionlist
- [x] Remove code from optionlist component

## Summary by CodeRabbit

* **Refactor**
  * Unified option selection across menus to a centralized event-driven flow, replacing per-item callbacks.
  * Standardized triggers for add, edit, delete, and archive actions for more consistent behavior.
  * Reduced reliance on utility bindings for cleaner, more maintainable code.

* **Improvements**
  * More reliable action handling from menus.
  * Bulk select toggles only when needed, avoiding unnecessary state changes.

* **Tests**
  * Updated component and integration tests to align with the new selection flow.
  * Added brief waits to stabilize list updates during test runs.

### DIFF
--- a/src/js/apps/patients/patient/flow/flow_app.js
+++ b/src/js/apps/patients/patient/flow/flow_app.js
@@ -1,4 +1,4 @@
-import { bind, get } from 'underscore';
+import { get } from 'underscore';
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 
@@ -146,16 +146,22 @@ export default SubRouterApp.extend({
         text: action.get('name'),
         itemType: action.type,
         hasOutreach: action.hasOutreach(),
-        onSelect: bind(this.triggerMethod, this, 'add:programAction', action),
+        programItem: action,
       };
     });
   },
 
   showAdd() {
-    this.showChildView('tools', new AddButtonView({
+    const addButtonView = new AddButtonView({
       headingText: i18n.addActionHeadingText,
       lists: [{ collection: new Backbone.Collection(this.addOpts) }],
-    }));
+    });
+
+    this.listenTo(addButtonView, 'add:programAction', programItem => {
+      this.triggerMethod('add:programAction', programItem);
+    });
+
+    this.showChildView('tools', addButtonView);
   },
 
   toggleBulkSelect() {
--- a/src/js/components/optionlist/index.js
+++ b/src/js/components/optionlist/index.js
@@ -30,7 +30,9 @@ export default Picklist.extend({
   },
   viewTriggers: {
     'close': 'close',
-    'picklist:item:select': 'select',
+  },
+  viewEvents: {
+    'picklist:item:select': 'onPicklistSelect',
   },
   position() {
     return this.uiView.getBounds(this.ui);
@@ -45,17 +47,10 @@ export default Picklist.extend({
   onClose() {
     this.destroy();
   },
-  onSelect({ model }) {
+  onPicklistSelect({ model }) {
     if (model.get('isDisabled')) return;
 
-    const onSelect = model.get('onSelect');
-
-    if (!onSelect) {
-      this.destroy();
-      return;
-    }
-
-    onSelect.call(this, model);
+    this.triggerMethod('select', model);
 
     this.destroy();
   },
--- a/src/js/views/patients/patient/sidebar/sidebar_views.js
+++ b/src/js/views/patients/patient/sidebar/sidebar_views.js
@@ -1,4 +1,3 @@
-import { bind } from 'underscore';
 import Radio from 'backbone.radio';
 import Backbone from 'backbone';
 import { View } from 'marionette';
@@ -69,23 +68,21 @@ const SidebarView = View.extend({
 
     const menuOptions = new Backbone.Collection([
       {
+        event: canEditPatient ? 'click:patientEdit' : 'click:patientView',
         text: canEditPatient ? i18n.menuOptions.edit : i18n.menuOptions.view,
-        onSelect: bind(this.triggerMethod, this, canEditPatient ? 'click:patientEdit' : 'click:patientView'),
       },
     ]);
 
     if (workspacePatient.canEdit()) {
       menuOptions.push({
+        event: 'click:activeStatus',
         text: workspacePatientStatus !== PATIENT_STATUS.ACTIVE ? i18n.menuOptions.activate : i18n.menuOptions.inactivate,
-        onSelect: bind(this.triggerMethod, this, 'click:activeStatus'),
       });
 
       if (workspacePatientStatus !== PATIENT_STATUS.ARCHIVED) {
         menuOptions.push({
+          event: 'click:shouldArchive',
           text: i18n.menuOptions.archive,
-          onSelect: () => {
-            this.showConfirmArchiveModal();
-          },
         });
       }
     }
@@ -100,9 +97,15 @@ const SidebarView = View.extend({
       popWidth: 248,
     });
 
+    this.listenTo(optionlist, 'select', ({ model }) => {
+      const event = model.get('event');
+
+      this.triggerMethod(event);
+    });
+
     optionlist.show();
   },
-  showConfirmArchiveModal() {
+  onClickShouldArchive() {
     const modal = Radio.request('modal', 'show:small', {
       bodyText: i18n.archiveModal.bodyText,
       headingText: i18n.archiveModal.headingText,
--- a/src/js/views/patients/patient/sidebar/sidebar_views.js
+++ b/src/js/views/patients/patient/sidebar/sidebar_views.js
@@ -97,7 +97,7 @@ const SidebarView = View.extend({
       popWidth: 248,
     });
 
-    this.listenTo(optionlist, 'select', ({ model }) => {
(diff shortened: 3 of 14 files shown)


src/js/views/patients/patient/sidebar/sidebar_views.js
```diff
@@ -100,6 +97,14 @@ const SidebarView = View.extend({
       popWidth: 248,
     });
 
+    this.listenTo(optionlist, 'select', ({ model }) => {
+      const optionItem = model.get('key');
```
paulfalgout: I think for this one rather than key you just put event and make it
```
event: canEditPatient ? 'click:patientEdit' : 'click:patientView',
...
event: 'click:activeStatus',
...
event: 'click:shouldArchive'
```
then make an `onShouldArchive` that either contains or calls `showConfirmArchiveModal`

paulfalgout (changes requested):
One small change.. ~~then I think we should remove the onSelect functionality from the optionlist component~~ I see this in the checklist now

src/js/views/patients/patient/sidebar/sidebar_views.js
nmajor25: That's a good way to do it 👍

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -146,16 +146,22 @@ export default SubRouterApp.extend({
...
       headingText: i18n.addActionHeadingText,
       lists: [{ collection: new Backbone.Collection(this.addOpts) }],
-    }));
+    });
+
+    this.listenTo(addButtonView, 'add:programAction', programItem => {
+      this.triggerMethod('add:programAction', programItem);
+    });
+
+    this.showChildView('tools', addButtonView);
```
nmajor25, line 164: This works when I test it myself in my local dev environment.

But it's failing in the cypress test. No idea why yet.

nmajor25: The `'select'` listener in the `add-workflow_views.js` isn't even be called when an option list item is clicked on in the test.

nmajor25: But if I pause the test, close the option list myself manually, and then click on an option list item. It works.

test/integration/patients/patient/patient-flow.js
```diff
@@ -1085,6 +1085,15 @@ context('patient flow page', function() {
...
       .should('not.contain', 'Draft');
 
+    cy
+      .get('body')
+      .type('{esc}');
+
+    cy
+      .get('.patient-flow__actions')
+      .contains('Add')
+      .click();
```
nmajor25: `this.listenTo(optionlist, 'select'...` doesn't fire in this test when the optionlist is rendered the first time.

Closing the optionlist and then re-opening it makes the listener work. I'm not sure why.

shadowhand: At least for standard `<select>` elements, the `onselect` only fires when the option is changed and does not fire onload.

src/js/components/optionlist/index.js
```diff
@@ -48,15 +48,6 @@ export default Picklist.extend({
   onSelect({ model }) {
     if (model.get('isDisabled')) return;
 
-    const onSelect = model.get('onSelect');
```
paulfalgout, line 51: I'm not sure this is working as expected.  if we're adding a different select trigger then `isDisabled` isn't doing anything unless we handle it in each event handler.  what we might need to do is trigger an internal event that then does:

```js
if (model.get('isDisabled')) return;
this.triggerMethod('select', model);
this.destroy();
```

src/js/apps/patients/patient/flow/flow_app.js
paulfalgout: need to figure this out before we merge this

paulfalgout (changes requested):

src/js/components/optionlist/index.js
nmajor25: What do you mean by triggering an internal event? 

Is the idea to remove this `onSelect()` all together? Because ya we need to do `this.destroy()` to close the picklist. Which I guess we could do in each `this.listenTo(optionlist...`.

src/js/apps/patients/patient/flow/flow_app.js
nmajor25: Still haven't figured out why that test isn't working. But I agree, its a sign that something is off with this.

src/js/components/optionlist/index.js
paulfalgout: Like
```js
viewEvents: {
    'picklist:item:select': 'onPicklistSelect',
},
onPicklistSelect({ model }) {
  if (model.get('isDisabled')) return;
  this.triggerMethod('select', model);
  this.destroy();
},
```

nmajor25: The difference being the use of `viewEvents` instead of `viewTriggers` that's currently being used?

paulfalgout: so the main change is that viewTriggers mean that `onSelect` and the `this.listenTo(myOptionList, 'select'` are the same event handler, so they would _both_ _alway_ need `if (model.get('isDisabled')) return;` at the top.  using viewEvents and triggering `select` means that the trigger only happens if the model isn't disabled and that check doesn't need to be added to every `this.listenTo(myOptionList` handler.

nmajor25: Okay gotcha, I just updated it to that. Seems to work.

Still need to figure out that test issue.

src/js/apps/patients/patient/flow/flow_app.js
nmajor25, line 177: The test issue has something to do with `showAdd()` being called in `toggleBulkSelect`. Because if I remove this line and do `this.showAdd()` up in the `onStart()`, the test passes. (but that breaks other things).

When the patient flow page is loaded, `toggleBulkSelect()` is called twice. Once via `this.toggleBulkSelect()` and a second time via `this.listenTo(this.editableCollection, 'reset', this.toggleBulkSelect);` when the listview triggers a `change:canEdit`.

This doesn't cause issues in the actual app. But for some reason this is goofing up the test. The `listenTo(optionlist, 'select'...` doesn't trigger after `this.showAdd()` is called a second time.

Still not sure how to resolve this. But I think this is the cause of the issue.

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -77,7 +77,6 @@ export default SubRouterApp.extend({
     }));
 
     this.listenTo(this.editableCollection, 'reset', this.toggleBulkSelect);
-    this.toggleBulkSelect();
```
nmajor25, line 80: These are both being called `onStart`, which is breaking the test. 

I'm not sure if this is the correct way to solve this. But we should make sure it's only called once on start somehow.

src/js/apps/patients/patient/flow/flow_app.js
nmajor25: Note: adding a `cy.wait(200)` to the beginning of the test makes the test pass.

src/js/apps/patients/patient/flow/flow_app.js
nmajor25: I don't think we want to do this. It doesn't break functionality. But it makes the elements load in a different order than they currently are.

nmajor25: I can think of two options:

1. Jerry-rig the test somehow to wait until after the list is rendered and the `this.listenTo(this.editableCollection, 'reset', this.toggleBulkSelect);` finishes.
2. Add logic to the `showAdd()` that bails out if the button is already being shown. Like:

```js
  showAdd() {
    // NEW LINES ADDED
    const currentView = this.getChildView('tools');
    if (currentView instanceof AddButtonView) {
      return;
    }

    const addButtonView = new AddButtonView({
      headingText: i18n.addActionHeadingText,
      lists: [{ collection: new Backbone.Collection(this.addOpts) }],
    });

    this.listenTo(addButtonView, 'add:programAction', programItem => {
      this.triggerMethod('add:programAction', programItem);
    });

    this.showChildView('tools', addButtonView);
  },
```

src/js/apps/patients/patient/flow/flow_app.js
paulfalgout: why does change:canEdit need to be triggered on the initial render?

src/js/apps/patients/patient/flow/flow_app.js
paulfalgout: I think maybe the solution is `if (!this.editableCollection.length) this.toggleBulkSelect();` @nmajor25

nmajor25: One thing with the `if (!this.editableCollection.length) this.toggleBulkSelect();` solution is that it will change how the elements load, causing a layout jump:

https://github.com/user-attachments/assets/a977caa8-d976-4ec0-87a1-3805ba8f7c38

paulfalgout (approved):

PR #1535, opened 2025-10-09
Update dependencies

nmajor25:
Shortcut Story ID: [sc-64301]

## Summary by CodeRabbit

- Chores
  - Updated multiple dependencies and tooling for compatibility and stability.
  - Added libraries to support offline caching/asset strategies and iconography.
  - Updated an authentication-related library.
  - Minor stylesheet tweak improving text-wrapping for more consistent rendering.
- New Features / Changes
  - Replaced front-end recording and in-browser profiler assets with updated implementations.
  - Updated the main front-end bundle reference.
- Notes
  - No breaking public API signature changes.


package-lock.json
nmajor25, line 1: `npm update` causing a 9k line deletion in the `package-lock.json`?

paulfalgout: OK well red flag.  tests passed, but I bet this has something to do with:
https://github.com/RoundingWell/care-ops-frontend/pull/1527/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R82

the service worker libraries aren't in dependencies.. nor is the fortawesome one..  I think we need to add these.  We were somehow getting them free with something else that is no longer providing them.

Not sure if that's it.. there are legitimate reasons that we could lose a bunch of sub dependencies, and that would be great. but I'm highly skeptical

paulfalgout (approved):

PR #1536, opened 2025-10-14
Use the callData.interactionId as the identifier

paulfalgout:
callData is available and therefore so is interactionId. The interactionId is created with `callStarted` so it is the most reasonable id to use anyhow, but is not by default available in the callLogData

Shortcut Story ID: [sc-64429]

## Summary by CodeRabbit

* **Improvements**
  * Call completion now reliably ties each call to its interaction identifier, improving association and post-call processing.
  * Faster, more consistent syncing of call outcomes with clearer payloads for backend updates.

* **Bug Fixes**
  * Fixed cases where session identifiers were omitted, preventing mismatched or missing call logs after calls end.
  * Reduced errors and improved stability when finalizing calls.

### DIFF
--- a/packages/care-ops-five9/five9_app.js
+++ b/packages/care-ops-five9/five9_app.js
@@ -54,9 +54,9 @@ export default App.extend({
       callEnded: () => {
         this.setState('callTime', null);
       },
-      callFinished: ({ callLogData }) => {
+      callFinished: ({ callLogData, callData }) => {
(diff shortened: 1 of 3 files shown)


paulfalgout:
@nmajor25 whelp the linter error is an issue with the editorconfig library.. there's a "fixed" issue on their repo.. but clearly not fixed..

nmajor25 (approved):

PR #1543, opened 2025-10-22
Fix bug: bulk edit states button label is blank when all selected states are the same

nmajor25:
Shortcut Story ID: [sc-64560]

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of patient bulk-edit when state information is missing or structured differently, reducing errors when changing states in bulk.

* **Tests**
  * Strengthened end-to-end checks for bulk-edit flows to ensure state, owner, due date/time, and duration display expected values during the workflow.

### DIFF
--- a/test/integration/patients/worklist/bulk-edit.js
+++ b/test/integration/patients/worklist/bulk-edit.js
@@ -519,6 +519,16 @@ context('Worklist bulk editing', function() {
     cy
       .get('@bulkEditSidebar')
       .find('[data-state-region]')
+      .contains('Done');
+
+    cy
+      .get('@bulkEditSidebar')
+      .find('[data-owner-region]')
+      .contains('Nurse');
+
+    cy
+      .get('@bulkEditSidebar')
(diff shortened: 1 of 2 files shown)


paulfalgout (changes requested):
Feels like there should be a test for this or no?

nmajor25:
Wasn't sure about testing these. Only the `Multiple Owners/States/Dates/etc.` versions of the buttons are tested. Not the versions where the individual state, owner, date, is shown. 

So we should probably test all of the other ones too. Not just the state button.

Which could be a pain... but I guess we could just add an extra action/flow to the list in the test, select just that one, and test the buttons.

test/integration/patients/worklist/bulk-edit.js
nmajor25, line 1: Easier than I originally thought. To verify the button labels are shown correctly when all states, owners, due dates, etc. are the same.

paulfalgout (approved):

PR #1546, opened 2025-10-27
On bulk editing flows, batch `PATCH /flows/:id/relationships/actions` requests

nmajor25:
Shortcut Story ID: [sc-64748]

Two changes made:

* Batch `PATCH /flows/:id/relationships/actions` requests when flows are being bulk edited. (`Apply Owner to Flow Actions` option selected)
* Create a new `batchInvoke()` method in `/js/base/collection` that we can use to send batch requests.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated internal batch processing utilities for collection operations to improve code maintainability and consistency. Collection save and owner application workflows now utilize centralized batch handling with no user-facing changes.

### DIFF
--- a/src/js/base/collection.js
+++ b/src/js/base/collection.js
@@ -37,4 +37,21 @@ export default Backbone.Collection.extend(extend({
 
     return Promise.all(destroys);
   },
+  async batchInvoke(methodName, batchSize, ...args) {
+    /* istanbul ignore next: Branch only for testing */
+    const size = _TEST_ ? 2 : batchSize;
+    const results = [];
+    let count = 0;
+
+    while (count < this.length) {
+      const batch = new this.constructor(this.slice(count, count + size));
+      const batchInvokes = batch.invoke(methodName, ...args);
+      const batchResults = await Promise.all(batchInvokes);
+
+      results.push(...batchResults);
+      count += size;
+    }
+
+    return results;
+  },
 }, JsonApiMixin));
--- a/src/js/entities-service/entities/actions.js
+++ b/src/js/entities-service/entities/actions.js
@@ -288,26 +288,7 @@ const Collection = BaseCollection.extend({
   url: '/api/actions',
   model: Model,
   save(attrs) {
-    return this._processBatches(attrs);
-  },
-  async _processBatches(
-    attrs,
-    /* istanbul ignore next: Branch only for testing */
-    batchSize = _TEST_ ? 2 : 25,
-  ) {
-    const results = [];
-    let count = 0;
-
-    while (count < this.length) {
-      const batch = new Collection(this.slice(count, count + batchSize));
-      const batchSaves = batch.invoke('saveAll', attrs);
-      const batchResults = await Promise.all(batchSaves);
-
-      results.push(...batchResults);
-      count += batchSize;
-    }
-
-    return results;
+    return this.batchInvoke('saveAll', 25, attrs);
   },
   groupByDate() {
     const groupedCollection = this.groupBy('due_date');
--- a/src/js/entities-service/entities/flows.js
+++ b/src/js/entities-service/entities/flows.js
@@ -131,19 +131,23 @@ const Collection = BaseCollection.extend({
   url: '/api/flows',
   model: Model,
   save(attrs) {
-    return this._processBatches(attrs);
+    return this._processBatches('saveAll', attrs);
+  },
+  applyOwner(owner) {
+    return this._processBatches('applyOwner', owner);
   },
   async _processBatches(
-    attrs,
-    /* istanbul ignore next: Branch only for testing */
-    batchSize = _TEST_ ? 2 : 25,
+    methodName,
+    ...args
   ) {
+    /* istanbul ignore next: Branch only for testing */
+    const batchSize = _TEST_ ? 2 : 20;
     const results = [];
     let count = 0;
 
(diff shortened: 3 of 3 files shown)


src/js/apps/patients/worklist/worklist_app.js
nmajor25, line 1: This sends all the apply owner requests in batches of 25 until they are finished. And then sends the PATCH flow requests in batches.

Opposed to sending all of the two request types together at the same time.

nmajor25: When sending them all together, we won't be able to 100% know when the entire process has finished. 

So if the PATCH flows requests finish first and applyOwner requests are still being sent, the UI loading state will be incorrect. And the bulk edit app will be stopped or restarted, which I assume would cause problems in that scenario?

paulfalgout: why does apply owner affect the loading state or the bulk-edit stuff? @nmajor25

nmajor25: A user selects 500 flows with the apply to owner option selected and we start batching both sets of requests together.

The `PATCH /flows` and `PATCH /api/flows/:id/relationships/actions` are currently sent simultaneously.

We currently track the `PATCH /flows` requests and change the UI when they are all finished. But the `PATCH /api/flows/:id/relationships/actions` requests don't have a promise or anything to track when they are finished.

So the UI loading state could be showing that the process is finished when there is still apply owner requests happening in the background.

If that doesn't matter, we can probably simplify this PR. And just add the batching in the flows entities file.

paulfalgout: I think it's safe to apply owner requests in the background.

paulfalgout: _but_ since both batches will happen simultaneously we should lower the max from 25 to 20

src/js/entities-service/entities/flows.js
```diff
@@ -131,19 +131,23 @@ const Collection = BaseCollection.extend({
...
   },
   async _processBatches(
-    attrs,
-    /* istanbul ignore next: Branch only for testing */
-    batchSize = _TEST_ ? 2 : 25,
+    methodName,
+    ...args
   ) {
+    /* istanbul ignore next: Branch only for testing */
+    const batchSize = _TEST_ ? 2 : 20;
```
nmajor25: This is updated to 20 per batch for flows. But the actions requests are still set to 25. Not sure if we want to update those to be consistent.

src/js/apps/patients/worklist/worklist_app.js
nmajor25: Okay that's been updated 👍

src/js/base/collection.js
```diff
@@ -37,4 +37,21 @@ export default Backbone.Collection.extend(extend({
 
     return Promise.all(destroys);
   },
+  async batchInvoke(methodName, args, options) {
+    /* istanbul ignore next: Branch only for testing */
+    const size = _TEST_ ? 2 : (options?.batchSize || 25);
```
nmajor25: Default set to 25 if no `options.batchSize` is passed.

src/js/entities-service/entities/actions.js
```diff
@@ -288,26 +288,7 @@ const Collection = BaseCollection.extend({
...
-      const batch = new Collection(this.slice(count, count + batchSize));
-      const batchSaves = batch.invoke('saveAll', attrs);
-      const batchResults = await Promise.all(batchSaves);
-
-      results.push(...batchResults);
-      count += batchSize;
-    }
-
-    return results;
+    return this.batchInvoke('saveAll', attrs, { batchSize: 2 });
```
paulfalgout: no options here

src/js/base/collection.js
```diff
@@ -37,4 +37,21 @@ export default Backbone.Collection.extend(extend({
 
     return Promise.all(destroys);
   },
+  async batchInvoke(methodName, batchSize, args) {
```
paulfalgout: I think this needs to be `...args` and `batch.invoke(methodName, ...args)`

src/js/base/collection.js
```diff
@@ -37,4 +37,21 @@ export default Backbone.Collection.extend(extend({
...
+    /* istanbul ignore next: Branch only for testing */
+    const size = _TEST_ ? 2 : batchSize;
+    const results = [];
+    let count = 0;
+
+    while (count < this.length) {
+      const batch = new this.constructor(this.slice(count, count + size));
```
paulfalgout, line 47: ah this line isn't something I considered.. hmm does this work?  I guess it does?

paulfalgout (changes requested):

src/js/base/collection.js
nmajor25: I've tested these manually in my local dev environment, everything looks to be working.

src/js/entities-service/entities/actions.js
nmajor25: This should be updated, think this is an outdated version of the commit.

src/js/base/collection.js
nmajor25: Yep, just updated that 👍

paulfalgout (approved):

PR #1547, opened 2025-10-29
Fix coverage branch miss when destroying modal animation

nmajor25:
Shortcut Story ID: [sc-64824]

To fix a coveralls branch miss: https://coveralls.io/builds/76185913/source?filename=src%2Fjs%2Fviews%2Fglobals%2Fmodal%2Fmodal_views.js#L47

### DIFF
--- a/src/js/views/globals/modal/modal_views.js
+++ b/src/js/views/globals/modal/modal_views.js
@@ -44,6 +44,7 @@ const SavingFooterView = View.extend({
     });
   },
   onBeforeDestroy() {
+    /* istanbul ignore else */
     if (this.animation) {
       this.animation.cancel();
     }


src/js/views/globals/modal/modal_views.js
```diff
@@ -44,6 +44,7 @@ const SavingFooterView = View.extend({
     });
   },
   onBeforeDestroy() {
+    /* istanbul ignore if */
```
nmajor25: I'm pretty sure we want to keep the protection on this. In scenarios where `this.animation` is `undefined`. When the modal view is destroyed before `onRender()` happens.

nmajor25: But not sure if we can test that effectively. So just adding a `istanbul ignore` to fix the coverage miss.

paulfalgout (approved):

PR #1548, opened 2025-10-29
Fix coverage miss for tooltip delay branches

nmajor25:
Shortcut Story ID: [sc-64824]

Fix coverage miss here: https://coveralls.io/builds/76111569/source?filename=src%2Fjs%2Fcomponents%2Ftooltip%2Findex.js#L36

### DIFF
--- a/src/js/components/tooltip/index.js
+++ b/src/js/components/tooltip/index.js
@@ -32,8 +32,7 @@ const TooltipView = View.extend({
 export default Component.extend({
   ViewClass: TooltipView,
   className: 'tooltip',
-  /* istanbul ignore next: Branch only for testing */
-  delay: _TEST_ ? 0 : 200,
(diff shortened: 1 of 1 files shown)


src/js/components/tooltip/index.js
```diff
@@ -32,8 +32,7 @@ const TooltipView = View.extend({
 export default Component.extend({
   ViewClass: TooltipView,
   className: 'tooltip',
-  /* istanbul ignore next: Branch only for testing */
-  delay: _TEST_ ? 0 : 200,
```
nmajor25, line 36: This `istanbul ignore next` isn't working here for some reason.

I think it's because it was being applied to the `delay` property key instead of `_TEST_ ? 0 : 200`

nmajor25: Looked like that is working now

paulfalgout (approved):

PR #1549, opened 2025-10-29
Remove programItem existance check in add workflow optionlist select

nmajor25:
Shortcut Story ID: [sc-64824]

To fix code coverage miss: https://coveralls.io/builds/76207892/source?filename=src%2Fjs%2Fviews%2Fpatients%2Fshared%2Fadd-workflow%2Fadd-workflow_views.js#L51

### DIFF
--- a/src/js/views/patients/shared/add-workflow/add-workflow_views.js
+++ b/src/js/views/patients/shared/add-workflow/add-workflow_views.js
@@ -48,8 +48,6 @@ const AddButtonView = View.extend({
     this.listenTo(optionlist, 'select', model => {
       const programItem = model.get('programItem');
 
-      if (!programItem) return;
-
       const isProgramAction = model.get('itemType') === 'program-actions';
       const messageType = isProgramAction ? 'add:programAction' : 'add:programFlow';
 


src/js/views/patients/shared/add-workflow/add-workflow_views.js
```diff
@@ -48,8 +48,6 @@ const AddButtonView = View.extend({
     this.listenTo(optionlist, 'select', model => {
       const programItem = model.get('programItem');
 
-      if (!programItem) return;
```
nmajor25, line 51: Don't think there is a scenario where this 'select' event occurs with an `undefined`/`null` `programItem`.

This is used in two places: pt dashboard & pt flow pages in the add workflow options list.

* Pt dasbhoard: when a program has no addable options, it creates `noResultsOpt` with `isDisabled: true`. So this onSelect isn't called.
* Pt flow: all items include `programItem: action`. If no addable actions exist, the collection is empty and there's no items to select.

paulfalgout (approved):
if it does it should be an easy RUM to track down

PR #1553, opened 2025-11-12
Silence some of the browser console logging done by five9

nmajor25:
Shortcut Story ID: [sc-64900]

This stops console.logs from happening from the `packages/care-ops-five9/sdk/five9.crm.sdk.js` file.

Note: there is other code that five9 loads in the iframe that logs a lot to the console. But that is cross origin content, so we can't stop it from logging to the console.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced excessive console logging from the Five9 SDK integration. Informational and debug messages are now suppressed, while critical warnings and errors remain visible for troubleshooting purposes. No changes to SDK functionality or API behavior.

### DIFF
--- a/packages/care-ops-five9/sdk/five9.crm.sdk.js
+++ b/packages/care-ops-five9/sdk/five9.crm.sdk.js
@@ -19,6 +19,16 @@ window.crmSdkVersion = '1.0.11';
     root.Five9.CrmSdk = factory();
   }
 }(globalThis, function () {
+  // Patched by care-ops: silence noisy Five9 logging
+  var __five9Console = globalThis.console || {};
+  var console = {
+    log: function () {},
+    debug: function () {},
+    info: function () {},
+    warn: __five9Console.warn ? __five9Console.warn.bind(__five9Console) : function () {},
+    error: __five9Console.error ? __five9Console.error.bind(__five9Console) : function () {},
+  };
+
 
 // We use RequireJS modules inside.
 // Since this library can be included into projects with JS/node modules support
--- a/packages/care-ops-five9/update-sdk.js
+++ b/packages/care-ops-five9/update-sdk.js
@@ -64,6 +64,31 @@ async function downloadSDK() {
       '}(globalThis, function',
     );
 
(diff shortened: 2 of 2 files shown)


packages/care-ops-five9/sdk/five9.crm.sdk.js
```diff
@@ -19,6 +19,16 @@ window.crmSdkVersion = '1.0.11';
...
 }(globalThis, function () {
+  // Patched by care-ops: silence noisy Five9 logging
+  var __five9Console = globalThis.console || {};
+  var console = {
+    log: function () {},
+    debug: function () {},
+    info: function () {},
+    warn: __five9Console.warn ? __five9Console.warn.bind(__five9Console) : function () {},
+    error: __five9Console.error ? __five9Console.error.bind(__five9Console) : function () {},
+  };
```
nmajor25, line 30: This will stop the packages/care-ops-five9/sdk/five9.crm.sdk.js file from doing any logging. I confirmed that works.

nmajor25: But there are other files that five9 runs that we can’t control. Most of those are logged by the https://app.five9.com/five9_clients/int.v13.0.464/app/all.js file. These are cross-origin, so there's no way to access or modify the JavaScript running inside that iframe for our app.

And I don’t think there’s a way to make them same-origin unless we host them ourselves somehow to make them same-origin. And then still figure out how to inject code to stop the logging in those files too.

Which seems like quite the effort. And against five9’s terms of service or something?

nmajor25: To me, it seems worth making the change in this PR though. It at least stops some of the logging. But most of them are coming from that `all.js` file.

paulfalgout: in the console you can filter the logs to just the window and not the iframe, so it's really about clearing the logs from our context that matters 👍

paulfalgout (approved):

PR #1555, opened 2025-11-19
Display program/patient action name icons in UI

nmajor25:
Shortcut Story ID: [sc-65172]

To be displayed inline with action names in the follow places where action data is displayed:

1. Worklists
2. Schedule/Reduced Schedule
3. Patient flow
4. Patient dashboard/archive
5. Action sidebar
6. Program workflows
7. Program flow

Example:

Screenshot 2025-11-19 at 11 11 59 AM

Corresponding BE PRs: https://github.com/RoundingWell/care-ops-backend/pull/10095, https://github.com/RoundingWell/care-ops-backend/pull/10107

## Summary by CodeRabbit

* **New Features**
  * Added support for displaying colored icons alongside action names throughout the application. Icons can now appear in patient views, programs, schedules, and worklists with available color options: black, grey, red, green, orange, and purple.

### DIFF
--- a/src/js/views/patients/patient/archive/action-item.hbs
+++ b/src/js/views/patients/patient/archive/action-item.hbs
@@ -1,5 +1,10 @@
 <div class="table-list__icon--large patient__action-icon">{{far icon}}</div>
-<div class="u-text--overflow-two-lines">{{ name }}</div>
+<div class="u-text--overflow-two-lines">
+  {{#if options.icon}}
+    <span class="action-icon--{{options.color}}">{{fa options.iconType options.icon}}</span>
+  {{/if}}
+  {{ name }}
+</div>
 <div class="table-list__meta">
   <span class="table-list__icon u-margin--r-8" data-details-region></span>
   <span data-state-region></span>
--- a/src/js/views/patients/patient/archive/archive_views.js
+++ b/src/js/views/patients/patient/archive/archive_views.js
@@ -16,6 +16,7 @@ import { ReadOnlyStateView, ReadOnlyOwnerView, ReadOnlyDueDateView, ReadOnlyDueT
 import ActionItemTemplate from './action-item.hbs';
 import FlowItemTemplate from './flow-item.hbs';
 
+import 'scss/domain/action-icons.scss';
 import '../patient.scss';
 
 const EmptyView = View.extend({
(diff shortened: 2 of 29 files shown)


src/js/entities-service/entities/actions.js
```diff
@@ -168,6 +168,13 @@ const _Model = BaseModel.extend({
   hasAttachments() {
     return !!this.getFiles().length;
   },
+  getNameIconOptions() {
+    const options = this.get('options');
+
+    if (!options.icon || !options.iconType || !options.color) return null;
```
nmajor25: I wasn't sure on this. Do we want all of these to be required?

I feel like `icon` and `iconType` should definitely be required. But `color` could be optional. Which would just default to black.

nmajor25: And there will be a coverage miss on this. Since `!options.iconType` and `!options.color` are not being hit in the tests.

I can update those once we decide on what we want to be required.

src/js/entities-service/entities/program-actions.js
```diff
@@ -64,6 +64,15 @@ const _Model = BaseModel.extend({
   disableAttachmentUploads() {
     this.save({ allowed_uploads: [] });
   },
+  getNameIconOptions() {
+    if (this.isNew()) return null;
```
nmajor25: Without this, the `Add Workflow`/`Add Action` buttons on the program pages would fail and show an action not found error. Not sure why.

paulfalgout (changes requested):
I think for now we should follow the state options pattern where color is a class like `action--${ options.color }` rather than modifying the fontawesome.  It also doesn't account for incorrect or missing values.  You may not need a `getNameIconOptions` either.. maybe just `{{#if options.icon}}`?

nmajor25:
@paulfalgout updates are made in new commits 👍

src/js/views/patients/patient/archive/action-item.hbs
```diff
@@ -1,5 +1,9 @@
 <div class="table-list__icon--large patient__action-icon">{{far icon}}</div>
-<div class="u-text--overflow-two-lines">{{ name }}</div>
+<div class="u-text--overflow-two-lines">
+  {{#if options.icon}}
```
paulfalgout, line 3: The {{/if}} needs to be outside of the span

paulfalgout (changes requested):

src/js/views/patients/patient/archive/action-item.hbs
nmajor25: Nice catch, this has been fixed 👍

paulfalgout (approved):

PR #1563, opened 2025-12-04
For workos auth, handle redirecting after `/logout`

nmajor25:
Shortcut Story ID: [sc-65433]

Instead of passing a URL to the `workos` sdk and having them do the redirect.

To fix a bug with the current implementation. Where the `/logout` page is being loaded and doesn't redirect, showing a blank screen to the user.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logout flow and navigation handling to ensure proper redirection after sign-out.

### DIFF
--- a/packages/care-ops-auth/providers/workos.js
+++ b/packages/care-ops-auth/providers/workos.js
@@ -74,7 +74,8 @@ export class WorkosAuthProvider extends AuthProvider {
 
     if (pathName === AuthProvider.PATH_LOGOUT) {
       this.token = null;
-      this.client.signOut({ returnTo: location.origin });
+      await this.client.signOut({ navigate: false });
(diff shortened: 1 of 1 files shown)


packages/care-ops-auth/providers/workos.js
```diff
@@ -74,7 +74,8 @@ export class WorkosAuthProvider extends AuthProvider {
 
     if (pathName === AuthProvider.PATH_LOGOUT) {
       this.token = null;
-      this.client.signOut({ returnTo: location.origin });
+      await this.client.signOut({ navigate: false });
+      window.location = AuthProvider.PATH_LOGIN;
```
nmajor25, line 78: Used `async/await` instead of `.then()`. To match patterns used in the rest of the file.

paulfalgout (approved):

PR #1565, opened 2025-12-08
For workos, wrap sign out in `try/finally` & use `window.location.replace()`

nmajor25:
Shortcut Story ID: [sc-65433]

Additional follow up fix for: #1563.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logout flow: sign-out and then force a redirect to the login page using a navigation method that clears session state and prevents returning to previous pages via the browser back button, yielding cleaner browser history handling.

### DIFF
--- a/packages/care-ops-auth/providers/workos.js
+++ b/packages/care-ops-auth/providers/workos.js
@@ -74,8 +74,13 @@ export class WorkosAuthProvider extends AuthProvider {
 
     if (pathName === AuthProvider.PATH_LOGOUT) {
       this.token = null;
-      await this.client.signOut({ navigate: false });
-      window.location = AuthProvider.PATH_LOGIN;
+
+      try {
+        await this.client.signOut({ returnTo: location.origin });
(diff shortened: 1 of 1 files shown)


packages/care-ops-auth/providers/workos.js
```diff
@@ -74,8 +74,13 @@ export class WorkosAuthProvider extends AuthProvider {
 
     if (pathName === AuthProvider.PATH_LOGOUT) {
       this.token = null;
-      await this.client.signOut({ navigate: false });
-      window.location = AuthProvider.PATH_LOGIN;
+
+      try {
+        this.client.signOut({ returnTo: location.origin });
```
shadowhand: Lost the `await` here... does that matter?

nmajor25: Yes it does, thanks 👍

paulfalgout (approved):
maybe?

PR #1566, opened 2025-12-08
Revert change for redirecting after workos logout

nmajor25:
Shortcut Story ID: [sc-65433]

This reverts commit 717fc8f098cb9a0e00fa7ca133aa0393a7b12240

### DIFF
--- a/packages/care-ops-auth/providers/workos.js
+++ b/packages/care-ops-auth/providers/workos.js
@@ -74,8 +74,7 @@ export class WorkosAuthProvider extends AuthProvider {
 
     if (pathName === AuthProvider.PATH_LOGOUT) {
       this.token = null;
-      await this.client.signOut({ navigate: false });
-      window.location = AuthProvider.PATH_LOGIN;
+      this.client.signOut({ returnTo: location.origin });
       return;
     }
 


packages/care-ops-auth/providers/workos.js
```diff
@@ -74,8 +74,7 @@ export class WorkosAuthProvider extends AuthProvider {
 
     if (pathName === AuthProvider.PATH_LOGOUT) {
       this.token = null;
-      await this.client.signOut({ navigate: false });
-      window.location = AuthProvider.PATH_LOGIN;
+      this.client.signOut({ returnTo: location.origin });
```
shadowhand, line 77: Does this need `await`?

paulfalgout: nope..  unless navigate: false is passed, no Promise is returned

paulfalgout (approved):

paulfalgout (changes requested):
need to full revert

packages/care-ops-auth/providers/workos.js
```diff
@@ -74,8 +74,10 @@ export class WorkosAuthProvider extends AuthProvider {
 
     if (pathName === AuthProvider.PATH_LOGOUT) {
       this.token = null;
```
shadowhand, line 76: @paulfalgout this is what I'm saying though... the token gets cleared right here. So why can't we have an `if` condition to handle the "oops, already logged out" state:


```suggestion
      if (! this.token) {
          window.location.replace(AuthProvider.PATH_LOGIN);
          return;
      }
      this.token = null;
```

shadowhand: Assuming the user is already logged out, why wouldn't this solve the original problem that was reported?

nmajor25:
@paulfalgout I cleared the setTimeout commit from this PR. So now it is only the revert.

packages/care-ops-auth/providers/workos.js
paulfalgout: It's possible for the token to be null here and still be logged in, and if you're actually logged in and we run window.locaton it just auto logs-in.

paulfalgout (approved):
I think we just revert this and hope workos makes the easy fix to their library

nmajor25:
@paulfalgout we are good to merge this now, right?

PR #1567, opened 2025-12-09
Fix bug where add workflow droplist input filter is not working

nmajor25:
Shortcut Story ID: [sc-65497]

Caused by [this change](https://github.com/RoundingWell/care-ops-frontend/commit/90a0ea93201e411ecbb0eff53f506afa8c27f138#diff-fab3cde5629392f2a3d9cf9dbe2c6a74f43e38465db2cc18b29a6e44ac2b84fbR34-R36).

Setting a `viewEvents` on the `OptionList` overrides the `viewEvents` on the `Picklist` component. So we lost the `'watch:change': 'onWatchChange',` listener.

### DIFF
--- a/src/js/components/optionlist/index.js
+++ b/src/js/components/optionlist/index.js
@@ -32,6 +32,7 @@ export default Picklist.extend({
     'close': 'close',
   },
   viewEvents: {
+    'watch:change': 'onWatchChange',
     'picklist:item:select': 'onPicklistSelect',
   },
   position() {
--- a/src/js/components/optionlist/optionlist.cy.js
+++ b/src/js/components/optionlist/optionlist.cy.js
@@ -20,6 +20,7 @@ context('Optionlist', function() {
         ui: this.ui.button,
         uiView: this,
         lists: [{ collection: this.collection }],
+        isSelectlist: true,
       });
 
       optionlist.show();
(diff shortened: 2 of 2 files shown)


src/js/components/optionlist/index.js
```diff
@@ -32,6 +32,7 @@ export default Picklist.extend({
     'close': 'close',
   },
   viewEvents: {
+    'watch:change': 'onWatchChange',
```
paulfalgout, line 35: can we add a test in the optionlist.cy.js for this?

paulfalgout (changes requested):

src/js/components/optionlist/index.js
nmajor25: I thought about it originally, wasn't sure. The filter input functionality is done by the picklist, not optionlist. So I decided to just add it to the integration test where the optionlist is used.

But it's probably worth adding it to the optionlist component test, as another layer of defense from `'watch:change': 'onWatchChange',` being removed in the future.

nmajor25: Okay, that's been updated. I added it to the component test and removed it from the integration test.

src/js/components/optionlist/optionlist.cy.js
```diff
@@ -88,5 +89,20 @@ context('Optionlist', function() {
...
+
+    cy
+      .get('.picklist')
+      .find('.js-input')
+      .type('Test 1');
+
+    cy
+      .get('.picklist')
+      .find('.picklist__item')
+      .should('have.length', 1);
```
nmajor25, line 106: If `'watch:change': 'onWatchChange',` is removed from the optionlist component, this fails.

paulfalgout (approved):

PR #1568, opened 2025-12-09
Optionally disable `pointerdown` tooltip event listener

nmajor25:
Shortcut Story ID: [sc-65463]

This PR adds a `shouldShowOnClick` option that can be passed to the tooltip component. Set to `true` by default.

### DIFF
--- a/src/js/views/forms/form/form_views.js
+++ b/src/js/views/forms/form/form_views.js
@@ -105,6 +105,7 @@ const FormStateActionsView = View.extend({
       message,
       uiView: this,
       ui: this.ui.sidebarButton,
+      shouldShowOnClick: false,
     });
   },
   renderExpandTooltip() {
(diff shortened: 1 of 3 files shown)


paulfalgout (changes requested):
I think we should modify the tooltip component with some sort of `shouldShowOnHover` / `shouldShowOnClick` which is true by default or something

nmajor25:
@paulfalgout yep that is the way to go. I updated to that and added a test in `tooltip.cy.js` as well.

src/js/views/forms/form/form_views.js
```diff
@@ -105,6 +105,7 @@ const FormStateActionsView = View.extend({
       message,
       uiView: this,
       ui: this.ui.sidebarButton,
+      shouldShowOnClick: false,
```
nmajor25, line 108: This does mean these won't be shown on touch devices at all (*I think*).

paulfalgout (approved):

PR #1570, opened 2025-12-15
Disable `pointerdown` show event on additional tooltips

nmajor25:
Shortcut Story ID: [sc-65591]

Additional follow up work from https://github.com/RoundingWell/care-ops-frontend/pull/1568/.

Screen recordings of the current tooltip behaviors on the shortcut story.

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted tooltip interaction behavior in date filters and activity comment sections—tooltips will no longer respond to clicks, improving interaction consistency.

### DIFF
--- a/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
+++ b/src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
@@ -157,6 +157,7 @@ const CommentView = View.extend({
       messageHtml: renderTemplate(template, { edited }),
       uiView: this,
       ui: this.ui.edit,
+      shouldShowOnClick: false,
     });
   },
   onClickEdit() {
--- a/src/js/views/patients/shared/components/date-filter/date-filter_views.js
+++ b/src/js/views/patients/shared/components/date-filter/date-filter_views.js
@@ -118,12 +118,14 @@ const ControllerView = View.extend({
       message: tooltipMessages.prevMessage,
       uiView: this,
       ui: this.ui.prev,
+      shouldShowOnClick: false,
     });
 
     new Tooltip({
       message: tooltipMessages.nextMessage,
       uiView: this,
       ui: this.ui.next,
+      shouldShowOnClick: false,
     });
   },
   getTooltipMessages() {


src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
```diff
@@ -157,6 +157,7 @@ const CommentView = View.extend({
       messageHtml: renderTemplate(template, { edited }),
       uiView: this,
       ui: this.ui.edit,
+      shouldShowOnClick: false,
```
paulfalgout, line 160: OK now I'm concerned with this change entirely.
We haven't always had tooltip issues across the app.  So what changed?

But if we remove this, then on a touch device we can't see the edited time?

paulfalgout (changes requested):

src/js/views/patients/sidebar/action/action-sidebar-activity-views.js
nmajor25: My guess is that it was happening, but didn't become noticeable until we added the delay: https://github.com/RoundingWell/care-ops-frontend/pull/1519

nmajor25: Do we still need to consider touch devices?

The `pointerenter` and `pointerdown` listeners were originally added for Ipad users: https://github.com/RoundingWell/care-ops-frontend/pull/648.

paulfalgout: Maybe not, but we might.   It just seems like we're regressing.  Is the solution to change the delay to 0 in these cases?

nmajor25: In which cases? I'm not sure how to determine when the delay should be `0` or `200`.

At this point, my opinion is to revert all these tooltip changes. Get rid of the delay we added in #1519. Which is what's causing these issues. And remove the `shouldShowOnClick` stuff.

The original reason the delay was added is outlined in [this story](https://app.shortcut.com/roundingwell/story/63848/add-a-200ms-delay-to-tooltip-showing-on-hover). It doesn't seem worth the trouble it's causing.

paulfalgout: it may be that we should revert back and just add the delay on row items?

I'm not sure really.  My spidey sense just thinks we're moving the problem around..

But I dunno.. maybe worth trying a few options.. maybe getting Mark to weigh in.. I don't want to add a feature to a generic unless we're really clear on the problem/pattern, and I'm less confident that I was...

nmajor25: Okay, I agree. I created a new shortcut story: https://app.shortcut.com/roundingwell/story/65602/tooltip-200ms-show-delay-causing-ui-bugs

nmajor25:
Closing this pull request. We're not moving forward with this and reverting related changes: https://app.shortcut.com/roundingwell/story/65602/tooltip-200ms-show-delay-causing-ui-bugs

PR #1574, opened 2025-12-18
On five9 inbound/transfer calls, ability to jump to patient dashboard

nmajor25:
Shortcut Story ID: [sc-65412]

## Summary by CodeRabbit

* **New Features**
  * Patient "Jump to" buttons added to the call panel (hidden for current patient).
  * Call flow now resolves agent numbers and surfaces matching patients automatically.
  * Init accepts a patient list to drive in-panel patient controls.

* **Bug Fixes / Behavior**
  * Call panel positioning, sizing, visibility, and open/close behavior refined; iframe visibility/pointer handling improved.
  * Patient buttons clear on call end and restore correctly during navigation/back actions.

* **Tests**
  * Integration tests expanded to cover patient-button behaviors and updated call payload shapes.

### DIFF
--- a/packages/care-ops-five9/five9.scss
+++ b/packages/care-ops-five9/five9.scss
@@ -5,23 +5,25 @@ $five9-call-active-hover: color(green, dark);
 $five9-call-ended: color(red);
 $five9-call-ended-hover: color(red, dark);
 
+.five9-wrapper {
+  bottom: 0;
+  position: fixed;
+  right: 40px;
+  width: 300px;
+}
+
 .five9-panel {
   background: white;
-  border: 1px solid #{$black-90};
   border-radius: 8px 8px 0 0;
-  bottom: 0;
   box-shadow: 0 4px 10px rgb(0 0 0 / 20%);
   display: flex;
   flex-direction: column;
   height: 36px;
   overflow: hidden;
-  position: fixed;
-  right: 40px;
   transition: height 0.2s ease;
-  width: 300px;
 
-  &.is-open {
-    height: 90vh;
+  .five9-wrapper.is-open & {
+    height: 80vh;
   }
 }
 
@@ -62,14 +64,12 @@ $five9-call-ended-hover: color(red, dark);
 }
 
 .five9-panel__iframe {
-  flex: 1 1 auto;
-  height: auto;
   min-height: 0;
   pointer-events: none;
   visibility: hidden;
 
-
-  .five9-panel.is-open & {
+  .five9-wrapper.is-open & {
+    flex: 1 1 auto;
     pointer-events: auto;
     visibility: visible;
   }
--- a/packages/care-ops-five9/five9.scss
+++ b/packages/care-ops-five9/five9.scss
@@ -6,6 +6,7 @@ $five9-call-ended: color(red);
 $five9-call-ended-hover: color(red, dark);
 
 .five9-wrapper {
+  border-radius: 8px 8px 0 0;
   bottom: 0;
   position: fixed;
   right: 40px;
@@ -60,6 +61,35 @@ $five9-call-ended-hover: color(red, dark);
     &:hover {
       background-color: $five9-call-ended-hover;
     }
+  }
+}
+
+.five9-panel__patient-btn {
+  align-items: center;
+  background-color: white;
+  border: 1px solid $five9-color;
+  border-radius: 8px;
+  color: $five9-color;
+  cursor: pointer;
+  display: flex;
+  font-size: 14px;
+  font-weight: 600;
+  justify-content: center;
+  margin-bottom: 8px;
+  padding: 8px 16px;
+  text-overflow: ellipsis;
+  transition: background-color 0.15s;
+  transition-timing-function: linear;
+  white-space: nowrap;
+  width: 100%;
+
+  &:hover {
+    background-color: color(light_blue, light);
+  }
+
+  svg {
+    flex-shrink: 0;
+    margin: 0 6px;
   }
 }
 
--- a/packages/care-ops-five9/five9_app.js
+++ b/packages/care-ops-five9/five9_app.js
@@ -52,7 +52,7 @@ export default App.extend({
 
         if (!actionId) return;
 
-        Radio.request('dialer', 'five9CallComplete', callData.interactionId, { actionId });
+        Radio.request('dialer', 'five9Call', { callData, actionId });
 
         await interaction.setCav({
           interactionId: callData.interactionId,
@@ -66,6 +66,7 @@ export default App.extend({
       },
       callAccepted: ({ callData }) => {
         this.setState('callTime', dayjs());
+
         Radio.request('dialer', 'callNumber', { actionId: this.getState('actionId'), number: callData.number });
       },
       callEnded: () => {
@@ -75,7 +76,7 @@ export default App.extend({
       callFinished: ({ callLogData, callData }) => {
         this.setState('isCalling', false);
         this.setState('actionId', null);
-        Radio.request('dialer', 'five9CallComplete', callData.interactionId, { callLog: callLogData });
+        Radio.request('dialer', 'five9Call', { callData, callLogData });
       },
     });
   },
--- a/packages/care-ops-five9/five9_app.js
+++ b/packages/care-ops-five9/five9_app.js
@@ -64,11 +64,13 @@ export default App.extend({
           ],
         });
       },
-      callAccepted: () => {
+      callAccepted: ({ callData }) => {
         this.setState('callTime', dayjs());
+        Radio.request('dialer', 'callNumber', { actionId: this.getState('actionId'), number: callData.number });
       },
       callEnded: () => {
         this.setState('callTime', null);
+        Radio.request('dialer', 'callNumber', null);
       },
       callFinished: ({ callLogData, callData }) => {
         this.setState('isCalling', false);
--- a/packages/care-ops-five9/five9_app.js
+++ b/packages/care-ops-five9/five9_app.js
@@ -14,14 +14,18 @@ export default App.extend({
   stateEvents: {
     'change:isLoggedIn': 'onLoginChange',
   },
-  initialize({ providerName }) {
+  initialize({ providerName, patients }) {
+    this.patients = patients;
     this.registerApi(providerName);
     this.handleLogin();
     this.subscribe();
   },
   startAfterInitialized: true,
   onStart() {
-    this.showView(new LayoutView({ model: this.getState() }));
+    this.showView(new LayoutView({
+      model: this.getState(),
+      collection: this.patients,
+    }));
   },
   onLoginChange() {
     this._call();
--- a/packages/care-ops-five9/five9_app.js
+++ b/packages/care-ops-five9/five9_app.js
@@ -1,5 +1,7 @@
+import { get } from 'underscore';
 import Radio from 'backbone.radio';
 import dayjs from 'dayjs';
+import fetcher, { handleJSON } from 'js/base/fetch';
 import { applicationApi, crmApi, interactionApi } from './sdk/index';
 
 import App from 'js/base/app';
@@ -64,14 +66,16 @@ export default App.extend({
           ],
         });
       },
-      callAccepted: ({ callData }) => {
+      callAccepted: async({ callData }) => {
         this.setState('callTime', dayjs());
 
-        Radio.request('dialer', 'callNumber', { actionId: this.getState('actionId'), number: callData.number });
+        const number = await this._getCallNumber(callData);
+
+        Radio.request('dialer', 'showPatientLinks', { actionId: this.getState('actionId'), number });
       },
       callEnded: () => {
         this.setState('callTime', null);
-        Radio.request('dialer', 'callNumber', null);
+        Radio.request('dialer', 'showPatientLinks', null);
       },
       callFinished: ({ callLogData, callData }) => {
         this.setState('isCalling', false);
@@ -97,5 +101,25 @@ export default App.extend({
     if (!this.getState('isLoggedIn') || !number) return;
     interaction.click2dial({ click2DialData: { clickToDialNumber: number } });
     this.setState('pendingCall', null);
+  },
+  async _getCallNumber(callData) {
+    const { number, agent } = callData;
+
+    if (number.includes('agent:')) {
+      const { data } = await fetcher('/api/artifacts/search', {
+        data: {
+          filter: {
+            type: 'five9-call-log',
+            path: 'callData.agent',
+            term: agent,
+            limit: 1,
+          },
+        },
+      }).then(handleJSON);
+
+      return get(data, '0.attributes.callData.number');
+    }
+
+    return number;
   },
 });
--- a/packages/care-ops-five9/five9_views.js
+++ b/packages/care-ops-five9/five9_views.js
@@ -1,8 +1,9 @@
 import { delay } from 'underscore';
+import Radio from 'backbone.radio';
 import dayjs from 'dayjs';
 
 import hbs from 'handlebars-inline-precompile';
-import { View } from 'marionette';
+import { View, CollectionView } from 'marionette';
 
 import './five9.scss';
 
@@ -48,10 +49,31 @@ const StatusView = View.extend({
   },
 });
 
+const PatientButtonItemView = View.extend({
+  tagName: 'button',
+  className: 'five9-panel__patient-btn',
+  template: hbs`
+    <span>Jump to</span>
+    {{far "address-card"}}
+    <span class="u-text--overflow">{{ name }}</span>
+  `,
+  triggers: {
+    'click': 'click',
+  },
+  onClick() {
+    Radio.trigger('event-router', 'patient:dashboard', this.model.id);
+  },
+});
+
+const PatientButtonsView = CollectionView.extend({
+  childView: PatientButtonItemView,
+});
+
 // https://app.five9.com/clients/integrations/adt.li.main.html?f9crmapi=true&f9verticalthreshold=300px#login/showLogin
 const LayoutView = View.extend({
   className: 'five9-wrapper',
   template: hbs`
+    <div data-patient-buttons-region></div>
     <div class="five9-panel">
       <div class="five9-panel__header js-header">
         {{fas "phone"}}
@@ -69,6 +91,7 @@ const LayoutView = View.extend({
     </div>
   `,
   regions: {
+    patientButtons: '[data-patient-buttons-region]',
     heading: '[data-heading-region]',
     status: '[data-status-region]',
   },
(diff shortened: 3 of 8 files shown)


src/js/services/dialer.js
```diff
@@ -34,6 +35,44 @@ export default App.extend({
   call(number, action) {
     this._call(number, action);
   },
+  five9CallNumber(values) {
+    if (!values) {
```
nmajor25: If `null` is passed like: `Radio.request('dialer', 'five9CallNumber', null);`, empty the patients collection. Removing any patient dashboard buttons visible in the UI. As it's configured now, that would be on call end.

src/js/services/dialer.js
```diff
@@ -34,6 +35,44 @@ export default App.extend({
...
+
+    const { actionId, number } = values;
+    const currentUrl = window.location.href;
+    const action = Radio.request('entities', 'actions:model', actionId);
+
+    if (action) {
+      const patient = action.getPatient();
+
+      if (patient) {
+        if (!patients.get(patient.id) && !currentUrl.includes(patient.id)) {
```
nmajor25: Don't add the patient button if the patient is already being shown or the user is already on that patient's dashboard (via url containing patient id).

src/js/services/dialer.js
```diff
@@ -34,6 +35,44 @@ export default App.extend({
...
+    searchCollection.fetch({ data: { 'filter[search]': number } }).then(() => {
+      searchCollection.each(patient => {
+        if (!patients.get(patient.id) && !currentUrl.includes(patient.id)) {
+          patients.add({
+            id: patient.id,
+            name: `${ patient.get('first_name') } ${ patient.get('last_name') }`,
+          });
+        }
+      });
+    });
```
nmajor25: Should maybe break this off into a separate function or find a cleaner way to make this api request.

src/js/services/dialer.js
```diff
@@ -1,7 +1,10 @@
+import Backbone from 'backbone';
 import Radio from 'backbone.radio';
 
 import App from 'js/base/app';
 
+const patients = new Backbone.Collection([]);
```
nmajor25, line 7: Collection that the buttons are based on.

`{ id: String, name: String },`

Seems to automatically show/hide buttons when patients are added and removed. And when the entire collection is emptied.

src/js/services/dialer.js
```diff
@@ -34,6 +35,44 @@ export default App.extend({
   call(number, action) {
     this._call(number, action);
   },
+  five9CallNumber(values) {
```
nmajor25: Scenarios for the `five9CallNumber` radio request, depending on what data is passed:

1. `null`: clear the patients collection, which removes all patient dashboard buttons from the UI.
2. `actionId`: get the patient based on the action model and add to collection. If not successful, attempt to get the patient via the phone number 👇 
3. `number`: do patient-search api fetch to get the patient/s with that phone number. And add to the collection. Could be more than one patient.

packages/care-ops-five9/five9.scss
```diff
@@ -61,15 +65,40 @@ $five9-call-ended-hover: color(red, dark);
...
+  align-items: center;
+  background-color: white;
+  border: 1px solid $five9-color;
+  border-radius: 8px;
+  color: $five9-color;
+  cursor: pointer;
+  display: flex;
+  font-size: 14px;
+  font-weight: 600;
+  height: 36px;
```
paulfalgout: don't love having to set the height.  I think it's an indication that some other setting is incorrect.

packages/care-ops-five9/five9.scss
```diff
@@ -61,15 +65,40 @@ $five9-call-ended-hover: color(red, dark);
...
+  justify-content: center;
+  margin-bottom: 8px;
+  overflow: hidden;
+  padding: 10px 16px;
```
paulfalgout: 10?

src/js/services/dialer.js
```diff
@@ -19,14 +23,56 @@ export default App.extend({
       const currentOrg = Radio.request('bootstrap', 'organization');
       const providerName = currentOrg.get('name');
 
+      this.listenTo(Radio.channel('event-router'), 'patient:dashboard', patientId => {
+        patients.remove(patientId);
+      });
```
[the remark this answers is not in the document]
paulfalgout: I do think this won't be specific to five9 so there's no need for it to be inside the if.  Could probably call a function that does this and starts by this.stopListening(Radio.channel('event-router'), 'patient:dashboard');

src/js/services/dialer.js
```diff
@@ -19,14 +23,56 @@ export default App.extend({
...
+
+    const { actionId, number } = values;
+    const currentUrl = window.location.href;
+    const action = Radio.request('entities', 'actions:model', actionId);
+
+    if (action) {
+      const patient = action.getPatient();
+
+      if (patient) {
+        if (!patients.get(patient.id) && !currentUrl.includes(patient.id)) {
```
paulfalgout: patients.has(patient) is enough

src/js/services/dialer.js
paulfalgout: yep should be searchCollection.each(patient, this.addPatient) or something

src/js/services/dialer.js
paulfalgout: I'm starting to wonder if we should just always add the patients to the collection and then just use a viewFilter on the collection that does this url check against the id.  that's probably cleaner..  don't have to worry about when to add/remove.  FWIW we can even just always add.. the collection will prevent duplicates

src/js/services/dialer.js
paulfalgout: This is actually generic enough we can just make it callNumber.  nothing five9 specific in here

packages/care-ops-five9/five9_app.js
```diff
@@ -48,6 +52,7 @@ export default App.extend({
 
         if (!actionId) return;
 
+        Radio.request('dialer', 'five9CallNumber', { actionId, number: callData.number });
```
paulfalgout: I think we want this inside callAccepted and the other in callEnded.  callAccepted also gets ({ callData })

paulfalgout (changes requested):
Looks close

packages/care-ops-five9/five9.scss
nmajor25: Switched that to `padding: 8px 16px`.

packages/care-ops-five9/five9.scss
nmajor25: Removed this and used the top padding instead. Height property isn't needed.

src/js/services/dialer.js
nmajor25: This is renamed now to `callNumber`.

packages/care-ops-five9/five9_views.js
```diff
@@ -48,25 +49,58 @@ const StatusView = View.extend({
...
+  },
+  onClick() {
+    Radio.trigger('event-router', 'patient:dashboard', this.model.id);
+  },
+});
+
+const PatientButtonsView = CollectionView.extend({
+  childView: PatientButtonItemView,
+  initialize() {
+    this.listenTo(Radio.channel('event-router'), 'patient:dashboard', patientId => {
```
paulfalgout: I don't think we need to remove it at all.  the filter will hide it if we're on the page... if we go somewhere else, might as well show the button again

paulfalgout (changes requested):

packages/care-ops-five9/five9_views.js
nmajor25: That filter isn't hiding it when navigating to the pt dashboard (with the pt five9 button already showing).

It does stop the button from showing if user is on that patient's dashboard.

paulfalgout: yeah but I think we can
```js
this.listenTo(Radio.channel('history'), 'change:route', this.render);
```

nmajor25: Nice, yeah that works 👍

src/js/services/dialer.js
```diff
@@ -21,17 +25,46 @@ export default App.extend({
...
+  callNumber(values) {
+    if (!values) {
+      patients.reset();
+      return;
+    }
+
+    const { actionId, number } = values;
+    const action = Radio.request('entities', 'actions:model', actionId);
+
+    if (action) {
```
nmajor25: `if (action)` isn't doing anything here. This is either a valid action model or `{}`. So the code inside is always running.

So we could check for `action.id` or remove the `if` block completely:

```js
. . .

const { actionId, number } = values;

const action = Radio.request('entities', 'actions:model', actionId);
const patient = action.getPatient();

if (patient) {
  this._addPatient(patient);
  return;
}

. . .
```

paulfalgout (approved):

src/js/services/ws.js
```diff
@@ -176,9 +177,10 @@ export default App.extend({
   },
 
   // TODO: We likely want to support a more reboust way of maintaining filters
-  subscribe(resources, { shouldPersist, filters } = {}) {
-    resources = this._getResources(resources);
+  subscribe(resources, { shouldPersist, filters, agent } = {}) {
+    resources = resources ? this._getResources(resources) : null;
```
nmajor25: So we can pass a `null` resources when subscribing: `Radio.request('ws', 'subscribe', null, { data });`.

src/js/services/dialer.js
```diff
@@ -55,6 +58,12 @@ export default App.extend({
 
       searchCollection.fetch({ data: { 'filter[search]': number } })
         .then(() => searchCollection.each(this._addPatient, this));
+
+      return;
+    }
+
+    if (number.includes('agent:')) {
+      Radio.request('ws', 'subscribe', null, { agent });
```
nmajor25: To create a ws subscription object like this:

```js
{
    "name": "Subscribe",
    "data": {
        "clientKey": "3a20dfdb-0997-49de-9745-9fe0a4a3c99c",
        "workspace": "785f926c-4449-11ea-9bf5-a3840c399337",
        "resources": [],
        "agent": "nurse@email.com"
    }
}
```

src/js/services/dialer.js
```diff
@@ -77,6 +86,12 @@ export default App.extend({
       });
     }
   },
+  handleWsMessage(data) {
+    if (data.category !== 'CallTransferred') return;
+
+    const { patient_id, name } = data.payload;
+    patients.add({ id: patient_id, name });
+  },
```
nmajor25, line 68: Do we need to unsubscribe after this?

packages/care-ops-five9/five9_app.js
nmajor25, line 1: This situation is for when a "from" agent transfers a call to a "to" agent. In that scenario, the patient's phone number is not available in the call data.

The following will happen:

1. "from" agent starts call with patient and then transfers the call to another agent.
    - "from" agent: on `'five9CallComplete'`, send `POST /api/five9/call-transferred`.
    - "to" agent: detects that the call was transferred because `callData.number` contains `'agent:<>'` instead of an actual phone number. And subscribes to a ws channel on the agent email.
2. BE gets `POST /api/five9/call-transferred` and sends ws message to FE.
    - Ws message contains patient id and name.
    - "to" agent: when recieved, adds patient to buttons collection.

nmajor25: Corresponding BE changes: https://github.com/RoundingWell/care-ops-backend/pull/10342

src/js/services/ws.js
nmajor25: I removed this change, instead just do `Radio.request('ws', 'subscribe', [], { data });`. Same result.

src/js/services/dialer.js
```diff
@@ -57,12 +57,25 @@ export default App.extend({
...
     Radio.request('entities', 'save:artifacts:model', {
       artifact: 'five9-call-log',
-      identifier,
-      values,
+      identifier: callData.interactionId,
+      values: callLogData,
     });
+
+    if (reason === 'TRANSFERRED') {
+      fetch('/api/five9/call-transferred', {
```
paulfalgout: Pretty sure this needs to be `import fetcher from 'js/base/fetch';` so that you're authorized?  surely this isn't a public API.  I think the fetcher default headers are fine

src/js/services/dialer.js
```diff
@@ -55,6 +58,12 @@ export default App.extend({
 
       searchCollection.fetch({ data: { 'filter[search]': number } })
         .then(() => searchCollection.each(this._addPatient, this));
+
+      return;
+    }
+
+    if (number.includes('agent:')) {
+      Radio.request('ws', 'subscribe', [], { agent });
```
paulfalgout: this is going to clear the subscription resources / any filter on whatever page the user is currently on.  I don't think we can do it this way at all

paulfalgout (changes requested):
Changes needed but we're going to have to chat

src/js/services/ws.js
```diff
@@ -74,6 +74,7 @@ export default App.extend({
     };
 
     if (this.filters) data.filters = this.filters;
+    if (this.agent) data.agent = this.agent;
```
paulfalgout: I do not want to attach a five9 specific variable to the ws service

src/js/services/dialer.js
```diff
@@ -57,12 +58,24 @@ export default App.extend({
...
+
     Radio.request('entities', 'save:artifacts:model', {
       artifact: 'five9-call-log',
-      identifier,
-      values,
+      identifier: callData.interactionId,
+      values: callLogData,
     });
+
+    if (reason === 'TRANSFERRED') {
```
nmajor25: Note: this needs to be `callData.reason`.

packages/care-ops-five9/five9_app.js
```diff
@@ -67,11 +67,15 @@ export default App.extend({
       callAccepted: ({ callData }) => {
         this.setState('callTime', dayjs());
 
-        Radio.request('dialer', 'callNumber', { actionId: this.getState('actionId'), number: callData.number });
+        Radio.request('dialer', 'showPatientLinks', {
```
paulfalgout: the async getCallNumber should happen here so that `showPatientLinks` remains dialer-service neutral.    dialer service doesn't need to know about agent.  safe to do here as either it's a quick return or the number is being transferred to someone who didn't start the call so they're not on that patient's dashboard or the action... 

But we don't want to introduce the phone validation here.. that can stay in the dialer service

src/js/services/dialer.js
```diff
@@ -32,31 +32,41 @@ export default App.extend({
...
     }
   },
+  async getCallNumber(callData) {
+    const { number } = callData;
+    const phone = parsePhoneNumber(number, 'US');
+
+    if (phone && phone.isValid()) return phone;
+
+    if (number.includes('agent:')) {
+      const artifact = await Radio.request('entities', 'fetch:artifacts:model', callData.agent);
```
paulfalgout: this isn't going to be the fetch.  I suspect we'll need a new endpoint/request here that's somehow five9 specific.   hopefully we can make it generic in the app but call it with five9 specific filters from the five9 package.

src/js/services/dialer.js
```diff
@@ -32,31 +32,41 @@ export default App.extend({
...
+
+      if (artifact.number) return artifact.number;
```
paulfalgout: this is going to be something like artifact.values.callData.number

paulfalgout (changes requested):

packages/care-ops-five9/five9_app.js
nmajor25: Makes sense, moved that to the `five9_app`

src/js/services/dialer.js
nmajor25: Ya I figured, just have that there for now. Figured we could figure this out once we get the endpoint from BE.

src/js/services/dialer.js
nmajor25: 👍

packages/care-ops-five9/five9_app.js
```diff
@@ -98,4 +101,18 @@ export default App.extend({
     interaction.click2dial({ click2DialData: { clickToDialNumber: number } });
     this.setState('pendingCall', null);
   },
+  async _getCallNumber(callData) {
+    const { number, agent } = callData;
+
+    if (number.includes(':agent')) {
+      const { data } = await fetcher(`/api/artifacts/search?type=five9-call-log&path=agent&term=${ agent }&limit=1`)
```
nmajor25: Not sure what `path=` should be.

packages/care-ops-five9/five9_app.js
```diff
@@ -98,4 +101,18 @@ export default App.extend({
...
+        .then(handleJSON);
+
+      if (!data.length) return null;
+
+      return data[0].attributes.callData.number;
```
nmajor25: Need to handle scenarios where `callData` or `callData.number` are `undefined`?

packages/care-ops-five9/five9_app.js
```diff
@@ -91,4 +102,18 @@ export default App.extend({
     interaction.click2dial({ click2DialData: { clickToDialNumber: number } });
     this.setState('pendingCall', null);
   },
+  async _getCallNumber(callData) {
+    const { number, agent } = callData;
+
+    if (number.includes(':agent')) {
```
paulfalgout: shouldn't this be `agent:`

packages/care-ops-five9/five9_app.js
```diff
@@ -91,4 +102,18 @@ export default App.extend({
...
+      const { data } = await fetcher(`/api/artifacts/search?filter[type]=five9-call-log&filter[path]=callData.agent&filter[term]=${ agent }&filter[limit]=1`)
```
paulfalgout:
```js
await fetcher('/api/artifacts/search', {
  data: {
    filter: {
      type: 'five9-call-log',
      path: 'callData.agent',
      term: agent,
      limit: 1,
    },
  },
});
```

packages/care-ops-five9/five9_app.js
```diff
@@ -91,4 +102,18 @@ export default App.extend({
...
+        .then(handleJSON);
+
+      if (!data.length) return null;
+
+      return get(data[0], ['attributes', 'callData', 'number']);
```
paulfalgout: no if needed.. can just do `get(data, '0.attributes.callData.number');`

paulfalgout (changes requested):

packages/care-ops-five9/five9_app.js
nmajor25: Updated

packages/care-ops-five9/five9_app.js
nmajor25: Updated

packages/care-ops-five9/five9_app.js
nmajor25: Yep 🤦‍♂️

paulfalgout (approved):

PR #1580, opened 2026-01-14
Fix flakey `schedule` => `click+shift multiselect` test

nmajor25:
Shortcut Story ID: [sc-65956]

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability by adding explicit route synchronization points around navigation and back actions, reducing flakiness and making workflow navigation checks more stable. No user-facing behavior changes.

### DIFF
--- a/test/integration/patients/worklist/schedule.js
+++ b/test/integration/patients/worklist/schedule.js
@@ -1533,10 +1533,12 @@ context('schedule page', function() {
       .click();
 
     cy
-      .navigate('/worklist');
+      .navigate('/worklist')
+      .wait('@routeActions');
 
     cy
-      .go('back');
+      .go('back')
+      .wait('@routeActions');
 
     cy
       .get('@scheduleList')
--- a/test/integration/patients/worklist/worklist.js
+++ b/test/integration/patients/worklist/worklist.js
@@ -4328,10 +4328,12 @@ context('worklist page', function() {
       .click();
 
     cy
-      .navigate('/schedule');
+      .navigate('/schedule')
+      .wait('@routeActions');
 
     cy
-      .go('back');
+      .go('back')
+      .wait('@routeActions');
 
     cy
       .get('@lastTableListItem')


test/integration/patients/worklist/schedule.js
```diff
@@ -1538,6 +1538,9 @@ context('schedule page', function() {
     cy
       .go('back');
 
+    cy
+      .wait('@routeActions');
```
nmajor25, line 1541: This part of the test verifies that `lastSelectedIndex` is reset to `null` when navigating away from the schedule. So when the user comes back to the page, click+shift multiselect doesn't work on the first click.

I think the flake is because the `.navigate('/worklist')` wasn't actually navigating to the worklist page. And therefore `lastSelectedIndex` was sometimes not getting reset to `null`.

nmajor25: I made the same change in the worklist test too. Although I haven't noticed the same level of flake on that one.

paulfalgout: why is navigate not working though?  because the url is incorrect or is there fundamentally something wrong with `navigate` and can it be fixed?

nmajor25: That's a good question. The `/worklist` URL is correct, works for what the test is trying to do.

But this isn't something I can reproduce locally. Only when running in ci. And only a percentage of the time. So it's hard to diagnose.

paulfalgout: I just assume the issue is not having the .`wait('@routeActions');` after the navigate and not the navigate itself

nmajor25: Definitely could be. We could do only that and see if it stops the flake.

nmajor25: Okay, updated 👍

paulfalgout (approved):

PR #1584, opened 2026-01-15
five9 updates for handling transferred calls

nmajor25:
## Summary by CodeRabbit

* **New Features**
  * Transferred calls are detected and shown with a dedicated "Transferred Call" heading, simplifying the UI for transferred sessions.
  * Transferred-call state is cleared when a call ends to restore the normal call view.

* **Bug Fixes**
  * More reliable extraction of call numbers from nested API responses for consistent call information.

### DIFF
--- a/packages/care-ops-five9/five9_app.js
+++ b/packages/care-ops-five9/five9_app.js
@@ -117,7 +117,7 @@ export default App.extend({
         },
       }).then(handleJSON);
 
-      return get(data, '0.attributes.values.callData.number');
+      return get(data, ['0', 'attributes', 'values', 'callData', 'number']);
     }
 
     return number;
--- a/packages/care-ops-five9/five9_app.js
+++ b/packages/care-ops-five9/five9_app.js
@@ -74,6 +74,7 @@ export default App.extend({
         Radio.request('dialer', 'showPatientLinks', { actionId: this.getState('actionId'), number });
       },
       callEnded: () => {
+        this.setState('isTransferredCall', null);
         this.setState('callTime', null);
         Radio.request('dialer', 'showPatientLinks', null);
       },
@@ -106,6 +107,8 @@ export default App.extend({
     const { number } = callData;
 
     if (number.includes('agent:')) {
+      this.setState('isTransferredCall', true);
+
       const { data } = await fetcher('/api/artifacts', {
         data: {
           filter: {
--- a/packages/care-ops-five9/five9_views.js
+++ b/packages/care-ops-five9/five9_views.js
@@ -105,6 +105,7 @@ const LayoutView = View.extend({
   modelEvents: {
     'change:isOpen': 'onChangeIsOpen',
     'change:isCalling': 'showCallState',
+    'change:isTransferredCall': 'showCallState',
     'change:callTime': 'showCallState',
   },
   ui: {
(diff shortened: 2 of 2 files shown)


packages/care-ops-five9/five9_app.js
```diff
@@ -117,7 +117,7 @@ export default App.extend({
         },
       }).then(handleJSON);
 
-      return get(data, '0.attributes.values.callData.number');
```
nmajor25, line 120: This works in lodash, but not underscore. Today I learned...

packages/care-ops-five9/five9_app.js
```diff
@@ -117,7 +117,7 @@ export default App.extend({
...
+      return get(data, ['0', 'attributes', 'values', 'callData', 'number']);
```
nmajor25: I'm going to add a test for this `_getCallNumber()` in general. In a separate PR.

packages/care-ops-five9/five9_app.js
nmajor25: Derek confirmed this works now in his testing.

paulfalgout: kind of surprising.  split('.'), would seemingly produce the same..  might be worth getting underscore to add this.

packages/care-ops-five9/five9_views.js
```diff
@@ -134,17 +135,23 @@ const LayoutView = View.extend({
     this.$el.toggleClass('is-open', this.model.get('isOpen'));
   },
   showCallState() {
+    const isTransferredCall = this.model.get('isTransferredCall');
     const callTime = this.model.get('callTime');
     const isCalling = !!this.model.get('isCalling');
 
-    this.ui.header.toggleClass('is-call-active', isCalling);
+    this.ui.header.toggleClass('is-call-active', isTransferredCall || isCalling);
```
paulfalgout: this is really `isTransferredCall || callTime` as if both of these are false then the style will be is-call-ended based on isCalling.

packages/care-ops-five9/five9_views.js
```diff
@@ -134,17 +135,23 @@ const LayoutView = View.extend({
...
 
     if (callTime) {
       this.showChildView('heading', new TimerView({ startTime: callTime }));
       return;
     }
 
-    const isCallEnded = isCalling && !callTime;
+    const isCallEnded = isCalling && !isTransferredCall && !callTime;
 
     this.ui.header.toggleClass('is-call-ended', isCallEnded);
```
paulfalgout: really this line is `this.ui.header.toggleClass('is-call-ended', isCalling);` isTransferredCall and callTime are always falsy by this point

paulfalgout (changes requested):

packages/care-ops-five9/five9_app.js
paulfalgout: oh ha.. it's not the array index that's the issue.. underscore get doesn't support nested paths as a string at all.  we just use it so much in formio I forget.

packages/care-ops-five9/five9_views.js
nmajor25: Updated 👍

packages/care-ops-five9/five9_views.js
nmajor25: Updated 👍

packages/care-ops-five9/five9_app.js
nmajor25: Yep, I was assuming underscore's methods work the same.

paulfalgout (approved):

PR #1590, opened 2026-01-26
In dialer service, handle `null`/`undefined` phone numbers in `showPatientLinks()`

nmajor25:
Shortcut Story ID: [sc-66108]

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling in the dialer service when phone numbers are not provided
  * The dialer now gracefully handles null or missing phone number parameters

* **Tests**
  * Added test coverage for edge case scenarios where the dialer is called without phone number data
  * Normalized test data for consistency

### DIFF
--- a/src/js/services/dialer.js
+++ b/src/js/services/dialer.js
@@ -48,6 +48,8 @@ export default App.extend({
       return;
     }
 
+    if (!number) return;
+
     const phone = parsePhoneNumber(number, 'US');
 
     if (phone && phone.isValid()) {
--- a/test/integration/services/dialer.js
+++ b/test/integration/services/dialer.js
@@ -164,6 +164,18 @@ context('Dialer Service', function() {
 
     cy
       .getRadio(Radio => {
+        Radio.request('dialer', 'showPatientLinks', {
+          actionId: null,
+          number: null,
+        });
+      });
+
+    cy
+      .get('@patientButtons')
+      .should('have.length', 1);
+
+    cy
+      .getRadio(Radio => {
         Radio.request('dialer', 'showPatientLinks', null);
       });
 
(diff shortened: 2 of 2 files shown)


src/js/services/dialer.js
```diff
@@ -48,6 +48,8 @@ export default App.extend({
       return;
     }
 
+    if (!number) return;
+
     const phone = parsePhoneNumber(number, 'US');
```
nmajor25, line 53: If `number` is `null` or `undefined`, this is throwing a console error. 

[Example.](https://app.datadoghq.com/error-tracking?query=service%3Acare-ops-frontend%20env%3Aprod%2A%20-env%3A%2Asandbox&link_source=monitor_notif&monitor_id=221105666&monitor_sub_type=.new%28%29&source=all&sp=%5B%7B%22p%22%3A%7B%22issueId%22%3A%2230122946-f246-11f0-93a3-da7ad0900002%22%7D%2C%22i%22%3A%22error-tracking-issue%22%7D%5D&from_ts=1769453426000&to_ts=1769455226000&live=false)

test/integration/services/dialer.js
```diff
@@ -162,6 +162,18 @@ context('Dialer Service', function() {
       .get('@patientButtons')
       .should('have.length', 1);
 
+    cy
+      .getRadio(Radio => {
+        Radio.request('dialer', 'showPatientLinks', {
+          actionId: null,
+          number: null,
```
nmajor25, line 169: Without the `if (!number) return;` this test will fail entirely because cypress fails immediately on those.

So the `.should('have.length', 1);` after isn't 100% necessary. But it verifies the existing patient buttons are unchanged.

paulfalgout (approved):

PR #1591, opened 2026-01-26
Fix flakey `dashboards all list` => `display dashboards list` test

nmajor25:
Shortcut Story ID: [sc-65969]

## Summary by CodeRabbit

* **Tests**
  * Updated test data for dashboard naming consistency.
  * Enhanced dashboard test support utilities with automatic ID generation for test fixtures.

### DIFF
--- a/test/integration/dashboards/dashboards-all.js
+++ b/test/integration/dashboards/dashboards-all.js
@@ -4,7 +4,7 @@ context('dashboards all list', function() {
   specify('display dashboards list', function() {
     const testDashboards = [
       getDashboard({
-        attributes: { name: 'Daily Dashboards' },
+        attributes: { name: 'Daily Dashboard' },
       }),
       getDashboard({
         attributes: { name: 'Weekly Dashboard' },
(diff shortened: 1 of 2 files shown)


test/integration/dashboards/dashboards-all.js
```diff
@@ -4,7 +4,7 @@ context('dashboards all list', function() {
   specify('display dashboards list', function() {
```
nmajor25, line 4: This test sometimes fails because not all three of the dashboards show in the list. Appears to caused by an `id` collision.

Where `getDashboards()` is returning one or more dashboards from the fixtures with the same `id`. And therefore the UI only shows one of those duplicate dashboards in the list.

test/support/api/dashboards.js
```diff
@@ -9,6 +10,11 @@ const TYPE = 'dashboards';
 export function getDashboard(data) {
   const resource = getResource(_.sample(fxTestDashboards), TYPE);
 
+  data = _.extend({ id: uuid() }, data);
```
nmajor25, line 13: Makes sure each dashboard has a unique `id`. And the `embed_url` contains that `id`, so we need to set that too.

paulfalgout (approved):

PR #1595, opened 2026-02-06
Expose patient search by identifier api to forms

nmajor25:
Shortcut Story ID: [sc-66394]

Related BE PR: https://github.com/RoundingWell/app-backend/pull/10469

## Summary by CodeRabbit

* **New Features**
  * Added patient lookup so forms can search and populate patient data by identifier type (e.g., MRN).

* **Tests**
  * Added integration test validating patient lookup flow and results rendering.

* **Chores**
  * Updated frontend assets to load the new module build.

* **Bug Fixes**
  * Removed an unused public recording export to clean up exposed functionality.


src/js/services/forms.js
```diff
@@ -57,6 +57,7 @@ export default App.extend({
     'clear:storedSubmission': 'clearStoredSubmission',
     'fetch:field': 'fetchField',
     'fetch:fieldHistory': 'fetchFieldHistory',
+    'fetch:patientByIdentifier': 'fetchPatientByIdentifier',
```
nmajor25: I went back and forth on the naming. `fetchPatientByIdentifier` vs. `fetchPatientsByIdentifier`.

The intent of this is to get one patient given an identifier value. But the api endpoint is used for getting multiple patients and will return a collection (containing zero patients or one patient).

src/js/services/forms.js
```diff
@@ -163,6 +164,21 @@ export default App.extend({
           this.send(message, { error: responseData }, requestId);
         });
   },
+  fetchPatientByIdentifier({ type, identifier }, requestId) {
```
paulfalgout: I think we should make this a more generic interface that just happens to be currently set to take indentifier.  Otherwise we'd just hardcord type.
And making it plural makes sense too.. it just-so-happens that identifier will always filter down to one.

paulfalgout (changes requested):

src/js/services/forms.js
shadowhand: It might make sense to just expose patient search generally to forms? I could see it being useful in other scenarios.

paulfalgout: Yeah maybe, halfway would be not destructuring and just passing the filter through..  we currently have a search implementation I'd want to refactor if we expose this entirely..

src/js/services/forms.js
```diff
@@ -163,6 +164,21 @@ export default App.extend({
           this.send(message, { error: responseData }, requestId);
         });
   },
+  fetchPatientsBy({ type, identifier }, requestId) {
```
paulfalgout: can you change this to `fetchPatientsBy(filter, requestId)`?  Removing the desructuring will make it more generic

paulfalgout (changes requested):

src/js/services/forms.js
nmajor25: Updated 👍

paulfalgout (approved):

PR #1596, opened 2026-02-06
Deploy with aws

DrechslerDerek:
Shortcut Story ID: [sc-66353]

## Summary by CodeRabbit

* **Chores**
  * Added tag-based release workflows with gated approvals for sandbox and production deployments.
* **New Features**
  * CI-driven multi-stack deployment tooling with optional CDN cache invalidation.
  * Automated generation of environment-specific app configuration from secure secrets and CI/git metadata.
* **Documentation**
  * New deployment runbook and updated README with tag-based release instructions.
* **Tests**
  * Unit tests added for app configuration utilities.


.circleci/config.yml
```diff
@@ -148,6 +165,50 @@ jobs:
...
+          command: npm run build
+      - save_cache:
+          key: node-modules-{{ checksum "package-lock.json" }}
+          paths:
+            - node_modules
+      - persist_to_workspace:
+          root: .
+          paths:
+            - .
+
```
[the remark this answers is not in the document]
paulfalgout: We should change this line to be:
```yml
            - dist
            - package.json
            - package-lock.json
```

.circleci/config.yml
```diff
@@ -148,6 +165,50 @@ jobs:
...
+        type: string
+        default: prod
+      sandbox:
+        type: string
+      deploy-name:
+        type: string
+        default: Deploy to S3
+    steps:
+      - attach_workspace:
+          at: .
```
paulfalgout, line 212: then re L189 add here:
```yml
      - restore_cache:
          key: node-modules-{{ checksum "package-lock.json" }}
      - run:
          name: Install dependencies
          command: npm ci
```
The cache reinstall is better than persisting the node_modules

.circleci/config.yml
```diff
@@ -256,3 +317,45 @@ workflows:
             --parallel --record
             --group edge-component
             --config retries=4
+
+  deploy-sandboxes:
+    when:
```
paulfalgout: This line is a circleci syntax issue.  should be:

Image

```yml
    when: *when-release
```

.circleci/config.yml
```diff
@@ -256,3 +317,45 @@ workflows:
...
+          context: aws-prod
+      - release-approval:
+          <<: *tag-filter
+          type: approval
+          requires:
+            - build
+      - deploy:
+          <<: *tag-filter
+          <<: *prod-deploy-job
+          name: deploy sandboxes
```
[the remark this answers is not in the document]
paulfalgout: yep line 335 should be replaced with
```yml
         context: aws-prod
          requires:
            - release-approval
```

.circleci/config.yml
```diff
@@ -256,3 +317,45 @@ workflows:
...
+          bucket-pattern: websitebucket
+          stage: prod
+          sandbox: "true"
+          deploy-name: Deploy sandbox stacks to S3
+
+  deploy-clients:
+    when:
```
paulfalgout: same here.. one line `    when: *when-release`

.circleci/config.yml
```diff
@@ -256,3 +317,45 @@ workflows:
...
+          <<: *tag-filter
+          context: aws-prod
+      - release-approval:
+          <<: *tag-filter
+          type: approval
+          requires:
+            - build
+      - deploy:
+          <<: *tag-filter
+          <<: *prod-deploy-job
```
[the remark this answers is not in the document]
paulfalgout: yep.. similar to above.. replace this with
```yml
          context: aws-prod
          requires:
            - release-approval
```

scripts/generate-appconfig.js
```diff
@@ -0,0 +1,281 @@
...
+function generateAppConfig(secrets, stack) {
+  const version = getVersion();
+  const stage = getStageFromStack(stack);
+
+  // Base config matching PHP AppConfig::toArray()
+  const config = {
+    app: {
+      env: stage,
+      stack,
+      name: secrets.Auth0Name || '',
```
paulfalgout: Is this always the organization name, even if auth0 isn't being used?  this name is what fills out the Organization in the main menu of the frontend app

scripts/deploy.js
```diff
@@ -0,0 +1,269 @@
...
+ * @param {string} filePath - The local file path
+ * @param {string} key - The S3 key (path)
+ */
+async function uploadFile(bucketName, filePath, key) {
+  const region = process.env.AWS_REGION || 'us-west-2';
+  const client = new S3Client({
+    region,
+    credentials: getCredentials(),
+  });
+
```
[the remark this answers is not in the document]
paulfalgout: I think this is legit.  If you move the client creation down into `deployToStack` and then pass the client into uploadFile and uploadDirectory it'll be significantly more performant.

paulfalgout (changes requested):

.circleci/config.yml
DrechslerDerek: what does that mean?

scripts/generate-appconfig.js
DrechslerDerek: This is actually an issue, because Auth0Name isn't used on some stacks, we don't have for a secret for Organization. Let me see how BE used to do it

DrechslerDerek: `'name' => $this->secrets->authZeroName,`
BE does this

scripts/deploy.js
```diff
@@ -0,0 +1,344 @@
...
+  args.forEach(arg => {
+    if (arg.startsWith('--')) {
+      const [key, value] = arg.slice(2).split('=');
+      params[key] = value;
+    }
+  });
+
+  return params;
+}
```
shadowhand: I'm absolutely sure there is a built-in way to handle this with Node.

scripts/deploy.js
```diff
@@ -0,0 +1,344 @@
...
+ * Extract stack name from bucket name
+ * Format: careops-<stage>-<stack>-<bucket-pattern>-<hash>
+ * @param {string} bucketName - The S3 bucket name
+ * @param {string} stage - The deployment stage
+ * @param {string} bucketPattern - The bucket pattern to match
+ * @returns {string|null} Stack name or null if doesn't match
+ */
+function extractStackFromBucket(bucketName, stage, bucketPattern) {
+  const prefix = `careops-${ stage }-`;
+  const suffix = `-${ bucketPattern }-`;
```
shadowhand: I think we probably want to limit this to just `${stage}-${stage}`.

shadowhand (changes requested):

.circleci/config.yml
paulfalgout: Well so if you persist the ` - .` above on line 189, it'll be all 1 gig of code and dependencies on the workspace..   _but_ if you only persist the package files and the dist files you can just re-npm i here.. persist the < 1 mb of build data and reinstall from the cache which will be faster than loading the massive persisted workspace

scripts/deploy.js
DrechslerDerek: and just assume websitebucket?

shadowhand: TBH I think we should pull the CloudFormation Output value named `WebsiteBucket`.

.circleci/config.yml
```diff
@@ -148,6 +165,57 @@ jobs:
...
+      - save_cache:
+          key: node-modules-{{ checksum "package-lock.json" }}
+          paths:
+            - node_modules
+      - persist_to_workspace:
+          root: .
+          paths:
+            - dist
+            - package.json
+            - package-lock.json
```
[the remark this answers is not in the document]
paulfalgout: oh.. ha. yup

paulfalgout:
@DrechslerDerek 👀  can't add you as a reviewer since it's your PR

scripts/deploy.js
```diff
@@ -0,0 +1,314 @@
...
+ * @param {string} stage - Deployment stage (dev, qa, prod, sandbox)
+ * @param {string} filterStack - Optional specific stack identifier to filter
+ * @returns {Promise<Map>} Map of stack identifiers to bucket names
+ */
+async function getStageStacks(cfClient, stage, filterStack) {
+  const stackBuckets = new Map();
+
+  if (filterStack) {
+    // Query specific stack
+    const stackName = `careops-${ stage }-${ filterStack }`;
```
paulfalgout, line 52: is this stackName still relevant or are we changing it currently?

.circleci/config.yml
```diff
@@ -4,6 +4,25 @@ orbs:
...
+  .tag-filter: &tag-filter
+    filters:
+      tags:
+        only: /.*/
+
+  when-release: &when-release
+    and:
+      - matches:
+          pattern: '^release-[0-9]{8}\.[0-9]+$'
+          value: << pipeline.git.tag >>
```
shadowhand: @paulfalgout you're going to tag things as `release00000.0000` ? Why not standard [semver](https://semver.org/)?

.circleci/config.yml
```diff
@@ -17,6 +36,41 @@ commands:
...
+      - run:
+          name: Verify Node version matches .nvmrc
+          command: |
+            if [ ! -f .nvmrc ]; then
+              echo "Missing .nvmrc at repo root"
+              exit 1
+            fi
+            expected="$(tr -d '[:space:]' < .nvmrc | sed 's/^v//')"
+            expected_major_minor="$(echo "$expected" | cut -d. -f1,2)"
+            actual_major_minor="$(node -p 'process.versions.node.split(".").slice(0, 2).join(".")')"
```
shadowhand: Rather than this monkey business, you can use `$NODE_VERSION` env:

https://github.com/CircleCI-Public/cimg-node/blob/main/22.16/Dockerfile

.circleci/config.yml
```diff
@@ -256,3 +327,37 @@ workflows:
...
+      - release-approval:
+          <<: *tag-filter
+          type: approval
+          requires:
+            - build
+      - deploy:
+          <<: [*tag-filter, *prod-deploy-job]
+          name: deploy sandboxes
+          stage: sandbox
+          deploy-name: Deploy sandbox stacks to S3
```
shadowhand: This name ...

.circleci/config.yml
```diff
@@ -256,3 +327,37 @@ workflows:
...
+      - release-approval:
+          <<: *tag-filter
+          type: approval
+          requires:
+            - build
+      - deploy:
+          <<: [*tag-filter, *prod-deploy-job]
+          name: deploy prod
+          stage: prod
+          deploy-name: Deploy production clients to S3
```
shadowhand: ... and this name are inconsistently phrased.

docs/deploy.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Local Deploy: All Stacks In A Stage
+
+Deploy every stack in a stage by omitting `--stack`:
+
+```bash
+AWS_PROFILE=<your-profile> npm run deploy -- --stage=sandbox
+```
+
+This discovers all `careops-<stage>-*` stacks (with pagination) and deploys each one.
```
shadowhand, line 92: `careops` phrase doesn't need to be included, does it?

scripts/deploy.js
```diff
@@ -0,0 +1,314 @@
...
+}
+
+/**
+ * Fetch stack secrets from AWS Secrets Manager
+ * @param {SecretsManagerClient} secretsClient - Secrets Manager client instance
+ * @param {string} stack - Stack identifier
+ * @returns {Promise<Object>} Parsed secret object
+ */
+async function fetchStackSecrets(secretsClient, stack) {
+  const secretName = `careops/customer/${ stack }`;
```
shadowhand: The secret names are changing to `customer/${stage}/${stack}`.

scripts/generate-appconfig.js
```diff
@@ -0,0 +1,144 @@
...
+
+  throw err;
+}
+
+/**
+ * Fetch stack secrets from AWS Secrets Manager using AWS SDK
+ * @param {string} stack - Stack identifier
+ * @returns {Promise<Object>} Parsed secret object
+ */
+async function fetchStackSecrets(stack) {
```
shadowhand: This method also exists in `deploy.js` ?

.circleci/config.yml
paulfalgout: semver IMO is useful for libraries or things that consume them to know how the dependency might affect the consumer.  But semver conventions are difficult to assess for the app and I don't think provide useful information when attempt to determine what's different on this.. a timestamp showing up in datadog is mildly more useful IMO to give a little bit of context

scripts/deploy.js
paulfalgout: when?

docs/deploy.md
paulfalgout: I don't know the answer to this.. I kept it as it was..  what should it be?

.circleci/config.yml
paulfalgout: going to try it, but codex was pushing back because we also use cypress containers and it doesn't think it's exported.. but I kinda think it likely is.

.circleci/config.yml
shadowhand: Okay, and why put `release` in the version?

docs/deploy.md
shadowhand: Eh, let's leave it for now.

scripts/deploy.js
shadowhand: The new secrets are all created, and for FE needs, the relevant information is already available.

.circleci/config.yml
paulfalgout: doesn't have to be release, but it should probably be _something_ even `v` maybe so that we aren't overmatching on potential other future tags?  not that we've used tags thus far.

paulfalgout: changed it to just `v`

shadowhand: Backend doesn't use any prefix, just (eg) `26.5.1`.

.circleci/config.yml
paulfalgout: whelp.. codex was correct, but I still simplified it a bit with `node -v`

paulfalgout (approved):

.circleci/config.yml
```diff
@@ -256,3 +324,37 @@ workflows:
...
+      - release-approval:
+          <<: *tag-filter
+          type: approval
+          requires:
+            - build
+      - deploy:
+          <<: [*tag-filter, *prod-deploy-job]
+          name: deploy sandboxes
+          stage: sandbox
+          deploy-name: Deploy sandbox stacks to S3
```
shadowhand: Technically, this would be better worded as:


```suggestion
          deploy-name: Deploy assets to sandbox stacks
```

scripts/lib/appconfig.js
```diff
@@ -0,0 +1,158 @@
+/**
+ * Determine the stage from stack name
+ * Matches backend Stage::fromStack() logic
```
shadowhand: https://github.com/RoundingWell/app-backend/pull/10508 changes this logic:

- There is a new stage called `sandbox`
- The default `stage` is now `sandbox`
- The stage `prod` MUST be explicitly set

shadowhand: TBH, I would rather NOT have this code here at all, and require that `--stack` or `STACK` or `DEPLOY_STACK` be explicitly defined.

scripts/lib/appconfig.js
```diff
@@ -0,0 +1,158 @@
...
+  deploymentSource,
+}) {
+  const stage = getStageFromStack(stack);
+
+  // Base config matching PHP AppConfig::toArray()
+  const config = {
+    app: {
+      env: stage,
+      stack,
+      name: secrets.Auth0Name || '',
```
shadowhand: This should be changed to `secrets.OrganizationName`.

shadowhand (changes requested):

scripts/lib/appconfig.js
paulfalgout: yep.. this is unnecessary..  ultimately also only used for datadog.. enforced correctly by CI and so at worst the FE logs on dev boxes or qa will be incorrect.. although I do not think we use the `env` on the FE logging anyhow.

docs/deploy.md
```diff
@@ -0,0 +1,131 @@
...
+
+## CircleCI Deploys
+
+CircleCI deploy workflows run only for release tags matching:
+- `vYYYYMMDD.N`
+
+Current deploy workflows:
+- `deploy-sandboxes` (stage `sandbox`)
+- `deploy-clients` (stage `prod`)
```
shadowhand: Naming seems inconsistent.

scripts/lib/appconfig.js
```diff
@@ -0,0 +1,134 @@
...
+  deploymentSource,
+}) {
+  // Base config matching PHP AppConfig::toArray()
+  const config = {
+    app: {
+      env: stage,
+      stack,
+      name: secrets.OrganizationName || '',
+      disableLoginPrompt: parseBool(secrets.DisableLoginPrompt || 'false'),
+      enableWebsockets: parseBool(secrets.WebsocketsEnabled || 'no'),
```
shadowhand, line 111: Why `false` for one default, and `no` for the other?

paulfalgout: I don't know.  this was added prior to my refactoring and I'm unsure about the format of the secrets

shadowhand (approved):

PR #1606, opened 2026-02-20
Update dependencies

nmajor25:
Shortcut Story ID: [sc-66119]

Minor versions updated.

Major version updates:
* `eslint`
* `jquery`
* `stylelint`
* `@cypress/code-coverage`

## Summary by CodeRabbit

* **Chores**
  * Updated development dependencies (jQuery v4, ESLint v10, Stylelint v17, Cypress code-coverage and related tooling).
* **Tests**
  * Simplified test scripts by removing explicit coverage environment flags so coverage now uses default behavior.
  * Adjusted test helpers to remove unintended variable side effects, improving test stability.
* **Maintenance**
  * Minor module import adjustment for compatibility with updated tooling.


package.json
```diff
@@ -28,9 +28,9 @@
     "formio:build": "npm run build --prefix packages/care-ops-formio",
     "coverage": "npm run coverage:clear && npm run coverage:component && npm run coverage:e2e && npm stop",
     "precoverage:e2e": "vite build -m test",
-    "coverage:e2e": "vite preview -m test & CYPRESS_COVERAGE=1 cypress run",
+    "coverage:e2e": "vite preview -m test & cypress run",
     "postcoverage:e2e": "npm run coverage:report",
-    "coverage:component": "CYPRESS_COVERAGE=1 cypress run --component",
```
nmajor25, line 33: I was investigating this change in the new `@cypress/code-coverage` version: https://github.com/cypress-io/code-coverage/pull/972. Potentially needing to change how the `CYPRESS_COVERAGE=1` env variable is set.

But I don't think we need `CYPRESS_COVERAGE=1` if `@cypress/code-coverage` is configured. The default is to run coverage, so I don't think these were doing anything before?

test/support/api/patient-fields.js
```diff
@@ -25,7 +25,7 @@ export function getPatientField(data) {
 }
 
 export function getPatientFields({ attributes, relationships, meta } = {}, { sample = 3, depth } = {}) {
-  if (depth++ > 1) return;
```
nmajor25, line 28: New `eslint` version picked a couple of these up on `npm run lint`.

nmajor25: `This assigned value is not used in subsequent statements  no-useless-assignment`

package.json
```diff
@@ -122,7 +122,7 @@
     "dayjs": "^1.10.6",
     "handlebars": "^4.7.7",
     "handlebars-intl": "^1.1.2",
-    "jquery": "^3.6.0",
+    "jquery": "^4.0.0",
```
nmajor25, line 125: Full change log here: https://github.com/jquery/jquery/releases/tag/4.0.0

Didn't spot anything that is breaking for us.

test/support/api/patient-fields.js
paulfalgout: ~it should be depth + 1 then~ nope nevermind

.circleci/config.yml
```diff
@@ -78,7 +78,6 @@ executors:
     environment:
       DISPLAY: :99
       CYPRESS_DISPLAY: :99
-      CYPRESS_COVERAGE: true
```
paulfalgout, line 81: there's also a couple of --env COVERAGE=true in here

package.json
paulfalgout: --env COVERAGE=true

paulfalgout (changes requested):

package.json
nmajor25: Updated 👍

.circleci/config.yml
nmajor25: Updated 👍

paulfalgout (approved):

PR #1607, opened 2026-02-20
Implement `subscriptionVersion` on web socket messages

nmajor25:
Shortcut Story ID: [sc-66686]

To fix a race condition bug where we were receiving ws messages from stale ws subscriptions.

1. When FE sends "Subscribe" ws message, create a uuid that is cached on the ws service and include that on the subscription data sent to the BE.
2. BE stores that and attaches it to future ws messages sent to the FE.
3. When FE receives ws messages, don't handle messages if the version is set on the message AND doesn't match what's cached on ws service. (if no `subscription_version` is set, handle message)

Corresponding BE PR: https://github.com/RoundingWell/app-backend/pull/10556

To do:

- [x] Implement `subscriptionVersion`
- [x] Modify component test
- [x] Test that it works on my dev sandbox

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced WebSocket reliability by implementing subscription version tracking to prevent processing of outdated or mismatched messages.

* **Tests**
  * Added comprehensive tests for subscription version validation in WebSocket messaging.

### DIFF
--- a/src/js/services/ws.cy.js
+++ b/src/js/services/ws.cy.js
@@ -98,18 +98,6 @@ context('WS Service', function() {
 
     const closedTest = { id: 'foo', type: 'bar' };
 
-    const notification = {
-      state: {},
-      data: {
-        name: 'Subscribe',
-        data: {
-          clientKey,
-          workspace,
-          resources: [closedTest],
-        },
-      },
-    };
-
     cy
       .startService()
       .then(() => {
@@ -127,7 +115,18 @@ context('WS Service', function() {
       .should('be.calledTwice')
       .then(spy => {
         const secondCall = spy.getCall(1);
-        expect(secondCall.args[0]).to.deep.equal(notification);
+        expect(secondCall.args[0]).to.deep.equal({
+          state: {},
+          data: {
+            name: 'Subscribe',
+            data: {
+              clientKey,
+              workspace,
+              resources: [closedTest],
+              subscriptionVersion: service.subscriptionVersion,
+            },
+          },
+        });
       });
   });
 
@@ -147,7 +146,13 @@ context('WS Service', function() {
       });
 
     cy
-      .sendWs({ category: 'Test', resource: { id: 'id', type: 'flows' } })
+      .then(() =>
+        cy.sendWs({
+          category: 'Test',
+          resource: { id: 'id', type: 'flows' },
+          subscription_version: service.subscriptionVersion,
+        }),
+      )
       .then(() => {
         expect(handler).to.be.calledOnce;
         const callArgs = handler.getCall(0).args;
@@ -156,11 +161,33 @@ context('WS Service', function() {
         expect(handler2).to.be.calledOnce;
       });
 
+    // subscription_version intentially left undefined to test that message is still handled
     cy
-      .sendWs({ category: 'Test' })
       .then(() => {
-        expect(handler).to.be.calledTwice;
-        expect(handler2).to.be.calledOnce;
+        handler.reset();
+        handler2.reset();
+
+        cy.sendWs({ category: 'Test' });
+      })
+      .then(() => {
+        expect(handler).to.be.calledOnce;
+        expect(handler2).to.not.be.called;
+      });
+
+    // should not handle this message because subscription_version doesn't match
+    cy
+      .then(() => {
+        handler.reset();
+        handler2.reset();
+
+        cy.sendWs({
+          category: 'Test',
+          subscription_version: 'stale-version',
+        });
+      })
+      .then(() => {
+        expect(handler).to.not.be.called;
+        expect(handler2).to.not.be.called;
       });
   });
 
@@ -191,6 +218,7 @@ context('WS Service', function() {
           clientKey,
           workspace,
           resources,
+          subscriptionVersion: Cypress.sinon.match.string,
         },
       };
     }
--- a/src/js/services/ws.js
+++ b/src/js/services/ws.js
@@ -1,6 +1,7 @@
 import { each, map, values, isArray } from 'underscore';
 import Backbone from 'backbone';
 import Radio from 'backbone.radio';
+import { v4 as uuid } from 'uuid';
 
 import App from 'js/base/app';
 
@@ -64,6 +65,8 @@ export default App.extend({
   },
 
   _subscribe() {
+    this.subscriptionVersion = uuid();
+
     const currentUser = Radio.request('bootstrap', 'currentUser');
     const currentWorkspace = Radio.request('workspace', 'current');
 
@@ -71,6 +74,7 @@ export default App.extend({
       clientKey: currentUser.clientKey,
       workspace: currentWorkspace.id,
       resources: this.resources.toJSON(),
+      subscriptionVersion: this.subscriptionVersion,
     };
 
     if (this.filters) data.filters = this.filters;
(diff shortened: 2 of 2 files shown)


src/js/services/ws.js
```diff
@@ -160,6 +164,8 @@ export default App.extend({
 
     if (!data.category || data.name === 'pong') return;
 
+    if (data.subscription_version !== this.subscriptionVersion) return;
```
nmajor25: Do we need to check `data.name === "Subscribe"` here? I think that will always be `false`.

nmajor25: From what I can tell, BE will be sending this to us as `subscription_version`.

nmajor25:
Interesting. Those failing e2e tests pass locally, but fail on circleci runs.

Edit: because each `cy.sendWs()` doesn't include a `subscribeVersion` value. Not sure how to handle this yet. The `uuid` is created in the app code, so the tests don't have access to it.

src/js/services/ws.js
```diff
@@ -160,6 +164,8 @@ export default App.extend({
...
+
```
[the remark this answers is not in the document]
nmajor25: @paulfalgout do we need to do this? Or just make sure BE deploys prior to FE.

src/js/services/ws.js
paulfalgout: data.category === 'Subscribe'

paulfalgout: we should update this once it is merged

src/js/services/ws.js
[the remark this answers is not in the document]
paulfalgout: we only need to deploy second

src/js/services/ws.cy.js
```diff
@@ -147,7 +146,13 @@ context('WS Service', function() {
       });
 
     cy
-      .sendWs({ category: 'Test', resource: { id: 'id', type: 'flows' } })
+      .then(() =>
+        cy.sendWs({
+          category: 'Test',
```
paulfalgout, line 151: there should only work for category 'Subscribe'

paulfalgout (changes requested):
however we got the clientKey to work in the tests.. that's also a uuid set by the frontend... so should be the same method for the tests

src/js/services/ws.js
nmajor25: `subscription_version` is confirmed to work using the BE branch. Deployed and tested via my dev sandbox.

nmajor25: Same thing for `data.category === 'Subscribe'`. I don't think that will ever be `true`.

On subscribe, we send a `name: "Subscribe"` message to the BE. 

And all of the ws messages we receive from the BE, the `category` is like `StateChanged`, `NameChanged`, `FlowStateChanged`, etc.. 

I believe the only other type we receive is the `name: "pong"` keep-alive messages. But we handle those by bailing out when `data.name === 'pong'` is in the line above.

nmajor25: At least that is my current understanding of this. I could be missing something.

nmajor25: Using this FE branch and the BE branch, I verified that web sockets work with these changes. But I'm not sure how to e2e test the race condition scenario.

src/js/services/ws.js
```diff
@@ -64,13 +65,16 @@ export default App.extend({
   },
 
   _subscribe() {
+    this.subscriptionVersion = _TEST_ ? 'test-id' : uuid();
```
nmajor25: Simplest solution I could think of. But I don't love it.

nmajor25: We could do something more complicated in the `/test/support/websockets.js` file:

```js
// ADDED LINE HERE
let capturedSubscriptionVersion;

const getServer = (url, messageHandler) => {
  return new Cypress.Promise(resolve => {
    if (mockServer) {
      mockServer.close();
    }

    mockServer = new Server(url, { mock: true });
  
    mockServer.on('connection', socket => {
      socket.on('message', message => {

        // ADDED CODE HERE
        const parsed = JSON.parse(message); 
        if (parsed.name === 'Subscribe') {
          capturedSubscriptionVersion = parsed.data.subscriptionVersion;
        }

        messageHandler(JSON.parse(message));
      });
    });

    resolve();
  });
};
```

But the patient flow => action sidebar web socket test was failing. I think  because that one has separate subscriptions for both the flow and the patient action? Not entirely sure.

nmajor25: We talked about `clientKey` and seeing what pattern is used for that, but there isn't anything that would pass/fail a test based its value. So I don't think it's applicable.

paulfalgout: this isn't doing anything.. the tests are all reading the id off of the service.  this can just be `= uuid()`

test/support/websockets.js
```diff
@@ -40,6 +40,7 @@ Cypress.Commands.add('mockWs', url => {
 
 Cypress.Commands.add('sendWs', message => {
   cy.wrap(socketReady).then(() => {
+    message.subscription_version = message.subscription_version ?? 'test-id';
```
paulfalgout: no need for this change

src/js/services/ws.js
paulfalgout: for future non-`name: 'Subscribe'` it may not include this version.. so I think at least we need this to be 
```
if (data.subscription_version && data.subscription_version !== this.subscriptionVersion) return;
```

Then in the tests we only need to set subscription_version if we're testing this

src/js/services/ws.cy.js
```diff
@@ -203,43 +227,43 @@ context('WS Service', function() {
 
     cy
       .get('@wsHandleMessage')
-      .should('be.calledWith', testData([notifications[0]]))
+      .should('be.calledWithMatch', testData([notifications[0]]))
```
paulfalgout: I don't think we need to loosen it here.  subscriptionVersion: Cypress.sinon.match.string, should allow for `calledWith`

paulfalgout (changes requested):
We can't test this in the e2e library.

src/js/services/ws.cy.js
nmajor25: Yep, updated 👍

test/support/websockets.js
nmajor25: Removed 👍

src/js/services/ws.js
nmajor25: Removed 👍

src/js/services/ws.cy.js
```diff
@@ -156,8 +161,24 @@ context('WS Service', function() {
         expect(handler2).to.be.calledOnce;
       });
 
+    // subscription_version intentially left undefined to test that message is stilled handled
+    cy
```
paulfalgout, line 165: let's reset the handler and only test the one we're calling:
```js
    cy
      .then(() => {
        handler.reset();
        cy.sendWs({
          category: 'Test',
          subscription_version: service.subscriptionVersion,
        });
      })
      .then(() => {
        expect(handler).to.be.calledOnce;
      });

    // should not handle this message because subscription_version doesn't match
    cy
      .then(() => {
        handler.reset();
        cy.sendWs({
          category: 'Test',
          subscription_version: 'stale-version',
        });
      })
      .then(() => {
        expect(handler).to.not.be.called;
      });
```

paulfalgout (changes requested):
bah I missed one comment

src/js/services/ws.cy.js
nmajor25: Ya that's better, updated that. But I needed to do a `reset()` on `handler2` as well.

paulfalgout (approved):

PR #1610, opened 2026-02-27
Utilize circleci deploy / rollback pipelines for frontend

paulfalgout:
Shortcut Story ID: [sc-66559]

I think this needs to be merged for me to fully test it.

Currently I have release artifacts saving to S3 and showing up in the deploy ui.

This will allow us to promote releases to any non-dev environment or all sandboxes/prod at once

## Summary by CodeRabbit

* **New Features**
  * Added npm commands to publish and download release artifacts with checksum verification.

* **Chores**
  * Switched CI release flow to publish and manage packaged build artifacts and release markers instead of direct deploy steps.
  * Added an artifact-based deploy path that downloads, verifies, and deploys packaged artifacts, with status updates for success/failure.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,11 +18,6 @@ aliases:
           pattern: '^v[0-9]{8}\.[0-9]+$'
           value: << pipeline.git.tag >>
 
-  prod-deploy-job: &prod-deploy-job
-    context: aws-prod
-    requires:
-      - release-approval
-
 commands:
   setup-fa-auth:
     steps:
@@ -253,6 +248,38 @@ jobs:
             fi
             circleci run release update "<< parameters.marker-deployment-name >>" --status=FAILED
 
+  publish-build-artifact:
+    docker:
+      - image: *node-image
+    steps:
+      - checkout
+      - attach_workspace:
+          at: .
+      - restore-npm-cache
+      - install-dependencies
+      - run:
+          name: Package build artifact
+          command: |
+            set -euo pipefail
+            tag="<< pipeline.git.tag >>"
+            if [ -z "$tag" ]; then
+              echo "No pipeline.git.tag; refusing to publish artifact"
+              exit 1
+            fi
+
+            tar -C dist -czf /tmp/dist.tar.gz .
+            sha256sum /tmp/dist.tar.gz > /tmp/dist.tar.gz.sha256
+      - run:
+          name: Upload build artifact once per tag
+          command: |
+            set -euo pipefail
+            npm run artifact:publish -- \
+              --bucket=rw-frontend-artifacts \
+              --prefix=app-frontend \
+              --tag="<< pipeline.git.tag >>" \
+              --artifact=/tmp/dist.tar.gz \
+              --checksum=/tmp/dist.tar.gz.sha256
+
 # ------------------------------------------------------------
 # WORKFLOWS
 # ------------------------------------------------------------
@@ -360,40 +387,14 @@ workflows:
             --group edge-component
             --config retries=4
 
-  deploy-sandboxes:
+  release-artifact:
     when: *when-release
     jobs:
       - build:
           <<: *tag-filter
-          context: aws-prod
-      - release-approval:
+      - publish-build-artifact:
           <<: *tag-filter
-          type: approval
+          name: publish release artifact
+          context: aws-prod
           requires:
             - build
-      - deploy:
-          <<: [*tag-filter, *prod-deploy-job]
-          name: deploy sandboxes
-          stage: sandbox
-          deploy-name: Deploy assets to sandbox stacks
-          marker-environment: sandbox
-          marker-deployment-name: deploy-sandbox
-
-  deploy-prod:
-    when: *when-release
-    jobs:
-      - build:
-          <<: *tag-filter
-          context: aws-prod
-      - release-approval:
-          <<: *tag-filter
-          type: approval
-          requires:
-            - build
-      - deploy:
-          <<: [*tag-filter, *prod-deploy-job]
-          name: deploy prod
-          stage: prod
-          deploy-name: Deploy assets to production stacks
-          marker-environment: prod
-          marker-deployment-name: deploy-prod
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,10 +273,6 @@ jobs:
           name: CircleCI Release Marker Plan
           command: |
             set -euo pipefail
-            if [ -z "<< pipeline.git.tag >>" ]; then
-              echo "No pipeline.git.tag; skipping release marker plan"
-              exit 0
-            fi
             circleci run release plan "release-<< pipeline.git.tag >>" \
               --environment-name="release" \
               --component-name="app-frontend" \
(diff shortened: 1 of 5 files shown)


paulfalgout:
https://status.coveralls.io/ coveralls is 😭 🔥

.circleci/deploy.yml
```diff
@@ -0,0 +1,160 @@
...
+                echo "Unsupported deploy environment: $target_env"
+                echo "Supported: qa, qa2, all-sandboxes, all-prod, <stack>-sandbox, <stack>-prod"
+                exit 1
+                ;;
+            esac
+
+            if [ -n "$deploy_stack" ] && [ "$deploy_stack" = "$target_env" ]; then
+              echo "Invalid deploy stack parsed from environment: $target_env"
+              exit 1
+            fi
```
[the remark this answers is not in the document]
paulfalgout: environments are added in the deploy UI of circleci.. they won't start with a hyphen

.circleci/config.yml
```diff
@@ -253,6 +248,70 @@ jobs:
...
+      - install-dependencies
+      - run:
+          name: Package build artifact
+          command: |
+            set -euo pipefail
+            tag="<< pipeline.git.tag >>"
+            if [ -z "$tag" ]; then
+              echo "No pipeline.git.tag; refusing to publish artifact"
+              exit 1
+            fi
```
shadowhand, line 268: Pretty sure this is only required one time, since the job will stop on the first failed step.

.circleci/config.yml
```diff
@@ -253,6 +248,70 @@ jobs:
...
+
+            tar -C dist -czf /tmp/dist.tar.gz .
+            sha256sum /tmp/dist.tar.gz > /tmp/dist.tar.gz.sha256
+      - run:
+          name: CircleCI Release Marker Plan
+          command: |
+            set -euo pipefail
```
shadowhand, line 275: Why is this on every step?

.circleci/deploy.yml
```diff
@@ -0,0 +1,160 @@
...
+              all-prod|prod)
+                deploy_stage="prod"
+                ;;
+              *-sandbox)
+                deploy_stage="sandbox"
+                deploy_stack="${target_env%-sandbox}"
+                ;;
+              *-prod)
+                deploy_stage="prod"
+                deploy_stack="${target_env%-prod}"
```
shadowhand: This doesn't look correct to me.

.circleci/deploy.yml
```diff
@@ -0,0 +1,160 @@
...
+                ;;
+              *)
+                echo "Unsupported deploy environment: $target_env"
+                echo "Supported: qa, qa2, all-sandboxes, all-prod, <stack>-sandbox, <stack>-prod"
```
shadowhand: I'm not a fan of this, I would rather see `stage` and `stack` passed separately. _Or_, if that is not possible, do `stage:stack` instead, as in:

```
qa:*
sandbox:*
prod:*
qa:qa2
sandbox:banana
prod:apple
```

.circleci/deploy.yml
```diff
@@ -0,0 +1,160 @@
...
+      - run:
+          name: Download release artifact
+          command: |
+            set -euo pipefail
+            npm run artifact:download -- \
+              --bucket=rw-frontend-artifacts \
+              --prefix=app-frontend \
+              --tag="$DEPLOY_TARGET_VERSION" \
+              --artifact=/tmp/dist.tar.gz \
+              --checksum=/tmp/dist.tar.gz.sha256
```
shadowhand: Not sure why these are two parameters, seems like the checksum should automatically download to `{artifact}.sha256`.

.circleci/deploy.yml
```diff
@@ -0,0 +1,160 @@
...
+              --target-version="$DEPLOY_TARGET_VERSION"
+      - run:
+          name: Deploy selected target
+          command: |
+            set -euo pipefail
+            if [ -n "$DEPLOY_STACK" ]; then
+              npm run deploy -- --stage="$DEPLOY_STAGE" --stack="$DEPLOY_STACK"
+            else
+              npm run deploy -- --stage="$DEPLOY_STAGE"
+            fi
```
shadowhand, line 135: Love this. 💯

scripts/download-artifact.js
```diff
@@ -0,0 +1,107 @@
...
+
+async function verifyChecksum(artifactPath, checksumPath) {
+  const checksumContent = await fs.readFile(checksumPath, 'utf8');
+  const expectedChecksum = parseChecksumFile(checksumContent);
+  const actualChecksum = await calculateSha256(artifactPath);
+
+  if (actualChecksum !== expectedChecksum) {
+    throw new Error(`Artifact checksum mismatch: expected=${ expectedChecksum }, actual=${ actualChecksum }`);
+  }
+}
```
shadowhand, line 73: I think this should be done as a command of `deploy.yml` for two reasons:

1. The checksum is created in YAML, so for parity, also check it in YAML.
2. The performance of raw filesystem will be much better than reading the file into Node memory.

scripts/download-artifact.js
```diff
@@ -0,0 +1,107 @@
...
+      artifact: { type: 'string' },
+      checksum: { type: 'string' },
+    },
+  });
+
+  const inputs = resolveInputs(values);
+  const artifactFileName = path.basename(inputs.artifactPath);
+  const checksumFileName = path.basename(inputs.checksumPath);
+  const artifactKey = `${ inputs.prefix }/${ inputs.tag }/${ artifactFileName }`;
+  const checksumKey = `${ inputs.prefix }/${ inputs.tag }/${ checksumFileName }`;
```
shadowhand, line 90: This is effectively static and doesn't need a parameter, imo.

scripts/publish-artifact.js
```diff
@@ -0,0 +1,107 @@
...
+  });
+
+  const inputs = resolveInputs(values);
+  await assertReadable(inputs.artifactPath);
+  await assertReadable(inputs.checksumPath);
+
+  const artifactFileName = path.basename(inputs.artifactPath);
+  const checksumFileName = path.basename(inputs.checksumPath);
+  const artifactKey = `${ inputs.prefix }/${ inputs.tag }/${ artifactFileName }`;
+  const checksumKey = `${ inputs.prefix }/${ inputs.tag }/${ checksumFileName }`;
```
shadowhand, line 86: Same here, I think the checksum should always be derived.

shadowhand (changes requested):

.circleci/deploy.yml
paulfalgout: giving it a shot.  codex isn't sure if the circleci deploy ux will like the `:`

scripts/download-artifact.js
paulfalgout: this is the only comment that codex pushed back on strongly saying, "kept in JS to encapsulate download+verify logic in one reusable place."

.circleci/config.yml
shadowhand: Marked resolved, but no answer? 😲

scripts/download-artifact.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+function parseChecksumFile(content) {
+  return content.trim().split(/\s+/)[0];
+}
+
+async function verifyChecksum(artifactPath, checksumPath) {
+  const checksumContent = await fs.readFile(checksumPath, 'utf8');
+  const expectedChecksum = parseChecksumFile(checksumContent);
+  const actualChecksum = await calculateSha256(artifactPath);
```
shadowhand, line 68: Surly this would be better done by invoking `sha256sum` as a console tool?

.circleci/config.yml
paulfalgout: sorry.. codex said it addressed it.. here's what it said when I asked about it just now:
```
I addressed the related redundancy by removing the repeated tag guards, but I left set -euo pipefail in each run step intentionally.

Reason: in CircleCI, each run step is a separate shell process, so those shell options do not carry across steps. Keeping it per-step is the reliable way to guarantee strict behavior everywhere.

If you want, I can still change style to reduce repetition (for example default shell options), but that won’t fully replace per-step -u unless we wire it carefully.
```

shadowhand (approved):

PR #1614, opened 2026-03-02
Shorten tag release and use in appconfig

paulfalgout:
Shortcut Story ID: [sc-###]

reduced version tag length for better circleci ux.

## Summary by CodeRabbit

* **Chores**
  * Updated release tag format to use a shorter YYMMDD pattern (vYYMMDD.N).
  * Enhanced deployment version detection to prioritize explicit deployment/tag-provided versions.
  * Adjusted deployment workflows and updated documentation to reflect the revised deploy stages and workflow names.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ aliases:
   when-release: &when-release
     and:
       - matches:
-          pattern: '^v[0-9]{8}\.[0-9]+$'
+          pattern: '^v[0-9]{6}\.[0-9]+$'
           value: << pipeline.git.tag >>
 
 commands:
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -182,7 +182,7 @@ function listRemoteTags(tagPattern) {
 }
 
 function computeNextReleaseTag() {
-  const datePart = dayjs.utc().format('YYYYMMDD');
+  const datePart = dayjs.utc().format('YYMMDD');
   const prefix = `v${ datePart }`;
   const matchingTags = listRemoteTags(`refs/tags/${ prefix }.*`)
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -277,6 +277,31 @@ jobs:
...
+              echo "No *.js.map files found in dist; failing release artifact publish"
+              exit 1
+            fi
+
+            site_arg=()
+            if [ -n "${DATADOG_SITE:-}" ]; then
+              site_arg+=(--datadog-site="${DATADOG_SITE}")
+            fi
+
+            npx --yes @datadog/datadog-ci@5.8.0 sourcemaps upload dist \
```
paulfalgout: I asked about why this was added here and not in package.json:

For this repo, it’s better because this is CI-only behavior.

Benefits of not adding @datadog/datadog-ci to package.json:

Keeps local dev installs lighter/faster (it’s a big CLI tree).
Avoids lockfile churn/noise in a public frontend repo.
Avoids introducing extra dependency/security surface for runtime code.
Still reproducible because we pin the exact CLI version in CI (@datadog/datadog-ci@5.8.0).
Tradeoff:

CI fetches that package at runtime (slightly slower step, needs npm/network access).
Given this is a release-only job, that tradeoff is usually worth it.

scripts/release.js
```diff
@@ -182,7 +182,7 @@ function listRemoteTags(tagPattern) {
 }
 
 function computeNextReleaseTag() {
-  const datePart = dayjs.utc().format('YYYYMMDD');
+  const datePart = dayjs.utc().format('YYMMDD');
```
shadowhand, line 185: This should be a separate PR, no?

shadowhand (changes requested):

shadowhand (approved):

PR #1621, opened 2026-03-05
Align appconfig/datadog config access

paulfalgout:
Shortcut Story ID: [sc-66943]

A small config clean up resulted in too many changes in too many places.  This will reduce that issue long term.

## Summary by CodeRabbit

* **New Features**
  * Improved service worker caching with network timeout and response validation for more reliable offline behavior.

* **Bug Fixes**
  * Updated configuration loading to fetch fresh app config at runtime and align startup behavior with the new config shape.

* **Chores**
  * Removed Kinde authentication provider.
  * Consolidated and simplified configuration APIs and introduced explicit app version/stage handling.


packages/care-ops-config/index.js
```diff
@@ -1,32 +1,56 @@
...
-  auth0Config,
-  e2eConfig,
   fetchConfig,
-  kindeConfig,
-  workosConfig,
-  datadogConfig,
-  appConfig,
-  versions,
+  getConfigSection,
+  getAppConfig,
```
shadowhand: I don't like these three functions being exported. I think there should be a function exported for specific values (like `getAppVersion`) for anything that needs to be accessible.

packages/care-ops-config/index.js
```diff
@@ -1,32 +1,56 @@
...
+      configState = config;
     });
 }
 
+function getConfigSection(name) {
+  return configState[name];
+}
+
+function getAppConfig() {
+  return getConfigSection('app');
```
shadowhand: Like Copilot said, I don't understand why this method exists, when it has no guards in place. It could just be `return configState.app ?? {};` ?

packages/care-ops-config/index.js
```diff
@@ -1,32 +1,56 @@
...
+}
+
+function getEnvName() {
+  const app = getAppConfig();
+  return `${ app.stage }.${ app.stack }`;
```
shadowhand: Like, this could be just as cleanly written as:


```suggestion
  return `${ configState.app.stage }.${ configState.app.stack }`;
```

packages/care-ops-config/index.js
```diff
@@ -1,32 +1,56 @@
...
     .then(response => response.json())
     .then(config => {
-      Object.assign(auth0Config, config.auth0);
-      Object.assign(e2eConfig, config.e2e);
-      Object.assign(kindeConfig, config.kinde);
-      Object.assign(workosConfig, config.workos);
-      Object.assign(datadogConfig, config.datadog);
-      Object.assign(appConfig, config.app);
-      Object.assign(versions, config.versions);
+      configState = config;
```
shadowhand: Personally, I would rather see this as:


```suggestion
      configState.app = config.app ?? {};
      configState.datadog = config.datadog ?? {};
      // etc...
```

shadowhand (commented):
Generally, I appreciate simplicity, but I think this refactor has lost a fair bit of type safety.

packages/care-ops-config/index.js
paulfalgout: enumerating the auth provider keys made this config package too coupled to auth.  But I have a solution which will solve both issues

packages/care-ops-config/index.js
paulfalgout: Updating, but FWIW I used to care more about these types of clean up, but I'm reducing my scrutiny over maintenance patterns because AI works with the code now more than I do.

src/js/auth.js
```diff
@@ -20,36 +17,28 @@ let authAgent;
...
+  if (getRootRoute() === 'outreach') return defaultAuthProvider;
 
-  if (!isEmpty(e2eConfig)) {
-    return new AuthProvider(e2eConfig);
-  }
+  const providerConfig = getAuthConfig();
+  const authProvider = getAuthProvider();
 
-  const LoginView = getLoginView();
+  if (authProvider === 'e2e') {
```
paulfalgout, line 30: @nmajor25 with this change
https://github.com/RoundingWell/app-tests/blob/master/cypress/support/e2e.js#L54
```js
      delete res.body.datadog;
      delete res.body.auth0;
      res.body.e2e = { token: `Basic ${ token }` };
      return res;
```
becomes
```js
      delete res.body.datadog;
      res.body.auth = {
        provider: 'e2e',
        config: { token: `Basic ${ token }` };
      };
      return res;
```

scripts/lib/appconfig.js
```diff
@@ -1,63 +1,75 @@
...
- * @returns {boolean}
- */
-export function parseBool(value) {
-  if (typeof value === 'boolean') return value;
-  if (typeof value !== 'string') return false;
-
-  const truthyValues = ['1', 'true', 'on', 'yes'];
-  return truthyValues.includes(value.toLowerCase());
+function getDisableLoginPrompt(secrets) {
+  return String(secrets.DisableLoginPrompt ?? '').toLowerCase() === 'true';
```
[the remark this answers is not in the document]
paulfalgout: we'll just use "true" from here on.. "yes" was used for the websockets flag we're no longer supporting.. a cleaner text-boolean is better

scripts/lib/appconfig.js
```diff
@@ -1,63 +1,75 @@
...
-    createClientOptions: {
-      apiHostname: secrets.WorkOsApiDomain || 'login.roundingwell.com',
+    provider: 'workos',
+    disableLoginPrompt: getDisableLoginPrompt(secrets),
+    config: {
+      clientId: secrets.WorkOsClientId || '',
+      createClientOptions: {
+        apiHostname: secrets.WorkOsApiDomain || 'login.roundingwell.com',
+      },
     },
```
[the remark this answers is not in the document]
paulfalgout: we never used this.  just going to remove it

packages/care-ops-auth/providers/workos.js
```diff
@@ -65,7 +53,7 @@ export class WorkosAuthProvider extends AuthProvider {
 
     const pathName = location.pathname;
 
-    const clientId = this._getClientId(pathName);
+    const { clientId } = this.config;
 
     this.client = await this._initClient(clientId, success);
```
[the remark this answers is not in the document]
paulfalgout: const { clientId } = this.config; would fail with pretty much just as useful of an error as the thrown error here as far as datadog is concerned.. I don't think we need the helper text

packages/care-ops-auth/providers/auth0.js
```diff
@@ -40,7 +40,6 @@ export class Auth0AuthProvider extends AuthProvider {
       ...this.config,
       authorizationParams: {
         redirect_uri: location.origin + AuthProvider.PATH_AUTHD,
-        audience: 'care-ops-backend',
```
shadowhand, line 43: I'm fairly certain this _can't_ be removed.

packages/care-ops-config/index.js
```diff
@@ -1,32 +1,67 @@
...
-      Object.assign(auth0Config, config.auth0);
-      Object.assign(e2eConfig, config.e2e);
-      Object.assign(kindeConfig, config.kinde);
-      Object.assign(workosConfig, config.workos);
-      Object.assign(datadogConfig, config.datadog);
-      Object.assign(appConfig, config.app);
-      Object.assign(versions, config.versions);
+      configState.app = config.app || {};
+      configState.auth = config.auth || {};
+      configState.datadog = config.datadog || {};
```
shadowhand, line 9: Why `||` over `??` ?

packages/care-ops-config/index.js
```diff
@@ -1,32 +1,67 @@
...
+function getAuthDisableLoginPrompt() {
+  return configState.auth.disableLoginPrompt;
+}
+
+function getAppName() {
+  return configState.app.name;
+}
+
+function getEnvName() {
+  return `${ configState.app.stage }.${ configState.app.stack }`;
```
shadowhand, line 30: Just FYI, devops has been discussing the need to change `stack` to `customer` as a business-wide terminology change. This primarily because the word "stack" has specific meaning in programming (e.g. "stack trace") and some services apply special meaning to `stack`. Most specifically, DataDog assumes that the log key `stack` contains a stack trace, and our usage causes some weird mixing of data types (string and object) in logging.

shadowhand (approved):

packages/care-ops-auth/providers/auth0.js
paulfalgout: I moved it over to the appconfig.json.  https://github.com/RoundingWell/app-frontend/pull/1621/changes#diff-82e13f72d43e6ff166c1ae8c72f8784a9be8c373218af8405a32c9119a712639R38

packages/care-ops-config/index.js
paulfalgout: I'm not sure the default is needed at all.  If any of these are empty it's a problem, but `0`, `false` `NaN` would probably be worse?  might as well default to `{}` over allowing other falsy values

PR #1623, opened 2026-03-17
Split frontend deploy artifacts from app deploys and migrate CircleCI AWS auth to OIDC

paulfalgout:
Shortcut Story ID: [sc-67219]

---

<a href="https://app.devin.ai/review/roundingwell/app-frontend/pull/1623" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **Chores**
  * CI now publishes release artifacts to a dev artifact bucket and uses AWS OIDC for role-based access; deploy job removed and workflows adjusted to the new artifact-centric flow.
  * Added npm cache restore/save around install to speed deploy-from-artifact runs.

* **Documentation**
  * Updated deployment docs and README to describe publishing release artifacts, artifact-based deploys, and OIDC role assumptions for dev/prod.


paulfalgout:
The [deploy-qa - 3204d418](https://github.com/RoundingWell/app-frontend/pull/1623/checks?check_run_id=67428872073) is failing because the release tag was using the prod s3 bucket.. so we need to get this merged, create a release tag and then it should successfully be able to push to qa

.circleci/config.yml
```diff
@@ -1,10 +1,13 @@
 version: 2.1
 
 orbs:
+  aws-cli: circleci/aws-cli@4.0
   cypress:   cypress-io/cypress@4
   coveralls: coveralls/coveralls@2
 
 aliases:
+  .aws-dev-role-arn: &aws-dev-role-arn arn:aws:iam::104566035342:role/CircleCIRole
+  .aws-prod-role-arn: &aws-prod-role-arn arn:aws:iam::543732963292:role/CircleCIRole
```
shadowhand: Whaaaaaat, these should not be publicly listed! These values need to be stored as secrets on CircleCI.

paulfalgout: removing.  for historical purposes I was given this config and I pressed multiple models on thinking this should be private, but the consensus was:

Not much point, unless you want operational indirection more than clarity.

For this case, I’d keep the role ARNs inline because:
- they are not secrets
- they make the deploy target obvious in code review
- CircleCI config validation stays self-contained
- moving them to env vars just shifts required knowledge into CircleCI settings
If the only goal is deduplication, env vars are a weak trade:
- the duplication is small
- the values live in two different config files anyway
- debugging gets worse because the effective role is no longer visible in the repo
Good reasons to use env vars instead:
- the ARN changes per org/environment outside repo control
- multiple repos share the same CircleCI contexts and you want one place to rotate values
- you want to keep account IDs/role names out of the repo for policy reasons

shadowhand: > you want to keep account IDs/role names out of the repo for policy reasons

✅

shadowhand (approved):

PR #1626, opened 2026-03-17
Configure `settings` service to use `currentClinician.settings`

nmajor25:
Shortcut Story ID: [sc-67178]

If setting is found in `currentClinician.settings: []`, that value should take priority over the same setting set on the workspace & organization level.

## Summary by CodeRabbit

* **New Features**
  * Settings management now supports clinician-specific overrides that automatically take priority over workspace defaults. Individual clinicians can maintain personalized preferences while workspace defaults continue to serve as fallbacks for settings not customized at the individual level.

* **Tests**
  * Expanded test suite to validate clinician-specific setting retrieval and ensure proper priority ordering between clinician and workspace settings.

### DIFF
--- a/src/js/services/settings.cy.js
+++ b/src/js/services/settings.cy.js
@@ -2,6 +2,7 @@ import Radio from 'backbone.radio';
 
 import SettingsService from './settings';
 
+import { Model as Clinician } from 'js/entities-service/entities/clinicians';
 import { Model as Workspace } from 'js/entities-service/entities/workspaces';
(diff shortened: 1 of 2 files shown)


src/js/services/settings.cy.js
nmajor25, line 1: Think it's fine to just test it here, and not in the integration tests yet. When we implement the new dialer functionality, we can test different scenarios there. Is my thinking.

paulfalgout (approved):

PR #1628, opened 2026-03-18
Fix logout hard-route

paulfalgout:
authkit 0.19 fixed session clean up, but in the case where the logout is reached from a hard refresh (which is how we've been testing), authkit assumes the client is unauthed without a local token and redirects as if already signed out.  Requesting the token seems to allow the signOut, and now that the sessions is cleaned up it actually goes to the login screen

Shortcut Story ID: [sc-65433]

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the logout process to gracefully handle edge cases where sessions may not exist, ensuring users can reliably sign out even in uncommon authentication scenarios.


shadowhand (approved):

nmajor25 (approved):
🤞 🤞 🤞

PR #1633, opened 2026-03-20
RingCentral call integration

nmajor25:
Shortcut Story ID: [sc-67295] [sc-67314]

https://github.com/user-attachments/assets/2cd5406b-0155-4703-a1a9-d54c004571cd

## Summary by CodeRabbit

* **New Features**
  * Add RingCentral call API to initiate calls from the app and prevent starting a new call while one is active
  * Patient quick-navigation buttons for jumping to patient dashboards
  * Call timer, transferred/ended indicators, and automatic call logging

* **Style**
  * Updated RingCentral UI header states and improved patient-button visuals

* **Tests**
  * Added unit and integration tests covering RingCentral call flows and patient-button behavior

### DIFF
--- a/packages/care-ops-ringcentral/ringcentral_app.js
+++ b/packages/care-ops-ringcentral/ringcentral_app.js
@@ -1,4 +1,3 @@
-import { delay } from 'underscore';
 import Radio from 'backbone.radio';
 import dayjs from 'dayjs';
 
@@ -42,11 +41,6 @@ export default App.extend({
 
       // when user creates a call from dial pad
       if (data.type === 'rc-call-init-notify') {
-        if (this._callEndTimer) {
-          clearTimeout(this._callEndTimer);
-          this._callEndTimer = null;
-        }
-
         this.setState('callState', null);
         this.setState('actionId', null);
       }
@@ -68,16 +62,11 @@ export default App.extend({
       if (data.type === 'rc-call-end-notify') {
         Radio.request('dialer', 'ringcentralCall', { callData: data.call });
 
-        this.setState('callState', 'ended');
+        this.setState('callState', null);
         this.setState('callTime', null);
+        this.setState('actionId', null);
 
         Radio.request('dialer', 'showPatientLinks', null);
-
-        this._callEndTimer = delay(() => {
-          this._callEndTimer = null;
-          this.setState('callState', null);
-          this.setState('actionId', null);
-        }, 10000);
       }
     });
   },
@@ -86,13 +75,6 @@ export default App.extend({
 
     // If there's an active call, only show the panel
     if (['active', 'transferred'].includes(this.getState('callState'))) return;
-
-    if (this._callEndTimer) {
-      clearTimeout(this._callEndTimer);
-      this._callEndTimer = null;
-      this.setState('callState', null);
-      this.setState('actionId', null);
-    }
 
     this.setState('pendingCall', number);
     this.setState('actionId', action.id);
--- a/packages/care-ops-ringcentral/ringcentral_app.js
+++ b/packages/care-ops-ringcentral/ringcentral_app.js
@@ -18,9 +18,13 @@ export default App.extend({
   onLoginChange() {
     this._call();
   },
+  initialize({ patients }) {
+    this.patients = patients;
+  },
   onStart() {
     this.showView(new LayoutView({
       model: this.getState(),
+      collection: this.patients,
     }));
 
     window.addEventListener('message', ({ data }) => {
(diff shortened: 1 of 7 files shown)


packages/care-ops-ringcentral/ringcentral_app.js
```diff
@@ -1,12 +1,97 @@
...
+        Radio.request('dialer', 'ringcentralCall', { callData: data.call });
+
+        this.setState('callTime', null);
+
+        Radio.request('dialer', 'showPatientLinks', null);
+
+        delay(() => {
+          this.setState('isCalling', false);
+          this.setState('actionId', null);
+        }, 10000);
```
nmajor25: This is configured to show the Call Ended panel header UI for 10 seconds (red bg color, call duration). And then switch back to the blue default.

I'm not sure how five9 currently handles this. How long this should be shown? Until the next call is made?

packages/care-ops-ringcentral/ringcentral_views.js
```diff
@@ -11,29 +77,35 @@ const StatusView = View.extend({
...
+      <div class="ringcentral-panel__header js-header">
         {{fas "phone"}}
-        <span>RingCentral</span>
+        <span data-heading-region>RingCentral</span>
         <span data-status-region></span>
       </div>
       <iframe
-        class="rc-panel__iframe"
+        class="ringcentral-panel__iframe"
         src="https://apps.ringcentral.com/integration/ringcentral-embeddable/latest/app.html?clientId=e2M8xGmJjcGcHFuFe7epUC"
```
nmajor25, line 81: Having the `clientId` on here doesn't let me login. I think we need to wait for RingCentral to give our registered app full permissions? I'm not sure.

But to test this locally, that needs to be removed for now. Which runs in demo mode.

nmajor25: This seems to be fixed now. I had to tweak the permissions for our app in the RingCentral dashboard.

packages/care-ops-ringcentral/ringcentral_app.js
```diff
@@ -53,12 +53,14 @@ export default App.extend({
 
       // when a user accepts a ringing inbound call or outbound call is connected
       if (data.type === 'rc-call-start-notify') {
+        const isTransferredCall = data.call?.direction === 'Inbound' && data.call?.isForwarded === true;
```
nmajor25, line 49: I'm fairly confident this will be a good indicator for a transferred call. But not 100% sure until it's tested.

packages/care-ops-ringcentral/ringcentral_app.js
```diff
@@ -53,12 +53,14 @@ export default App.extend({
...
+
         this.setState('callTime', dayjs());
-        this.setState('callState', 'active');
+        this.setState('callState', isTransferredCall ? 'transferred' : 'active');
 
         Radio.request('dialer', 'showPatientLinks', {
           actionId: this.getState('actionId'),
-          number: data.call?.to,
+          number: isTransferredCall ? data.call?.from : data.call?.to,
```
nmajor25, line 55: Not as confident here, for getting the patient's phone number on transferred calls.

For a transferred call, the `callData.from` may just contain the patient's number. But not sure until tested.

If not, a solution would be to grab the `callData.partyData.sessionId` => [make api request to ringcentral](https://developers.ringcentral.com/api-reference/Call-Control/readCallSessionStatus) => extract patient phone number from `responseData.parties` array.

packages/care-ops-ringcentral/ringcentral_app.js
paulfalgout: five9 had a state after the call ended prior to the user choosing a disposition that this state reflects.  I don't think we need the state if it doesn't exist on ringcentral.  I guess what I don't know is if we're collecting useful data once the call ended from ringcentral?  Is that a thing?

paulfalgout (changes requested):
I'm not convinced that we need to attempt to add state handling to ring central if it works different to five9.  We need to figure out how/when the disposition information is set.  That is what would clear the call ended state, rather than an arbitrary timeout.

packages/care-ops-ringcentral/ringcentral_app.js
nmajor25: Oh okay, the user choosing a disposition step is in the five9 UI? I wasn't aware of that.

So the question is whether or not we show the `Call Ended` (red) UI version of the panel header at all? And just switch it back to the default blue version on `rc-call-end-notify`.

nmajor25: We could remove it in this PR and have Mark test it without it. And come back and add it if needed.

I don't think my ringcentral account is setup to make actual calls. All calls I try to make fail and give me a message to contact the account representative. And I don't understand the user flow enough to test this properly.

paulfalgout: Right so.. in five9 `Radio.request('dialer', 'five9Call', { callData, callLogData });` occurs in `callFinished` which is after the disposition is chosen and so it is included in `callLogData`.  If we were to create the artifact at `callEnded` it would be before the disposition, and it would not include the data we're trying to capture.  so we really need to find out when/how a disposition is selected and how to capture that data.  Because likely `Radio.request('dialer', 'ringcentralCall', { callData: data.call });` is pointless to do here

paulfalgout (changes requested):
i think we need to take a pause here.  this isn't going to accomplish capturing the disposition

paulfalgout (approved):

PR #1650, opened 2026-04-06
Improve CircleCI deploy marker fan-out for stage-wide deploys

paulfalgout:
Shortcut Story ID: [sc-new-story]

## Summary

Improve CircleCI deploy markers for stage-wide deploys.

## Changes

- keep the wildcard marker for `qa:*`, `sandbox:*`, and `prod:*`
- fan out concrete environment markers from the same resolved target snapshot used by the deploy
- continue through all resolved environments and fail at the end if any target fails
- preserve per-environment marker outcomes on partial success
- add focused tests and update deploy docs

## Why

A stage-wide deploy should update the wildcard marker and the concrete environment markers it actually touched.

This keeps CircleCI Deploy closer to the real rollout state, especially when a stage-wide deploy only partially succeeds.

## Validation

- `node --test scripts/deploy.test.js scripts/deploy-markers.test.js scripts/dispatch-qa2-e2e.test.js scripts/lib/appconfig.test.js`
- CircleCI config validated for `.circleci/deploy.yml`

## Summary by CodeRabbit

* **Documentation**
  * Clarified deploy runbook: stage-wide deploys emit per-environment markers; semantics for partial/total success and overall stage reporting are defined.

* **New Features**
  * Deploys now record and publish per-environment outcomes for both stage-wide and specific targets, improving visibility of successes/failures.

* **Tests**
  * Added unit tests covering marker naming, update logic, and deploy-target/status persistence.

* **Chores**
  * CI deploy workflow updated to resolve marker environments and expose marker/status artifacts for downstream steps.

### DIFF
--- a/.circleci/deploy.yml
+++ b/.circleci/deploy.yml
@@ -114,9 +114,6 @@ jobs:
               deploy_organization=""
             fi
 
-            marker_env_slug="${target_env//:/-}"
-            marker_env_slug="${marker_env_slug//\*/all}"
-            marker_name="deploy-${marker_env_slug}-${target_version}"
             deploy_run_url="${CIRCLE_BUILD_URL:-https://app.circleci.com/pipelines/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/${CIRCLE_PIPELINE_ID}}"
 
             {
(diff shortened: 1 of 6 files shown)


.circleci/deploy.yml
```diff
@@ -114,20 +114,38 @@ jobs:
...
+              npm run deploy -- --stage="$DEPLOY_STAGE" --write-targets-file="$deploy_targets_file" --list-marker-environments > "$marker_env_file"
+            fi
+
+            if [ ! -s "$marker_env_file" ]; then
+              echo "No deploy marker environments were resolved"
+              exit 1
+            fi
+
+            printf 'export DEPLOY_MARKER_ENV_FILE=%q\n' "$marker_env_file" >> "$BASH_ENV"
+            printf 'export DEPLOY_MARKER_STATUS_FILE=%q\n' "/tmp/deploy-marker-status.json" >> "$BASH_ENV"
```
shadowhand: Why doesn't this one get a variable?

scripts/deploy-markers.js
```diff
@@ -0,0 +1,179 @@
...
+}) {
+  if (deployResult === 'success') {
+    return plannedEnvironments.map(environment => ({
+      environment,
+      status: 'SUCCESS',
+    }));
+  }
+
+  const updates = [];
+  const updateKeys = new Set();
```
shadowhand: This is weird to me... why use array + set when an object would accomplish the same thing?

scripts/deploy-markers.js
```diff
@@ -0,0 +1,179 @@
...
+  const addUpdate = (environment, status) => {
+    if (!environment) return;
+    const key = `${ environment }:${ status }`;
+    if (updateKeys.has(key)) return;
+    updateKeys.add(key);
+    updates.push({ environment, status });
```
shadowhand:
```suggestion
    const key = `${ environment }:${ status }`;
    if (updates.hasOwn(environment)) return;
    updates[environment] = status;
```

shadowhand (commented):
Gut reaction is that this feels a bit complicated to me and that there should be an easier solution. I'm unsure _what_ that solution would be, though.

.circleci/deploy.yml
```diff
@@ -114,20 +114,41 @@ jobs:
...
       - setup-aws-oidc
+      - run:
+          name: Resolve deploy marker environments
+          command: |
+            set -euo pipefail
+
+            deploy_tmp_dir="$(mktemp -d)"
+            marker_env_file="$deploy_tmp_dir/deploy-marker-environments.txt"
+            deploy_targets_file="$deploy_tmp_dir/deploy-targets.json"
+            deploy_marker_status_file="$deploy_tmp_dir/deploy-marker-status.json"
```
shadowhand, line 135: Instead of all of this, what about using [lowdb](https://github.com/typicode/lowdb) or even a temporary SQLite database?

paulfalgout:
> Gut reaction is that this feels a bit complicated to me and that there should be an easier solution. I'm unsure _what_ that solution would be, though.

Yeah a circleci rep reached out to me about deploys earlier, so I responded about this situation too..  it's pretty unfortunate that per-environment deploys get stale after our most common deploy stream.. so I think it's likely worth it to make the deploy UI correct.. But I don't know that people use it like we are.

Told this to codex:

"I simplified a few local pieces in the latest commits: the marker update accumulator is now a single keyed object, the deploy path reuses one AWS client bundle, and the snapshot-file errors are clearer. I looked for a materially simpler shape, but the remaining complexity is mostly from CircleCI’s marker lifecycle itself: resolve the exact target set once, plan markers, run the deploy, then update wildcard and concrete markers from the actual per-environment outcomes. The simpler alternatives I tried either hardcoded environment lists or lost partial-success accuracy"

.circleci/deploy.yml
paulfalgout: codex's thought: I considered that, but I think lowdb or SQLite would be heavier than the problem here. The state we need to hand off is just three short-lived step artifacts inside one CircleCI job: resolved target list, marker env list, and marker status. Plain files keep the dataflow explicit in the YAML and avoid introducing a dependency plus a second state model just to pass a few values between steps. I did simplify the local bookkeeping further in the latest commit, though.

shadowhand (approved):
I still think that `lowdb` would be a reasonable addition here, being that it would be dev-only, lightweight (only JS/JSON), and provide some safety around storage. 🤷🏼‍♂️ But, this is also a clean approach and unlikely to need much more in the way of features.

PR #1655, opened 2026-04-08
In cypress tests, stub forms iframe instead of using formio

nmajor25:
Shortcut Story ID: [sc-67687]
Closes FE-9

## Summary by CodeRabbit

* **Tests**
  * Migrated iframe tests from fragile in-iframe DOM ops to message-driven validations using a browser-side iframe stub.
  * Updated multiple integration suites to assert message traffic/state instead of reading iframe DOM; removed several deprecated specs.
  * Added a stable iframe fixture, global intercept, and a Cypress helper command to serve and use the iframe stub.
* **Chores**
  * Consolidated form iframe URL references to a single route used across the app.


test/integration/patients/patient/sidebar.js
```diff
@@ -337,11 +337,6 @@ context('patient sidebar', function() {
       .contains('Test Modal Form')
       .click();
 
-    cy
-      .get('.modal--large')
-      .find('.js-submit')
-      .should('be.disabled');
```
nmajor25, line 343: I believe this was testing that the submit button is disabled while the form inside the iframe is loading. But I don't think we can test that with the new stubbed iframe.

test/integration/formservice/formservice.js
nmajor25, line 1: Removed this file, since they were tests of the formio package. Not of the parent app's behavior.

nmajor25:
Some coverage branch drops I need to look into for this PR.

https://coveralls.io/builds/78750144

test/integration/forms/action.js
```diff
@@ -175,9 +177,10 @@ context('Patient Action Form', function() {
...
-      .iframe()
-      .find('[name="data[fields.foo]"]')
-      .type('bar');
+      .get('iframe')
+      .then(iframe => {
+        iframe[0].contentWindow.iframeStub.send('update:storedSubmission', { fields: { foo: 'bar' } });
+      });
 
     cy
       .wait(300) // NOTE: must wait due to debounce in iframe
```
paulfalgout: all of these waits shouldn't be necessary.. they were waiting on formio

test/support/defaults.js
```diff
@@ -20,6 +20,9 @@ beforeEach(function() {
     },
   );
 
+  cy
+    .intercept('GET', '/formapp/**', { fixture: 'formio-stub.html' });
```
paulfalgout: this is a surprisingly easy way to do this

test/integration/patients/patient/sidebar.js
paulfalgout: I think if you overwrite cy
    .intercept('GET', '/formapp/**', { fixture: 'formio-stub.html' }); and add a delay it should be false during the delay

test/integration/forms/action.js
```diff
@@ -2831,9 +2826,13 @@ context('Patient Action Form', function() {
...
-      .iframe()
-      .find('textarea[name="data[familyHistory]"]')
-      .should('contain', 'Form draft work done in another tab.');
+      .get('iframe')
+      .should(([iframe]) => {
+        const { receivedMessages } = iframe.contentWindow.iframeStub;
+        const formData = receivedMessages.find(m => m.message === 'fetch:form:data');
+
+        expect(formData.args.value.storedSubmission.familyHistory).to.equal('Form draft work done in another tab.');
+      });
```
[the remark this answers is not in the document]
paulfalgout: I think this review is legit

paulfalgout (changes requested):
I think we should be able to get the coverage back up.

src/js/entities-service/entities/forms.js
```diff
@@ -19,8 +19,7 @@ const _Model = BaseModel.extend({
     return get(this.get('options'), 'submit_hidden');
   },
   getFormUrl(params = {}) {
-    // NOTE: /formapp/ is legacy formio
-    const baseUrl = this.get('url') || '/formapp/index.html';
+    const baseUrl = this.get('url') || '/forms/formio/index.html';
     const url = new URL(baseUrl, window.location.origin);
```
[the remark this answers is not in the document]
paulfalgout: expected

paulfalgout (approved):

PR #1656, opened 2026-04-09
Dialer UI updates

nmajor25:
Shortcut Story ID: [sc-67722] [sc-67702]

Summary of changes in this PR:

* `RingCentral`: show "Incoming Call" in UI panel when incoming call is ringing.
* `RingCentral`: fixed issue where patient jump buttons not shown on inbound calls.
* `RingCentral`/`five9`: don't show patient jump buttons on flows that belong to the current patient id/s.

## Summary by CodeRabbit

* **New Features**
  * Enhanced call state detection to distinguish transferred calls from regular inbound calls
  * Improved patient tracking when navigating between patient flows

* **Bug Fixes**
  * Fixed call state display logic to properly show ringing and active states
  * Improved patient button filtering to avoid duplicate display in flow views

* **Tests**
  * Expanded integration tests for dialer functionality with flow patient tracking

### DIFF
--- a/packages/care-ops-five9/five9_views.js
+++ b/packages/care-ops-five9/five9_views.js
@@ -67,12 +67,11 @@ const PatientButtonItemView = View.extend({
 
 const PatientButtonsView = CollectionView.extend({
   childView: PatientButtonItemView,
-  initialize() {
-    this.listenTo(Radio.channel('history'), 'change:route', this.render);
+  collectionEvents: {
+    'change:currentPatientId': 'render',
   },
   viewFilter({ model }) {
-    const currentUrl = window.location.href;
-    return !currentUrl.includes(model.id);
+    return this.collection.currentPatientId !== model.id;
   },
 });
 
--- a/packages/care-ops-ringcentral/ringcentral_app.js
+++ b/packages/care-ops-ringcentral/ringcentral_app.js
@@ -46,20 +46,20 @@ export default App.extend({
 
       // when an inbound call is ringing
       if (data.type === 'rc-call-ring-notify') {
-        if (['active', 'transferred'].includes(this.getState('callState'))) return;
+        if (this.getState('callState') === 'active') return;
 
         this.setState('callState', 'ringing');
       }
 
       // when a user accepts a ringing inbound call or outbound call is connected
       if (data.type === 'rc-call-start-notify') {
-        const isTransferredCall = data.call?.direction === 'Inbound' && data.call?.isForwarded === true;
+        const isInboundCall = data.call?.direction === 'Inbound';
 
-        this.setState('callState', isTransferredCall ? 'transferred' : 'active');
+        this.setState('callState', 'active');
 
         Radio.request('dialer', 'showPatientLinks', {
           actionId: this.getState('actionId'),
-          number: isTransferredCall ? data.call?.from : data.call?.to,
+          number: isInboundCall ? data.call?.from : data.call?.to,
         });
       }
 
@@ -78,7 +78,7 @@ export default App.extend({
     this.setState('isOpen', true);
 
     // If there's an active call, only show the panel
-    if (['active', 'transferred'].includes(this.getState('callState'))) return;
+    if (this.getState('callState') === 'active') return;
 
     this.setState('pendingCall', number);
     this.setState('actionId', action.id);
--- a/packages/care-ops-ringcentral/ringcentral_app.js
+++ b/packages/care-ops-ringcentral/ringcentral_app.js
@@ -44,6 +44,13 @@ export default App.extend({
         this.setState('actionId', null);
       }
 
+      // when an inbound call is ringing
+      if (data.type === 'rc-call-ring-notify') {
+        if (['active', 'transferred'].includes(this.getState('callState'))) return;
+
+        this.setState('callState', 'ringing');
+      }
+
       // when a user accepts a ringing inbound call or outbound call is connected
       if (data.type === 'rc-call-start-notify') {
         const isTransferredCall = data.call?.direction === 'Inbound' && data.call?.isForwarded === true;
--- a/src/js/apps/patients/patient/flow/flow_app.js
+++ b/src/js/apps/patients/patient/flow/flow_app.js
@@ -66,6 +66,9 @@ export default SubRouterApp.extend({
     this.currentRoute = currentRoute;
     this.editableCollection = actions.clone();
     this.patient = patient;
+
+    Radio.trigger('dialer', 'change:currentPatientId', patient.id);
+
     this.addOpts = this.getAddOpts(this.flow.getProgramFlow());
 
     this.subscribe();
--- a/src/js/apps/patients/reduced-patients-main_app.js
+++ b/src/js/apps/patients/reduced-patients-main_app.js
@@ -1,3 +1,5 @@
+import Radio from 'backbone.radio';
+
 import RouterApp from 'js/base/routerapp';
 
 import FlowApp from 'js/apps/patients/patient/flow/flow_app';
@@ -44,6 +46,18 @@ export default RouterApp.extend({
       action: 'showFlow',
       route: 'flow/:id/action/:id',
     },
+  },
+
+  onBeforeAppRoute(event, patientId) {
+    // if routing to flow route, currentPatientId is set by flow_app's onStart()
+    if (event.startsWith('flow')) return;
+
+    // determines if dialer patient buttons are shown or hidden
+    const isPatientRoute = event.startsWith('patient');
+    Radio.trigger('dialer', 'change:currentPatientId', isPatientRoute ? patientId : null);
+  },
+  onStop() {
+    Radio.trigger('dialer', 'change:currentPatientId', null);
   },
   showPatient(patientId) {
     this.startRoute('patient', { patientId });
--- a/test/integration/services/dialer.js
+++ b/test/integration/services/dialer.js
@@ -5,81 +5,105 @@ import { getPatient } from 'support/api/patients';
 import { getFlow } from 'support/api/flows';
 import { stateTodo } from 'support/api/states';
 import { getCurrentClinician } from 'support/api/clinicians';
+import { workspaceOne } from 'support/api/workspaces';
+import { roleReducedEmployee } from 'support/api/roles';
+
+const STATE_VERSION = 'v6';
+
+const testPatient = getPatient({
+  attributes: {
+    first_name: 'Test',
+    last_name: 'Patient',
+  },
+});
+
+const otherTestPatient = getPatient({
+  attributes: {
+    first_name: 'Other',
+    last_name: 'Patient',
+  },
+});
+
+const testFlow = getFlow({
+  attributes: {
+    name: 'Test Flow',
+  },
+  relationships: {
+    state: getRelationship(stateTodo),
+    patient: getRelationship(testPatient),
+  },
+});
+
+const testAction = getAction({
+  attributes: {
+    name: 'Test Action',
+  },
+  relationships: {
+    state: getRelationship(stateTodo),
+    flow: getRelationship(testFlow),
+    patient: getRelationship(testPatient),
+  },
+});
+
+const searchResults = [
+  {
+    id: testPatient.id,
+    type: 'patient-search-results',
+    attributes: {
+      ...testPatient.attributes,
+      match: {
+        label: 'Phone Number',
+        value: '+16513216543',
+      },
+    },
+    relationships: {
+      patient: getRelationship(testPatient, 'patients'),
+    },
+  },
+  {
+    id: otherTestPatient.id,
+    type: 'patient-search-results',
+    attributes: {
+      ...otherTestPatient.attributes,
+      match: {
+        label: 'Phone Number',
+        value: '+16513216543',
+      },
+    },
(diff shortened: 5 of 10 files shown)


packages/care-ops-ringcentral/ringcentral_app.js
```diff
@@ -97,4 +104,11 @@ export default App.extend({
 
     this.setState('pendingCall', null);
   },
+  isTransferredCall(call) {
```
paulfalgout: ### ⚠️ Heuristic: `partyId` suffix for transfer detection

```js
isTransferredCall(call) {
  if (call?.direction !== 'Inbound') return false;

  const suffix = Number(call?.partyData?.partyId?.split('-').at(-1));
  return Number.isFinite(suffix) && suffix >= 3;
}
```

### What this actually detects

* **Not** “this call was transferred”
* It’s closer to: **“this is likely not the first leg of the session”**

### Why it sometimes works

* RingCentral telephony sessions have multiple **party legs**
* These often appear as:

  * `-1`, `-2`, `-3`, ...
* Higher suffix ⇒ later leg ⇒ could be transfer, conference, etc.

### Why it’s unsafe

* ❌ **Not documented** — no guarantee suffix is sequential or meaningful
* ❌ **False positives** — conferences, consult transfers, queues, IVR hops
* ❌ **Embeddable ≠ telephony layer** — internal structure may change
* ❌ `direction: "Inbound"` does not distinguish transfers

### Recommendation

Treat this as a **best-effort heuristic only**, not a reliable signal.

If used, document it clearly:

> “Identifies likely secondary call legs (may include transfers, conferences, or routing artifacts).”

### Better approach (still heuristic)

Combine multiple signals:

* same `from`
* close timestamps
* new `callId` / `sessionId`
* suffix ≥ 2 or 3

### Bottom line

* ✅ Useful hint
* ❌ Not reliable
* ❌ Not supported by RingCentral docs

src/js/apps/patients/patient/flow/flow_app.js
nmajor25, line 1: One thing I noticed with this solution.

If you navigate from the patient dashboard => patient flow. The patient jump button/s will show when the patient flow is loading and then hide again once fetch finishes. I'm not sure how to avoid that.

For example:

https://github.com/user-attachments/assets/5b342f1c-e3a1-4938-9cb4-51e1a89af844

nmajor25:
I still need to update the tests on this, for the jump to patient button change.

src/js/apps/patients/patient/patient_app.js
```diff
@@ -77,6 +77,8 @@ export default SubRouterApp.extend({
   onStart({ currentRoute }, patient) {
     this.patient = patient;
 
+    Radio.trigger('dialer', 'change:currentPatientId', patient.id);
```
paulfalgout: what if we move this up to the onBeforeStart and use `onBeforeStart({ currentRoute: { eventArgs: [patientId] } }) {`

there's no flow equivalent though

packages/care-ops-five9/five9_views.js
```diff
@@ -69,10 +69,11 @@ const PatientButtonsView = CollectionView.extend({
   childView: PatientButtonItemView,
   initialize() {
     this.listenTo(Radio.channel('history'), 'change:route', this.render);
+    this.listenTo(this.collection, 'change:currentPatientId', this.render);
   },
   viewFilter({ model }) {
     const currentUrl = window.location.href;
```
paulfalgout, line 78: no need to check the url anymore?

paulfalgout (changes requested):

packages/care-ops-five9/five9_views.js
nmajor25: If we move the `Radio.trigger('dialer', 'change:currentPatientId', patientId);` to the `onBeforeStart` in the patient app, I think we can remove the url check.

nmajor25: When I test that, it behaves the same with the url check removed.

src/js/apps/patients/patient/patient_app.js
nmajor25: That works well 👍

src/js/apps/patients/patient/flow/flow_app.js
nmajor25: I went with a different solution, to fix this: http://github.com/RoundingWell/app-frontend/compare/368afcb4db0dcfa81285e8747b9ad45356fb9605..97127c3a51a538cc6a2993bfd9e3b41274938fbd.

packages/care-ops-five9/five9_views.js
```diff
@@ -68,11 +68,15 @@ const PatientButtonItemView = View.extend({
 const PatientButtonsView = CollectionView.extend({
   childView: PatientButtonItemView,
   initialize() {
-    this.listenTo(Radio.channel('history'), 'change:route', this.render);
+    this.listenTo(Radio.channel('history'), 'change:route', fragment => {
+      if (!fragment.includes('/flow/')) this.render();
+    });
+
+    this.listenTo(this.collection, 'change:currentFlowPatientId', this.render);
   },
```
nmajor25, line 76: Unless the user is navigating to the flow page, call `this.render()` to filter the buttons by simply checking the url string.

If the user is navigating to the flow page, use the `this.collection.currentFlowPatientId` value to determine if the current flow belongs to the patient. `currentFlowPatientId` is set every time the flow app starts and is cleared every time it stops.

nmajor25: To fix this issue with the previous solution: https://github.com/RoundingWell/app-frontend/pull/1656#discussion_r3074741881.

nmajor25: I've manually tested this pretty extensively for RingCentral. But I still need to update the cypress tests. So something wrong with this may be revealed during that work.

packages/care-ops-ringcentral/ringcentral_app.js
nmajor25: Ya, I agree :(

packages/care-ops-five9/five9_views.js
```diff
@@ -68,11 +68,15 @@ const PatientButtonItemView = View.extend({
...
   viewFilter({ model }) {
     const currentUrl = window.location.href;
-    return !currentUrl.includes(model.id);
+    return !currentUrl.includes(model.id) && this.collection.currentFlowPatientId !== model.id;
```
paulfalgout, line 79: still checking the url?  It really seems like there should be a singular check here that shouldn't rely on the url

test/integration/services/dialer.js
nmajor25, line 1: These tests are updated now, to verify the patient buttons don't show on the patient dashboard or patient flow.

And that the buttons remain hidden when navigating between them: patient dashboard <=> patient flow.

test/integration/services/dialer.js
```diff
@@ -105,7 +127,21 @@ context('Dialer Service', function() {
...
+        fx.data = [testFlow];
+
+        return fx;
+      })
+      .routePatientActions(fx => {
+        fx.data = [];
+
+        return fx;
+      })
       .visit(`/flow/${ testFlow.id }/action/${ testAction.id }`)
```
nmajor25, line 145: I wanted to just start this test out by loading the worklist. But we still need to start this test on the flow page with the action sidebar open. 

Otherwise, the action won't be available on the first `showPatientLinks` radio. And therefore, the patient won't be found and no buttons will show for the dialer at all without doing a patient search request.

nmajor25: So I load `/flow/${ testFlow.id }/action/${ testAction.id }` initially. And then navigate to the worklist.

packages/care-ops-ringcentral/ringcentral_app.js
```diff
@@ -71,7 +78,7 @@ export default App.extend({
     this.setState('isOpen', true);
 
     // If there's an active call, only show the panel
-    if (['active', 'transferred'].includes(this.getState('callState'))) return;
+    if (this.getState('callState') === 'active') return;
```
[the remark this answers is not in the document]
paulfalgout: why not the suggested solution?

packages/care-ops-five9/five9_views.js
paulfalgout: the changes in this file should be:
```
collectionEvents: {
  'change:currentPatientId': 'render',
},
viewFilter({ model }) {
  return this.collection.currentPatientId !== model.id;
},
```

and figuring out how to make currentPatientId correct should happen in the dialer service

src/js/apps/patients/patient/flow/flow_app.js
```diff
@@ -95,6 +98,9 @@ export default SubRouterApp.extend({
 
     this.startRoute(currentRoute);
   },
+  onStop() {
+    Radio.trigger('dialer', 'change:currentFlowPatientId', null);
```
paulfalgout, line 102: I think the solution for this is to move this up into the patient and reduced patient app.  we'll still need to set it in this onStart, but the onStop and patient related ids can be set in those two apps.

paulfalgout (changes requested):

packages/care-ops-ringcentral/ringcentral_app.js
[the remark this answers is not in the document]
nmajor25: I don't think it would hurt. Probably worth adding.

I didn't add it originally because of this check. Which stops the `ringing` state from being set if another call is currently active.

```
if (data.type === 'rc-call-ring-notify') {
  if (this.getState('callState') === 'active') return;

  this.setState('callState', 'ringing');
}
```

So I'm not sure if this scenario is possible. But not 100% sure.

packages/care-ops-five9/five9_views.js
nmajor25: Ya, to avoid this:

If you navigate from the patient dashboard => patient flow. The patient jump button/s will show when the patient flow is loading and then hide again once fetch finishes.

If the flow page loads quickly, the button flashing visible => not-visible is not going to look good.

https://github.com/user-attachments/assets/5b342f1c-e3a1-4938-9cb4-51e1a89af844

nmajor25: So my solution to fix that was:

1. Listen for every route change and parse it for a matching patient id. (unless the route is changing to the flow page)
2. If the route is changing to the flow page, rely on the `currentFlowPatientId` instead. Which is set/cleared by the flow_app's `onStart`/`onStop`.

nmajor25: But ya, maybe there is a cleaner way to avoid the issue in the screen recording above.

packages/care-ops-five9/five9_views.js
```diff
@@ -68,11 +68,15 @@ const PatientButtonItemView = View.extend({
 const PatientButtonsView = CollectionView.extend({
   childView: PatientButtonItemView,
   initialize() {
-    this.listenTo(Radio.channel('history'), 'change:route', this.render);
+    this.listenTo(Radio.channel('history'), 'change:route', fragment => {
+      if (!fragment.includes('/flow/')) this.render();
```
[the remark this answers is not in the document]
nmajor25: in our app, the fragments contain the workspace name. Which leads off the string. Example: `workspace-one/flow/:id`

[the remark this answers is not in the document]
paulfalgout: I don't love that if we want to change the urls for flows we have to know that each dialer we've integrated needs to be updated.  We really need to find a way to push only the information needed generically such that each dialer can respond to it.  which is why I think the dialer needs to be patientId only.. and we need to harden when the patientId gets updated by moving that into the patients-main_app and reduced-patients-main_app.. or alternatively making some sort of shared patient resolver that both the flow and any other patient context shares..  however I suspect the router apps are cleaner in the end

src/js/apps/patients/patient/flow/flow_app.js
nmajor25: Ya this seems to be the way to go. Just doing some manual testing to fully verify this works.

test/integration/services/dialer.js
nmajor25, line 1: I still need to incorporate the reduced schedule into one of these tests.

nmajor25: Okay that is updated now.

test/support/api/flows.js
```diff
@@ -44,7 +44,7 @@ export function getFlows({ attributes, relationships, meta } = {}, { sample = 3,
   return _.times(sample, () => getFlow({ attributes, relationships, meta }, { depth }));
 }
 
-Cypress.Commands.add('routeFlow', (mutator = _.identity) => {
+Cypress.Commands.add('routeFlow', (mutator = _.identity, options = {}) => {
```
paulfalgout, line 47: actually a good pattern for all of these probably

src/js/apps/patients/reduced-patients-main_app.js
```diff
@@ -45,6 +47,18 @@ export default RouterApp.extend({
       route: 'flow/:id/action/:id',
     },
   },
+
+  onBeforeAppRoute(event, patientId) {
```
paulfalgout, line 51: cleaner than I expected

paulfalgout (approved):

PR #1659, opened 2026-04-13
FE-93 remove legacy appconfig compatibility output

paulfalgout:
Closes FE-93

Can just wait to merge this until after the release goes out for convenience. 

## Summary
Remove the legacy `appconfig.json` compatibility output now that the app has migrated to the nested config format.

## What changed
- stopped generating legacy `app.env`, `app.stack`, and `app.disableLoginPrompt`
- stopped generating `versions.frontend`
- stopped generating top-level `auth0` and `workos` compatibility blocks
- kept the nested `auth` block unchanged
- updated unit tests to assert the new shape and verify the legacy keys are absent

## Verification
- `node --test /Users/paulfalgout/Projects/care-ops-frontend/scripts/lib/appconfig.test.js`

## Summary by CodeRabbit

* **Refactor**
  * Removed legacy configuration compatibility layer and deprecated fields to simplify the application configuration process and reduce technical debt.

* **Tests**
  * Updated test assertions to validate removal of deprecated configuration fields and ensure correct behavior of the streamlined configuration system.

### DIFF
--- a/scripts/lib/appconfig.js
+++ b/scripts/lib/appconfig.js
@@ -61,24 +61,6 @@ function addAuthProvider(config, secrets) {
     createAuth0AuthConfig(secrets);
 }
 
-function addLegacyAuthKeys(config) {
-  if (!config.auth || !config.auth.provider) return;
-  const legacyAuth = {
-    provider: config.auth.provider,
-    disableLoginPrompt: config.auth.disableLoginPrompt,
-    ...(config.auth.config || {}),
-  };
-
-  if (config.auth.provider === 'workos') {
-    config.workos = legacyAuth;
-    return;
-  }
-
-  if (config.auth.provider === 'auth0') {
-    config.auth0 = legacyAuth;
-  }
-}
-
 /**
  * Remove null/undefined values from object (matches PHP nullifyAny)
  * @param {Object} obj - Object to clean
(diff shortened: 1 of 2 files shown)


nmajor25 (approved):

scripts/lib/appconfig.js
```diff
@@ -140,20 +122,6 @@ export function buildAppConfig({
   // Add WorkOS or Auth0 (WorkOS takes priority)
```
shadowhand, line 122: Auth0 is no longer viable.

scripts/lib/appconfig.test.js
```diff
@@ -126,15 +101,6 @@ test('buildAppConfig prioritizes WorkOS and sets auth.provider=workos', () => {
...
       },
     },
-    workos: {
-      provider: 'workos',
-      disableLoginPrompt: false,
-      clientId: 'workos-client',
-      createClientOptions: {
-        apiHostname: 'workos.example.com',
-        devMode: true,
-      },
```
shadowhand, line 136: Is this correct, or should it be the `auth` block that gets deleted?

shadowhand (changes requested):

scripts/lib/appconfig.js
paulfalgout: we'll remove auth0 in one go.. appconfig and code

scripts/lib/appconfig.test.js
paulfalgout: The auth block supports whichever auth we're using, it just happens to be only workos right now.  this is instead of having separate keys per provider.  the only reason this was left in config was for any unlucky release time where the user downloaded the old src but the new config

shadowhand (approved):

PR #1669, opened 2026-04-20
Fix WorkOS session revocation during auth recovery

paulfalgout:
## Summary

Fixes CS-32.

This updates the WorkOS AuthKit frontend flow so normal auth failures no longer call logout and revoke the user session.

Changes:
- Stop calling `logout()` from WorkOS token acquisition failures
- Move 401 recovery into the WorkOS auth provider
- Clear only the local bearer token on recoverable auth failures
- Force-refresh through AuthKit once, retry the failed request once, then show login prompt if recovery fails
- Keep explicit user logout as the only direct logout path
- Preserve hard-linked `/logout` support without manufacturing sessions for signed-out users
- Add a cookie-gated `/logout` round trip for the new-tab AuthKit session gap
- Route lightweight auth events to Datadog RUM actions instead of console logs
- Add focused WorkOS auth and fetch recovery tests
- Update the 401 e2e expectation so it no longer expects redirect to `/logout`

## Validation

- `npm run lint`
- `npm run test:unit`
- `npm run coverage:e2e -- --spec test/integration/globals/error.js`
- `npx vitest run src/js/auth/workos.test.js src/js/base/fetch.test.js`
- `git diff --check`

---
## Summary by cubic
Prevents WorkOS session revocation on token errors and during 401 recovery, and adds a centralized refresh-and-retry flow that keeps users on their current route. Resolves CS-32.

- **Bug Fixes**
  - Stop calling logout on `getAccessToken` failures; clear only the local bearer token and emit `AUTH_GET_TOKEN_FAILED`.
  - Add `auth.handleUnauthorized` that force-refreshes via `@workos-inc/authkit-js`, retries once, then prompts login if still 401; coalesces concurrent recoveries; skips when offline.
  - Update base `fetch` to delegate 401s to auth and use the retry response; remove automatic redirect behavior.
  - Refine `/logout`: if a user exists, call `signOut({ returnTo: origin })`; if only a WorkOS session cookie exists, start a logout round trip; otherwise return to app root. Explicit `logout()` signs out without routing through `/logout`.
  - Send auth events to Datadog RUM using `addAction` from `@roundingwell/care-ops-datadog`.
  - Add focused WorkOS auth and fetch recovery tests; update the 401 e2e to keep the current workspace route.

<sup>Written for commit 0ca35280437914664ede821d624fd00a0ff01d62. Summary will update on new commits.</sup>

### DIFF
--- a/src/js/base/fetch.js
+++ b/src/js/base/fetch.js
@@ -76,6 +76,14 @@ async function buildFetcher(url, options = {}) {
   return registerFetcher(baseUrl, fetch(url, options), controller);
 }
 
+async function handleUnauthorized(url, options = {}) {
+  return Radio.request('auth', 'handleUnauthorized', async() => {
+    // recoverAuth force-refreshes through AuthKit; the retry uses the SDK-cached fresh token.
+    return buildFetcher(url, options)
+      .finally(() => removeFetcher(url));
+  });
+}
+
 function getValue(value) {
   return encodeURIComponent(value ?? '');
 }
@@ -145,14 +153,18 @@ export default async(url, options) => {
   const fetcher = getActiveFetcher(url, options) || buildFetcher(url, options);
 
   return fetcher
-    .then(async response => {
+    .then(response => {
       removeFetcher(url);
 
-      if (!response.ok) {
-        if (response.status === 401) {
-          Radio.request('auth', 'logout');
-        }
+      if (response.status === 401) {
+        return handleUnauthorized(url, options)
+          .then(retryResponse => retryResponse || response);
+      }
 
+      return response;
+    })
+    .then(response => {
+      if (!response.ok) {
         if (response.status >= 400) {
           const contentType = String(response.headers.get('Content-Type'));
 
--- a/packages/care-ops-auth/AuthProvider.js
+++ b/packages/care-ops-auth/AuthProvider.js
@@ -68,6 +68,10 @@ export class AuthProvider {
     window.location = AuthProvider.PATH_LOGOUT;
   }
 
+  clearToken() {
+    this.token = null;
+  }
+
   setToken(tokenString) {
     this.token = tokenString;
   }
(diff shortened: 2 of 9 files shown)


src/js/base/fetch.js
```diff
@@ -145,14 +153,18 @@ export default async(url, options) => {
...
+    .then(response => {
       removeFetcher(url);
 
-      if (!response.ok) {
-        if (response.status === 401) {
-          Radio.request('auth', 'logout');
-        }
+      if (response.status === 401) {
+        return handleUnauthorized(url, options)
+          .then(retryResponse => retryResponse || response);
```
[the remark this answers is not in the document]
paulfalgout: auth0 not in use

src/js/base/fetch.test.js
nmajor25, line 1: Should these new tests be `cy.js` files?

src/js/auth/workos.test.js
nmajor25, line 1: Is this file intended to test all paths for workos? Like `getToken` happy path, `auth()` lifecycle, `onRedirectCallback()` scenarios. Looks like those are the main areas not covered.

nmajor25: This is the only place workos is tested, if I understand correctly. Integration and e2e tests bypass regular auth.

src/js/base/fetch.test.js
paulfalgout: this isn't a cypress test.. it's just one that AI is running when changing the file

src/js/auth/workos.test.js
paulfalgout: similar to above it's just an AI test file that the agent uses.  not in any CI or npm runner.

paulfalgout: I'm having it convert it to cypress test cause why not

nmajor25 (approved):

PR #1691, opened 2026-05-13
fix(ui): overflow ellipsis not working on custom filter buttons

nmajor25:
Closes FE-127

To fix text ellipsis overflow not working in the custom filters dropdown buttons:

Screenshot 2026-05-13 at 2 42 08 PM

---
## Summary by cubic
Fix overflow ellipsis on custom filter dropdown buttons so long labels truncate instead of wrapping, addressing FE-127. Removes the `flex` class from the button, which blocked `text-overflow: ellipsis` in production.

<sup>Written for commit 6d86ea60cd97c43ea89e439f3e29ccda9034845a. Summary will update on new commits.</sup>

### DIFF
--- a/src/js/views/patients/sidebar/filters/filters-sidebar_views.js
+++ b/src/js/views/patients/sidebar/filters/filters-sidebar_views.js
@@ -30,7 +30,7 @@ const CustomFilterDropList = Droplist.extend({
     return this.getView().$el.outerWidth();
   },
   viewOptions: {
-    className: 'button-secondary w-100 flex',
+    className: 'button-secondary w-100',
     template: hbs`{{ value }}{{#unless value}}{{ defaultText }}{{/unless}}`,
     templateContext: {
       defaultText: i18n.customFilterView.defaultText,


src/js/views/patients/sidebar/filters/filters-sidebar_views.js
```diff
@@ -30,7 +30,7 @@ const CustomFilterDropList = Droplist.extend({
     return this.getView().$el.outerWidth();
   },
   viewOptions: {
-    className: 'button-secondary w-100 flex',
+    className: 'button-secondary w-100',
```
nmajor25, line 33: This is an interesting one.

The overflow ellipsis works in development. But not in production.

It must be something with the css rule ordering vite does differently in dev vs. prod builds?

paulfalgout (approved):

src/js/views/patients/sidebar/filters/filters-sidebar_views.js
paulfalgout: And the totals continue to work regardless of length with this change?

nmajor25: Not sure what you mean, this is just on the button where the filter item's label/text is shown.

PR #1705, opened 2026-05-28
perf(entities): cache bootstrap and workspace fetches

paulfalgout:
Closes FE-143

## What
- Add a fail-soft IndexedDB response cache for JSON:API entity fetches.
- Expose the authenticated user id for cache partitioning and prune other-user cache entries after auth.
- Cache selected bootstrap and workspace entity-service fetches with user/workspace-aware keys and stale-while-revalidate behavior.

## Why
- Startup fetches repeatedly request mostly stable bootstrap/workspace data before the app can render.
- Cached entity responses let the app hydrate from local state quickly while the server remains the source of truth through background revalidation.

## Scope
- Impacts shared entity fetch behavior in `src/js/base/**` and selected entity services for bootstrap/workspace data.
- Updates WorkOS auth provider plumbing to expose user id for partitioning.
- Does not cache patient navigation data or action/patient PHI fetched after startup.

## Deployment Impact
- Normal app deployment required.
- No form bundle redeploy is required beyond the normal app asset deployment.
- Shared entity fetch behavior changes affect bootstrap/workspace startup paths globally.

## Testing
- `npx cypress run --component --spec "src/js/base/cache/*.cy.js,src/js/auth.cy.js"`
- `npm run eslint -- src/js/auth.js src/js/auth.cy.js src/js/base/cache/entity-cache.js src/js/base/cache/entity-cache.cy.js src/js/base/cache/idb.js src/js/base/cache/idb.cy.js src/js/base/cache/response-cache.cy.js src/js/base/collection.js src/js/base/entity-service.js src/js/base/fetch.js src/js/base/jsonapi-mixin.js src/js/base/model.js src/js/entities-service/clinicians.js src/js/entities-service/forms.js src/js/entities-service/programs.js src/js/entities-service/roles.js src/js/entities-service/settings.js src/js/entities-service/states.js src/js/entities-service/teams.js src/js/entities-service/widgets.js src/js/entities-service/workspaces.js src/js/services/bootstrap.js`
- Deployed current branch state to QA2 for manual smoke validation before final commit cleanup.

---
## Summary by cubic
Adds a fail-soft IndexedDB cache for JSON:API entity fetches to speed startup with stale-while-revalidate; cache is partitioned by user/workspace and cleaned up on auth/logout. Addresses FE-143.

- **New Features**
  - IndexedDB response cache with SWR and a 1-week TTL; fails soft if IDB is unavailable.
  - User/workspace-scoped keys; logout clears cache and auth prunes other-user entries.
  - Exposes `auth:getUserId` (sync Radio reply) and implements `getUserId` in `@roundingwell/care-ops-auth` WorkOS provider.
  - Adds `fetchCollectionCache`, `fetchModelCache`, and `fetchByCache`; enabled for bootstrap/workspace data (roles, settings, teams, widgets, workspaces, forms, states, programs by workspace, clinicians by workspace, and `clinicians/me`).
  - Uses canonical GET URLs for cache keys; stamps `__cached_ts` in the fetch layer for consistency.
  - Adds `idb` for IndexedDB access.

- **Bug Fixes**
  - Preload `/appconfig.json` with `crossorigin` to match credentials.
  - Ensure bootstrap always settles by wrapping `getUserId` and cache pruning in try/finally.
  - Do not memoize a failed IndexedDB open; retry later so transient failures don’t disable caching.
  - Stabilize Cypress cache/auth specs with proper Radio cleanup and command queue chaining.

<sup>Written for commit 9208fc2298b661e0bc7a62aff6db2ecd30190910. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/RoundingWell/app-frontend/pull/1705?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Persistent, user-scoped response caching with background refresh for collections and models; cache keys include user/workspace and exposed API to read cached results.
  * App exposes a user-id accessor and keeps a cached user id for scoping.

* **Chores**
  * Added IndexedDB-backed caching and idb dependency; preload now includes crossorigin.
  * JSON:API responses are stamped with cache timestamps for reliable replay.

* **Tests**
  * Added tests covering cache persistence, invalidation, partitioning, background refresh, fail‑soft IndexedDB, and auth cleanup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RoundingWell/app-frontend/pull/1705?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/js/auth.js
+++ b/src/js/auth.js
@@ -99,8 +99,12 @@ async function auth() {
 
   return new Promise(resolve => {
     agent.auth(async() => {
-      cachedUserId = await agent.getUserId();
-      await pruneOtherPartitions(cachedUserId);
+      try {
+        cachedUserId = await agent.getUserId();
+        await pruneOtherPartitions(cachedUserId);
+      } catch {
+        // swallow — boot without cache rather than crash auth.
+      }
       resolve();
     });
   });
--- a/src/js/auth.js
+++ b/src/js/auth.js
@@ -10,11 +10,14 @@ import { AuthProvider } from '@roundingwell/care-ops-auth/AuthProvider.js';
 
 import { addAction } from 'js/datadog';
 
+import { clearCache, pruneOtherPartitions } from 'js/base/cache/entity-cache';
+
 import 'scss/app-root.scss';
 
 import { LoginPromptView } from 'js/views/globals/prelogin/prelogin_views';
 
 let authAgent;
+let cachedUserId;
 
 function trackAuthEvent(name, context) {
   addAction(name, context);
@@ -75,17 +78,31 @@ async function logout() {
   return agent.logout();
 }
 
+function getUserId() {
+  return cachedUserId;
+}
+
 Radio.reply('auth', {
   handleUnauthorized,
   logout,
   setToken,
   getToken,
+  getUserId,
 });
 
 async function auth() {
+  if (location.pathname === AuthProvider.PATH_LOGOUT) {
+    await clearCache();
+  }
+
   const agent = await getAuthAgent();
+
   return new Promise(resolve => {
-    agent.auth(resolve);
+    agent.auth(async() => {
+      cachedUserId = await agent.getUserId();
+      await pruneOtherPartitions(cachedUserId);
+      resolve();
+    });
   });
 }
 
--- a/src/js/base/cache/entity-cache.js
+++ b/src/js/base/cache/entity-cache.js
@@ -1,0 +1,56 @@
+// Per-fetch JSON:API response cache. On hit, cached docs replay through
+// Backbone's parse pipeline so state is byte-identical to a fresh fetch.
+// SWR orchestration lives in BaseEntity.fetch*Cache.
+
+import dayjs from 'dayjs';
+
+import idb from './idb';
+
+export const ENTRY_VERSION = 1;
+// SWR refreshes on every online load, so TTL only bounds offline / cold-start staleness.
+const TTL_HOURS = 24;
+const STORE = 'entities';
+const KEY_SEPARATOR = '|';
+
+// Single source of truth for the cache key format.
+export function cacheKey(userId, workspaceId, url) {
+  return `${ userId }${ KEY_SEPARATOR }${ workspaceId || '' }${ KEY_SEPARATOR }${ url }`;
+}
+
+export async function getResponse(key) {
+  if (!key) return null;
+  const entry = await idb.get(STORE, key);
+  if (!entry) return null;
+  if (entry.entryVersion !== ENTRY_VERSION) return null;
+  if (dayjs.utc().diff(dayjs.utc(entry.ts), 'hour') >= TTL_HOURS) return null;
+  return entry.response;
+}
+
+export async function setResponse(key, response) {
+  if (!key || !response) return;
+  // Synchronous deep-snapshot: later mutations to `response` (e.g. by
+  // Backbone's parse pipeline) cannot leak into the cached form.
+  const snapshot = JSON.parse(JSON.stringify(response));
+  await idb.put(STORE, key, {
+    entryVersion: ENTRY_VERSION,
+    ts: dayjs.utc().format(),
+    response: snapshot,
+  });
+}
+
+// Wipe every cached response. Used on explicit logout.
+export async function clearCache() {
+  await idb.clear(STORE);
+}
+
+// On successful auth, drop any responses keyed for a different user.
+export async function pruneOtherPartitions(currentUserId) {
+  if (!currentUserId) return;
+  const prefix = `${ currentUserId }${ KEY_SEPARATOR }`;
+  const keys = await idb.keys(STORE);
+  await Promise.all(
+    keys
+      .filter(k => typeof k === 'string' && !k.startsWith(prefix))
+      .map(k => idb.delete(STORE, k)),
+  );
+}
--- a/src/js/base/cache/entity-cache.js
+++ b/src/js/base/cache/entity-cache.js
@@ -8,7 +8,7 @@ import idb from './idb';
 
 export const ENTRY_VERSION = 1;
 // SWR refreshes on every online load, so TTL only bounds offline / cold-start staleness.
-const TTL_HOURS = 24;
+const TTL_HOURS = 24 * 7;
 const STORE = 'entities';
 const KEY_SEPARATOR = '|';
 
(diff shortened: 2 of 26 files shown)


src/js/base/entity-service.js
```diff
@@ -35,25 +47,81 @@ export default MnObject.extend({
...
+    const userId = Radio.request('auth', 'getUserId');
+    if (!userId) return this.fetchBy(url, options);
+
+    const { cacheScope, ...requestOptions } = options;
+    const dbKey = cacheKey(userId, this._cacheWorkspaceId(cacheScope), getUrl(url, requestOptions.data));
+    const fetchOptions = { ...requestOptions, _cacheKey: dbKey };
+
+    return getResponse(dbKey).then(cached => {
+      if (cached) {
+        const model = new this.Entity.Model({ id: cached.data.id });
```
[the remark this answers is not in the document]
paulfalgout: it's a backbone.store model

src/js/entities-service/clinicians.js
```diff
@@ -27,7 +27,7 @@ const Entity = BaseEntity.extend({
     const url = `/api/workspaces/${ workspaceId }/relationships/clinicians`;
     const workspace = Radio.request('entities', 'workspaces:model', workspaceId);
 
-    return this.fetchCollection({ url })
+    return this.fetchCollectionCache({ url })
```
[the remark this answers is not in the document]
paulfalgout: acceptable

src/js/entities-service/programs.js
```diff
@@ -16,7 +16,7 @@ const Entity = BaseEntity.extend({
   },
   fetchProgramsByWorkspace(workspaceId) {
     const url = `/api/workspaces/${ workspaceId }/relationships/programs`;
-    return this.fetchCollection({ url });
+    return this.fetchCollectionCache({ url });
```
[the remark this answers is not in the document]
paulfalgout: Verified the scenario is real but observable impact is zero in practice: users in this app effectively never get new workspaces mid-session, and the existing flow already requires a refresh for workspace-membership changes to surface. The SWR background refresh writes the fresh response back to IDB, so a reload picks up new workspaces the same as today. Accepting as a documented SWR limitation; no code change.

src/js/base/entity-service.js
```diff
@@ -35,25 +47,81 @@ export default MnObject.extend({
 
     return collection.fetch(options);
   },
+  fetchCollectionCache(options = {}) {
```
nmajor25, line 50: Something claude found, seems like an edge case. So not sure if it's worth exploring:

> Background SWR fetch swallows 403 silently
> 
> On a cache hit, the background refresh is fired as collection.fetch(fetchOptions).catch(() => {}). 
> 
> For server errors (5xx) this is fine — fetch.js triggers unknownError via Radio before the promise chain, so the error view shows regardless.
> 
> But for a JSON:API 403, neither of those paths fires. A user whose account is revoked mid session would keep browsing on stale cached data with no error, no redirect, nothing. The cache TTL is a week, so this could persist across many visits.
>
>Worth at least checking response.status in the background .catch and triggering unknownError (or forcing a logout) on 401/403, rather than catching everything.

src/js/base/model.js
nmajor25, line 119: Finding from claude, seems like a nitpick.

> isCached() name is misleading
>
> The method now correctly means "has this model been populated from a fetch" (true for both network and IDB responses, false for stubs). That's actually an improvement over develop where the set() override made it always true — so the intent is right.
>
> But the name reads as "is this model currently serving cached/stale data," which is a different thing. A future reader could plausibly branch on it for staleness detection and be wrong. Something like isPopulated() or hasData() would be unambiguous.

src/js/base/entity-service.js
```diff
@@ -35,25 +47,81 @@ export default MnObject.extend({
...
+
+    return getResponse(dbKey).then(cached => {
+      if (cached) {
+        collection.set(cached, { parse: true });
+        collection.fetch(fetchOptions).catch(() => {});
+        return collection;
+      }
+      return collection.fetch(fetchOptions);
+    });
+  },
```
nmajor25, line 70: Another claude finding, not sure if legit.

> SWR cache replay reverts in-session writes on workspace switch
> 
> On switching back to workspace A, fetchCollectionCache replays the IDB snapshot from when the user last loaded workspace A — before any writes made during that visit. A program created in workspace A will be missing from the list for however long the background refresh takes (~200–500ms on a normal connection). This isn't a race — timing of the switch doesn't matter. Any normal workspace switch after a write produces it.
>
> Whether this needs write-through to IDB after model.save() is a product call, but it's worth being explicit about the trade-off.

src/js/base/cache/entity-cache.js
```diff
@@ -0,0 +1,56 @@
...
+// Single source of truth for the cache key format.
+export function cacheKey(userId, workspaceId, url) {
+  return `${ userId }${ KEY_SEPARATOR }${ workspaceId || '' }${ KEY_SEPARATOR }${ url }`;
+}
+
+export async function getResponse(key) {
+  if (!key) return null;
+  const entry = await idb.get(STORE, key);
+  if (!entry) return null;
+  if (entry.entryVersion !== ENTRY_VERSION) return null;
```
nmajor25: Small thing found by claude:

> Version-mismatched entries are never cleaned up — they become permanent dead weight in IDB until the user explicitly logs out (which calls clearCache()). Could opportunistically delete here:
>
> if (entry.entryVersion !== ENTRY_VERSION) {
>    idb.delete(STORE, key);
>    return null;
> }
>
> Low priority since the entry count is small and logout clears everything, but it's a one-liner.

src/js/auth.js
```diff
@@ -75,17 +78,34 @@ async function logout() {
...
+
   const agent = await getAuthAgent();
+
   return new Promise(resolve => {
-    agent.auth(resolve);
+    agent.auth(async() => {
+      try {
+        cachedUserId = await agent.getUserId();
+        await pruneOtherPartitions(cachedUserId);
+      } finally {
```
nmajor25: Claude finding:

> No catch here means if getUserId() throws, the async callback's returned promise rejects unhandled — agent.auth doesn't await it. 
>
> The app boots fine regardless (resolve() still fires), but it'd show up as a spurious unhandled rejection in the console and potentially as a false Datadog alert. 
>
> Since the intent is already "swallow and boot without cache," a catch {} just makes that explicit.

nmajor25 (commented):
Looks good to me, ran it through claude and it found some small things.

src/js/base/entity-service.js
paulfalgout: they'll hit a 403 on any of the page loads after the bootstrap.. so like any worklist or patient or whatever

src/js/base/model.js
paulfalgout: I agree, but for a different day.  it was already called this

src/js/base/entity-service.js
paulfalgout: this is technically true, but we don't make programs very often... and the program lists have to do with filtering clinicians.. I don't think it can cause any real problem, and it's really no different than what happens already.

nmajor25 (approved):

PR #1715, opened 2026-06-09
feat(forms): replace draft prompt with persistent draft status indicator

nmajor25:
Closes FE-150

Summary of changes:

- Always load the user's draft when opening a form
- Replace the `StoredSubmissionView` interstitial prompt (full-page "you have unsubmitted work" screen with Load/Discard buttons, [screenshot](https://github.com/user-attachments/assets/b94bac9d-c618-46cd-ac8a-bb8995bb46f4)) with a persistent shield icon button in the form controls bar
- Draft status is now always visible while editing — clicking the shield opens a dropdown showing when the draft was last saved and a "Discard draft..." option
- Discarding from the dropdown shows a confirmation modal before clearing the draft (same as before)

Screenshot 2026-06-09 at 2 20 22 PM

Screenshot 2026-06-09 at 2 20 28 PM

---
## Summary by cubic
Replaced the full-screen draft prompt with a non-blocking shield button that shows live draft status and a discard option. Drafts now auto-load so users can start editing right away (FE-150).

- **New Features**
  - Draft status button (shield) shows “Last saved …” and auto-refreshes every 45s.
  - Click opens a dropdown with “Discard draft...” and a confirm modal.
  - Drafts auto-load; removed the interstitial load/discard screen.
  - Only shows when a draft exists and the form is editable (not read-only/locked, no response selected).

- **Refactors**
  - Removed `StoredSubmissionView` and `LastUpdatedView`; added `DraftStatusView` and `DraftMenuView` with new `data-draft-status-region`.
  - Updated i18n copy/keys (“Your draft is saved automatically.” / “Last saved ...”); removed obsolete keys.
  - CSS cleanup: removed prompt styles; added draft menu styles and new zero-margin utilities.
  - Updated integration tests for auto-load and the new dropdown flow.

<sup>Written for commit 68fc7799488d4d8ab661e8928f3d0c4e65e36d92. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/RoundingWell/app-frontend/pull/1715?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **New Features**
  * Added draft status menu with "Last saved" timestamp display that auto-refreshes every 45 seconds
  * Added draft discard functionality with confirmation modal

* **Style**
  * Updated styling for draft status menu and related form controls
  * Added zero-value margin utility classes

* **Tests**
  * Updated integration tests to verify draft status UI interactions and timestamp updates

### DIFF
--- a/src/js/apps/forms/form/form-patient_app.js
+++ b/src/js/apps/forms/form/form-patient_app.js
@@ -17,7 +17,7 @@ import {
   StatusView,
   ReadOnlyView,
   SaveView,
-  LastUpdatedView,
+  DraftStatusView,
 } from 'js/views/forms/form/form_views';
 
 export default App.extend({
@@ -116,11 +116,12 @@ export default App.extend({
     this.showFormSave();
   },
   onFormServiceUpdateSubmission(updated) {
-    this.showLastUpdated(updated);
+    this.setState({ updated });
   },
   stateEvents: {
     'change': 'onChangeState',
     'change:isExpanded': 'showSidebar',
+    'change:updated': 'onChangeDraftStatus',
   },
   onChangeState(state) {
     localStore.set(`form-state_${ this.currentUser.id }`, state.pick('isExpanded', 'saveButtonType'));
@@ -141,7 +142,16 @@ export default App.extend({
     this.toggleState('isExpanded');
   },
   showContent() {
+    if (!this.isReadOnly) this.loadDraftStatus();
     this.showForm();
+  },
+  async loadDraftStatus() {
+    const { updated } = await Radio.request(`form${ this.form.id }`, 'get:storedSubmission');
+
+    /* istanbul ignore if: difficult to force stale async render */
+    if (this.isDestroyed()) return;
+
+    this.setState({ updated });
   },
   showForm(responseId) {
     this.showChildView('form', new IframeView({
@@ -171,10 +181,26 @@ export default App.extend({
   showReadOnly() {
     this.showChildView('formAction', new ReadOnlyView());
   },
-  showLastUpdated(updated) {
-    const lastUpdatedView = new LastUpdatedView({ updated });
+  onChangeDraftStatus() {
+    const updated = this.getState('updated');
 
-    this.showChildView('formUpdated', lastUpdatedView);
+    if (!updated) {
+      this.getRegion('draftStatus').empty();
+      return;
+    }
+
+    if (this.getRegion('draftStatus').hasView()) return;
+
+    const draftStatusView = new DraftStatusView({ model: this.getState() });
+    this.showChildView('draftStatus', draftStatusView);
+
+    this.listenTo(draftStatusView, {
+      async 'discard:submission'() {
+        await Radio.request(`form${ this.form.id }`, 'clear:storedSubmission');
+        this.showForm();
+        this.showFormActions();
+      },
+    });
   },
   showFormSaveDisabled() {
     if (this.isSubmitHidden) return;
--- a/src/js/apps/forms/form/form-patient_app.js
+++ b/src/js/apps/forms/form/form-patient_app.js
@@ -13,7 +13,6 @@ import WidgetsHeaderApp from 'js/apps/forms/widgets/widgets_header_app';
 import {
   LayoutView,
   IframeView,
-  StoredSubmissionView,
   FormStateActionsView,
   StatusView,
   ReadOnlyView,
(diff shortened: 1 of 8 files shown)


src/scss/core/_space.scss
```diff
@@ -2,25 +2,29 @@
...
 
 // Aligns with small button text
+.u-margin--t-0 { margin-top: 0 !important; }
 .u-margin--t-4 { margin-top: 4px !important; }
 .u-margin--t-8 { margin-top: 8px !important; }
 .u-margin--t-16 { margin-top: 16px !important; }
 .u-margin--t-24 { margin-top: 24px !important; }
 .u-margin--t-32 { margin-top: 32px !important; }
 
+.u-margin--r-0 { margin-right: 0 !important; }
```
nmajor25, line 12: I needed just this one for this PR, but figured I should add the others as well for consistency.

src/js/apps/forms/form/form_app.js
```diff
@@ -307,10 +316,27 @@ export default App.extend({
...
+
+    this.showChildView('draftStatus', draftStatusView);
+
+    this.listenTo(draftStatusView, {
+      async 'discard:submission'() {
+        await Radio.request(`form${ this.form.id }`, 'clear:storedSubmission');
+        this.showForm();
+        this.showFormActions();
+      },
+    });
```
nmajor25, line 339: `showLastUpdated(updated)` guard breakdown:

### Guard 1: `if (!updated)` — clear the shield

`clearStoredSubmission()` fires `update:submission` with no argument after a successful submit or discard. No draft means no shield — the region is emptied and `_draftStatusView` is nulled so guard 2 doesn't read a stale reference on the next call.

### Guard 2: `if (this._draftStatusView && this._draftStatusView.getState('isActive')) return` — don't replace an open dropdown

`update:submission` fires at load time (via `fetchFormData` → `getStoredSubmission`) and again whenever the user types (debounced 15s), so it can arrive while the shield is already showing. Replacing a `Droplist` component mid-interaction via `showChildView` destroys it, which strips the `click:discard` event binding and causes a TypeError in `onChangeIsActive` (`getView()` is undefined on a destroyed component). The guard skips replacement while the menu is open.

`_draftStatusView` is held on the app because `region.currentView` is the shield button View (the component's `ViewClass` instance), not the `DraftStatusView` component — there's no other path to `getState('isActive')`.

nmajor25:
I've got this up on my dev sandbox, it works well during my testing.

src/js/apps/forms/form/form-patient_app.js
```diff
@@ -141,38 +140,14 @@ export default App.extend({
...
-        },
-        async 'discard:submission'() {
-          await Radio.request(`form${ this.form.id }`, 'clear:storedSubmission');
-          this.showForm();
-          this.showFormActions();
-        },
-      });
-
-      return;
+      if (updated) this.showDraftStatus(updated);
```
paulfalgout: I think we could move this into a separate async function that doesn't block the form render.  so showContent doesn't need to be async.. also I think we can drop the `if (updated)` it's not saving much that isn't already handled in the method corrrect?

src/js/apps/forms/form/form_app.js
```diff
@@ -342,10 +316,30 @@ export default App.extend({
...
-    const lastUpdatedView = new LastUpdatedView({ updated });
+  showDraftStatus(updated) {
+    if (!updated) {
+      this.getRegion('draftStatus').empty();
+      this._draftStatusView = null;
+      return;
+    }
+
+    if (this._draftStatusView && this._draftStatusView.getState('isActive')) {
+      this._draftStatusView.updated = updated;
```
paulfalgout: I wonder if we should have this app's state model deal with fetching the updated and then we can pass the state model down to the view rather than attaching the view and updating a view variable... it can then render when the data updates as well which deals with copilots issue.  Really you could probably just always render the view, but if in the view !this.model.get('updated) you could render an empty template?
then modelEvents: { 'change:updated': 'renderUpdated' } handles when it has a value or doesn't.  Then the state model can have a method added that clears the data and calls the clearSubmission service request?  renderUpdated would also check as to whether it should only update the timestamp is the option isactive.. or if it's show/hiding the whole thing with a full render.

src/js/views/forms/form/form_views.js
```diff
@@ -290,35 +251,64 @@ const SaveButtonTypeDroplist = Droplist.extend({
...
-    return {
-      updated: this.updated,
-    };
+    return { updated: this.getOption('updated') };
+  },
+  triggers: {
+    'click .js-discard': 'click:discard',
+  },
+  initialize() {
+    this.renderInterval = setInterval(() => {
```
paulfalgout, line 271: I think if we move updated to the state model and pass it down, then we're just rendering things when the data on the model changes

paulfalgout (changes requested):

src/js/apps/forms/form/form_app.js
nmajor25: Yes this works well, like it much better 👍

src/js/apps/forms/form/form-patient_app.js
nmajor25: Makes sense, updated to that 👍

src/js/views/forms/form/form_views.js
nmajor25: 👍

src/js/views/forms/form/form_views.js
```diff
@@ -290,38 +251,67 @@ const SaveButtonTypeDroplist = Droplist.extend({
...
+      headingText: i18n.draftStatusView.discardModal.headingText,
+      submitText: i18n.draftStatusView.discardModal.submitText,
+      buttonClass: 'button--red',
+      onSubmit: () => {
+        modal.destroy();
+        this.triggerMethod('discard:submission');
+      },
+    });
+  },
+});
```
nmajor25, line 313: The coverage drop was on the `onBeforeDestroy()`:

```js
  onBeforeDestroy() {
    if (this.getState('isActive')) this.popRegion.empty();
  },
```

I removed it completely. It was there to protect against the droplist menu potentially staying open when the shield icon is removed from the UI. And getting into a weird UI state. But I don't think it's possible for that to happen.

nmajor25: If it could happen, it would be on a `clearStoredSubmission` event. Which happens when the user discards the submission via the confirmation modal (droplist menu is always closed in that scenario). Or if the form is submitted (even from within the form), which the droplist menu will always be closed in that situations.

paulfalgout (approved):

PR #1724, opened 2026-06-15
fix(cache): keep blocked IndexedDB upgrades from hanging startup

paulfalgout:
Closes FE-166

## What
- Adds a timeout around the memoized IndexedDB open path so blocked cache opens resolve to no-cache mode.
- Closes an existing IDB connection when it blocks a newer version upgrade.
- Adds component coverage for unavailable IndexedDB, blocked opens, and fail-fast behavior after timeout.

## Why
One environment could hang during startup when `idb.openDB()` never settled during a blocked version upgrade.

## Scope
Impacts the shared `src/js/base/cache/idb.js` wrapper used by entity response caching and form draft storage. No unrelated cache key, entity-service, or draft behavior was changed.

## Deployment Impact
Normal app frontend deployment only. No form redeploy or separate bundle redeploy is required.

## Testing
- `npx cypress run --component --spec src/js/base/cache/idb.cy.js`

---
## Summary by cubic
Prevents startup and logout hangs from a blocked `idb.openDB()` upgrade by capping the open at 3s and falling back to no-cache. Fixes FE-166.

- **Bug Fixes**
  - Memoizes a single open with a 3s timeout; blocked/failed or sync-unavailable opens resolve to null, and later calls fail fast without re-waiting.
  - Adds `blocking()` to close the current connection and clear the memo; `terminated()` drops the memo so the next read reopens.
  - Closes any connection that completes after the timeout so it can’t linger or block upgrades; expands tests for unavailable IndexedDB, blocked opens, fail-fast, late-open close, and `versionchange`/`close` using `cy.clock()`/`cy.tick()`.

<sup>Written for commit 017b5472b948c3c6dc8ce831100693233f64abe6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/RoundingWell/app-frontend/pull/1724?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **Bug Fixes**
  * Improved IndexedDB cache initialization with a bounded wait, preventing startup deadlocks when a version upgrade is blocked in another tab.
  * Enhanced cache connection lifecycle handling: timed-out or “orphaned” connections are closed, and later reads automatically recover by reopening as needed.

* **Tests**
  * Added Cypress coverage for blocked IndexedDB opens, timeout-based fail-fast memoization, orphaned connection cleanup after late completion, and versionchange blocking/termination behaviors.

### DIFF
--- a/src/js/base/cache/idb.cy.js
+++ b/src/js/base/cache/idb.cy.js
@@ -1,5 +1,68 @@
 import idb from './idb';
 
+// A stand-in for an IDBOpenDBRequest whose open never settles — what a blocked
+// version upgrade (another tab holding an older version open) looks like to idb.
+function neverSettlingRequest() {
+  const request = {
+    then() {
+      return request;
+    },
+    catch() {
+      return request;
+    },
+    addEventListener() {},
+    removeEventListener() {},
+  };
+  return request;
+}
+
+// Like the above, but the open can be completed later via resolveWith(db).
+// Backed by a real promise so every subscriber (idb internals, the orphan
+// guard, Promise.race) resolves correctly regardless of when it subscribed.
+function pendingRequest() {
+  let resolveOpen;
+  const opened = new Promise(resolve => {
+    resolveOpen = resolve;
+  });
+  return {
+    then(onFulfilled, onRejected) {
+      return opened.then(onFulfilled, onRejected);
+    },
+    catch(onRejected) {
+      return opened.catch(onRejected);
+    },
+    addEventListener() {},
+    removeEventListener() {},
+    resolveWith(db) {
+      resolveOpen(db);
+    },
+  };
+}
+
+// A stand-in for the live IDBDatabase the idb library hands back after a
+// successful open. It records the versionchange/close listeners idb wires up so
+// a test can fire them and exercise the blocking()/terminated() handlers.
+function openedDatabase() {
+  const handlers = {};
+  const close = cy.stub();
+  const db = {
+    close,
+    get() {
+      return Promise.resolve(undefined);
+    },
+    addEventListener(type, handler) {
+      (handlers[type] || (handlers[type] = [])).push(handler);
+    },
+  };
+  return {
+    db,
+    close,
+    fire(type) {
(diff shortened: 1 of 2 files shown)


src/js/base/cache/idb.js
```diff
@@ -4,29 +4,75 @@ const DB_NAME = 'careops-cache';
...
+  // orphaned connection so it cannot linger or block a later upgrade.
+  opening.then(db => {
+    if (timedOut && db) db.close();
+  }).catch(() => {});
+
+  dbPromise = Promise.race([opening, timeout])
+    .catch(() => null)
+    .then(db => {
+      clearTimeout(timer);
+      return db || null;
```
nmajor25: Nitpick: think this could just be `return db`. Since `.catch(() => null)` already ensures the db is the db object or null by the time it arrives to the `.then()`?

src/js/base/cache/idb.cy.js
nmajor25, line 1: Are `blocking()` and `terminated()` tested directly in this file? If I delete them the component tests still pass.

From claude:

> `blocking()` and `terminated()` have no test coverage. Both handlers are wired on the live `db` object after a successful open (`db.addEventListener(...)` inside the `idb` library), so they're unreachable in the current specs since `indexedDB.open` is stubbed before the DB ever opens. Deleting either handler, or removing the `dbPromise = undefined` assignment, would pass all specs.
>
> `blocking()` is the cross-tab deadlock fix named in the commit message — worth a spec. The existing mock pattern extends cleanly: return a fake db from `pendingRequest.resolveWith()` that captures `addEventListener` listeners, then fire `versionchange` manually and assert that a subsequent `idb.get()` re-opens rather than returning the stale memo.

nmajor25 (approved):

PR #1725, opened 2026-06-15
feat(forms): UI updates to draft status button & droplist

nmajor25:
FE-160

Follow up work to https://github.com/RoundingWell/app-frontend/pull/1718.

---
## Summary by cubic
Adds a “See draft status” tooltip to the draft status button and refines the draft droplist spacing and discard hover state for clarity. Addresses FE-160 by making draft status easier to find and aligning menu styles.

- **New Features**
  - Tooltip: shows “See draft status” on hover with a slight delay; auto-hides while the droplist is open. Added i18n key at `draftStatusView.tooltip`.
  - Draft menu UI: adjusted vertical and side padding for info/saved text; discard button hover now uses a blue background with white text.

<sup>Written for commit 33f80a6968e49c895a91a80980b2dca347ab5ed7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/RoundingWell/app-frontend/pull/1725?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a “See draft status” tooltip for the draft status control, including delayed display and proper dismissal behavior.

* **Style**
  * Refined draft menu spacing with updated padding for menu info and saved-state areas.
  * Updated discard button hover styling to use a blue background with white text.

* **Tests**
  * Improved integration coverage to validate the tooltip timing, text, dismissal, and draft menu content.

### DIFF
--- a/src/js/views/forms/form/form_views.js
+++ b/src/js/views/forms/form/form_views.js
@@ -286,6 +286,31 @@ const DraftStatusView = Droplist.extend({
   initialize({ model }) {
     this.model = model;
   },
+  onShow() {
+    this._showTooltip();
+
+    this.listenTo(this.getState(), 'change:isActive', (state, isActive) => {
+      if (isActive) {
+        this._tooltip.destroy();
+        this.getView().$el.off('.tooltip');
+        return;
+      }
+
+      this._showTooltip();
+    });
+  },
+  _showTooltip() {
+    const view = this.getView();
+
+    view.$el.off('.tooltip');
+    this._tooltip = new Tooltip({
+      message: i18n.draftStatusView.tooltip,
+      uiView: view,
+      ui: view.$el,
+      orientation: 'vertical',
+      shouldDelay: true,
+    });
+  },
   showPicklist() {
     const menuView = new DraftMenuView({ model: this.model });
 
--- a/src/js/i18n/en-US.yml
+++ b/src/js/i18n/en-US.yml
@@ -118,6 +118,7 @@ careOptsFrontend:
           currentVersionButton: Back to Current Version
         draftStatusView:
           discardDraft: Discard draft...
+          tooltip: See draft status
           discardModal:
             bodyText: Discarding your unsubmitted work on this form will load the latest data for this patient, but you'll lose your unsubmitted work.
             headingText: Are you sure?
--- a/src/js/views/forms/form/form.scss
+++ b/src/js/views/forms/form/form.scss
@@ -177,17 +177,19 @@
   border-radius: 4px;
   box-shadow: 0 2px 8px rgb(0 0 0 / 12%);
   font-size: 12px;
-  padding: 12px 16px;
+  padding: 12px 0;
   white-space: nowrap;
 }
 
 .form__draft-menu-info {
   color: #{$black-40};
+  padding: 0 16px;
 }
 
 .form__draft-menu-saved {
   font-size: 14px;
   margin-top: 8px;
+  padding: 0 16px;
 }
 
 .form__draft-menu-discard-button {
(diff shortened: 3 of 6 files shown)


src/js/views/forms/form/form_views.js
```diff
@@ -286,6 +286,31 @@ const DraftStatusView = Droplist.extend({
...
+
+    view.$el.off('.tooltip');
+    this._tooltip = new Tooltip({
+      message: i18n.draftStatusView.tooltip,
+      uiView: view,
+      ui: view.$el,
+      orientation: 'vertical',
+      shouldDelay: true,
+    });
+  },
```
nmajor25, line 313: Shows tooltip only when droplist is open.

https://github.com/user-attachments/assets/4420de11-9013-494c-b3d0-82540c6708ba

nmajor25, line 313: One issue with this is the tooltips on the other buttons can show when the droplist is open:

https://github.com/user-attachments/assets/af43da07-6873-424e-85c7-cfc83309e125

nmajor25: Which I guess is already an existing issue with the submit button preference dropdown:

Image

nmajor25: I'm not sure if it's worth it. But one way to solve it would be to add a `this.model.isDroplistOpen` or something similar. And add a guard to each tooltip to stop them from showing.

Example:

```js
renderSidebarTooltip() {
  if (this.model.get('isDroplistOpen')) return;

  . . .
}
```

paulfalgout: I'm not sure we want to fix it this broadly.  this is more about tooltips in proximity to the dropdown.  needs more research at best.. another day.

src/js/views/forms/form/form_views.js
```diff
@@ -286,6 +286,31 @@ const DraftStatusView = Droplist.extend({
...
+        return;
+      }
+
+      this._showTooltip();
+    });
+  },
+  _showTooltip() {
+    const view = this.getView();
+
+    view.$el.off('.tooltip');
```
paulfalgout, line 305: I think maybe we should update the base tooltip component:

```js
onBeforeDestroy() {
  clearTimeout(this.delayTimeout);
  if (this.ui) this.ui.off('.tooltip');
},
```

paulfalgout (changes requested):

src/js/views/forms/form/form_views.js
nmajor25: Can't get this to work. It breaks other tooltips in the app.

Claude thinks it's a timing issue:

> Marionette's `triggerMethod('render')` calls `onRender` before firing the `render` event, so a new tooltip's listeners are already attached to the element by the time the old tooltip's `onBeforeDestroy` runs. With the namespace approach, the new tooltip's listeners get wiped too. The specific-handler approach should avoid that, but something in `Component.apply` (called after `setListeners` in the constructor) appears to interfere with stored bound references as instance properties.

src/js/views/forms/form/form_views.js
nmajor25: I should have named the variable different. It would just be if the draft status or submit droplists are shown. But ya for another day.

paulfalgout (approved):

PR #1743, opened 2026-07-27
refactor(i18n): split locale files by domain

paulfalgout:
Closes FE-185

## What

- Split the monolithic `en-US.yml` locale into domain-owned files.
- Compose the domain files into the same runtime locale object and reject duplicate top-level namespaces.
- Preserve locale metadata for PhraseApp uploads and document locale ownership rules.

## Why

This is the first reviewable foundation slice of the action-page work. It isolates locale organization from the later patient application, routing, and view changes.

## Scope

Impacts only `src/js/i18n/**`. Translation keys and runtime locale data are unchanged. Patient application structure, routing, styling, and views are intentionally excluded. This PR targets `feature/action-page-review`, not `develop`.

## Deployment Impact

No form or form-bundle redeploy is needed. This partial review PR does not deploy or merge into `develop`; the complete feature will follow the normal app deployment path later.

## Testing

- Passed locally: `npm run lint`
- Passed locally: `npx cypress run --component --spec src/js/i18n/index.component.cy.js`
- Passed locally: exact `assert.deepStrictEqual` comparison of the recomposed locale object against `feature/action-page-review`
- Passed locally: final action-page runtime locale comparison against the pre-review branch
- Passed locally: `git diff --check origin/feature/action-page-review..HEAD`
- Full E2E, component, and coverage jobs are running in CircleCI for the current commit.


paulfalgout:
@coderabbitai review

src/js/i18n/index.js
```diff
@@ -4,11 +4,29 @@ import Handlebars from 'handlebars/dist/cjs/handlebars';
...
+const localEnUs = {
+  locales: 'en-US',
+  [localeKey]: Object.assign(
+    {},
+    cliniciansEnUs[localeKey],
+    dashboardsEnUs[localeKey],
+    globalsEnUs[localeKey],
+    patientsEnUs[localeKey],
+    programsEnUs[localeKey],
+    sharedEnUs[localeKey],
```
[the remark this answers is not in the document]
paulfalgout: No change needed. I verified the parsed top-level key sets: clinicians.yml contributes only `clinicians`, programs.yml only `programs`, and shared.yml contributes `components`, `shared`, and `regions`; all six files have zero overlapping top-level keys. The `shared` and `sidebar` nodes mentioned here are nested beneath their owning domain roots, so Object.assign does not overwrite them. I also recomposed all six YAML files and used `assert.deepStrictEqual` against the original locale object from `feature/action-page-review`; the comparison passed exactly.

src/js/i18n/index.js
```diff
@@ -4,11 +4,29 @@ import Handlebars from 'handlebars/dist/cjs/handlebars';
...
+  ),
```
paulfalgout: **The shallow merge makes "one namespace, one file" a silent, unguarded invariant.**

`Object.assign` is one level deep. If two locale files ever declare the same key under `careOptsFrontend` — someone adds `components:` to `clinicians.yml`, or splits `forms` out of `patients.yml` without deleting it there — one entire subtree is dropped silently. No build error, no lint error, just missing strings at runtime.

It's more likely than it looks, because the file↔namespace mapping is already not 1:1: `patients.yml` owns `forms` + `patients`, `shared.yml` owns `components` + `shared` + `regions`. A contributor can't infer the rule from the layout, and nothing in the PR encodes or documents it.

Minimum: state the invariant in the README. Better: a dev-only duplicate-key assertion — there's precedent in this exact file at the `/* istanbul ignore if: dev use only */` PhraseApp block below, so it stays consistent with the E2E-only coverage policy.

src/js/i18n/index.js
```diff
@@ -4,11 +4,29 @@ import Handlebars from 'handlebars/dist/cjs/handlebars';
...
+import globalsEnUs from './en-US/globals.yml';
+import patientsEnUs from './en-US/patients.yml';
+import programsEnUs from './en-US/programs.yml';
+import sharedEnUs from './en-US/shared.yml';
 import { registerWith } from './intl';
 
 const localeKey = 'careOptsFrontend';
 
+const localEnUs = {
+  locales: 'en-US',
```
paulfalgout: **`locales: en-US` moved out of the uploaded YAML into JS — check the PhraseApp side.**

Runtime-neutral, and in fact unread either way: `intl.js` reads `intlData.locales` off the exported `intl`, which is only the `careOptsFrontend` subtree — `locales` was never inside it. (The `key === 'locales'` guard in `phraseAppMessages` below is likewise pre-existing dead code, not a regression from this PR.)

The open question is PhraseApp: this key previously shipped in the uploaded file and now ships in no file. If PhraseApp keys off it, that's a behavior change on the translation side; if not, it shows up as a removed key on the next upload. The README's upload rewrite doesn't mention either outcome. Worth confirming before this branch hits translate.roundingwell.com — I can't verify PhraseApp's behavior from here.

src/js/i18n/en-US/patients.yml
```diff
@@ -1,101 +1,5 @@
...
-      emptyFindInListView:
-        noResults: No results match your Find in List search
-      emptyView: No Dashboards
-      layoutView:
-        nameHeader: Name
-        title: Dashboards
-    dashboardViews:
-      contextTrailView:
-        contextBackBtn: Back to List
   forms:
```
paulfalgout: **`forms` is an app root but has no locale file.**

`src/js/apps/` roots are `clinicians, dashboards, forms, globals, patients, programs`. The README says the files "correspond to the app ownership roots," but `careOptsFrontend.forms` is a *sibling* top-level namespace — not nested under `patients` — living in `patients.yml`. So the stated rule is inaccurate for one of six roots, and the one file whose name doesn't match its namespace is also the one most likely to collide later (see the `Object.assign` comment on `index.js`).

Splitting `forms.yml` costs zero key churn, restores 1:1 file↔namespace, and matches this PR's own rule that moving a module doesn't justify renaming keys. If keeping `forms` here is deliberate, the README should say so explicitly rather than claim root correspondence.

src/js/i18n/README.md
```diff
@@ -4,17 +4,25 @@
...
-is _doing_ in context, not what the text necessarily _says_. IE perhaps the text string we are
-replacing is "Save". Rather than name the context name it could be saveLabel (or just label
-if the View name is particularly descriptive.) This way the key is not ambiguous as far as,
-"Is it for the button or the label?" etc.
+under [`src/js/i18n/en-US`](https://github.com/RoundingWell/app-frontend/tree/develop/src/js/i18n/en-US).
+
+Locale files are divided by stable production ownership:
+
+- `clinicians.yml`, `dashboards.yml`, `globals.yml`, `patients.yml`, and `programs.yml`
+  correspond to the app ownership roots.
```
paulfalgout: This claim doesn't hold for `forms`: it's an app root under `src/js/apps/forms` with its own top-level `careOptsFrontend.forms` namespace, but no `forms.yml`. Either add the file or reword this to name the exception (see the comment on `patients.yml`).

src/js/i18n/README.md
```diff
@@ -4,17 +4,25 @@
...
-The key naming strategy currently under consideration is:
-`view.directory.fileViews.viewName.stringContextName` where the first part is the
-the directory of the view file or template delimited by "." followed by a camelCased version
-of the file in that directory; followed by a camelCased version of the ViewClass (if for a view);
-followed by the camelCased string contextual name. We should attempt to describe what the string
-is _doing_ in context, not what the text necessarily _says_. IE perhaps the text string we are
-replacing is "Save". Rather than name the context name it could be saveLabel (or just label
-if the View name is particularly descriptive.) This way the key is not ambiguous as far as,
-"Is it for the button or the label?" etc.
+under [`src/js/i18n/en-US`](https://github.com/RoundingWell/app-frontend/tree/develop/src/js/i18n/en-US).
```
paulfalgout, line 7: Nit: this link points at `tree/develop/...`, which 404s until the work reaches `develop`.

paulfalgout (commented):
Reviewed the split. Verified independently: the recomposed object is `deepStrictEqual` to `origin/feature/action-page-review:src/js/i18n/en-US.yml`, and no two files currently share a top-level namespace under `careOptsFrontend`. Mechanically clean.

Two things worth fixing before more slices stack on this layout (both get harder to correct once locale edits land on the split files), one question, and three nits — all inline.

Non-issues: no perf change (six static imports collapse into the same chunk), no security surface, and `index.js` composition runs on every import so E2E covers the happy path. What's absent is a guard against key *loss*, which is the first comment below.

src/js/i18n/en-US/clinicians.yml
```diff
@@ -0,0 +1,70 @@
...
+        workspaces: Workspaces
+      clinicianModalViews:
+        clinicianModal:
+          email: clinician@organization.com
+          headingText: Add Clinician
+          name: Clinician First Last Name
+    cliniciansAllViews:
+      emptyView: No Clinicians
+      emptyFindInListView:
+          noResults: No results match your Find in List search
```
paulfalgout: Nit: the double indentation under `emptyFindInListView:` is carried over verbatim from the old file. Harmless YAML, but this is a brand-new file — free cleanup.

src/js/i18n/en-US/shared.yml
```diff
@@ -0,0 +1,32 @@
...
+        listSearchViews:
+          placeholder: Find in List...
+      teamComponent:
+        defaultText: Select Team...
+        headingText: Update Team
+        placeholderText: Team...
+      workspacesManager:
+        workspacesDroplist:
+          addWorkspace: Add Workspace...
+  regions:
```
paulfalgout: Nit: `components`, `shared`, `regions` is neither alphabetical nor grouped. Sorting helps PhraseApp diffs and human scanning.

src/js/i18n/en-US/clinicians.yml
paulfalgout: Handled in `clinicians.yml`: corrected the nested `noResults` indentation.

nmajor25 (approved):

